PR #3, opened 2024-09-11
Examples of clean and smudge

seanses:
A complete example to use PointerFileTranslator to clean and smudge files.


rajatarya:
This looks good to me!

PR #5, opened 2024-09-12
CasObject + LocalClient

rajatarya:
cas_object crate with cas_client/local_client.rs updated to use cas_object crate.

All unit-tests pass.

Next up is remote_client.rs, but this should bring new CAS Object format into xet-core.

### DIFF
--- a/cas_client/src/local_client.rs
+++ b/cas_client/src/local_client.rs
@@ -298,7 +298,7 @@ impl Client for LocalClient {
 
     async fn get_length(&self, prefix: &str, hash: &MerkleHash) -> Result<u64> {
         let file_path = self.get_path_for_entry(prefix, hash);
-        match File::open(&file_path) {
+        match File::open(file_path) {
             Ok(file) => {
                 let mut reader = BufReader::new(file);
                 let cas = CasObject::deserialize(&mut reader)?;


rajatarya:
Most of this code has been reviewed in the xetpoc-monorepo but has been adapted so it can fit into this repo (crates moved around, creating CasObjectError). 

Happy to walk through this code since the changes all appear new. Just let me know.

Relevant PRs: 
* https://github.com/huggingface-internal/xetpoc-monorepo/pull/5
* https://github.com/huggingface-internal/xetpoc-monorepo/pull/10

cas_client/src/local_client.rs
```diff
@@ -404,199 +311,122 @@ impl Client for LocalClient {
...
+
+        // Act & Assert
+        assert!(client.put("", &hash, data, chunk_boundaries).await.is_ok());
+
+        let returned_data = client.get("", &hash).await.unwrap();
+        assert_eq!(data_again, returned_data);
     }
 
     #[tokio::test]
-    async fn test_failures() {
```
rajatarya, line 463: I will pull this test in, is a useful one and I should preserve it.

cas_object/src/error.rs
```diff
@@ -0,0 +1,60 @@
+use std::convert::Infallible;
+
+use http::uri::InvalidUri;
+use merklehash::MerkleHash;
+use xet_error::Error;
+
+#[non_exhaustive]
+#[derive(Error, Debug)]
+pub enum CasObjectError {
```
rajatarya, line 8: I will remove all the enum variants that CasObject is not using.

rajatarya:
I recommend looking at `local_client.rs` - it is the main glue between cas_object_format.rs and the rest of cas_client crate.

cas_object/src/cas_object_format.rs
```diff
@@ -0,0 +1,476 @@
...
+/// See details here: https://www.notion.so/huggingface2/Introduction-To-XetHub-Storage-Architecture-And-The-Integration-Path-54c3d14c682c4e41beab2364f273fc35?pvs=4#4ffa9b930a6942bd87f054714865375d
+pub struct CasObjectHeader {
+    /// CAS identifier: "XCASBLOB"
+    pub ident: [u8; 8],
+
+    /// Format version, expected to be 0 right now.
+    pub version: u8,
+    
+    /// Compression scheme for CAS block, 0 for uncompressed, 1 for LZ4
+    pub compression_method: u8,
```
hoytak, line 26: I put a note in the design doc about this field, but it was after I came back from vacation.   think this field should be larger, like u32.  The reason is that one of the things we considered but did not implement was to use content-specific dictionaries for the compression.  The zSTD API makes this quite doable, where we could build a common fixed compression dictionary based on an archive of files of a certain type, and then use that dictionary to compress and decompress all files of that type.  This means the number of possible compression “methods” could actually be much larger if we choose to do this.  

It's not a priority now, but allowing more options here now would allow that to be implemented in the future without changing the header format.

rajatarya: I'm happy to change this, do you think I should do it now or can I make this change after this one lands?

Mostly would be a pretty small and self-contained change to unit-tests and just this file. The intent with this code structure is that CasObject should be the interface/struct/class that all the Rust code uses for Xorb read/write.

cas_object/src/cas_object_format.rs
```diff
@@ -0,0 +1,476 @@
...
+const CAS_OBJECT_FORMAT_VERSION: u8 = 0;
+const CAS_OBJECT_HEADER_DEFAULT_LENGTH: u32 = 50;
+const CAS_OBJECT_COMPRESSION_UNCOMPRESSED: u8 = 0;
+const CAS_OBJECT_COMPRESSION_LZ4: u8 = 1;
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+/// Header struct for [CasObject]
+/// 
+/// See details here: https://www.notion.so/huggingface2/Introduction-To-XetHub-Storage-Architecture-And-The-Integration-Path-54c3d14c682c4e41beab2364f273fc35?pvs=4#4ffa9b930a6942bd87f054714865375d
+pub struct CasObjectHeader {
```
hoytak, line 18: A late comment in the design doc: ok for now, but down the road we may want to make all of this a footer instead of a header, as having this information in the header requires we build the whole CAS object in memory, then write this, then write the data.   Having this information as a footer allows us to write the CAS data in a stream and then tack this information on at the end.  All the http or s3 ranged requests allow reading from the end, so that won't be a problem.  (This is how we do it now in an mdb shard)

rajatarya: 100% planned. Just not needed for Steel Thread. My plan is to address header -> footer changes after we get Steel Thread going.

It goes into the next block of work (header -> footer, compression) after we get steel thread going. This should be a pretty small change because I'll change it way before we get to prod.

cas_object/src/cas_object_format.rs
rajatarya: Filed this as sub-task here: https://linear.app/xethub/issue/STO-1/cas-object-format

cas_object/src/cas_object_format.rs
rajatarya: Filed this as sub-task here: https://linear.app/xethub/issue/STO-1/cas-object-format

Cargo.toml
```diff
@@ -16,7 +16,7 @@ members = [
     "retry_strategy",
     "shard_client",
     "utils",
-    "xet_error",
+    "xet_error", "cas_object",
```
seanses, line 19: nit: put into a new line

rajatarya: This is fixed in my next commit already - working on RemoteClient.

PR #8, opened 2024-09-13
add cas_types

assafvayner:
Adding cas_types crate to xet-core for client usage, will be added as a dependency serverside

### DIFF
--- a/cas_types/Cargo.toml
+++ b/cas_types/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "cas_types"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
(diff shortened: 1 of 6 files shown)


cas_types/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "cas_types"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+anyhow = "1.0.86"
+serde = { version = "1.0.208", features = ["derive"] }
+merklehash = { git = "https://github.com/xetdata/xet-core" }
```
rajatarya, line 9: This should be fixed to path.

PR #10, opened 2024-09-14
Adding hf_xet library

jgodlew:
Adds an hf_xet library that exposes a python interface for reading/writing pointer files.


data/src/pointer_file.rs
```diff
@@ -12,7 +12,7 @@ const CURRENT_VERSION: &str = "0";
 /// A struct that wraps a Xet pointer file.
 /// Xet pointer file format is a TOML file,
 /// and the first line must be of the form "# xet version <x.y>"
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
```
seanses, line 15: If possible let's not have a Default for PointerFile.

PR #14, opened 2024-09-16
fix lints for stable

assafvayner:
Not sure what was causing these lint issues and maybe only for me, but this PR resolves those lints.


rajatarya:
This looks good to me!

@seanses or @hoytak : any concerns with merging this?

seanses:
> This looks good to me!
> 
> @seanses or @hoytak : any concerns with merging this?

Let's merge it!

PR #16, opened 2024-09-19
CasObject v2

rajatarya:
* Changes header to footer.
* Supports chunk compression and chunk headers
* Updated unit-tests

### DIFF
--- a/cas_object/src/cas_chunk_format.rs
+++ b/cas_object/src/cas_chunk_format.rs
@@ -75,6 +75,7 @@ fn write_chunk_header<W: Write>(w: &mut W, chunk_header: &CASChunkHeader) -> std
 
 #[inline]
 fn copy_three_byte_num(buf: &mut [u8; 3], num: u32) {
+    debug_assert!(num < 16_777_216); // verify that chunk is under 16MB
     let bytes = num.to_le_bytes(); // Convert u32 to little-endian bytes
     buf.copy_from_slice(&bytes[0..3]);
 }
--- a/cas_object/src/cas_object_format.rs
+++ b/cas_object/src/cas_object_format.rs
@@ -281,7 +281,7 @@ impl CasObject {
     ///
     /// Ex. If user requests range bytes 150-250 from a Xorb, and assume the following layout:
     /// ```
-    /// //              chunk: [   0  |    1    |    2    |    3    ]
+    /// //               chunk: [   0  |    1    |    2    |    3    ]
     /// // uncompressed chunks: [ 0-99 | 100-199 | 200-299 | 300-399 ]
     /// //   compressed chunks: [ 0-49 |   50-99 | 100-149 | 150-199 ]
     /// ```


cas_object/src/cas_chunk_format.rs
```diff
@@ -0,0 +1,209 @@
...
+    Ok(result)
+}
+
+pub fn deserialize_chunk<R: Read>(
+    reader: &mut R,
+) -> Result<(CASChunkHeader, Vec<u8>), CasObjectError> {
+    let header = deserialize_chunk_header(reader)?;
+    let mut compressed_buf = Vec::with_capacity(header.get_compressed_length() as usize);
+    let mut chunk = reader.take(header.get_compressed_length() as u64);
+    chunk.read_to_end(&mut compressed_buf)?;
```
hoytak: Issue with this approach is that if the file is truncated, this will succeed without an error.   Let's do 

```
let mut compressed_buf = vec![0 ; header.get_compressed_length()]; 
reader.read_all(&mut compressed_buf[..])?;
```

cas_object/src/cas_chunk_format.rs
```diff
@@ -0,0 +1,209 @@
...
+
+    let uncompressed_buf = match header.get_compression_scheme() {
+        CompressionScheme::None => compressed_buf,
+        CompressionScheme::LZ4 => {
+            let mut uncompressed_buf = Vec::new();
+            let mut dec = FrameDecoder::new(Cursor::new(compressed_buf));
```
hoytak, line 151: We could also just pass the reader in here?

cas_object/src/cas_chunk_format.rs
```diff
@@ -0,0 +1,209 @@
...
+            copy(&mut dec, &mut uncompressed_buf)?;
```
hoytak: Could also just return Vec::from_iter(dec)

cas_object/src/cas_object_format.rs
```diff
@@ -1,70 +1,81 @@
...
 
-const CAS_OBJECT_FORMAT_IDENT: [u8; 8] = [ b'X', b'C', b'A', b'S', b'B', b'L', b'O', b'B'];
+const CAS_OBJECT_FORMAT_IDENT: [u8; 7] = [b'X', b'E', b'T', b'B', b'L', b'O', b'B'];
 const CAS_OBJECT_FORMAT_VERSION: u8 = 0;
-const CAS_OBJECT_HEADER_DEFAULT_LENGTH: u32 = 50;
-const CAS_OBJECT_COMPRESSION_UNCOMPRESSED: u8 = 0;
-const CAS_OBJECT_COMPRESSION_LZ4: u8 = 1;
+const CAS_OBJECT_INFO_DEFAULT_LENGTH: u32 = 60;
 
+// TODO: #[repr(C, packed)] - this requires alignment stuff to get corrected for Vec & DataHash.
```
hoytak: Not really necessary here, this isn't a POD struct with the vec and stuff in it so doesn't really matter.

cas_object/src/cas_object_format.rs
```diff
@@ -74,45 +85,51 @@ impl CasObjectHeader {
...
+        // now seek back that many bytes + size of length (u32) and read sequentially.
+        reader.seek(std::io::SeekFrom::End(
+            -(size_of::<u32>() as i64 + info_length as i64),
+        ))?;
 
         // Helper function to read data and update the byte count
         let mut read_bytes = |data: &mut [u8]| -> Result<(), CasObjectError> {
             reader.read_exact(data)?;
-            total_bytes_read += data.len();
+            total_bytes_read += data.len() as u32;
```
hoytak, line 134: reader.read_exact(data)? returns the number of bytes read, could just use that

cas_object/src/cas_object_format.rs
```diff
@@ -123,198 +140,336 @@ impl CasObjectHeader {
...
 
         let mut num_chunks = [0u8; 4];
         read_bytes(&mut num_chunks)?;
         let num_chunks = u32::from_le_bytes(num_chunks);
 
-        let mut chunk_uncompressed_len = Vec::with_capacity(num_chunks as usize);
+        let mut chunk_size_info = Vec::with_capacity(num_chunks as usize);
         for _ in 0..num_chunks {
-            let mut buf = [0u8; 4];
+            let mut buf = [0u8; 8];
```
hoytak: This is correct, but for clarity, could use sizeof here

cas_object/src/cas_object_format.rs
```diff
@@ -123,198 +140,336 @@ impl CasObjectHeader {
...
+                _buffer,
+            },
+            info_length,
+        ))
     }
 }
 
+impl CasChunkInfo {
+    pub fn as_bytes(&self) -> Result<Vec<u8>, CasObjectError> {
+        let mut serialized_bytes = Vec::with_capacity(2 * size_of::<u32>()); // 8 bytes, 2 u32
```
hoytak: This allocation in an inner loop is going to get expensive...  I'd simply return a [u8 ; 8] object by value instead of a Vec<u8>.   Also, it can never fail, so no need to return a Result.  (as_bytes() usually doesn't.)

```
let ret = [self.start_byte_index.to_le_bytes(), self.cumulative_uncompressed_len.to_le_bytes()];
unsafe { std::mem::transmute(ret) }
```

cas_object/src/cas_chunk_format.rs
assafvayner: What if the reader is longer than the rest of the chunk? Maybe best to just do a `.take()`

cas_object/src/cas_object_format.rs
assafvayner: I modified this section to not use Vec, but I'll leave it to Rajat to decide if we want to use the unsafe option here. I think it's fine, but it doesn't save us that much work here.

cas_object/src/cas_object_format.rs
rajatarya: `reader.read_exact()` returns a `Result<()>` so that's why I'm doing it this way. Do you mean a different read API? My understanding with read_exact() is that it will keep going back until it has filled the buffer provided.

cas_object/src/cas_object_format.rs
rajatarya: I've left as @assafvayner wrote - @hoytak what do you think? I am not familiar with transmute and unsafe - will look for your guidance here.

cas_object/src/cas_chunk_format.rs
rajatarya: I think the concern @assafvayner is raising is that we only want to read/decode the length of this chunk, so is there a way to only copy a specific number of bytes from reader -> writer without introducing an in-memory buffer?

assafvayner: yes this is the second piece, the lz4_flex library doesn't give you a good way to know how much was written. That being said there is a really easy change to make the library make the info available, there was a PR for it, no progress since. We could fork it and do that.

cas_object/src/cas_chunk_format.rs
```diff
@@ -0,0 +1,281 @@
...
+
+fn write_chunk_header<W: Write>(w: &mut W, chunk_header: &CASChunkHeader) -> std::io::Result<()> {
+    w.write_all(&[chunk_header.version])?;
+    w.write_all(&chunk_header.compressed_length)?;
+    w.write_all(&[chunk_header.compression_scheme])?;
+    w.write_all(&chunk_header.uncompressed_length)
+}
+
+#[inline]
+fn copy_three_byte_num(buf: &mut [u8; 3], num: u32) {
```
seanses, line 77: Ideally we should check `num < 2^24`, maybe in debug mode only `debug_assert!()`

rajatarya: Done!

cas_object/src/cas_object_format.rs
hoytak: Ah, most of them do, but I guess read_exact does not.   I stand corrected.

PR #19, opened 2024-09-23
Plumb endpoint and access token

bpronan:
Plumbing the endpoint and JWT access token from the python interface through to the request headers for the reconstruction and shard POST|PUT calls.

Partially resolves:
https://linear.app/xethub/issue/STO-35/secure-read-path
https://linear.app/xethub/issue/STO-36/secure-write-path

### DIFF
--- a/cas_types/src/lib.rs
+++ b/cas_types/src/lib.rs
@@ -22,7 +22,9 @@ pub struct CASReconstructionTerm {
     pub hash: HexMerkleHash,
     pub unpacked_length: u32,
     pub range: Range,
-    pub range_start_offset: u32,
+    // TODO: disabling until https://github.com/huggingface-internal/xetcas/pull/31/files
+    // is merged.
+    // pub range_start_offset: u32,
     pub url: String,
 }
 


hf_xet/src/lib.rs
```diff
@@ -7,29 +7,29 @@ use pyo3::prelude::*;
 use data::PointerFile;
 
 #[pyfunction]
-#[pyo3(signature = (file_paths), text_signature = "(file_paths: List[str]) -> List[PyPointerFile]")]
-pub fn upload_files(file_paths: Vec<String>) -> PyResult<Vec<PyPointerFile>> {
+#[pyo3(signature = (file_paths, endpoint, token), text_signature = "(file_paths: List[str], endpoint: Optional[str], token: Optional[str]) -> List[str]")]
```
jgodlew: nit: this should still be returning a `PyPointerFile`

data/src/configurations.rs
```diff
@@ -25,6 +25,7 @@ pub struct CacheConfig {
 #[derive(Debug)]
 pub struct StorageConfig {
     pub endpoint: Endpoint,
+    pub token: Option<String>,
     pub auth: Auth,
```
jgodlew: nit: should the token be a part of the `Auth` struct instead?

data/src/cas_interface.rs
```diff
@@ -49,7 +49,7 @@ pub(crate) async fn create_cas_client(
 
     // Raw remote client.
     let remote_client = Arc::new(
-        RemoteClient::from_config(endpoint.to_string()).await,
+        RemoteClient::from_config(endpoint.to_string(), None).await,
```
jgodlew: Should this be the token stored in the config (i.e. `cas_storage_config.token`)?

data/src/configurations.rs
seanses: Feel free to drop other fields in `Auth` if we don't need them. @bpronan

data/src/cas_interface.rs
bpronan: Indeed it should.

data/src/configurations.rs
bpronan: Done. I've moved token into auth and removed the other fields. Even though it only has one field, I'm leaving auth as a separate struct because we may have other fields we need to add.

PR #20, opened 2024-09-23
test-utils crate

assafvayner:
helper crate to share test utilities like generating cas_objects and shards, mainly for use in cas_server

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ members = [
     "utils",
     "xet_error",
     "cas_object",
-    "cas_types",
+    "cas_types", "test-utils",
 ]
 
 exclude = [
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -20,7 +20,7 @@ members = [
     "utils",
     "xet_error",
     "cas_object",
-    "cas_types",
+    "cas_types", "test-utils",
```
seanses, line 23: Looks like there's a merge conflict, notice double occurrences of `cas_object` and `cas_types`

assafvayner: I'm going to close this PR and move the code to xetcas. using these inside xet-core is causing circular dep issues. (not exactly, but too many to try to remedy)

PR #21, opened 2024-09-24
CasObject validation with updated tests

rajatarya:
- also added Copy trait to HexMerkleHash struct

### DIFF
--- a/cas_object/src/cas_object_format.rs
+++ b/cas_object/src/cas_object_format.rs
@@ -553,14 +553,18 @@ impl CasObject {
         let mut hash_chunks: Vec<Chunk> = Vec::new();
         let mut cumulative_uncompressed_length: u32 = 0;
         let mut cumulative_compressed_length: u32 = 0;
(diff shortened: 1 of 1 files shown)


cas_object/src/cas_object_format.rs
```diff
@@ -538,6 +539,78 @@ impl CasObject {
...
+        let mut cumulative_uncompressed_length: u32 = 0;
+        let mut cumulative_compressed_length: u32 = 0;
+        for (idx, c) in cas.info.chunk_size_info[..cas.info.chunk_size_info.len() - 1].iter().enumerate() {
+
+            // fence post conditions - starting chunk and final chunk
+            if idx == 0 && c.start_byte_index != 0 {
+                // for 1st chunk verify that its start_byte_index is 0
+                warn!("XORB Validation: Byte 0 does not contain 1st chunk.");
+                return Ok(false);
+            }
```
assafvayner: can we move this outside the loop to check the 1 value?

PR #23, opened 2024-09-24
s3 range propagation

assafvayner:
needed to succeed in making s3 presigned url range pull requests.

### DIFF
--- a/cas_client/src/remote_client.rs
+++ b/cas_client/src/remote_client.rs
@@ -270,6 +270,13 @@ impl CASAPIClient {
 
 async fn get_one(term: &CASReconstructionTerm) -> Result<Bytes> {
     debug!("term: {term:?}");
+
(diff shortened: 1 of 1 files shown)


cas_client/src/remote_client.rs
```diff
@@ -275,11 +286,13 @@ async fn get_one(term: &CASReconstructionTerm) -> Result<Bytes> {
         .await
         .map_err(CasClientError::ReqwestError)?;
     let mut readseek = Cursor::new(xorb_bytes.to_vec());
+    let data = cas_object::deserialize_chunks(&mut readseek)?;
+    let len = (term.range.end - term.range.start) as usize;
```
rajatarya, line 296: Probably want to check that term.range.end > term.range.start so this convert to usize doesn't panic.

PR #25, opened 2024-09-24
Passing the JWT token to all xetcas requests

bpronan:
Client side changes for https://github.com/huggingface-internal/xetcas/pull/35.

Resolves (part of) STO-40.


bpronan:
Good point. I'll add a task to track that.

PR #26, opened 2024-09-26
HMAC implementation for keyed hashing of a MerkleHash

hoytak:
Implements `hmac_hash(key : [u8 ; 32], hash : MerkleHash)` to generate a protected keyed hash.   It uses the standard HMAC algorithm with blake3 as the cryptographic hash function.

### DIFF
--- a/merklehash/src/data_hash.rs
+++ b/merklehash/src/data_hash.rs
@@ -176,6 +176,13 @@ impl DataHash {
         }
         Ok(hash)
     }
+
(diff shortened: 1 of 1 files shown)


ylow:
Is there not already a HMAC we can use? We should not be implementing our own crypto.

ylow:
We are already using keyed_hash elsewhere. We should just use that https://docs.rs/blake3/latest/blake3/fn.keyed_hash.html

PR #28, opened 2024-09-26
Adding the auth to the default headers instead of to each call

bpronan:
This clean the headers up from the different reqwest::send calls. 

Middleware required a little too much boilerplate, though it's likely going to be the final solution here when we add token refresh logic.

Resolves STO-63.


bpronan:
Closing pull request as the refresh middleware removes the need for this change.

PR #29, opened 2024-09-27
CAS client API refactor; Shard and CAS object format changes to use chunk indices

seanses:
- Delete grpc cas and shard code.
- Rewrite CAS client api as there will be no API to get cas block by hash.
- Delete CachingClient and StagingClient (need to rewrite them in the future).
- Use chunk indices in `FileDataSequenceEntry` and cas object format.
- Misc code clean up and reorganization.

Related server-side PR:

- https://github.com/huggingface-internal/xetcas/pull/46
- https://github.com/huggingface-internal/xetcas/pull/53


cas_object/src/cas_chunk_format.rs
```diff
@@ -145,14 +145,23 @@ pub fn deserialize_chunk_to_writer<R: Read, W: Write>(
...
-        CompressionScheme::None => writer.write_all(&compressed_buf)?,
+    let uncompressed_len = match header.get_compression_scheme() {
+        CompressionScheme::None => {
+            writer.write_all(&compressed_buf)?;
+            compressed_buf.len() as u32
+        }
         CompressionScheme::LZ4 => {
             let mut dec = FrameDecoder::new(Cursor::new(compressed_buf));
-            copy(&mut dec, writer)?;
+            copy(&mut dec, writer)? as u32
```
assafvayner, line 155: IIRC this will return the compressed len!

assafvayner: Actually, not sure anymore.

cas_client/src/remote_client.rs
```diff
@@ -280,14 +199,13 @@ async fn get_one(term: &CASReconstructionTerm) -> Result<Bytes> {
         .bytes()
         .await
         .map_err(CasClientError::ReqwestError)?;
+    if xorb_bytes.len() as u32 != term.url_range.end - term.url_range.start {
+        error!("got back a smaller range than requested");
```
rajatarya, line 254: Should this return an Err instead of just logging?

shard_client/src/http_shard_client.rs
```diff
@@ -182,17 +195,66 @@ impl ShardDedupProber<ShardClientError> for HttpShardClient {
...
 
         let response_body = response.bytes().await?;
-        let response_info: QueryChunkResponse = serde_json::from_reader(response_body.reader())?;
+        let mut reader = response_body.reader();
+
+        let mut downloaded_shards = vec![];
+
+        let mut write_join_set = JoinSet::<Result<()>>::new();
+
+        // Return in the format of
```
rajatarya, line 204: What concerns me with this loop is that the server can "DOS" the client by just continuously sending 'bytes'. Instead, could this protocol start with num_shards: usize and then for each shard: [key, shard] (like it is here)?

rajatarya: Also, what happens if the `read_u32` call fails for another reason (ex. 'broken pipe') - shouldn't that error propagate up to trigger a retry?

utils/src/singleflight.rs
rajatarya, line 1: Should we delete this file?

cas_object/src/cas_chunk_format.rs
seanses: The number of bytes read and that written should equal: `copy` doesn't do any magic, it just does byte-by-byte copy. It's the `impl Read for FrameDecoder` that generates more bytes than what's read from `compressed_buf`.

utils/src/singleflight.rs
seanses: Maybe not so early. We may need it again.

cas_client/src/remote_client.rs
seanses: Fixed.

shard_client/src/http_shard_client.rs
seanses: > What concerns me with this loop is that the server can "DOS" the client by just continuously sending 'bytes'. Instead, could this protocol start with num_shards: usize and then for each shard: [key, shard] (like it is here)?

Rather, maybe the client can define some threshold that "I have had enough dedup info downloaded", and stop reading from the socket.

seanses: > Also, what happens if the `read_u32` call fails for another reason (ex. 'broken pipe') - shouldn't that error propagate up to trigger a retry?

Actually we didn't implement streaming: the `response.bytes()` above already collected all bytes so there's no pipe here.

seanses: I checked multiple places and looks like we never retry on corrupted response. I guess that will be severe bug and it's not worthy to retry?

PR #30, opened 2024-09-27
Add token refresh to CAS/Shard Client

jgodlew:
Fixes STO-71

* Update parameters for python interface to group the token with an expiry and adds a callback function to refresh tokens.
* Moves `Auth` config to utils crate for sharing between our crates (until CAS client + Shard client are merged together)
* Adds a Trait: `TokenRefresher` to wrap the functionality for getting a refreshed token.
* In hf_xet, implemented the `TokenRefresher` with a `WrappedTokenRefresher` that wraps a python function for refreshing the token.
* Adds a middleware to CAS clients to apply the auth token to request headers.

### DIFF
--- a/cas_client/src/auth.rs
+++ b/cas_client/src/auth.rs
@@ -13,6 +13,15 @@ pub struct AuthMiddleware {
 }
 
 impl AuthMiddleware {
+    /// Fetches a token from our TokenProvider. This locks the TokenProvider as we might need
+    /// to refresh the token if it has expired.
+    ///
+    /// In the common case, this lock is held only to read the underlying token stored
+    /// in memory. However, in the event of an expired token (e.g. once every 15 min),
+    /// we will need to hold the lock while making a call to refresh the token
+    /// (e.g. to a remote service). During this time, no other CAS requests can proceed
+    /// from this client until the token has been fetched. This is expected/ok since we
+    /// don't have a valid token and thus any calls would fail.
     fn get_token(&self) -> Result<String, anyhow::Error> {
         let mut provider = self
             .token_provider
--- a/utils/src/auth.rs
+++ b/utils/src/auth.rs
@@ -22,6 +22,17 @@ impl TokenRefresher for NoOpTokenRefresher {
     }
 }
 
+#[derive(Debug)]
+pub struct ErrTokenRefresher;
+
+impl TokenRefresher for ErrTokenRefresher {
+    fn refresh(&self) -> Result<TokenInfo, AuthError> {
+        Err(AuthError::RefreshFunctionNotCallable(
+            "Token refresh not expected".to_string(),
+        ))
+    }
+}
+
 /// Shared configuration for token-based auth
 #[derive(Debug, Clone)]
 pub struct AuthConfig {
@@ -51,7 +62,7 @@ impl AuthConfig {
             (Some(token), expiry, None) => Some(Self {
                 token,
                 token_expiration: expiry.unwrap_or(u64::MAX),
-                token_refresher: Arc::new(NoOpTokenRefresher),
+                token_refresher: Arc::new(ErrTokenRefresher),
             }),
             (_, _, _) => None,
         }


cas_client/src/auth.rs
```diff
@@ -0,0 +1,57 @@
...
+use tracing::info;
+
+/// AuthMiddleware is a thread-safe middleware that adds a CAS auth token to outbound requests.
+/// If the token it holds is expired, it will automatically be refreshed.
+pub struct AuthMiddleware {
+    token_provider: Option<Arc<Mutex<TokenProvider>>>,
+}
+
+impl AuthMiddleware {
+    fn get_token(provider_ref: &Arc<Mutex<TokenProvider>>) -> Result<String, anyhow::Error> {
```
sirahd: [nit] can't we make `get_token` operate on `self.token_provider` and return `Option<Result<String, anyhow::Error>` based on its existence? 
```suggestion
    fn get_token(&self) -> Option<Result<String, anyhow::Error>> {
```

hf_xet/src/lib.rs
```diff
@@ -1,37 +1,78 @@
...
+    files: Vec<PyPointerFile>,
+    endpoint: Option<String>,
+    token_info: Option<(String, u64)>,
+    token_refresher: Option<Py<PyAny>>,
+) -> PyResult<Vec<String>> {
+    let pfs = files.into_iter().map(PointerFile::from).collect();
+    let refresher = token_refresher
+        .map(WrappedTokenRefresher::from_func)
+        .transpose()?
+        .map(to_arc_dyn);
```
sirahd, line 57: does the compiler complain if we do this?
```suggestion
        .map(Arc::new);
```

hf_xet/src/lib.rs
```diff
@@ -77,15 +118,52 @@ impl PyPointerFile {
...
+            .enable_all()
+            .build()?
+            .block_on(async move { try_refresh(refresher).await })
+            .map_err(|e| PyException::new_err(format!("{e:?}")))
+    })
+}
+
+async fn try_refresh(refresher: WrappedTokenRefresher) -> errors::Result<(String, u64)> {
+    let refresher = Arc::new(refresher);
+    let tokens = tokio_par_for_each(
```
sirahd: Does this mean every time `try_refresh` is called, there will be `data_client::MAX_CONCURRENT_UPLOADS` request to moon-landing? Is this necessary?

utils/src/auth.rs
```diff
@@ -0,0 +1,88 @@
...
+pub trait TokenRefresher: Debug + Send + Sync {
+    /// Get a new auth token for CAS and the unixtime (in seconds) for expiration
+    fn refresh(&self) -> Result<(String, u64), AuthError>;
+}
+
+#[derive(Debug)]
+pub struct NoOpTokenRefresher;
+
+impl TokenRefresher for NoOpTokenRefresher {
+    fn refresh(&self) -> Result<(String, u64), AuthError> {
```
sirahd: we should define this tuple as a separate type for readability (only if it does not complicate the pyo3 interface too much)

utils/src/auth.rs
```diff
@@ -0,0 +1,88 @@
...
+        Ok(("token".to_string(), 0))
```
sirahd, line 21: Should the `NoOpTokenRefresher` return an error when `refresh` is called so that the error gets propagated, instead of returning invalid token?

utils/src/auth.rs
```diff
@@ -0,0 +1,88 @@
...
+                token_refresher: Arc::new(NoOpTokenRefresher),
+            }),
+            (_, _, _) => None,
+        }
+    }
+}
+
+pub struct TokenProvider {
+    token: String,
+    expiration: u64,
```
assafvayner, line 74: Any reason not to keep this as a SystemTime? minor optimization

shard_client/src/http_shard_client.rs
```diff
@@ -43,14 +50,14 @@ fn retry_http_status_code(stat: &reqwest::StatusCode) -> bool {
     stat.is_server_error() || *stat == reqwest::StatusCode::TOO_MANY_REQUESTS
 }
 
-fn is_status_retriable_and_print(err: &reqwest::Error) -> bool {
+fn is_status_retriable_and_print(err: &reqwest_middleware::Error) -> bool {
     let ret = if let Some(code) = err.status() {
```
assafvayner: style nit, maybe change this to err.status().map(... retry_http_status_code(...).ok_or(true)

hf_xet/src/lib.rs
assafvayner: you might need the extra step to coerce the Arc's inner to a dyn trait

hf_xet/src/config.rs
```diff
@@ -51,3 +58,10 @@ pub fn default_config(endpoint: String, token: Option<String>) -> errors::Result
 
     Ok(translator_config)
 }
+
+fn convert(opt: Option<(String, u64)>) -> (Option<String>, Option<u64>) {
+    match opt {
+        Some((s, n)) => (Some(s), Some(n)),
+        None => (None, None),
+    }
+}
```
assafvayner: look at https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.unzip

data/src/shard_interface.rs
```diff
@@ -39,7 +39,7 @@ pub async fn create_shard_client(
 ) -> Result<Arc<dyn ShardClientInterface>> {
     info!("Shard endpoint = {:?}", shard_storage_config.endpoint);
     let client: Arc<dyn ShardClientInterface> = match &shard_storage_config.endpoint {
-        Server(endpoint) => Arc::new(HttpShardClient::new(endpoint, shard_storage_config.auth.token.clone())),
+        Server(endpoint) => Arc::new(HttpShardClient::new(endpoint, shard_storage_config.auth.as_ref())),
```
assafvayner: as a slight style nit, I prefer to just take a reference (prepend &) if it works. if it doesn't keep this, also if you prefer as_ref, I'll use it elsewhere when needed for consistency across the code base

cas_client/src/remote_client.rs
assafvayner, line 1: there might be conflicts with [PR](https://github.com/huggingface-internal/xet-core/pull/28)

cas_client/src/auth.rs
```diff
@@ -0,0 +1,57 @@
...
+            .map_err(|e| anyhow!("lock error: {e:?}"))?;
+        provider
+            .get_valid_token()
+            .map_err(|e| anyhow!("couldn't get token: {e:?}"))
+    }
+}
+
+impl From<Option<&AuthConfig>> for AuthMiddleware {
+    fn from(value: Option<&AuthConfig>) -> Self {
+        if value.is_none() {
```
assafvayner: this makes me want to add ErrorPrinter trait to Option<T> if you're up for it, should be quick, actually probably OptionPrinter trait.

hf_xet/src/lib.rs
jgodlew: Yeah, Assaf is right, the compiler doesn't know how to coerce the `Arc<WrappedTokenRefresher>` inside the map to a `Arc<dyn TokenRefresher>`...

So you either need a helper function that does that (e.g. the `to_arc_dyn` we have in the code) or expand the map statement and explicitly cast the new Arc:
```
.map(|r| Arc::new(r) as Arc<dyn TokenRefresher>)
```
I personally find the function more readable.

hf_xet/src/lib.rs
jgodlew: Sorry, this was a testing function to help de-risk the design of calling a Python Callable from inside rust async code. I'll remove this function and the associated python-exposed function (`try_refresh_token`) since they aren't used.

cas_client/src/remote_client.rs
jgodlew: Yeah, this should replace that PR.

cas_client/src/auth.rs
```diff
@@ -0,0 +1,54 @@
...
+    token_provider: Arc<Mutex<TokenProvider>>,
+}
+
+impl AuthMiddleware {
+    fn get_token(&self) -> Result<String, anyhow::Error> {
+        let mut provider = self
+            .token_provider
+            .lock()
+            .map_err(|e| anyhow!("lock error: {e:?}"))?;
+        provider
```
port8080, line 30: Can you add a comment here about the likely duration of `get_valid_token`? I want to understand the impact of this lock being held across a remote call: I think no other operation can go ahead if the token is invalid, but would like to see that commented.

hf_xet/src/lib.rs
```diff
@@ -1,37 +1,76 @@
...
+pub fn upload_files(
+    py: Python,
+    file_paths: Vec<String>,
+    endpoint: Option<String>,
+    token_info: Option<(String, u64)>,
+    token_refresher: Option<Py<PyAny>>,
+) -> PyResult<Vec<PyPointerFile>> {
+    let refresher = token_refresher
+        .map(WrappedTokenRefresher::from_func)
+        .transpose()?
```
port8080, line 26: 👏

cas_client/src/auth.rs
jgodlew: I've refactored this to only build an `AuthMiddleware` when there is a config specified, making this function a method on the struct.

utils/src/auth.rs
jgodlew: SystemTime is not useful for storing the date time that we would need (e.g. no method for getting a SystemTime from an epoch timestamp). I could import the chrono crate to parse and store `DateTime` structs, but would rather work with epoch time as a u64 instead.

data/src/shard_interface.rs
jgodlew: I don't have a strong opinion about `&Option<T>` vs `Option<&T>` as they both have pretty minimal pros and cons. I've swapped out my code to use the `&Option<T>` and translate when needed (i.e. constructing the middleware).

cas_client/src/auth.rs
jgodlew: I've added an OptionPrinter to the `error_printer` crate.

utils/src/auth.rs
jgodlew: I've made this a type alias for the time being, I am not sure if we need a full struct type as it is pretty simple, but I don't have a strong opinion on that.

cas_client/src/auth.rs
jgodlew: I've added a comment to the method. Most of the time (i.e. when the token is valid), this token is just an in-memory lookup (<1us) and when the token is expired, everyone has to wait anyways for a valid token to come back from Hub.

utils/src/auth.rs
jgodlew: This is mainly used for testing/if we have a long-lived credential. I've added an `ErrTokenRefresher` that responds with an error (for cases where there is no refresher specified).

PR #33, opened 2024-10-03
Shard management and searching with HMAC keying 

hoytak:
Ready for Review: 

- Adds hmac keys to the shard format and shard management infrastructure.
- Adds function to do a streaming rebuild of a shard with hmac keys applied.

### DIFF
--- a/mdb_shard/src/shard_format.rs
+++ b/mdb_shard/src/shard_format.rs
@@ -145,9 +145,9 @@ impl MDBShardFileFooter {
 
         // Do a version check as a simple guard against using this in an old repository
         if version != MDB_SHARD_FOOTER_VERSION {
-            //return Err(MDBShardError::ShardVersionError(
-            panic!("Error: Expected footer version {MDB_SHARD_FOOTER_VERSION}, got {version}");
-            //));
+            return Err(MDBShardError::ShardVersionError(format!(
+                "Error: Expected footer version {MDB_SHARD_FOOTER_VERSION}, got {version}"
+            )));
         }
 
         let mut obj = Self {


mdb_shard/src/shard_file_handle.rs
```diff
@@ -130,6 +132,33 @@ impl MDBShardFile {
         Ok(ret)
     }
 
+    /// Write out the current shard, re-keyed with an hmac key, to the output directory in question, returning
+    /// the foll path to the new shard.
```
seanses: typo
```suggestion
    /// the full path to the new shard.
```

mdb_shard/src/shard_format.rs
```diff
@@ -144,9 +146,9 @@ impl MDBShardFileFooter {
 
         // Do a version check as a simple guard against using this in an old repository
         if version != MDB_SHARD_FOOTER_VERSION {
-            return Err(MDBShardError::ShardVersionError(format!(
-                "Error: Expected header version {MDB_SHARD_FOOTER_VERSION}, got {version}"
-            )));
+            //return Err(MDBShardError::ShardVersionError(
```
seanses: we can drop this commented snippet

mdb_shard/src/shard_format.rs
```diff
@@ -643,14 +668,22 @@ impl MDBShardInfo {
...
+        unkeyed_query_hashes: &[MerkleHash],
         cas_entry_index: u32,
         cas_chunk_offset: u32,
     ) -> Result<Option<(usize, FileDataSequenceEntry)>> {
-        if query_hashes.is_empty() {
+        if unkeyed_query_hashes.is_empty() {
             return Ok(None);
         }
 
+        let keyed_hash = |h: MerkleHash| {
```
seanses: Is this identical to the above `pub fn keyed_chunk_hash`?

hoytak: Yes, will consolidate.

mdb_shard/src/shard_format.rs
hoytak: The panic below was for catching an annoying error during debugging, reverted that.

PR #34, opened 2024-10-04
Collect telemetry for upload path

port8080:
Get the telemetry data to the py layer. This will be sent over the telemetry endpoint later.

Related to STO-7

### DIFF
--- a/cas_client/src/interface.rs
+++ b/cas_client/src/interface.rs
@@ -42,7 +42,7 @@ pub trait UploadClient {
 #[async_trait]
 pub trait ReconstructionClient {
     /// Get a entire file by file hash.
-    async fn get_file(&self, hash: &MerkleHash, writer: &mut Box<dyn Write + Send>) -> Result<()>;
+    async fn get_file(&self, hash: &MerkleHash, writer: &mut Box<dyn Write + Send>) -> Result<u64>;
 
     /// Get a entire file by file hash at a specific bytes range.
     async fn get_file_byte_range(
@@ -51,7 +51,7 @@ pub trait ReconstructionClient {
         offset: u64,
         length: u64,
         writer: &mut Box<dyn Write + Send>,
-    ) -> Result<()>;
+    ) -> Result<u64>;
 }
 
 pub trait Client: UploadClient + ReconstructionClient {}
--- a/cas_client/src/remote_client.rs
+++ b/cas_client/src/remote_client.rs
@@ -102,13 +102,13 @@ impl UploadClient for RemoteClient {
 
 #[async_trait]
 impl ReconstructionClient for RemoteClient {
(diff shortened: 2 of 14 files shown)


cas_client/src/interface.rs
```diff
@@ -42,7 +42,7 @@ pub trait UploadClient {
 #[async_trait]
 pub trait ReconstructionClient {
     /// Get a entire file by file hash.
-    async fn get_file(&self, hash: &MerkleHash, writer: &mut Box<dyn Write + Send>) -> Result<()>;
+    async fn get_file(&self, hash: &MerkleHash, writer: &mut Box<dyn Write + Send>) -> Result<u64>;
```
assafvayner, line 45: add a comment for what this return value is, presumably amount written to writer

cas_client/src/remote_client.rs
```diff
@@ -215,7 +215,8 @@ impl RemoteClient {
                 .map_err(|e| CasClientError::InternalError(anyhow!("join error {e}")))??;
             writer.write_all(&piece)?;
         }
-        Ok(total_len as usize)
+        // Todo: return the bytes which were read from the cache for telemetry
+        Ok(total_len as u64)
```
assafvayner, line 219: this total len is the uncompressed length of the section, not the number of bytes over the network which is what I think you're attempting to track that number should be the difference in the url_range end - start for each piece.

data/src/data_processing.rs
```diff
@@ -283,29 +288,37 @@ pub(crate) async fn register_new_cas_block(
     cas_data.chunks.clear();
     cas_data.pending_file_info.clear();
 
-    Ok(cas_hash)
+    Ok((cas_hash, compressed_bytes_len as u64))
```
seanses, line 291: nit: using `compressed_bytes_len` from client side is unreliable, people can inject any number here and that does not impact clean process correctness.

data/src/clean.rs
```diff
@@ -577,13 +585,14 @@ impl Cleaner {
             if cas_data_accumulator.data.len() >= TARGET_CAS_BLOCK_SIZE {
                 let mut new_cas_data = take(cas_data_accumulator.deref_mut());
                 drop(cas_data_accumulator); // Release the lock.
-                register_new_cas_block(
+                let (_cas_hash, compressed_bytes) = register_new_cas_block(
```
seanses, line 588: This `compressed_bytes` 
- doesn't include all bytes produced by this file after dedup: they are also produced in `fn dedup(...)`. I noticed you modified that function but the return value is not used? 
- data in this `cas_data_accumulator` may also include bytes produced by other files cleaned before this file using the same PointerFileTranslator.

rajatarya:
Lots of things have changed since the time of this PR, closing it out. If we choose to add telemetry later on we will likely have to start from scratch anyway.

PR #35, opened 2024-10-04
Using blake3 hmac for protected chunk support

rajatarya:
Making the protected chunk range hash more secure - using `blake3::keyed_hash`.

**Is it OK that this function returns a Vec<u8>?**

### DIFF
--- a/cas_object/src/cas_object_format.rs
+++ b/cas_object/src/cas_object_format.rs
@@ -6,7 +6,7 @@ use crate::{
 use anyhow::anyhow;
 use bytes::Buf;
 use merkledb::{prelude::MerkleDBHighLevelMethodsV1, Chunk, MerkleMemDB};
-use merklehash::MerkleHash;
+use merklehash::{DataHash, MerkleHash};
 use std::{
     cmp::min,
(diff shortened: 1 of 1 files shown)


cas_object/src/cas_object_format.rs
```diff
@@ -378,8 +378,8 @@ impl CasObject {
         &self,
         chunk_start_index: u32,
         chunk_end_index: u32,
-        key: &[u8],
-    ) -> Result<MerkleHash, CasObjectError> {
+        key: &[u8; 32],
+    ) -> Result<Vec<u8>, CasObjectError> {
```
seanses: Probably not returning as `Vec<u8>`, that suggests arbitrary length. Let's convert it to MerkleHash.

rajatarya: OK that is what I was thinking, but it sort of weird because doesn't MerkleHash imply that it was generated via our merklehash.

The other thought was to return blake3::Hash directly, but that felt like it was leaking the abstraction.

Should I create a basic type (ex. HMACHash) and return that instead?

PR #36, opened 2024-10-08
Creation timestamp and key expiry for shards

hoytak:
Shards now have a creation timestamp and an expiry for the key.

All shards now have a timestamp denoting SystemTime::now() for the shard on writing of the footer.

As for key expiration, local shards do not have a key expiration by default, and keyed shards must be creating with have a key expiration time.

The shard manager will skip loading shards with an expired key, and by default shards that are more than 7 days expired are deleted.

### DIFF
--- a/mdb_shard/src/shard_format.rs
+++ b/mdb_shard/src/shard_format.rs
@@ -11,7 +11,6 @@ use std::io::{copy, Read, Seek, SeekFrom, Write};
 use std::mem::size_of;
 use std::ops::Add;
 use std::sync::Arc;
-use std::time::Duration;
 use std::time::UNIX_EPOCH;
 use tracing::debug;
 use utils::serialization_utils::*;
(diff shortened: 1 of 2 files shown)


mdb_shard/src/shard_format.rs
```diff
@@ -89,8 +100,15 @@ pub struct MDBShardFileFooter {
...
+    // The creation time of this shard, in seconds since the epoc
+    pub shard_creation_timestamp: u64,
+
+    // The time, in seconds since the epoc, after which this shard is no longer assumed to be valid.
+    // Locally created shards do not have an expiry.
+    pub shard_key_expiry: u64,
+
     // More locations to stick in here if needed.
-    _buffer: [u64; 7],
+    _buffer: [u64; 6],
```
seanses, line 111: I guess we don't have to shrink _buffer size, there's no need to maintain backward compatibility now.

PR #37, opened 2024-10-08
chunk cache impl 1

assafvayner:
Adds documented and tested implementation of the chunk cache aligned with design: https://www.notion.so/huggingface2/Client-Data-Cache-10b1384ebcac80b59516d5ae258a17a7

### DIFF
--- a/chunk_cache/src/disk.rs
+++ b/chunk_cache/src/disk.rs
@@ -26,7 +26,7 @@ pub mod test_utils;
 
 // consistently use URL_SAFE (also file path safe) base64 codec
 pub(crate) const BASE64_ENGINE: GeneralPurpose = URL_SAFE;
-pub const _DEFAULT_CAPACITY: u64 = 1 << 30; // 1 GB
+pub const DEFAULT_CAPACITY: u64 = 1 << 30; // 1 GB
 
 type CacheState = HashMap<Key, SortedVec<CacheItem>>;
 
@@ -147,6 +147,10 @@ impl DiskCache {
                 if !md.is_file() {
                     continue;
                 }
+                if md.len() > DEFAULT_CAPACITY {
+                    return Err(ChunkCacheError::general(format!("Cache directory contains a file larger than {} GB, cache directory state is invalid", (DEFAULT_CAPACITY as f64 / (1 << 30) as f64))));
+                }
+
                 // don't track an item that takes up the whole capacity
                 if md.len() > capacity {
                     continue;
@@ -577,9 +581,9 @@ impl ChunkCache for DiskCache {
 
 #[cfg(test)]
 mod tests {
-    use std::collections::BTreeSet;
+    use std::{collections::BTreeSet, fs::OpenOptions};
 
-    use crate::disk::{parse_key, test_utils::*};
+    use crate::disk::{key_dir, parse_key, test_utils::*};
 
     use cas_types::{Key, Range};
     use rand::thread_rng;
@@ -587,13 +591,13 @@ mod tests {
 
     use crate::ChunkCache;
 
-    use super::{DiskCache, _DEFAULT_CAPACITY};
+    use super::{DiskCache, DEFAULT_CAPACITY};
 
     #[test]
     fn test_get_cache_empty() {
         let mut rng = thread_rng();
         let cache_root = TempDir::new("empty").unwrap();
-        let cache = DiskCache::initialize(cache_root.into_path(), _DEFAULT_CAPACITY).unwrap();
+        let cache = DiskCache::initialize(cache_root.into_path(), DEFAULT_CAPACITY).unwrap();
         assert!(cache
             .get(&random_key(&mut rng), &random_range(&mut rng))
             .unwrap()
@@ -605,7 +609,7 @@ mod tests {
         let mut rng = thread_rng();
         let cache_root = TempDir::new("put_get_simple").unwrap();
         let cache =
-            DiskCache::initialize(cache_root.path().to_path_buf(), _DEFAULT_CAPACITY).unwrap();
+            DiskCache::initialize(cache_root.path().to_path_buf(), DEFAULT_CAPACITY).unwrap();
 
         let key = random_key(&mut rng);
         let range = Range { start: 0, end: 4 };
@@ -631,7 +635,7 @@ mod tests {
         let mut rng = thread_rng();
         let cache_root = TempDir::new("put_get_subrange").unwrap();
         let cache =
-            DiskCache::initialize(cache_root.path().to_path_buf(), _DEFAULT_CAPACITY).unwrap();
+            DiskCache::initialize(cache_root.path().to_path_buf(), DEFAULT_CAPACITY).unwrap();
 
         let key = random_key(&mut rng);
         let range = Range { start: 0, end: 4 };
@@ -685,7 +689,7 @@ mod tests {
     fn test_same_puts_noop() {
         let cache_root = TempDir::new("same_puts_noop").unwrap();
         let cache =
-            DiskCache::initialize(cache_root.path().to_path_buf(), _DEFAULT_CAPACITY).unwrap();
+            DiskCache::initialize(cache_root.path().to_path_buf(), DEFAULT_CAPACITY).unwrap();
         let mut it = RandomEntryIterator::default().with_range_len(1000);
         let (key, range, offsets, data) = it.next().unwrap();
         assert!(cache.put(&key, &range, &offsets, &data).is_ok());
@@ -698,7 +702,7 @@ mod tests {
             let mut it = RandomEntryIterator::default();
             let cache_root = TempDir::new("overlap_range_data_mismatch_fail").unwrap();
             let cache =
-                DiskCache::initialize(cache_root.path().to_path_buf(), _DEFAULT_CAPACITY).unwrap();
+                DiskCache::initialize(cache_root.path().to_path_buf(), DEFAULT_CAPACITY).unwrap();
             let (key, range, offsets, data) = it.next().unwrap();
             assert!(cache.put(&key, &range, &offsets, &data).is_ok());
             (cache_root, cache, key, range, offsets, data)
@@ -744,7 +748,7 @@ mod tests {
     fn test_initialize_non_empty() {
         let cache_root = TempDir::new("initialize_non_empty").unwrap();
         let cache =
-            DiskCache::initialize(cache_root.path().to_path_buf(), _DEFAULT_CAPACITY).unwrap();
+            DiskCache::initialize(cache_root.path().to_path_buf(), DEFAULT_CAPACITY).unwrap();
         let mut it = RandomEntryIterator::default();
 
         let mut keys_and_ranges = Vec::new();
@@ -755,7 +759,7 @@ mod tests {
             keys_and_ranges.push((key, range));
         }
         let cache2 =
-            DiskCache::initialize(cache_root.path().to_path_buf(), _DEFAULT_CAPACITY).unwrap();
+            DiskCache::initialize(cache_root.path().to_path_buf(), DEFAULT_CAPACITY).unwrap();
         for (i, (key, range)) in keys_and_ranges.iter().enumerate() {
             let get_result = cache2.get(&key, &range);
             assert!(get_result.is_ok(), "{i} {get_result:?}");
@@ -784,7 +788,7 @@ mod tests {
         const LARGE_FILE: u64 = 1000;
         let cache_root = TempDir::new("initialize_too_large_file").unwrap();
         let cache =
-            DiskCache::initialize(cache_root.path().to_path_buf(), _DEFAULT_CAPACITY).unwrap();
+            DiskCache::initialize(cache_root.path().to_path_buf(), DEFAULT_CAPACITY).unwrap();
         let mut it = RandomEntryIterator::default().with_range_len(LARGE_FILE as u32);
 
         let (key, range, offsets, data) = it.next().unwrap();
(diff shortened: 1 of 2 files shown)


chunk_cache/src/disk/cache_item.rs
```diff
@@ -0,0 +1,114 @@
...
+// sorted CacheItems using the range field to match a range for search
+// see logic
+impl Ord for CacheItem {
+    fn cmp(&self, other: &Self) -> Ordering {
+        match self.range.cmp(&other.range) {
+            Ordering::Less => return Ordering::Less,
+            Ordering::Equal => {}
+            Ordering::Greater => return Ordering::Greater,
+        }
+        match self.len.cmp(&other.len) {
```
port8080: When does the case occur where the range is equal and the lengths are different? Pls add a comment to clarify and maybe a pointer to the test case if there exists one

chunk_cache/src/disk.rs
```diff
@@ -0,0 +1,718 @@
...
+                return (key.clone(), random_item - count);
+            }
+            count += items.len();
+        }
+
+        panic!("should have returned")
+    }
+}
+
+fn total_bytes(state: &CacheState) -> u64 {
```
port8080, line 526: Can total_bytes and num_items be memoized? i.e. kept as fields inside CacheState and updated in the `initialize` and `put_impl` calls. Maybe take as a optimization todo.

chunk_cache/src/disk.rs
```diff
@@ -0,0 +1,718 @@
...
+            capacity,
+        })
+    }
+
+    fn get_impl(&self, key: &Key, range: &Range) -> Result<Option<Vec<u8>>, ChunkCacheError> {
+        if range.start >= range.end {
+            return Err(ChunkCacheError::InvalidArguments);
+        }
+
+        let mut state = self.state.lock()?;
```
port8080: Blocker: Are you holding this lock across a file read from disk? I would think that a get should be able to get the memory metadata it needs (e.g. items.clone()) and NOT hold the lock for the entire duration for all the keys

port8080: confirmed with AV - a lock being held during the get_impl across the file read from disk and I cannot put key2/range0-1 while key1/range0-1000 is being read from disk

this is too coarse a locking mechanism - let us come up with something better

chunk_cache/src/disk.rs
```diff
@@ -0,0 +1,718 @@
...
+    }
+
+    #[allow(dead_code)]
+    fn total_bytes(&self) -> Result<u64, ChunkCacheError> {
+        let state = self.state.lock()?;
+        let value = total_bytes(&state);
+        Ok(value)
+    }
+
+    pub fn initialize(cache_root: PathBuf, capacity: u64) -> Result<Self, ChunkCacheError> {
```
port8080, line 64: Add a sane large limit on the number of entries, and break when you reach that limit - estimate based on a cache size of 100GB or something

chunk_cache/src/disk.rs
```diff
@@ -0,0 +1,718 @@
...
+            return Err(ChunkCacheError::InvalidArguments);
+        }
+
+        let mut state = self.state.lock()?;
+        let total_bytes = total_bytes(&state);
+
+        // check if we already contain the range
+        if let Some(item_set) = state.get(key) {
+            if item_set.binary_search_by(range_contained_fn(range)).is_ok() {
+                return Ok(());
```
port8080, line 290: nit: Worth checking that the hash is identical in this case of a hit

chunk_cache/src/disk.rs
```diff
@@ -0,0 +1,718 @@
...
+            .update(data)
+            .finalize();
+
+        let item = CacheItem {
+            range: range.clone(),
+            len: (header_buf.len() + data.len()) as u64,
+            hash,
+        };
+
+        if self.capacity < total_bytes + item.len {
```
port8080: Do the eviction after the file has been written to the disk - you want to make sure that the file writes succeed

chunk_cache/src/disk.rs
```diff
@@ -0,0 +1,718 @@
...
+        fw.write_all(data)?;
+        fw.close()?;
+
+        item_set.insert(item);
+
+        Ok(())
+    }
+
+    /// removed items from the cache (including deleting from file system)
+    /// until at least to_remove number of bytes have been removed
```
port8080, line 396: comment or assert that self.state.lock is held by the caller before evict is called

chunk_cache/src/disk.rs
```diff
@@ -0,0 +1,718 @@
...
+    ) -> Result<(), ChunkCacheError> {
+        if range.start >= range.end
+            || chunk_byte_indicies.len() != (range.end - range.start + 1) as usize
+            // chunk_byte_indices is guarenteed to be more than 1 element at this point
+            || chunk_byte_indicies[0] != 0
+            || *chunk_byte_indicies.last().unwrap() as usize != data.len()
+        {
+            return Err(ChunkCacheError::InvalidArguments);
+        }
+
```
port8080, line 286: add a check that the size is at most a certain percentage (constant - 10%?) of the capacity

chunk_cache/src/disk.rs
```diff
@@ -0,0 +1,718 @@
...
+    use super::DiskCache;
+
+    #[test]
+    fn test_get_cache_empty() {
+        let cache_root = TempDir::new("empty").unwrap();
+        let cache = DiskCache::initialize(cache_root.into_path(), DEFAULT_CAPACITY).unwrap();
+        assert!(cache.get(&random_key(), &random_range()).unwrap().is_none());
+    }
+
```
port8080, line 606: test suggestion:  root directory with a file larger than DEFAUL_CAPACITY should throw an error during initialize

chunk_cache/src/disk.rs
```diff
@@ -0,0 +1,718 @@
...
+    #[test]
+    fn test_same_puts_noop() {
+        let cache_root = TempDir::new("puts_eviction").unwrap();
+        let cache =
+            DiskCache::initialize(cache_root.path().to_path_buf(), DEFAULT_CAPACITY).unwrap();
+        let (key, range, offsets, data) = RandomEntryIterator.next().unwrap();
+        assert!(cache.put(&key, &range, &offsets, &data).is_ok());
+
+        assert!(cache.put(&key, &range, &offsets, &data).is_ok());
```
port8080, line 696: test suggestion: related to a prev comment, error when the data content is different

chunk_cache/src/disk/test_utils.rs
```diff
@@ -0,0 +1,81 @@
...
+use rand::Rng;
+use sorted_vec::SortedVec;
+
+pub const DEFAULT_CAPACITY: u64 = 1 << 30;
+#[cfg(test)]
+pub const RANGE_LEN: u32 = 16 << 10;
+#[cfg(not(test))]
+pub const RANGE_LEN: u32 = 16 << 19;
+
+pub fn print_directory_contents(path: &Path) {
```
port8080, line 12: 👏 This is very useful, I would encourage you to pull this out as a helper app and not only keep this for tests.

chunk_cache/src/disk/cache_item.rs
```diff
@@ -0,0 +1,114 @@
...
+        write_u32(&mut w, self.range.start)?;
+        write_u32(&mut w, self.range.end)?;
+        write_u64(&mut w, self.len)?;
+        write_hash(&mut w, &self.hash)?;
+        Ok(BASE64_ENGINE.encode(buf))
+    }
+
+    pub(crate) fn parse(file_name: &[u8]) -> Result<CacheItem, ChunkCacheError> {
+        let buf = BASE64_ENGINE.decode(file_name)?;
+        let mut r = Cursor::new(buf);
```
port8080, line 69: Validate that buf is of the expected length CACHE_ITEM_FILE_NAME_BUF_SIZE

chunk_cache/src/disk.rs
assafvayner: setting this to be 2 * capacity, if we initialize to 100GB on a 1 GB capacity we'll be spending a lot of time evicting on the first put.

chunk_cache/src/disk.rs
assafvayner: If you're referring to the merklehash in the key, then state.get(key) returning a Some is an exact hit.

If you're referring to verifying the file exists and matches the in-memory state (i.e. not evicted by another process), that's not checked and can be added.

chunk_cache/src/bin/cache_resiliancy_test.rs
```diff
@@ -0,0 +1,145 @@
...
+};
+
+use cas_types::{Key, Range};
+use chunk_cache::{ChunkCache, DiskCache, RandomEntryIterator};
+use clap::{Args, Parser, Subcommand};
+use tempdir::TempDir;
+
+#[derive(Parser, Debug)]
+#[command(about)]
+struct ResiliancyTestArgs {
```
jgodlew: nit/TODO: it might make sense to have this as an integration test in a `tests` directory.

chunk_cache/src/disk.rs
```diff
@@ -0,0 +1,759 @@
...
+            };
+            let md = match key_dir.metadata() {
+                Ok(md) => md,
+                Err(e) => {
+                    if e.kind() == ErrorKind::NotFound {
+                        continue;
+                    }
+                    return Err(e.into());
+                }
+            };
```
jgodlew, line 101: nit: it seems that we are using this pattern in many places in these for-loops and could be a good candidate for a macro if you are inclined. For example, replacing this with something like:

`let md = continue_if_not_found!(key_dir.metadata())?;`

chunk_cache/src/disk.rs
```diff
@@ -0,0 +1,759 @@
...
+                    });
+                }
+                return Err(e.into());
+            }
+        };
+
+        for key_dir in readdir {
+            let key_dir = match key_dir {
+                Ok(kd) => kd,
+                Err(_) => continue,
```
jgodlew: should we follow the same pattern here as other io::Result checks (e.g. continue if NotFound, but error otherwise)?

chunk_cache/src/disk.rs
```diff
@@ -0,0 +1,759 @@
...
+                Ok(md) => md,
+                Err(e) => {
+                    if e.kind() == ErrorKind::NotFound {
+                        continue;
+                    }
+                    return Err(e.into());
+                }
+            };
+            if !md.is_dir() {
+                warn!(
```
jgodlew: TODO: before shipping, we will probably want this to be a debug log since the user probably doesn't want to see this log entry (as it isn't an actual error).

chunk_cache/src/disk.rs
```diff
@@ -0,0 +1,759 @@
...
+                        continue;
+                    }
+                    _ => return Err(e.into()),
+                },
+            };
+            let hash = blake3::Hasher::new().update_reader(&mut file)?.finalize();
+            if hash != item.hash {
+                error!("file hash mismatch on path: {path:?}, key: {key}, item: {item}");
+                items.remove_index(idx);
+                remove_file(&path)?;
```
jgodlew: Do we want to check for `items` being empty and removing it as well (from state and the fs)? 

Also, it seems like this "remove item+file" functionality would be good to extract into a helper method to use in here, put_impl, and evict.

chunk_cache/src/disk/cache_item.rs
```diff
@@ -0,0 +1,104 @@
...
+use cas_types::Range;
+use std::{
+    cmp::Ordering,
+    io::{Cursor, Read, Write},
+    mem::size_of,
+};
+use utils::serialization_utils::{read_u32, read_u64, write_u32, write_u64};
+
+const CACHE_ITEM_FILE_NAME_BUF_SIZE: usize =
+    size_of::<u32>() * 2 + size_of::<u64>() + blake3::OUT_LEN;
```
jgodlew, line 14: nit: probably want a unit test to verify that CacheItem::to_file_name() writes this many bytes.

chunk_cache/src/disk.rs
```diff
@@ -0,0 +1,759 @@
...
+        let mut bytes_removed = 0;
+        while to_remove > bytes_removed {
+            let (key, idx) = self.random_item(state);
+            let items = state.get_mut(&key).ok_or(ChunkCacheError::Infallible)?;
+            let item = &items[idx];
+            let len = item.len;
+            let path = self
+                .cache_root
+                .join(key_dir(&key))
+                .join(item.to_file_name()?);
```
jgodlew: nit: probably want to have a helper function for getting the path of a `(key, item)`.

chunk_cache/src/disk.rs
jgodlew: Isn't this already implied by state being wrapped by a `MutexGuard`? In order to get a MutexGuard, you need to have `locked` the resource since creating one (e.g. `MutexGuard::new()`) is not publicly accessible and is unsafe?

chunk_cache/src/disk.rs
assafvayner: oh, this raises a good point, I never actually check that the data is the same. I will add this check, but this will make void puts more expensive.

chunk_cache/src/disk.rs
assafvayner: I don't think this is a case to error, I think this is a case to ignore, i.e. don't load this file into the cache. I can think of a case where we intentionally specify a small cache < size_of(xorb) and a xorb sized file can exist and should probably remain in the cache unless we explicitly wipe the cache (rm -rf).

Added case and test, but waiting to get an ack to resolve.

assafvayner: corresponded with JG, agreeing to add this error case.

PR #39, opened 2024-10-09
Add file verification entry into shard

seanses:
Add FileVerificationEntry according to https://www.notion.so/huggingface2/File-metadata-and-dedup-10a1384ebcac80b98db7cfdba2a6fe83?pvs=4#ceaf182ec9454a63a4b6b56f4926782c. If the most significant bit of `FileDataSequenceHeader::file_flags` is set, there are equal number of `FileVerificationEntry`s following `FileDataSequenceEntry`s.

When cleaning a file completes, one keyed-hash is computed for each range of chunk hashes in a cas object for verification, with key being the cas object hash.

### DIFF
--- a/cas_object/src/cas_object_format.rs
+++ b/cas_object/src/cas_object_format.rs
@@ -680,13 +680,7 @@ mod tests {
         // Arrange
         let (c, _cas_data, _raw_data, _raw_chunk_boundaries) =
             build_cas_object(3, ChunkSize::Fixed(100), CompressionScheme::None);
-        let key: &[u8; 32] = c
-            .info
-            .cashash
-            .as_bytes()
-            .try_into()
-            .map_err(|_| "The slice is not 32 bytes long")
-            .unwrap();
+        let key: [u8; 32] = c.info.cashash.into();
 
         let hashes: Vec<u8> = c
             .info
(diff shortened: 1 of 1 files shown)


cas_object/src/cas_object_format.rs
```diff
@@ -685,7 +680,13 @@ mod tests {
         // Arrange
         let (c, _cas_data, _raw_data, _raw_chunk_boundaries) =
             build_cas_object(3, ChunkSize::Fixed(100), CompressionScheme::None);
-        let key: &[u8; 32] = c.info.cashash.as_bytes().try_into().map_err(|_| "The slice is not 32 bytes long").unwrap();
+        let key: &[u8; 32] = c
```
hoytak: Nit:  let key : [u8 ; 32] = c.info.cashash.into(); works without the unwrap.

mdb_shard/src/file_structs.rs
```diff
@@ -76,6 +88,20 @@ impl FileDataSequenceHeader {
...
         })
     }
+
+    pub fn contains_verification(&self) -> bool {
+        (self.file_flags & MDB_FILE_FLAG_VERIFICATION_MASK) != 0
+    }
+
+    /// Get the number of info entries following the header in this shard,
+    /// this includes "FileDataSequenceEntry"s and "FileVerificationEntry"s.
+    pub fn num_info_entry_following(&self) -> u32 {
```
hoytak, line 98: I'm not sure what we would do with this information, as the structs have different sizes, yes?

Edit: They are the same size.  Got it.

PR #40, opened 2024-10-10
Refactor CAS Client - Stage 1

rajatarya:
Happy to walk through these changes if that is easier.

### Goals:

1. Lighter-weight cas_client, no more layering of clients, with improved maintainability by only having one crate that talks to CAS.
1. Remove parallelism from within cas_client, have threads/parallel tasks be managed by the caller. This allows simplified maintenance/testing and performance optimizations.

### Stage 1: Refactor cas_client

* Remove CachingClient
* "Flatten" Clients and make the implementations 'free/static' methods that take (optional Cache, AuthMiddleware, HttpMiddleware)
* Have PointerFileTranslator drive cas_client (upload_async() calls PointerFileTranslator.start_clean() and .finalize_cleaning() and download_async() calls PointerFileTranslator.smudge_file_from_pointer

### DIFF
--- a/cas_client/src/remote_client.rs
+++ b/cas_client/src/remote_client.rs
@@ -12,7 +12,7 @@ use merklehash::MerkleHash;
 use reqwest::{StatusCode, Url};
 use reqwest_middleware::ClientWithMiddleware;
 use std::io::{Cursor, Write};
-use tracing::{debug, error, warn};
+use tracing::{debug, error};
 use utils::auth::AuthConfig;
 
 pub const CAS_ENDPOINT: &str = "http://localhost:8080";
@@ -20,29 +20,6 @@ pub const PREFIX_DEFAULT: &str = "default";
 
 const NUM_RETRIES: usize = 5;
 const BASE_RETRY_DELAY_MS: u64 = 3000;
-
-fn retry_http_status_code(stat: &reqwest::StatusCode) -> bool {
-    stat.is_server_error() || *stat == reqwest::StatusCode::TOO_MANY_REQUESTS
-}
-
-fn is_status_retryable_and_print(err: &reqwest::Error) -> bool {
-    let ret = err
-        .status()
-        .as_ref()
-        .map(retry_http_status_code)
-        .unwrap_or(true); // network issues should be retried
-    if ret {
-        warn!("{err:?}. Retrying...");
-    }
-    ret
-}
-
-fn is_middleware_status_retryable_and_print(err: &reqwest_middleware::Error) -> bool {
-    match err {
-        reqwest_middleware::Error::Reqwest(error) => is_status_retryable_and_print(error),
-        _ => false,
-    }
-}
 
 #[derive(Debug)]
 pub struct RemoteClient {


rajatarya:
Added logging and initial delay to retry policy. Will add some http mocking unit tests in `http_client.rs` to test out the retry & logging parts.

cas_client/README.md
```diff
@@ -1,2 +1,34 @@
...
+### Caching
+
+Caching happens locally using the cache crate, specifically using the ChunkCache. When RemoteClient is provided a ChunkCache then it
+will use this on download calls (the ReconstructionClient trait `get_file` and `get_file_byte_range`).
+
+## Overall CAS Communication Design
+
+### Authentication
+
+Authentication is done using AuthMiddleware, which sets an Authorization Header and refreshes it periodically with CAS. See [src/auth.rs].
```
seanses: `src/auth.rs` is deleted by this PR, let's update the reference

cas_client/src/interface.rs
```diff
@@ -29,24 +30,29 @@ pub trait UploadClient {
...
 
 /// A Client to the CAS (Content Addressed Storage) service to allow reconstructing a
 /// pointer file based on FileID (MerkleHash).
 #[async_trait]
 pub trait ReconstructionClient {
     /// Get a entire file by file hash.
-    async fn get_file(&self, hash: &MerkleHash, writer: &mut Box<dyn Write + Send>) -> Result<()>;
+    async fn get_file(
+        &self,
+        http_client: &ClientWithMiddleware,
```
assafvayner, line 55: Is there a reason why this http client isn't simply provided upon initialization for the RemoteClient and is instead provided through this trait function? If you have a good reason, can you add a comment, and otherwise consider removing the http_client as a parameter to the trait function.

cas_client/src/lib.rs
```diff
@@ -2,17 +2,16 @@
...
 pub use local_client::tests_utils;
 pub use local_client::LocalClient;
-pub use remote_client::build_reqwest_client;
+pub use http_client::build_auth_http_client;
+pub use http_client::build_http_client;
 pub use remote_client::RemoteClient;
 
-mod auth;
-mod caching_client;
+// mod auth;
```
assafvayner, line 12: rm comment

cas_client/src/interface.rs
```diff
@@ -29,24 +30,29 @@ pub trait UploadClient {
...
+        hash: &MerkleHash,
+        writer: &mut Box<dyn Write + Send>,
+    ) -> Result<()>;
 
     /// Get a entire file by file hash at a specific bytes range.
     async fn get_file_byte_range(
         &self,
+        http_client: &ClientWithMiddleware,
```
assafvayner, line 66: same as above.

cas_client/src/interface.rs
```diff
@@ -29,24 +30,29 @@ pub trait UploadClient {
...
         hash: &MerkleHash,
         offset: u64,
         length: u64,
```
assafvayner, line 69: is the length always known or needed here?

Is the information provided by byte_range essentially the same as offset and length?

cas_client/src/interface.rs
```diff
@@ -29,24 +30,29 @@ pub trait UploadClient {
...
+    ) -> Result<bool>;
 }
 
 /// A Client to the CAS (Content Addressed Storage) service to allow reconstructing a
 /// pointer file based on FileID (MerkleHash).
 #[async_trait]
 pub trait ReconstructionClient {
     /// Get a entire file by file hash.
-    async fn get_file(&self, hash: &MerkleHash, writer: &mut Box<dyn Write + Send>) -> Result<()>;
+    async fn get_file(
```
assafvayner, line 53: Is there a need to provide 2 functions (document the difference in the information the 2 stubs need?

Is get_file(..) the same as get_file_byte_range(.., offset: 0, length: length of file)?

If it is the case above, then it may also make sense to instead of ask the implementor to implement both to provide an implementation of get_file based on get_file_byte_range i.e. in the trait definition

```rust
    async fn get_file(...) -> Result<()> {
       self.get_file_byte_range(..., <default args>).await
    }
```

cas_client/src/remote_client.rs
```diff
@@ -191,24 +180,25 @@ impl RemoteClient {
...
+        let response = self.http_auth_client.post(url).body(data).send().await?;
         let response_body = response.bytes().await?;
         let response_parsed: UploadXorbResponse = serde_json::from_reader(response_body.reader())?;
 
         Ok(response_parsed.was_inserted)
     }
 
     async fn get_ranges(
-        &self,
+        http_client: &ClientWithMiddleware,
```
seanses, line 168: To confirm, the second goal below is not part of this PR?
> Remove parallelism from within cas_client, have threads/parallel tasks be managed by the caller. This allows simplified maintenance/testing and performance optimizations.

rajatarya: Yes, this PR is setting things up so that can be done in a subsequent PR. I am going to make it so cas_client isn't spawning threads. Right now it is spawning threads in the download path for both files and shards. The idea is cas_client should be a dumb client, not something that controls threads.

cas_client/src/remote_client.rs
```diff
@@ -191,24 +180,25 @@ impl RemoteClient {
...
         reconstruction_response: QueryReconstructionResponse,
         _byte_range: Option<(u64, u64)>,
         writer: &mut Box<dyn Write + Send>,
     ) -> Result<usize> {
         let info = reconstruction_response.reconstruction;
         let total_len = info.iter().fold(0, |acc, x| acc + x.unpacked_length);
-        let futs = info
-            .into_iter()
-            .map(|term| tokio::spawn(async move { get_one_range(&term).await }));
+        let futs = info.into_iter().map(|term| {
```
seanses, line 175: Also note that this is not parallelism: `map()` is lazy and is driven by `next()` (e.g. `for .. in ..` or `.collect()`).

cas_client/src/remote_client.rs
```diff
@@ -26,7 +25,7 @@ fn retry_http_status_code(stat: &reqwest::StatusCode) -> bool {
     stat.is_server_error() || *stat == reqwest::StatusCode::TOO_MANY_REQUESTS
 }
 
-fn is_status_retriable_and_print(err: &reqwest::Error) -> bool {
+fn is_status_retryable_and_print(err: &reqwest::Error) -> bool {
```
seanses: I think we can delete this function because it's internally handled by retry middleware.

cas_client/src/remote_client.rs
```diff
@@ -38,21 +37,28 @@ fn is_status_retriable_and_print(err: &reqwest::Error) -> bool {
     ret
 }
 
-fn is_middleware_status_retriable_and_print(err: &reqwest_middleware::Error) -> bool {
+fn is_middleware_status_retryable_and_print(err: &reqwest_middleware::Error) -> bool {
```
seanses: Also delete this

cas_client/src/interface.rs
rajatarya: Main thing is that this is a non-authenticated http_client. The authenticated one is stored as that is used for all the RemoteClient operations.

cas_client/src/interface.rs
seanses: I guess the main obstacle is that we don't have a file range query on the cas server yet, then `get_file_byte_range` would be a limp implementation that doesn't respect offset and length so we can keep them separated for now.

rajatarya:
On Friday I manually verified that Steel Thread works with these changes. I will make the other PR feedback and update the PR.

cas_client/src/interface.rs
rajatarya: This isn't implemented yet, when we implement it we should revisit if this is the right interface.

cas_client/src/interface.rs
rajatarya: 100% agree that when we implement `get_file_byte_range` then we can move the logic there and have `get_file` call into it.

cas_client/src/remote_client.rs
rajatarya: Do you mean this only spawns one thread at a time? I would have assumed that `.map` would iterate through the vector and spawn a thread for each entry.

cas_client/src/interface.rs
rajatarya: Added more comments around the interface here. I will add more in the Readme as well - but outside of this commit. I think it is important to put the design in the Readme for this crate - so will take a task to do that later this week.

PR #46, opened 2024-10-16
Refactor CAS Client - Stage 2

rajatarya:
* Collapsed shard_client into cas_client.
* ShardClientError -> CasClientError

Expected to merge AFTER #40.

See context below:

## Goals:

1. Lighter-weight cas_client, no more layering of clients, with improved maintainability by only having one crate that talks to CAS.
1. Remove parallelism from within cas_client, have threads/parallel tasks be managed by the caller. This allows simplified maintenance/testing and performance optimizations.

### Stage 2: Refactor xet-core - **THIS**
* Remove shard_client crate, migrate code into cas_client crate


cas_client/src/lib.rs
```diff
@@ -15,3 +15,41 @@ mod interface;
...
+
+use crate::error::Result;
+use async_trait::async_trait;
+pub use http_shard_client::HttpShardClient;
+pub use local_shard_client::LocalShardClient;
+use mdb_shard::shard_dedup_probe::ShardDedupProber;
+use mdb_shard::shard_file_reconstructor::FileReconstructor;
+use merklehash::MerkleHash;
+
+/// A Client to the Shard service. The shard service
```
seanses, line 31: We can move these trait definitions into interface.rs

rajatarya:
Manually tested and verified Steel Thread continues to work with these changes. Will address remaining feedback and update PR.

cas_client/src/lib.rs
rajatarya: Sounds good, moved.

rajatarya:
Since #40 going in as a squashed commit, moved this to https://github.com/huggingface-internal/xet-core/pull/53. Same changes, plus moving trait definitions to interface.rs.

PR #48, opened 2024-10-18
Use hardcoded key to generate chunk range hash

seanses:


data/src/clean.rs
```diff
@@ -581,7 +581,7 @@ impl Cleaner {
                         .iter()
                         .map(|(hash, _)| *hash)
                         .collect();
-                    let range_hash = range_hash_from_chunks(&chunk_hashes, &entry.cas_hash.into());
+                    let range_hash = range_hash_from_chunks(&chunk_hashes);
```
seanses, line 584: At this point when we want to generate chunk range hash we previously use the cas hash as the key. The bug is that the cas hash in this entry may not be valid until that cas block is "full" and flushed out. We could store the corresponding list of chunk hashes for entries with invalid cas hash into the cas_data_accumulator.pending_file_info, but that's a lot of information and there seems no harm to just use a hardcoded key.

seanses:
> My only concern/consideration is around key rotation / management. Is there a risk if this key is leaked/known?

I don't think there is. This key is meant to be public so any client can generate a chunk range hash for verification.

PR #49, opened 2024-10-18
Adding changes needed by DDB PR

jgodlew:
- Need to export the Key->hex serde module so we can use it with serde_dynamo crate in CAS. 
- PR comment to change name of function to get global_dedup chunk listing from a shard: https://github.com/huggingface-internal/xetcas/pull/75#discussion_r1805246722


port8080:
🚢

PR #50, opened 2024-10-18
reconstruction api refactor for frag

assafvayner:


cas_client/src/remote_client.rs
```diff
@@ -204,29 +206,61 @@ impl RemoteClient {
...
-                .await
-                .map_err(|e| CasClientError::InternalError(anyhow!("join error {e}")))??;
-            writer.write_all(&piece)?;
+        let terms = reconstruction_response.terms;
+        let fetch_info = reconstruction_response.fetch_info;
+        let total_len = terms.iter().fold(0, |acc, x| acc + x.unpacked_length);
+
+        let mut cache: HashMap<&CASReconstructionFetchInfo, (Vec<u32>, Vec<u8>)> = HashMap::new();
+        for term in terms.iter() {
+            let fhash = fetch_info
```
bpronan, line 215: Minor nit: I would prefer these were named `fetch_hash` and `fetch_term` to improve readability.

cas_object/src/cas_chunk_format.rs
```diff
@@ -165,6 +165,26 @@ pub fn deserialize_chunk_to_writer<R: Read, W: Write>(
     Ok(header.get_compressed_length() as usize + CAS_CHUNK_HEADER_LENGTH)
 }
 
+pub fn deserialize_chunks2<R: Read>(reader: &mut R) -> Result<(Vec<u32>, Vec<u8>), CasObjectError> {
```
bpronan, line 168: Do we need both functions (`deserialize_chunks` and `deserialize_chunks2`) around? If we are keeping both around, let's rename this one to something like `deserialize_indexed_chunks`.

cas_object/src/cas_chunk_format.rs
```diff
@@ -165,6 +165,26 @@ pub fn deserialize_chunk_to_writer<R: Read, W: Write>(
...
+    let mut buf = Vec::new();
+    let mut indices = vec![0];
+    loop {
```
bpronan, line 171: Not a deal breaker for this check-in, but I don't like that this is a effectively hidden a 2x nested "infinite" loop. Can we collapse both functions into a single loop? 

Non-conditional loops are troublesome enough on their own merit.

cas_object/src/cas_chunk_format.rs
assafvayner: don't need this, Rajat modified the correct one to provide this.

PR #51, opened 2024-10-21
WIP: Caching Client integration

rajatarya:
Just look at the last commit in this PR - that is the Caching client integration.

### DIFF
--- a/cas_client/src/remote_client.rs
+++ b/cas_client/src/remote_client.rs
@@ -9,11 +9,13 @@ use bytes::Bytes;
 use cas_object::CasObject;
 use cas_types::{CASReconstructionTerm, Key, QueryReconstructionResponse, UploadXorbResponse};
 use chunk_cache::{CacheConfig, ChunkCache, DiskCache};
+use error_printer::ErrorPrinter;
 use merklehash::MerkleHash;
 use reqwest::{StatusCode, Url};
 use reqwest_middleware::ClientWithMiddleware;
-use tracing::info;
 use std::io::{Cursor, Write};
+use std::sync::Arc;
+use tracing::info;
 use tracing::{debug, error, warn};
 use utils::auth::AuthConfig;
 
@@ -46,16 +48,18 @@ fn is_middleware_status_retryable_and_print(err: &reqwest_middleware::Error) -> 
     }
 }
 
-#[derive(Debug)]
 pub struct RemoteClient {
     endpoint: String,
     http_auth_client: ClientWithMiddleware,
-    disk_cache: Option<DiskCache>,
+    disk_cache: Option<Arc<dyn ChunkCache>>,
 }
 
 impl RemoteClient {
-    pub fn new(endpoint: &str, auth: &Option<AuthConfig>, cache_config: &Option<CacheConfig>) -> Self {
-
+    pub fn new(
+        endpoint: &str,
+        auth: &Option<AuthConfig>,
+        cache_config: &Option<CacheConfig>,
+    ) -> Self {
         // use disk cache if provided.
         let disk_cache = if let Some(cache) = cache_config {
             info!(
@@ -63,9 +67,15 @@ impl RemoteClient {
                 cache.cache_directory, cache.cache_size, cache.cache_blocksize
             );
 
-            Some(DiskCache::initialize(
-                cache.cache_directory.clone(),
-                cache.cache_size).unwrap())
+            if let Ok(cache) =
+                DiskCache::initialize(cache.cache_directory.clone(), cache.cache_size)
+                    .log_error("failed to initialize cache, not using cache")
+            {
+                let cacheptr = Arc::new(cache) as Arc<dyn ChunkCache>;
+                Some(cacheptr)
+            } else {
+                None
+            }
         } else {
             None
         };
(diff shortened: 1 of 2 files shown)


data/Cargo.toml
```diff
@@ -13,14 +13,13 @@ path = "src/bin/example.rs"
 [dependencies]
 cas_client = { path = "../cas_client" }
 cas_object = { path = "../cas_object" }
+chunk_cache = { path = "../chunk_cache" }
```
jgodlew, line 16: Do we need this dependency on chunk_cache?

data/src/test_utils/local_test_client.rs
```diff
@@ -17,7 +18,7 @@ pub struct LocalTestClient {
 
 impl LocalTestClient {
```
jgodlew, line 19: nit: if this is only for testing, does it make sense to tag the module with a `#[cfg(test)]` to ensure it doesn't get picked up by prod code?

cas_object/src/cas_object_format.rs
```diff
@@ -306,13 +306,13 @@ impl CasObject {
         for idx in 0..cas.info.num_chunks {
             // deserialize each chunk
             reader.seek(std::io::SeekFrom::Start(start_offset as u64))?;
-            let (data, compressed_chunk_length) = deserialize_chunk(reader)?;
-            let chunk_uncompressed_length = data.len();
+            let (data, compressed_chunk_length, chunk_uncompressed_length) = deserialize_chunk(reader)?;
+            // let chunk_uncompressed_length = data.len();
```
jgodlew, line 310: nit: we can probably remove this comment

cas_client/src/remote_client.rs
```diff
@@ -38,21 +41,53 @@ fn is_status_retriable_and_print(err: &reqwest::Error) -> bool {
...
+                    .log_error("failed to initialize cache, not using cache")
+            {
+                let cacheptr = Arc::new(cache) as Arc<dyn ChunkCache>;
+                Some(cacheptr)
+            } else {
+                None
+            }
+        } else {
+            None
+        };
```
jgodlew, line 81: nit: I'm not keen on the multiple if/else blocks here. It can be refactored to the following:
```suggestion
        let disk_cache = cache_config.as_ref()
            .inspect(|cache| info!(
                "Using disk cache directory: {:?}, size: {}, blocksize: {}.",
                cache.cache_directory, cache.cache_size, cache.cache_blocksize
            ))
            .and_then(|cache| {
                DiskCache::initialize(cache.cache_directory.clone(), cache.cache_size)
                    .log_error("failed to initialize cache, not using cache").ok()
            })
            .map(|disk_cache| Arc::new(disk_cache) as Arc<dyn ChunkCache>);
```

jgodlew: This could probably be further simplified with some other refactors (e.g. DiskCache::initialize takes an `&CacheConfig` instead of the raw values).

cas_client/src/remote_client.rs
```diff
@@ -191,24 +210,29 @@ impl RemoteClient {
...
         let response_body = response.bytes().await?;
         let response_parsed: UploadXorbResponse = serde_json::from_reader(response_body.reader())?;
 
         Ok(response_parsed.was_inserted)
     }
 
     async fn get_ranges(
-        &self,
+        http_client: &ClientWithMiddleware,
+        disk_cache: Option<Arc<dyn ChunkCache>>,
```
jgodlew, line 222: I am not sure about this change. Considering this is is a private function and is only called within the `RemoteClient`, having `self` supply the http_client and disk_cache would make more sense?

cas_client/src/interface.rs
```diff
@@ -29,24 +30,29 @@ pub trait UploadClient {
...
 
 /// A Client to the CAS (Content Addressed Storage) service to allow reconstructing a
 /// pointer file based on FileID (MerkleHash).
 #[async_trait]
 pub trait ReconstructionClient {
     /// Get a entire file by file hash.
-    async fn get_file(&self, hash: &MerkleHash, writer: &mut Box<dyn Write + Send>) -> Result<()>;
+    async fn get_file(
+        &self,
+        http_client: &ClientWithMiddleware,
```
jgodlew, line 47: Having the `ReconstructionClient` take in an http_client is awkward as it couples the trait to a specific communication protocol and requires the caller to know how to talk to the CAS server (e.g. the client needs the appropriate auth).

As a specific example, it seems weird that we store a `ClientWithMiddleware` in the RemoteClient, but don't end up using it for these reconstruction calls because the caller is providing a [possibly different] http client?

rajatarya: Yeah, I tried to address with more comments but this is confusing.

There are two http_client objects used. One is Authenticated and responsible for talking with CAS. The other is unauthenticated and responsible for talking with S3 directly.

Basically the ReconstructionClient is talking with S3 directly, so doesn't need an authenticated http_client.

This stems from there being two sets of network calls for `get_file`. 1) is to CAS to get the `Vec<ReconstructionTerm>` and then 2) is to S3 through presigned URLs returned in the `ReconstructionTerm`.

So, I put the overall CAS communication http_client (the authenticated one) into the RemoteClient as it is needed for all RemoteClient APIs. The unauthenticated http_client is only used with presigned URLs.

Previously, the `get_ranges()` and `get_one_range()` would directly create a new http_client and make requests using it. I wanted the client it uses to be passed in, so there was consistent retry logic & logging.

assafvayner:
should this be closed?

PR #52, opened 2024-10-21
Add prefix level directory to chunk cache disk impl

assafvayner:
fixes STO-115.

ChunkCache implementation will use additional level of subdirectories to match updated diagram [here](https://www.notion.so/huggingface2/Client-Data-Cache-10b1384ebcac80b59516d5ae258a17a7?pvs=4#1141384ebcac805ca341ca80c3c6e443).

This is done to control the number of subdirectories under the cache root.

As a result of the added complexity some refactoring is done especially in the initialize function and removal of directories.

### DIFF
--- a/chunk_cache/src/disk.rs
+++ b/chunk_cache/src/disk.rs
@@ -118,7 +118,7 @@ impl DiskCache {
         let mut num_items = 0;
         let max_num_bytes = 2 * capacity;
 
-        let readdir = match read_dir(&cache_root) {
+        let readdir = match read_dir(cache_root) {
             Ok(Some(rd)) => rd,
             Ok(None) => return Ok(CacheState::new(state, 0, 0)),
             Err(e) => return Err(e),


chunk_cache/src/disk.rs
```diff
@@ -53,13 +56,117 @@ pub struct DiskCache {
...
+    if md.len() > capacity {
+        return Ok(None);
+    }
+
+    let cache_item = match CacheItem::parse(item.file_name().as_encoded_bytes())
+        .debug_error("failed to decode a file name as a cache item")
+    {
+        Ok(i) => i,
+        Err(_) => {
+            remove_file(item.path())?;
```
rajatarya: I think there should be a `warn!` here, this is unexpected.

chunk_cache/src/disk.rs
```diff
@@ -53,13 +56,117 @@ pub struct DiskCache {
...
+            return Ok(None);
+        }
+    };
+    if md.len() != cache_item.len {
+        // file is invalid, remove it
+        remove_file(item.path())?;
```
rajatarya: Add a `warn!` here, as this is unexpected.

chunk_cache/src/disk.rs
```diff
@@ -53,13 +56,117 @@ pub struct DiskCache {
...
+            if e.kind() == ErrorKind::NotFound {
+                Ok(None)
+            } else {
+                Err(e.into())
+            }
+        }
+    }
+}
+
+// returns Ok(Some(_)) if result dirent is a directory, Ok(None) if was removed
```
rajatarya: The comment Ok(None) if it was removed doesn't seem accurate with code in Line 102. Instead, it seems like Ok(None) is returned if the DirEntry is actually not a dir.

I think the comment should probably be updated to reflect why Ok(None) is returned.

chunk_cache/src/disk.rs
```diff
@@ -77,126 +184,81 @@ impl DiskCache {
...
-            Err(e) => {
-                if e.kind() == ErrorKind::NotFound {
-                    return Ok(Self {
-                        cache_root,
-                        capacity,
-                        state: Arc::new(Mutex::new(CacheState::new(state, 0, 0))),
-                    });
-                }
-                return Err(e.into());
+        let readdir = match read_dir(&cache_root) {
```
rajatarya: I think `initialize` should have a comment that describes the layout, or maybe the implementation should mention the layout. Kinda gnarly to read this to see that there are two directory levels before chunk entries. Something like:

```
DiskCache organizes chunks on disk using the following directory layout. This is done to ensure a large local chunk cache (10GB) does not result in 40K directories off the root.

More information available here: [link to Notion doc section].

cache_root/
  xorb_prefix/
    xorb_hash/
       xorb_chunk_range
```

PR #54, opened 2024-10-21
Cas Client ChunkCache integration

rajatarya:
Integrating ChunkCache into cas_client crate.

With this PR, local disk caching of chunks should be used for downloading files.

FYI: @seanses I know you are out today/tomorrow so not putting you on this PR, but wanted you to know it exists.

### DIFF
--- a/cas_client/src/remote_client.rs
+++ b/cas_client/src/remote_client.rs
@@ -37,22 +37,15 @@ impl RemoteClient {
         auth: &Option<AuthConfig>,
         cache_config: &Option<CacheConfig>,
     ) -> Self {
-        // use disk cache if provided.
-        let disk_cache = if let Some(cache) = cache_config {
+        // use disk cache if cache_config provided.
+        let disk_cache = if let Some(cache_config) = cache_config {
             info!(
-                "Using disk cache directory: {:?}, size: {}, blocksize: {}.",
-                cache.cache_directory, cache.cache_size, cache.cache_blocksize
+                "Using disk cache directory: {:?}, size: {}.",
+                cache_config.cache_directory, cache_config.cache_size
             );
-
-            if let Ok(cache) =
-                DiskCache::initialize(cache.cache_directory.clone(), cache.cache_size)
-                    .log_error("failed to initialize cache, not using cache")
-            {
-                let cacheptr = Arc::new(cache) as Arc<dyn ChunkCache>;
-                Some(cacheptr)
-            } else {
-                None
-            }
+            DiskCache::initialize(cache_config)
+                .log_error("failed to initialize cache, not using cache")
+                .ok().map(|disk_cache| Arc::new(disk_cache) as Arc<dyn ChunkCache>)
         } else {
             None
         };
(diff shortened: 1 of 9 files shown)


data/Cargo.toml
```diff
@@ -13,6 +13,7 @@ path = "src/bin/example.rs"
 [dependencies]
 cas_client = { path = "../cas_client" }
 cas_object = { path = "../cas_object" }
+chunk_cache = { path = "../chunk_cache" }
```
jgodlew: chunk_cache shouldn't be needed by the data crate, right?

cas_client/src/remote_client.rs
```diff
@@ -21,17 +25,42 @@ pub const PREFIX_DEFAULT: &str = "default";
...
+                    .log_error("failed to initialize cache, not using cache")
+            {
+                let cacheptr = Arc::new(cache) as Arc<dyn ChunkCache>;
+                Some(cacheptr)
+            } else {
+                None
+            }
+        } else {
+            None
+        };
```
jgodlew: nit: I'm not keen on the multiple if/else blocks here. It can be refactored to the following:
```
        let disk_cache = cache_config.as_ref()
            .inspect(|cache| info!(
                "Using disk cache directory: {:?}, size: {}, blocksize: {}.",
                cache.cache_directory, cache.cache_size, cache.cache_blocksize
            ))
            .and_then(|cache| {
                DiskCache::initialize(cache.cache_directory.clone(), cache.cache_size)
                    .log_error("failed to initialize cache, not using cache").ok()
            })
            .map(|disk_cache| Arc::new(disk_cache) as Arc<dyn ChunkCache>);
```

jgodlew: This could probably be further simplified with some other refactors (e.g. `DiskCache::initialize` takes an `&CacheConfig` instead of the raw values).

cas_client/src/remote_client.rs
```diff
@@ -166,6 +196,7 @@ impl RemoteClient {
 
     async fn get_ranges(
         http_client: &ClientWithMiddleware,
+        disk_cache: Option<Arc<dyn ChunkCache>>,
```
jgodlew: Considering this is a private function and is only called within the RemoteClient, having this be a method on the RemoteClient and using `self` to supply the `http_client` and `disk_cache` would make more sense?

data/Cargo.toml
rajatarya: Good catch, removing!

cas_client/src/remote_client.rs
rajatarya: Woah, didn't know about this syntax at all.

I like further refactoring this to take a ChunkConfig, that seems like a good idea.

cas_client/src/remote_client.rs
rajatarya: I totally agree for `disk_cache`. Let me make that change. The `self.http_client` is authenticated, where as the one passed to `get_ranges` is not (since it is just making http get calls using presigned urls provided in the `Vec<CASReconstructionTerm>`.

PR #55, opened 2024-10-22
reconstruction api refactor for fragmentation

assafvayner:
For STO-82.

Refactors the query reconstruction api response to contain a separate aggregated data fetching information field and thins the reconstruction term contents.

CAS server changes are required and will be merged along with this PR.

### DIFF
--- a/cas_types/src/lib.rs
+++ b/cas_types/src/lib.rs
@@ -13,6 +13,7 @@ pub struct UploadXorbResponse {
     pub was_inserted: bool,
 }
 
+// note that the standard PartialOrd/Ord impls will first check `start` then `end`
 #[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq, PartialOrd, Ord, Default, Hash)]
 pub struct Range {
     pub start: u32,
@@ -26,6 +27,11 @@ impl std::fmt::Display for Range {
     }
 }
 
+/// Describes a portion of a reconstructed file, namely the xorb and
+/// a range of chunks within that xorb that are needed.
+///
+/// unpacked_length is used for validation, the result data of this term
+/// should have that field's value as its length
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct CASReconstructionTerm {
     pub hash: HexMerkleHash,
@@ -47,7 +53,7 @@ pub struct CASReconstructionFetchInfo {
     // chunk index start and end in a xorb
     pub range: Range,
     pub url: String,
-    // byte index start and end in a xorb
+    // byte index start and end in a xorb, used exclusively for Range header
     pub url_range: Range,
 }
 


cas_client/Cargo.toml
```diff
@@ -52,7 +51,7 @@ tokio-rustls = "0.25.0"
 rustls-pemfile = "2.0.0"
 hyper-rustls = { version = "0.26.0", features = ["http2"] }
 lz4 = "1.24.0"
-reqwest = "0.12.7"
+reqwest = { version = "0.12.7", features = ["json"] }
```
rajatarya, line 54: Why are the json features needed?

cas_client/src/remote_client.rs
```diff
@@ -143,14 +147,18 @@ impl Reconstructable for RemoteClient {
...
-
-        Ok(response_parsed)
+        let response = self
+            .http_auth_client
+            .get(url)
+            .send()
+            .await?
+            .error_for_status()?;
+
+        let query_resonstruction_response: QueryReconstructionResponse = response
```
rajatarya: typo: query_resonstruction_response -> query_re**c**onstruction_response

cas_client/src/remote_client.rs
```diff
@@ -143,14 +147,18 @@ impl Reconstructable for RemoteClient {
...
+            .json()
```
rajatarya, line 164: Is this changing the json parsing away from serde?

cas_client/src/remote_client.rs
```diff
@@ -180,96 +188,155 @@ impl RemoteClient {
...
+    singleflight::Group<(Vec<u8>, Vec<u32>), CasClientError>;
+
+pub(crate) async fn get_one_term(
+    http_client: Arc<ClientWithMiddleware>,
     disk_cache: Option<Arc<dyn ChunkCache>>,
-    term: &CASReconstructionTerm,
-) -> Result<Bytes> {
+    term: CASReconstructionTerm,
+    fetch_info: Arc<HashMap<HexMerkleHash, Vec<CASReconstructionFetchInfo>>>,
+    sfg: Arc<ChunkDataSingleFlightGroup>,
```
rajatarya: Please rename parameter to something more descriptive. sfg -> thundering_herd_group - something like that.

cas_client/src/remote_client.rs
```diff
@@ -180,96 +188,155 @@ impl RemoteClient {
...
 
-    // check disk cache
-    if let Some(cache) = &disk_cache {
-        let key = Key {
-            prefix: PREFIX_DEFAULT.to_string(),
-            hash: term.hash.into(),
-        };
+    let hash = term.hash.into();
+
+    let key = Key {
```
rajatarya: key isn't used later on so I think it should move it into the `if let Some(cache) = &disk_cache` block.

cas_client/src/remote_client.rs
```diff
@@ -180,96 +188,155 @@ impl RemoteClient {
...
+        .ok_or(CasClientError::InvalidArguments)
+        .log_error("invalid response from CAS server: failed to get term hash in fetchables")?;
+    let fetch_term = hash_fetch_info
+        .iter()
+        .find(|fterm| fterm.range.start <= term.range.start && fterm.range.end >= term.range.end)
+        .ok_or(CasClientError::InvalidArguments)
+        .log_error("invalid response from CAS server: failed to match hash in fetch_info")?;
+
+    // fetch the range from blob store and deserialize the chunks
+    let sfg_key = &fetch_term.url;
```
rajatarya: Does this include the query string params? Won't the presigned URLs be unique even if the same range is listed because of the AWS v4 signature?

cas_client/src/remote_client.rs
```diff
@@ -180,96 +188,155 @@ impl RemoteClient {
...
+async fn download_range(
+    fetch_term: CASReconstructionFetchInfo,
+    http_client: Arc<ClientWithMiddleware>,
+) -> Result<(Vec<u8>, Vec<u32>)> {
+    let url = Url::parse(fetch_term.url.as_str())?;
     let response = http_client
         .get(url)
-        .header(
-            reqwest::header::RANGE,
-            format!("bytes={}-{}", term.url_range.start, term.url_range.end - 1),
```
rajatarya, line 247: Wait, is `range.end` now inclusive of the last byte?

`fn range_header(range: &Range)` does not have the -1 on range.end that was present previously.

cas_client/Cargo.toml
rajatarya: Or, rather, is this why serde isn't needed directly in parsing json?

cas_client/src/remote_client.rs
```diff
@@ -180,96 +188,155 @@ impl RemoteClient {
...
+    let (mut data, chunk_byte_indices) = sfg
+        .work_dump_caller_info(sfg_key, download_range(fetch_term.clone(), http_client))
+        .await?;
+
+    // now write it to cache, the whole fetched term
+    if let Some(cache) = disk_cache {
+        cache.put(&key, &fetch_term.range, &chunk_byte_indices, &data)?;
+    }
+
+    // else case data matches exact, save some work
```
rajatarya: This comment seems incorrect - I believe this code is when the fetch_term.range is bigger than the term.range - meaning what was downloaded was more bytes than requested - so this code trims the fetched bytes to what was requested.

rajatarya:
After reading all the context in the Linear issue and the Slack thread - I think the code should have more comments reflecting the desire/intention of this work.

Something should clarify that the reconstruction flow now - how the fetch_info hashmap is used to limit the number of requests made for the same xorb - and I am guessing creates one larger request for a larger range than needed?

I would suggest the comment also result in some renaming of the variables. My undesrtanding is: Vec<CASReconstructionTerm> is the xorb chunks necessary for reconstructing the file. HashMap<HashHex, ...> is the collection of xorb chunk ranges collected from Vec<CASReconstructionTerm> and the intention of this HashMap is to make fewer requests to block store.

cas_client/Cargo.toml
assafvayner: It's the latter, reqwest calls serde_json

cas_client/src/remote_client.rs
assafvayner: It does. Suppose another term matches this fetch info, then they will find the same fetch_info -> same exact url.

cas_client/src/remote_client.rs
assafvayner: yes. The CAS server will provide the inclusive range for the url range. This is better since then if we move to a blob store that does range-end-exclusive semantics CAS can determine what to send back. see [cas pr](https://github.com/huggingface-internal/xetcas/pull/81/files#diff-8684d863cb133e1a285659129a125192caf94d71a053eb5f4d3fe8909b189ef0R148-R161)

cas_client/src/remote_client.rs
assafvayner: fetch_term is guarenteed to be "greater than" or equal (encomapssing) from the search for it
i.e.
```
        .find(|fterm| fterm.range.start <= term.range.start && fterm.range.end >= term.range.end)
```
 
 I will update the comment to indicate this situation, but at this point if fetch_term.range != term.range it must be the case that it is larger.

cas_client/src/remote_client.rs
assafvayner: no, still serde, just cleaner, reqwest calls into serde_json. without needing to return an owned buffer

PR #57, opened 2024-10-23
add TryFrom string trait to range

sirahd:


cas_types/src/lib.rs
```diff
@@ -27,6 +27,37 @@ impl std::fmt::Display for Range {
...
+pub enum RangeParseError {
+    InvalidFormat,
+    ParseError(ParseIntError),
+}
+
+impl TryFrom<String> for Range {
+    type Error = RangeParseError;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        let parts: Vec<&str> = value.split('-').collect();
```
assafvayner: do you mind using splitn instead of split here? just minor optimization

cas_types/src/lib.rs
```diff
@@ -27,6 +27,37 @@ impl std::fmt::Display for Range {
     }
 }
 
+#[derive(Debug)]
+pub enum RangeParseError {
+    InvalidFormat,
+    ParseError(ParseIntError),
+}
+
+impl TryFrom<String> for Range {
```
assafvayner: I think we should implement the TryFrom<&str> body then make TryFrom<String> call into that.
The reverse forces a string copy which isn't needed.

assafvayner: to_owned forces copying the data.

PR #63, opened 2024-10-24
Add File SHA to shard info

jgodlew:
* Adds extended metadata to the `MDBFileInfo` (`FileMetadataExt`). Currently, this contains a sha256 for the file (needed by CAS)
* Propagates the SHA256 from the caller (i.e. python -> hf_xet -> cleaner).
* Updates shard set operations to allow merging MDBFileInfo when there are disjoint flags (e.g. one has validation and the other has a sha).

### DIFF
--- a/data/src/clean.rs
+++ b/data/src/clean.rs
@@ -2,12 +2,16 @@ use std::collections::HashMap;
 use std::mem::take;
 use std::ops::DerefMut;
 use std::path::{Path, PathBuf};
-use std::sync::Arc;
 use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Arc;
 use std::time::Instant;
 
 use cas_object::range_hash_from_chunks;
+use error_printer::ErrorPrinter;
 use lazy_static::lazy_static;
+use mdb_shard::file_structs::{
+    FileDataSequenceEntry, FileDataSequenceHeader, FileMetadataExt, FileVerificationEntry, MDBFileInfo,
+};
 use mdb_shard::shard_file_reconstructor::FileReconstructor;
 use mdb_shard::{hash_is_global_dedup_eligible, ShardFileManager};
 use merkledb::aggregate_hashes::file_node_hash;
@@ -17,10 +21,6 @@ use tokio::sync::mpsc::error::TryRecvError;
 use tokio::sync::mpsc::{channel, Receiver, Sender};
 use tokio::sync::Mutex;
 use tokio::task::{JoinHandle, JoinSet};
-use error_printer::ErrorPrinter;
-use mdb_shard::file_structs::{
-    FileDataSequenceEntry, FileDataSequenceHeader, FileMetadataExt, FileVerificationEntry, MDBFileInfo,
-};
 use tracing::{debug, error, info, warn};
 
 use crate::cas_interface::Client;


data/src/clean.rs
```diff
@@ -95,13 +101,22 @@ impl Cleaner {
...
     ) -> Result<Arc<Self>> {
         let (data_p, data_c) = channel::<BufferItem<Vec<u8>>>(buffer_size);
 
         let (chunk_p, chunk_c) = channel::<Option<ChunkYieldType>>(buffer_size);
 
         let chunker = chunk_target_default(data_c, chunk_p);
 
+        // TODO: implement sha calculation in chunker
+        let sha256 = sha256
+            .or_else(rand_sha)
```
bpronan: Can we assume this is to keep existing code paths from breaking if they don't supply a SHA? If so, could you add a comment explaining the need here?

mdb_shard/src/file_structs.rs
```diff
@@ -6,11 +6,14 @@ use merklehash::MerkleHash;
 use utils::serialization_utils::*;
 
 use crate::cas_structs::{CASChunkSequenceEntry, CASChunkSequenceHeader};
+use crate::error::MDBShardError;
 use crate::shard_file::MDB_FILE_INFO_ENTRY_SIZE;
 
 pub const MDB_DEFAULT_FILE_FLAG: u32 = 0;
 pub const MDB_FILE_FLAG_WITH_VERIFICATION: u32 = 1 << 31;
 pub const MDB_FILE_FLAG_VERIFICATION_MASK: u32 = 1 << 31;
+pub const MDB_FILE_FLAG_WITH_METADATA_EXT: u32 = 1 << 30;
```
bpronan, line 15: Is there a logic to why it's 30? If so, it may make sense to leave some room (eg. ` << 24`) in case we want to associate some other flags categorically closer to the `VERIFICATION` flag.

mdb_shard/src/file_structs.rs
```diff
@@ -230,11 +295,55 @@ impl FileVerificationEntry {
...
+            sha256,
+            _unused: Default::default(),
+        }
+    }
+
+    pub fn serialize<W: Write>(&self, writer: &mut W) -> Result<usize, std::io::Error> {
+        let mut buf = [0u8; size_of::<Self>()];
+
+        {
+            let mut writer = Cursor::new(&mut buf[..]);
```
bpronan, line 307: Is there a deeper reason to the block? If not, I would prefer something like the following here:

```
let mut buf = ...
let mut buf_writer = Cursor::new(&mut buf[..]);
write_hash(&mut buf_writer, &self.sha256)?;
write_u64s(&mut buf_writer, &self._unused)?;

write.write_all(&buf)?;
```

mdb_shard/src/shard_file_manager.rs
```diff
@@ -608,24 +608,7 @@ mod tests {
...
-
-            let metadata = FileDataSequenceHeader::new(file_hash, *file_block_size, false);
-
-            let file_info = MDBFileInfo {
-                metadata,
-                segments,
-                verification: vec![],
-            };
-
+            let file_info = gen_random_file_info(&mut rng, file_block_size, false, false);
```
bpronan, line 610: Much cleaner!

mdb_shard/src/file_structs.rs
jgodlew: No specific reason to put it at the 30th bit, just because it was the "next" bit to use after the verification flag, making it easier in the future to find an empty index to add a new flag (as opposed to having some "holes" in the u32).

mdb_shard/src/file_structs.rs
jgodlew: The block has the benefit of "dropping" the Cursor when we are done using it. This is also the same style as the other serialize functions for file info components.

data/src/clean.rs
jgodlew: Yeah, I was originally doing this to make it easier to migrate (as we would just generate a random SHA). However, that could lead to issues in the future and have now changed it to return an error if the SHA isn't present (until I can implement the SHA generation code within the chunker, then it can be truly optional).

mdb_shard/src/file_structs.rs
```diff
@@ -6,11 +6,14 @@ use merklehash::MerkleHash;
...
 use crate::shard_file::MDB_FILE_INFO_ENTRY_SIZE;
 
 pub const MDB_DEFAULT_FILE_FLAG: u32 = 0;
 pub const MDB_FILE_FLAG_WITH_VERIFICATION: u32 = 1 << 31;
 pub const MDB_FILE_FLAG_VERIFICATION_MASK: u32 = 1 << 31;
+pub const MDB_FILE_FLAG_WITH_METADATA_EXT: u32 = 1 << 30;
+pub const MDB_FILE_FLAG_METADATA_EXT_MASK: u32 = 1 << 30;
 
 /// Each file consists of a FileDataSequenceHeader following
 /// a sequence of FileDataSequenceEntry and maybe a sequence
```
seanses: Let's also update this comment to contain "maybe FileMetadataExt"

mdb_shard/src/file_structs.rs
```diff
@@ -87,21 +98,75 @@ impl FileDataSequenceHeader {
...
         } else {
-            self.num_entries
+            self.num_entries + num_metadata_ext
+        }
+    }
+
+    /// Checks that the two headers correspond to the same file. Checks that
+    /// the file hashes are the same and that the number of entries are the
+    /// same.
+    pub fn check_same_file(header1: &Self, header2: &Self) -> Result<(), MDBShardError> {
```
seanses: Is this function for debug assertion? If so I'd prefer to use `debug_assert!` or just return bool instead of returning errors. It's a bit confusing why two different files lead to an error from the perspective of this function alone.

mdb_shard/src/file_structs.rs
```diff
@@ -230,11 +295,55 @@ impl FileVerificationEntry {
     }
 }
 
+/// Extended metadata about the file (e.g. sha256). Stored in the FileInfo when the
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct FileMetadataExt {
```
seanses, line 290: Let's also add a static assertion for the size of this struct here: https://github.com/huggingface-internal/xet-core/blob/155adcc79b1c132fd347bad69e116590735e5b98/mdb_shard/src/shard_format.rs#L22

hf_xet/src/data_client.rs
```diff
@@ -71,12 +71,15 @@ pub async fn download_async(
     Ok(paths)
 }
 
-async fn clean_file(processor: &PointerFileTranslator, f: String) -> errors::Result<PointerFile> {
+async fn clean_file(processor: &PointerFileTranslator, (f, sha): (String, String)) -> errors::Result<PointerFile> {
     let mut read_buf = vec![0u8; READ_BLOCK_SIZE];
     let path = PathBuf::from(f);
     let mut reader = BufReader::new(File::open(path.clone())?);
+    if sha != "" {
```
bpronan: This should be deleted.

mdb_shard/src/file_structs.rs
jgodlew: I originally had it as returning an error instead of bool since I wanted to surface why the two headers aren't the same (different Hash or num_entries). I've changed to a debug assertion instead.

PR #64, opened 2024-10-24
Make Range type support generic number type

sirahd:
In response to this review comment https://github.com/huggingface-internal/xetcas/pull/93#discussion_r1815697669, this PR makes `Range` support generic number type, and renames the existing `Range` -> `ChunkRange` and introduces `FileRange`

### DIFF
--- a/cas_client/src/remote_client.rs
+++ b/cas_client/src/remote_client.rs
@@ -6,8 +6,8 @@ use anyhow::anyhow;
 use async_trait::async_trait;
 use cas_object::CasObject;
 use cas_types::{
-    CASReconstructionFetchInfo, CASReconstructionTerm, ChunkRange, HexMerkleHash, HttpRange, Key,
-    QueryReconstructionResponse, UploadXorbResponse,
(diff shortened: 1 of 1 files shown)


cas_types/src/lib.rs
```diff
@@ -75,10 +78,10 @@ pub struct CASReconstructionTerm {
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
 pub struct CASReconstructionFetchInfo {
     // chunk index start and end in a xorb
-    pub range: Range,
+    pub range: ChunkRange,
     pub url: String,
     // byte index start and end in a xorb, used exclusively for Range header
-    pub url_range: Range,
+    pub url_range: ChunkRange,
```
assafvayner: It might make sense to make this it's own kind of Range as well or use a FileRange. I'm thinking when we pass around file ranges they are also end inclusive so it will be easier to keep the end-exclusivity consistent.

sirahd: Good point. I added a new `HttpRange` type that represents the inclusive-end range. `FileRange` is kept end exclusive since it's used throughout the codebase to calculate offset, etc.

PR #68, opened 2024-10-26
Fix dedup query

seanses:
- Fix dedup query response parse logic according to the server implementation. See https://github.com/huggingface-internal/xetcas/pull/99
- Fix config setting so that global dedup is actually enabled. [context](https://github.com/huggingface-internal/xet-core/blob/8364c9337ee7eda5faf4209693b8fb6afcf5368a/data/src/data_processing.rs#L112)
- Add clean metrics reporting for debugging (also useful for future telemetry setup). [Related PR](https://github.com/huggingface-internal/xet-core/pull/34)

Tested with hfxet in the below steps:
1. Upload a file, see no dedup
2. Remove local shards (`rm -rf .xet`)
3. Upload the same file, see 100% dedup due to global dedup

```
(.venv) di@di-mbp ~/tt/download % ipython
In [1]: from huggingface_hub import upload_folder, login

In [2]: login(token="...")
...

In [3]: upload_folder(folder_path="/Users/di/tt/bsf13/data/", repo_id="di/m1", repo_type="model")
...
{"timestamp":"2024-10-26T00:00:21.858541Z","level":"INFO","fields":{"message":"Cleaning file None finished in 0 s 217 ms, processed 9815362 bytes, produced 9815362 bytes after dedup."},"filename":"/Users/di/hf/xet-core/data/src/clean.rs","line_number":177}
...
Out[3]: CommitInfo(commit_url='http://localhost:5564/di/m1/commit/83f9a4fc1392e475006893cc8219270aec663820', commit_message='Upload folder using huggingface_hub', commit_description='', oid='83f9a4fc1392e475006893cc8219270aec663820', pr_url=None, repo_url=RepoUrl('http://localhost:5564/di/m1', endpoint='http://localhost:5564', repo_type='model', repo_id='di/m1'), pr_revision=None, pr_num=None)

In [4]: !rm -rf .xet

In [5]: !ls -la .xet
ls: .xet: No such file or directory

In [6]: upload_folder(folder_path="/Users/di/tt/bsf13/data/", repo_id="di/m1", repo_type="model")
...
{"timestamp":"2024-10-26T00:00:52.002688Z","level":"INFO","fields":{"message":"Cleaning file None finished in 0 s 217 ms, processed 9815362 bytes, produced 0 bytes after dedup."},"filename":"/Users/di/hf/xet-core/data/src/clean.rs","line_number":177}
...
Out[6]: CommitInfo(commit_url='http://localhost:5564/di/m1/commit/6042cd060241b5062dc86de2b259f47e9a14e7cb', commit_message='Upload folder using huggingface_hub', commit_description='', oid='6042cd060241b5062dc86de2b259f47e9a14e7cb', pr_url=None, repo_url=RepoUrl('http://localhost:5564/di/m1', endpoint='http://localhost:5564', repo_type='model', repo_id='di/m1'), pr_revision=None, pr_num=None)
```

### DIFF
--- a/cas_client/src/http_shard_client.rs
+++ b/cas_client/src/http_shard_client.rs
@@ -142,7 +142,7 @@ impl ShardDedupProber<CasClientError> for HttpShardClient {
             .get(url)
             .send()
             .await
-            .map_err(|e| CasClientError::Other(format!("request failed with code {e}")))?;
+            .map_err(|e| CasClientError::Other(format!("request failed with error {e}")))?;
 
         // Dedup shard not found, return empty result
         if !response.status().is_success() {
@@ -162,7 +162,7 @@ impl ShardDedupProber<CasClientError> for HttpShardClient {
         let file_name = local_shard_name(&shard_hash);
         let file_path = shard_cache_dir.join(file_name);
         let mut writer = hashed_writer.into_inner();
-        writer.dest_at(file_path);
+        writer.set_dest_path(file_path);
         writer.close()?;
 
         Ok(vec![shard_hash])
--- a/file_utils/src/safe_file_creator.rs
+++ b/file_utils/src/safe_file_creator.rs
@@ -21,7 +21,7 @@ impl SafeFileCreator {
     /// and a temporary file is created then renamed on close.
     pub fn new<P: AsRef<Path>>(dest_path: P) -> io::Result<Self> {
         let dest_path = dest_path.as_ref().to_path_buf();
-        let temp_path = Self::temp_file_path(Some(&dest_path));
+        let temp_path = Self::temp_file_path(Some(&dest_path))?;
 
         // This matches the permissions and ownership of the parent directory
         let file = create_file(&temp_path)?;
@@ -37,11 +37,11 @@ impl SafeFileCreator {
 
     /// Safely creates a new file while a destination name can't be decided now. Users need to call
     /// ```ignore
-    /// pub fn dest_at<P: AsRef<Path>>(dest_path: P)
+    /// pub fn set_dest_path<P: AsRef<Path>>(dest_path: P)
     /// ```
     /// to set the destination before closing the file.
     pub fn new_unnamed() -> io::Result<Self> {
-        let temp_path = Self::temp_file_path(None);
+        let temp_path = Self::temp_file_path(None)?;
 
         // This matches the permissions and ownership of the parent directory
         let file = create_file(&temp_path)?;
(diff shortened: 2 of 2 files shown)


file_utils/src/safe_file_creator.rs
```diff
@@ -44,19 +64,28 @@ impl SafeFileCreator {
...
+        let parent = path.parent().map(|p| p.to_owned()).unwrap_or_else(std::env::temp_dir);
+        let file_name = path.file_name().unwrap_or_default().to_str().unwrap_or_default();
 
         let mut rng = thread_rng();
         let random_hash: String = (0..10).map(|_| rng.sample(Alphanumeric)).map(char::from).collect();
         let temp_file_name = format!(".{}.{hash}.tmp", file_name, hash = random_hash);
         parent.join(temp_file_name)
     }
 
+    pub fn dest_at<P: AsRef<Path>>(&mut self, dest_path: P) {
```
port8080: nit: `dest_at` -> `set_dest_path`

cas_client/src/http_shard_client.rs
```diff
@@ -141,59 +137,35 @@ impl ShardDedupProber<CasClientError> for HttpShardClient {
 
         let url = Url::parse(&format!("{0}/chunk/{key}", self.endpoint))?;
 
-        let response = self
+        let mut response = self
             .client
             .get(url)
             .send()
             .await
             .map_err(|e| CasClientError::Other(format!("request failed with code {e}")))?;
```
assafvayner: while you're here, can you fix this log, this log will fail if there was a network error or failure to send request/receive response. We need to add
```
.error_for_status().map_err(|e| CasClientError::Other(<status code error>)
```
for errors from the API

assafvayner: I see you handle the bad error code below, so just change this map err, rm `with code`

cas_client/src/http_shard_client.rs
```diff
@@ -141,59 +137,35 @@ impl ShardDedupProber<CasClientError> for HttpShardClient {
...
+
+        let shard_hash = hashed_writer.hash();
+        let file_name = local_shard_name(&shard_hash);
+        let file_path = shard_cache_dir.join(file_name);
+        let mut writer = hashed_writer.into_inner();
+        writer.dest_at(file_path);
+        writer.close()?;
 
-        Ok(downloaded_shards)
+        Ok(vec![shard_hash])
```
assafvayner, line 168: should this function just return 1 shard hash instead of a vec?

file_utils/src/safe_file_creator.rs
```diff
@@ -44,19 +64,28 @@ impl SafeFileCreator {
     }
 
     /// Generates a temporary file path in the same directory as the destination file
-    pub fn temp_file_path<P: AsRef<Path>>(dest_path: P) -> PathBuf {
-        let path = dest_path.as_ref();
-        let parent = path.parent().unwrap_or_else(|| Path::new("."));
-        let file_name = path.file_name().unwrap().to_str().unwrap();
+    fn temp_file_path(dest_path: Option<&Path>) -> PathBuf {
+        let path = dest_path.map(|p| p.to_owned()).unwrap_or_default();
+        let parent = path.parent().map(|p| p.to_owned()).unwrap_or_else(std::env::temp_dir);
```
assafvayner: this seems to me like an infallible error rather than a temp_dir, but if you think it's fine then it's fine

seanses: You are right, will change to return Err

cas_client/src/http_shard_client.rs
seanses: It's reserved for batch request: the function takes a list of chunk hashes and return a list of shards, though the batch part is not implemented yet.

PR #71, opened 2024-10-29
Shard format update

hoytak:
This PR switches the order of the sections in the shard format so all the lookup tables are at the end.  This allows us to only read the first part of the shard server-side to get all the information needed, without needing the footer to know where the sections are.  

To make this more robust, I also introduced explicit functions to call to get the byte offset ranges of each of the sections instead of having numerous places that depend on the end of a section being the start of the next.  This makes the code more readable and robust.  

Because this is a big change, it also updates the magic tag for the format to be `b'HFMetadataBeta'` plus some magic numbers.  This will cause failures to happen earlier rather than later.

### DIFF
--- a/mdb_shard/src/shard_format.rs
+++ b/mdb_shard/src/shard_format.rs
@@ -571,8 +571,12 @@ impl MDBShardInfo {
 
         reader.seek(SeekFrom::Start(self.metadata.cas_info_offset))?;
 
-        while let Some(cas_block) = CASChunkSequenceHeader::deserialize(reader)? {
-            let pos = reader.stream_position()? - size_of::<CASChunkSequenceHeader>();
+        loop {
+            let pos = reader.stream_position()?;
+            let cas_block = CASChunkSequenceHeader::deserialize(reader)?;
+            if cas_block.is_bookend() {
+                break;
+            }
             let n = cas_block.num_entries;
             cas_blocks.push((cas_block, pos));
 
@@ -756,8 +760,8 @@ impl MDBShardInfo {
 
             let mut cas_index = 0;
 
-            reader.seek(SeekFrom::Start(cas_lookup_start))?;
-            while reader.stream_position()? < cas_lookup_end {
+            reader.seek(SeekFrom::Start(cas_info_start))?;
+            while reader.stream_position()? < cas_info_end {
                 let cas_header = CASChunkSequenceHeader::deserialize(reader)?;
 
                 for chunk_index in 0..cas_header.num_entries {
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,4 @@
 {
-  "rust-analyzer.rustfmt.extraArgs": ["+nightly"],
   "[rust]": {
     "editor.defaultFormatter": "rust-lang.rust-analyzer"
   }


mdb_shard/src/shard_format.rs
```diff
@@ -101,14 +104,19 @@ impl MDBShardFileHeader {
 pub struct MDBShardFileFooter {
     pub version: u64,
     pub file_info_offset: u64,
+    pub cas_info_offset: u64,
+
+    // Lookup tabels.  These come after the info sections to allow the shard to be partially
+    // read without including the header.
```
seanses: without including the footer?

mdb_shard/src/shard_format.rs
```diff
@@ -309,6 +317,12 @@ impl MDBShardInfo {
             Self::convert_and_save_file_info(writer, &mdb.file_content)?;
         bytes_pos += bytes_written;
```
seanses, line 312: Let's also update this comment to reflect the new order: https://github.com/huggingface-internal/xet-core/blob/64d71312c7c0e81857f4891d6d4e8255a8648606/mdb_shard/src/shard_format.rs#L226

mdb_shard/src/shard_format.rs
jgodlew: Also typo: tabels -> tables

mdb_shard/src/shard_format.rs
```diff
@@ -747,20 +755,19 @@ impl MDBShardInfo {
...
-                / size_of::<CASChunkSequenceEntry>();
+            let n_elements_cap = (cas_info_end - cas_info_start) as usize / size_of::<CASChunkSequenceEntry>();
 
             ret = Vec::with_capacity(n_elements_cap);
 
             let mut cas_index = 0;
 
-            reader.seek(SeekFrom::Start(self.metadata.cas_info_offset))?;
-            while reader.stream_position()? < self.metadata.cas_lookup_offset {
+            let (cas_lookup_start, cas_lookup_end) = self.cas_info_byte_range();
```
seanses: This is identical to the above `let (cas_info_start, cas_info_end) = self.cas_info_byte_range();` so we can reuse those values? The variable names are also a bit misleading

mdb_shard/src/shard_format.rs
```diff
@@ -989,13 +1036,31 @@ impl MDBShardInfo {
...
+        debug_assert_eq!(reader.stream_position()?, cas_lookup_start);
+
         out_footer.cas_lookup_offset = byte_pos as u64;
 
         if include_cas_lookup_table {
             // The cas lookup table should be the same, so just copy it directly
-            byte_pos +=
-                copy(&mut reader.take(self.metadata.chunk_lookup_offset - self.metadata.cas_lookup_offset), writer)?
-                    as usize;
+            byte_pos += copy(&mut reader.take(cas_lookup_end - cas_info_start), writer)? as usize;
```
seanses: `cas_lookup_end - cas_lookup_start`?

mdb_shard/src/shard_format.rs
hoytak: Good catch.

mdb_shard/src/shard_format.rs
hoytak: Yes, good catch.

mdb_shard/src/shard_format.rs
hoytak: Great catch.  Will look at why this passed tests... :-/...

PR #72, opened 2024-10-29
Generate sha instead of supplying

jgodlew:
Fixes: STO-124

* Adds a sha256 hasher to the cleaner to generate a sha256 while a file is being cleaned, storing that sha256 in the shard.
* Removes the ability for clients (i.e. python) to supply a sha256.

### DIFF
--- a/data/src/clean.rs
+++ b/data/src/clean.rs
@@ -184,8 +184,8 @@ impl Cleaner {
 
         self.metrics.file_size.fetch_add(data.len() as u64, Ordering::Relaxed);
 
+        self.sha_generator.update(&data)?;
         if !self.check_passthrough_status(&data).await? {
-            self.sha_generator.update(&data)?;
             self.add_data_to_chunking(BufferItem::Value(data)).await?
         }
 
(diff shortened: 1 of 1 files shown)


data/src/clean.rs
```diff
@@ -167,6 +185,7 @@ impl Cleaner {
         self.metrics.file_size.fetch_add(data.len() as u64, Ordering::Relaxed);
 
         if !self.check_passthrough_status(&data).await? {
+            self.sha_generator.update(&data)?;
```
seanses: Looks like there's a bug: if `check_passthrough_status` returns true (meaning data is already processed inside), and buffer eventually becomes larger than the small file threshold so those bytes are sent to chunking, those data are not updated to the sha generator. The `update` inside the `fn finish` only accounts for the case that the buffer size didn't get pass the small file threshold.

So instead of inserting sha update to multiple places, I recommend put it only once before this `if` block so we don't have to deal with internal buffering logic (like how we track total file size above).

jgodlew: Good catch! I was trying to only calculate the sha when we actually do chunking (since passthrough files don't need us to calculate the sha256), but I see that I missed a place with this buffer inside the `check_passthrough_status` call. 

I've changed the code to just update the sha regardless of passthrough status as you've mentioned.

PR #73, opened 2024-10-29
Support range get in PointerFileTranslator

sirahd:
Related PR: https://github.com/huggingface-internal/xetcas/pull/93

### DIFF
--- a/cas_client/src/remote_client.rs
+++ b/cas_client/src/remote_client.rs
@@ -373,11 +373,7 @@ async fn download_range(
 
 #[cfg(test)]
 mod tests {
-    use std::fs::File;
     use std::io::Read;
-    use std::ops::Deref;
-    use std::ptr::write;
-    use std::rc::Rc;
     use std::sync::Mutex;
 
     use cas_object::test_utils::{build_cas_object, ChunkSize};


cas_client/src/remote_client.rs
```diff
@@ -202,32 +197,43 @@ impl RemoteClient {
...
         let fetch_info = Arc::new(reconstruction_response.fetch_info);
 
         let single_flight_group = Arc::new(singleflight::Group::new());
 
-        let total_len = terms.iter().fold(0, |acc, x| acc + x.unpacked_length);
+        let total_len = byte_range
+            .as_ref()
+            .map(|range| range.end - range.start)
+            // use the full reconstruction terms if fetching full range
+            .unwrap_or(terms.iter().fold(0, |acc, x| acc + x.unpacked_length as u64));
```
assafvayner: style nit, since the unwrap_or op is relatively complex and heavy I think it would be easier to read with:
```
let total_len = if let Some(range) = &byte_range {
    range.end - range.start
} else {
    terms.iter().fold(0, |acc, x| acc + x.unpacked_length as u64)
}
```

cas_client/src/remote_client.rs
```diff
@@ -202,32 +197,43 @@ impl RemoteClient {
...
-            writer.write_all(&term_data)?;
+            let start = if i == 0 {
+                // use offset_into_first_range for the first term if needed
+                max(0, reconstruction_response.offset_into_first_range as usize)
+            } else {
+                0
+            };
+            let end: usize = min(remaining_len, term_data.len() as u64) as usize + start;
+            writer.write_all(&term_data[start..end])?;
+            remaining_len -= (end - start) as u64;
```
assafvayner, line 232: this works, for readability I'd maybe be more happy with just handling the start and end case (at the start and end) and then all the middle "the easy way".

e.g. for the start you can handle the first case then loop over `term_data_futures.iter().skip(1)` and for the last case if your calculation for the `end` variable is not `term_data.len()` then do some kind of splicing.

cas_client/src/remote_client.rs
```diff
@@ -202,32 +197,43 @@ impl RemoteClient {
...
                 .map_err(|e| CasClientError::InternalError(anyhow!("join error {e}")))?
                 .log_error("error getting one term")?;
-            writer.write_all(&term_data)?;
+            let start = if i == 0 {
+                // use offset_into_first_range for the first term if needed
+                max(0, reconstruction_response.offset_into_first_range as usize)
+            } else {
+                0
+            };
+            let end: usize = min(remaining_len, term_data.len() as u64) as usize + start;
```
assafvayner: suppose start != 0 (first range) and remaining_len is high, `end` becomes `term_data.len() + start` so when you slice later `start..end` goes past the buffer. Correct me if I misunderstand, but it's a minor bug. I think adding more comments on these calculations will help.

data/src/test_utils/local_test_client.rs
```diff
@@ -76,24 +78,14 @@ impl ReconstructionClient for LocalTestClient {
             else {
                 return Err(CasClientError::InvalidRange);
             };
+            let start = byte_range.as_ref().map(|range| range.start as usize).unwrap_or(0);
+            let end = byte_range.as_ref().map(|range| range.end as usize).unwrap_or(one_range.len());
 
-            writer.write_all(&one_range)?;
+            writer.write_all(&one_range[start..end])?;
```
assafvayner, line 84: stylewise optionally I like `if let Some(range) = byte_range{ ... } else { ... }`. Though I don't care as much about this case.

hf_xet/src/data_client.rs
```diff
@@ -4,6 +4,7 @@ use std::io::{BufReader, Read, Write};
 use std::path::PathBuf;
 use std::sync::Arc;
 
+use cas_types::FileRange;
```
assafvayner: does this file use it?

cas_client/src/remote_client.rs
sirahd: Good catch. I mistakenly added the `+ start` at the end after running into a bug while testing, while it should correctly be added as `remaining_len + start`.

hf_xet/src/data_client.rs
jgodlew: Make sure to remove the cas_types from the hf_xet Cargo.toml and lock files as well.

cas_client/src/remote_client.rs
sirahd: I'm not sure if that would make it any more readable than the current form. We still need to keep track of `remaining_len` for the term. The `min`, `max` function to correctly determine the offset is still needed. The only difference would be the initial offset calculation is pulled out of a loop into the first case.

PR #74, opened 2024-10-31
Explicit threadpool construction and usage

rajatarya:
hf_xet now creates a single tokio::runtime multi-threaded threadpool and it lives through the life of the python process.

This threadpool is passed into data crate, in PointerFileTranslator, and then passed into cas_client::remote_client::RemoteClient::new(), so it is explicit who owns the concurrent execution primitives.

Other changes:
* Updates tokio crate to 1.41 (latest stable)
* hf_xet initialization configures .xet directory to ~/.xet, and cache root to ~/.cache/huggingface/xet - using dirs crate so works across all OSes.

### DIFF
--- a/data/src/remote_shard_interface.rs
+++ b/data/src/remote_shard_interface.rs
@@ -1,5 +1,4 @@
 use std::ffi::OsStr;
-use std::os::unix::thread;
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
 


hoytak:
For reference, check out https://github.com/xetdata/xet-core/blob/main/xetldfs/src/runtime.rs

utils/src/threadpool.rs
```diff
@@ -0,0 +1,116 @@
...
+        )
+    }
+}
+
+/// Intended to be used as a singleton threadpool for the entire application.
+/// This is a simple wrapper around tokio's runtime, with some default settings.
+/// Intentionally unwrap this because if it fails, the application should not continue.
+fn new_threadpool() -> tokio::runtime::Runtime {
+    tokio::runtime::Builder::new_multi_thread()
+        .worker_threads(4) // 4 active threads
```
jgodlew: nit: Some constants for these at the top would be good.

PR #85, opened 2024-11-05
post investigation edits for download improvements

assafvayner:
- removes hash checking on cache files
- changes locking on cache reads and writes
- buffered stream of term gets rather than all spawned at once.

### DIFF
--- a/chunk_cache/src/disk.rs
+++ b/chunk_cache/src/disk.rs
@@ -581,7 +581,7 @@ impl DiskCache {
 }
 
 #[inline]
-fn index_of<T: PartialEq>(list: &Vec<T>, value: &T) -> Option<usize> {
+fn index_of<T: PartialEq>(list: &[T], value: &T) -> Option<usize> {
     for (i, list_value) in list.iter().enumerate() {
         if list_value == value {
             return Some(i);


cas_client/src/remote_client.rs
```diff
@@ -58,7 +59,7 @@ impl RemoteClient {
         Self {
             endpoint: endpoint.to_string(),
             http_auth_client: http_client::build_auth_http_client(auth, &None).unwrap(),
-            disk_cache,
+            chunk_cache: disk_cache,
```
rajatarya: I would rename `disk_cache` in the `new()` function to `chunk_cache` if renaming the struct field.

cas_client/src/remote_client.rs
```diff
@@ -275,27 +276,22 @@ impl RemoteClient {
...
         } else {
             terms.iter().fold(0, |acc, x| acc + x.unpacked_length as u64)
         };
+
+        let mut futs = Vec::with_capacity(terms.len());
+        for term in terms.into_iter() {
+            futs.push(get_one_term(http_client.clone(), self.chunk_cache.clone(), term, fetch_info.clone()));
+        }
+
+        let mut futs = futures::stream::iter(futs).buffered(4);
```
rajatarya: Channeling my inner @jgodlew : should have this magic number constant be defined as a constant.

cas_client/src/remote_client.rs
```diff
@@ -275,27 +276,22 @@ impl RemoteClient {
...
-            self.threadpool
-                .spawn(get_one_term(http_client, disk_cache, term, fetch_info, single_flight_group))
-        });
-        for (i, fut) in term_data_futures.enumerate() {
-            let term_data = fut
-                .await
-                .map_err(|e| CasClientError::InternalError(anyhow!("join error {e}")))?
-                .log_error("error getting one term")?;
+        while let Some(term_data_result) = futs.next().await {
+            let term_data = term_data_result.log_error("error fetching 1 term")?;
```
rajatarya: Nit: Shouldn't the error log emit the term as well, so it can support debugging if there is corruption in the term itself, or something went wrong in transit?

cas_client/src/remote_client.rs
```diff
@@ -305,17 +301,14 @@ impl RemoteClient {
             let end: usize = min(remaining_len + start as u64, term_data.len() as u64) as usize;
             writer.write_all(&term_data[start..end])?;
             remaining_len -= (end - start) as u64;
+            i += 1;
```
rajatarya: `i` should be renamed something useful - maybe `term_idx` or something like that.

chunk_cache/src/disk.rs
```diff
@@ -219,27 +253,44 @@ impl DiskCache {
...
+                },
             };
-            let hash = compute_hash_from_reader(&mut file_buf)?;
-            if hash != cache_item.hash {
-                debug!("file hash mismatch on path: {path:?}, key: {key}, item: {cache_item}");
-                self.remove_item(key, &cache_item)?;
-                continue;
-            }
+
+            // TODO: reintroduce hash validation of cache file, but not for every get, memoize success status per cache
```
rajatarya, line 284: Nit: I'd remove commented code that isn't in use. Just adds tech debt.

chunk_cache/src/disk.rs
```diff
@@ -219,27 +253,44 @@ impl DiskCache {
...
-            let mut file_buf = {
-                let mut file = match File::open(&path) {
-                    Ok(file) => file,
-                    Err(e) => match e.kind() {
-                        ErrorKind::NotFound => {
-                            self.remove_item(key, &cache_item)?;
-                            continue;
-                        },
-                        _ => return Err(e.into()),
+            // OLD, needed for hash validation, read file to buffer to do validation
```
rajatarya, line 256: Nit: Remove unnecessary commented code.

chunk_cache/src/disk.rs
```diff
@@ -519,9 +599,10 @@ fn compute_hash(header: &[u8], data: &[u8]) -> blake3::Hash {
     blake3::Hasher::new().update(header).update(data).finalize()
 }
 
-fn compute_hash_from_reader(r: &mut impl Read) -> Result<blake3::Hash, ChunkCacheError> {
-    Ok(blake3::Hasher::new().update_reader(r)?.finalize())
-}
+// OLD
```
rajatarya, line 626: Nit: Remove unnecessary commented code.

hf_xet/src/lib.rs
```diff
@@ -122,7 +122,6 @@ impl PyPointerFile {
 #[pymodule]
 pub fn hf_xet(m: &Bound<'_, PyModule>) -> PyResult<()> {
     log::initialize_logging();
-    let threadpool = ThreadPool::new()?;
```
rajatarya, line 125: Can you explain deleting this? Are you sure you wanted to delete this?

Where is the Threadpool created then? It is important that the Threadpool is only initialized _once_ for the Python process. By initializing it here only one threadpool for the `import hfxet` call.

rajatarya: Sorry, realized I broke the build. Because of the singleton-esque pattern for ThreadPool initialization this is no longer needed. Please delete it :)

cas_client/src/remote_client.rs
assafvayner: we don't have access to the term at this point, whatever information that went wrong should be in the error returned

chunk_cache/src/disk.rs
assafvayner: we plan to bring it back, I'm picking this up as a TODO later. Commented out for perf reasons at the moment.

assafvayner: this is the TODO: https://linear.app/xethub/issue/STO-142/hash-validation-on-cache-item-files

cas_client/src/remote_client.rs
```diff
@@ -275,27 +276,22 @@ impl RemoteClient {
...
         let total_len = if let Some(range) = byte_range {
             range.end - range.start
         } else {
             terms.iter().fold(0, |acc, x| acc + x.unpacked_length as u64)
         };
+
+        let mut futs = Vec::with_capacity(terms.len());
+        for term in terms.into_iter() {
+            futs.push(get_one_term(http_client.clone(), self.chunk_cache.clone(), term, fetch_info.clone()));
+        }
```
jgodlew: nit: I would prefer a `into_iter().map()` style iteration. Also, since you are pushing the collection into a futures::stream, we can just use the mapped iterator directly instead of collecting to a Vec.

PR #86, opened 2024-11-05
Fix build, erroneously line was merged.

rajatarya:
I broke the build, this fixes it (the `maturin develop`).


assafvayner:
this is now already in, we should close this?

rajatarya:
Yes will close.

PR #91, opened 2024-11-20
stream version of deserialize_chunks

assafvayner:
Adding a feature gated streamable implementation of deserializing xorb chunks. Intended use is in both CAS server and client to stream chunk deserialization (including decompression) while data is coming from S3.

e.g.

https://github.com/huggingface-internal/xet-core/blob/main/cas_client/src/remote_client.rs#L444-L454
https://github.com/huggingface-internal/xetcas/blob/main/xetcas/cas_server/src/core.rs#L685-L696

### DIFF
--- a/cas_object/src/cas_chunk_format/async_deserialize.rs
+++ b/cas_object/src/cas_chunk_format/async_deserialize.rs
@@ -123,16 +123,10 @@ mod tests {
         compression_scheme: CompressionScheme,
     ) -> impl Stream<Item = Result<Bytes, std::io::Error>> + Unpin {
         let data = get_chunks(rng, num_chunks, compression_scheme);
-        if data.len() < 100 {
-            return futures::stream::iter(vec![Ok(Bytes::from(data))]);
-        }
-        let chunks = data.chunks(data.len() / (2 + rng.gen::<usize>() % 8));
-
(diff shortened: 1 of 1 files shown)


cas_object/src/cas_chunk_format/async_deserialize.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    fn get_stream(
+        rng: &mut impl Rng,
+        num_chunks: u32,
+        compression_scheme: CompressionScheme,
+    ) -> impl Stream<Item = Result<Bytes, std::io::Error>> + Unpin {
+        let data = get_chunks(rng, num_chunks, compression_scheme);
+        if data.len() < 100 {
+            return futures::stream::iter(vec![Ok(Bytes::from(data))]);
+        }
```
sirahd: will this case ever be true? `CHUNK_SIZE` is set to 1000 and all test cases have >=1 chunk

assafvayner: I guess not

assafvayner:
Also tested against dev, feels slightly downloading ylow/testparquet, but no data to back it up. Code to update remote client used for test will come in another PR.

PR #93, opened 2024-11-21
re-add singleflight to cas client downloads and clean

assafvayner:
Tested adding singleflight over s3 downloads, found improvement.

Found in an experiment counting the number of duplicate downloads that singleflight does reduce the number of duplicate downloads.

experiment details:
- downloading [ylow/testparquet](https://xethub-poc.us.dev.moon.huggingface.tech/datasets/ylow/testparquet) from my m2 mac on a 100 Mbps home network (effective download speed 40-60 Mbps)
- without singleflight, duplicate download count was 368 requests and with singleflight, duplicate count was 90 requests

### DIFF
--- a/cas_client/src/remote_client.rs
+++ b/cas_client/src/remote_client.rs
@@ -1,7 +1,7 @@
 use std::cmp::{max, min};
 use std::collections::{HashMap, HashSet};
 use std::io::{Cursor, Write};
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
 
 use anyhow::anyhow;
 use async_trait::async_trait;
@@ -17,8 +17,9 @@ use http::header::RANGE;
 use merklehash::MerkleHash;
 use reqwest::{StatusCode, Url};
 use reqwest_middleware::ClientWithMiddleware;
-use tracing::{debug, error, info};
+use tracing::{debug, error, info, trace};
 use utils::auth::AuthConfig;
+use utils::singleflight::Group;
 use utils::ThreadPool;
 
 use crate::error::Result;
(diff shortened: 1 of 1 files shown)


cas_client/src/remote_client.rs
```diff
@@ -358,7 +371,9 @@ pub(crate) async fn get_one_term(
 
     // fetch the range from blob store and deserialize the chunks
     // then put into the cache if used
-    let (mut data, chunk_byte_indices) = download_range(http_client, &fetch_term).await?;
+    let (mut data, chunk_byte_indices) = range_download_single_flight
+        .work_dump_caller_info(&fetch_term.url, download_range(http_client, fetch_term.clone(), term.hash))
```
seanses, line 375: Is `fetch_term.url` the key? Will there be a case that two terms share the same URL but different byte range? In that case the singleflight will return the incorrect result.

rajatarya: I thought the range is passed in as a header (Bytes=) so not in the URL, right @assafvayner?

assafvayner: ## For Di's question:

For any term that needs to be downloaded, we look in the fetch_info section of the reconstruction response. If any 2 terms are fulfilled by adjacent or overlapping chunk ranges in the same xorb then they will be coalesced into an encompassing range in the fetch_info map. so even if the 2 `term.range` are different they may map to the same fetch_info term which will mean they will use the same presigned url.

The singleflight will return the whole data section for that fetch term, for either real term that uses the given fetch term, any trimming from the fetch term range will occur after singleflight returns.

## For Rajat's question

The S3 presigned url also contains the byte range information, it is encrypted in the signature. This is how we are able to restrict the byte range that url is valid for, otherwise it would allow access to the whole xorb, which is in breach of the auth model (since even if you dedup against some data in some xorb, you may not have access to the whole xorb).

S3 requires that the signature generated with the auth secret (that AWS has) with the params in the headers (e.g. range) match the secret that was generated by the signer (who also has the secret), hence we also have to pass in the Range header, and it must match what the signer (CAS) put when signing the url.

seanses: Makes sense to me now that the returned data will be trimmed appropriately.

PR #96, opened 2024-11-26
Add wheel package/release for hf_xet

jgodlew:
Should allow us to have GitHub automatically build/push wheels for the different envs to GitHub releases (and pypi when we have it / creds)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,11 +48,12 @@ jobs:
           args: --release --out dist
           sccache: 'true'
           manylinux: auto
+        working_directory: hf_xet
       - name: Upload wheels
         uses: actions/upload-artifact@v4
         with:
           name: wheels-linux-${{ matrix.platform.target }}
-          path: dist
+          path: hf_xet/dist
 
   musllinux:
     runs-on: ${{ matrix.platform.runner }}
@@ -79,11 +80,12 @@ jobs:
           args: --release --out dist
           sccache: 'true'
           manylinux: musllinux_1_2
+        working_directory: hf_xet
       - name: Upload wheels
         uses: actions/upload-artifact@v4
         with:
           name: wheels-musllinux-${{ matrix.platform.target }}
-          path: dist
+          path: hf_xet/dist
 
   windows:
     runs-on: ${{ matrix.platform.runner }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,191 @@
...
+#
+name: CI
+
+on:
+  push:
+    #    branches:
+    #      - main
+    #      - master
+    tags:
+      - 'v20*'
```
bpronan, line 14: Not a blocker for this feature, but is there any reason we aren't using semver for our version tags?

.github/workflows/release.yml
```diff
@@ -0,0 +1,191 @@
...
+    pull_request:
```
bpronan, line 15: If I read this correctly, we're doing a PyPi upload even for pull requests. Is that correct?

If so, maybe we want to split that out into a separate workflow.

.github/workflows/release.yml
```diff
@@ -0,0 +1,191 @@
...
+      - name: Upload sdist
+        uses: actions/upload-artifact@v4
+        with:
+          name: wheels-sdist
+          path: hf_xet/dist
+
+  release-pypi:
+    name: Release PyPi
+    runs-on: ubuntu-latest
+    if: false #"startsWith(github.ref, 'refs/tags/')"
```
bpronan, line 164: Ok. So, we aren't ever doing a PyPi push.

bpronan: As long as we use the commented code when enabling, we should be good. Sorry for the confusion.

.github/workflows/release.yml
jgodlew: I was basing this off of the existing CM workflow and [tagging convention](https://www.notion.so/huggingface2/CM-20241122-Sandbox-Refresh-Moonlanding-CAS-1461384ebcac80b3aa32f65ce9b14742?pvs=4#14f3de5b92104b2d800da820d75b20e9) (e.g. v20241122-00-http-bridge). 

We could change our CM to use a semantic version. I think we are doing this instead of semantic versioning is to have all the repos (xetcas, xet-core, moonlanding-xetpoc, huggingface_hub_xetpoc) use the same tag as opposed to having each have a potentially different semantic version.

jgodlew: IMO: once we get our forks of moon-landing and [maybe] huggingface_hub merged into their respective repos, we should switch to tags with a semantic version.

rajatarya: I agree - once we are in 'prod' then I think semantic versioning makes sense. Right now we are being really clear that nothing is stable / to be trusted.

PR #97, opened 2024-11-26
rm singleton, add cache_manager layer

assafvayner:
Removes the PointerFileTranslator singleton.

Adds a CacheManager layer, exposed by a `get_cache` function that dedupes Cache initializations based on config so if 2 RemoteClients want a cache with the same config, the same instance is returned. Uses weak pointers to allow caches to be dropped (and recreated when necessary).

Manually tested multiple repositories on dev/ephemeral with 1 upload & 1 download.

### DIFF
--- a/chunk_cache/src/lib.rs
+++ b/chunk_cache/src/lib.rs
@@ -10,6 +10,7 @@ pub use disk::test_utils::*;
 pub use disk::DiskCache;
 use error::ChunkCacheError;
 use mockall::automock;
+
 use crate::disk::DEFAULT_CAPACITY;
 
 /// ChunkCache is a trait for storing and fetching Xorb ranges.


chunk_cache/src/cache_manager.rs
```diff
@@ -0,0 +1,52 @@
...
+    /// get takes a CacheConfig and checks if there exists a valid `DiskCache` with a matching
+    /// cache_directory then it will return an Arc to that `DiskCache` instance. If it doesn't exist
+    /// or the `DiskCache` instance has been deallocated (CacheManager only holds a weak pointer)
+    /// then it creates a new instance based on the provided config.
+    fn get(&self, config: &CacheConfig) -> Result<Arc<dyn ChunkCache>, ChunkCacheError> {
+        let mut vals = self.vals.lock()?;
+        if let Some(v) = vals.get_mut(&config.cache_directory) {
+            let w = v.borrow().clone();
+            let result: Arc<dyn ChunkCache> = if let Some(value) = w.upgrade() {
+                value
```
rajatarya: When is this branch expected to be executed? 

If the hashmap entry exists and the cloned borrowed pointer cannot be upgraded? When would that happen?

assafvayner: added comments to address this. if the pointer cannot be upgraded (i.e. the cache instance has been deallocated) a new one is created and replaces the pointer in the map.

PR #98, opened 2024-11-27
Parallelize xorb upload

seanses:
Parallelize xorb upload and registration. Calls to register xorbs return immediately after computing a xorb hash so callers can continue with other work (e.g. dedup), and xorb data is queued internally to be uploaded and registered.

Achieving up to 1.8 Gbps upload speed against the ephemeral CAS.

### DIFF
--- a/data/src/parallel_xorb_uploader.rs
+++ b/data/src/parallel_xorb_uploader.rs
@@ -28,9 +28,8 @@ type XorbUploadValueType = (MerkleHash, Vec<u8>, Vec<(MerkleHash, usize)>);
 /// Calls to registering xorbs return immediately after computing a xorb hash so callers
 /// can continue with other work, and xorb data is queued internally to be uploaded and registered.
 ///
-/// It is critical to call [`flush`] before `ParallelXorbUploader` is dropped. Though
-/// dropping will attempt to wait for all transfers to finish, any errors
-/// that happen in the process of dropping will be ignored.
+/// It is critical to call [`flush`] before `ParallelXorbUploader` is dropped. Dropping will
+/// cancel all ongoing transfers automatically.
 pub(crate) struct ParallelXorbUploader {
     // Configurations
     cas_prefix: String,
@@ -132,12 +131,6 @@ impl XorbUpload for ParallelXorbUploader {
         }
 
         Ok(())
-    }
-}
-
-impl Drop for ParallelXorbUploader {
-    fn drop(&mut self) {
-        let _ = self.threadpool.block_in_place(async { self.flush().await });
     }
 }
 
--- a/utils/src/threadpool.rs
+++ b/utils/src/threadpool.rs
@@ -94,10 +94,6 @@ impl ThreadPool {
     pub fn get_handle(&self) -> tokio::runtime::Handle {
         self.inner.handle().clone()
     }
-
-    pub fn block_in_place<F: std::future::Future>(&self, future: F) -> F::Output {
-        tokio::task::block_in_place(|| self.inner.handle().block_on(future))
-    }
 }
 
 impl Display for ThreadPool {


data/src/parallel_xorb_uploader.rs
```diff
@@ -0,0 +1,201 @@
...
+            upload_worker: Mutex::new(None),
+            threadpool,
+        });
+
+        Self::run(uploader.clone(), xorb_data_c).await;
+
+        uploader
+    }
+
+    async fn run(uploader: Arc<Self>, xorbs: Receiver<XorbUploadInputType>) {
```
hoytak: I feel like this design could be a lot clearer if we use a JoinSet and then a Semaphore to handle the max parallel uploads...  Let's chat about the design when you're better?

seanses: That was my first consideration, but later I think the current approach makes it convenient to collect upload run time for each file, and more importantly, makes a multiple file upload task resumable such that we don't re-upload files whose xorbs are already in the remote. Both of them need to know when a specific file has all its xorbs sent to the remote successfully (by calling `flush` after adding all xorbs to the queue). JoinSet is not ordered so it would require more bookkeeping.

In addition to that, I think window based guard works better at congestion control on poor quality networks in that no more xorb will be put onto link when the oldest one is stuck for a long time, this resembles TCP.

data/src/parallel_xorb_uploader.rs
```diff
@@ -0,0 +1,212 @@
...
+                    // Uploaded a xorb successfully.
+                    _ => (),
+                }
+            }
+        });
+        let mut worker = uploader.upload_worker.lock().await;
+        *worker = Some(upload_task);
+    }
+
+    pub async fn register_new_cas_block(&self, cas_data: CASDataAggregator) -> Result<MerkleHash> {
```
assafvayner: can this & flushed be wrapped in a trait?

data/src/parallel_xorb_uploader.rs
```diff
@@ -0,0 +1,212 @@
...
+            .map(|(h, len)| {
+                let result = CASChunkSequenceEntry::new(*h, *len, pos);
+                pos += *len;
+                result
+            })
+            .collect();
+        let cas_info = MDBCASInfo { metadata, chunks };
+
+        shard_manager.add_cas_block(cas_info).await?;
+    }
```
assafvayner, line 177: why does this need to be in its own scope?
the function scope ends right after this scope ends

seanses: There was no special reason to put them into a scope -- only to organize code and contrast two blocks of logic.

data/src/parallel_xorb_uploader.rs
seanses: I don't expect we'll have multiple variations of "ParallelXorbUploader" and dynamically configure which one to use, so probably not necessary at the moment?

assafvayner: I'd like to be able to plug and play with in a non-tokio version

seanses: OK if you need it. Added a trait.

data/src/parallel_xorb_uploader.rs
```diff
@@ -0,0 +1,224 @@
...
+
+        signal_rx.await.map_err(|e| InternalError(format!("{e}")))?;
+
+        Ok(())
+    }
+}
+
+impl Drop for ParallelXorbUploader {
+    fn drop(&mut self) {
+        let _ = tokio::task::block_in_place(|| Handle::current().block_on(async move { self.flush().await }));
```
rajatarya: Can you use the self.threadpool to do this instead of tokio::task?

seanses: Added block_in_place fn inside ThreadPool struct and use it instead.

data/src/parallel_xorb_uploader.rs
```diff
@@ -0,0 +1,223 @@
...
+                .map(|item| process_xorb_data_queue_item(item, &shard_manager, &cas, &cas_prefix))
+                .buffered(*MAX_CONCURRENT_XORB_UPLOADS);
+
+            while let Some(ret) = buffered_upload.next().await {
+                match ret {
+                    // All xorbs added to the queue before this signal were all uploaded successfully,
+                    // send out this signal.
+                    Ok(QueueItem::Flush(signal)) => signal.send(()).expect("Upload flush signal error"),
+                    Err(e) => {
+                        panic!("Uploading and registering Xorb failed with {e}");
```
hoytak: Should this be a panic here?

data/src/parallel_xorb_uploader.rs
```diff
@@ -0,0 +1,223 @@
...
+            upload_worker: Mutex::new(None),
+            threadpool,
+        });
+
+        Self::run(uploader.clone(), xorb_data_c).await;
+
+        uploader
+    }
+
+    async fn run(uploader: Arc<Self>, xorbs: Receiver<XorbUploadInputType>) {
```
hoytak: might be easier if `uploader` was called `self`

data/src/parallel_xorb_uploader.rs
```diff
@@ -0,0 +1,223 @@
...
+                    // Uploaded a xorb successfully.
+                    _ => (),
+                }
+            }
+        });
+        let mut worker = uploader.upload_worker.lock().await;
+        *worker = Some(upload_task);
+    }
+
+    async fn task_is_running(&self) -> Result<()> {
```
hoytak: Would this be better as a bool?  seems it's only used externally to poll this...

data/src/parallel_xorb_uploader.rs
```diff
@@ -0,0 +1,223 @@
...
+
+        Ok(cas_hash)
+    }
+
+    /// Flush makes sure all xorbs added to queue before this call are sent successfully
+    /// to remote. This function can be called multiple times and should be called at
+    /// least once before `ParallelXorbUploader` is dropped.
+    async fn flush(&self) -> Result<()> {
+        // no need to flush if task already finished
+        if self.task_is_running().await.is_err() {
```
hoytak: The is_err() makes me think that we should propagate the error.

data/src/parallel_xorb_uploader.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+
+async fn process_xorb_data_queue_item(
+    item: XorbUploadInputType,
+    shard_manager: &Arc<ShardFileManager>,
+    cas: &Arc<dyn Client + Send + Sync>,
+    cas_prefix: &str,
+) -> Result<XorbUploadOutputType> {
+    let (cas_hash, data, chunks) = match item {
```
hoytak: nit -- took a bit to follow the flush logic here, so I would rather this logic were consolidated into the main control loop.  I would think this function would handle the xorb upload, not be part of the logic of the control loop.

data/src/parallel_xorb_uploader.rs
```diff
@@ -0,0 +1,186 @@
...
+}
+
+#[async_trait]
+impl XorbUpload for ParallelXorbUploader {
+    async fn register_new_cas_block(&self, cas_data: CASDataAggregator) -> Result<MerkleHash> {
+        self.status_is_ok().await?;
+
+        let cas_hash = cas_node_hash(&cas_data.chunks[..]);
+
+        // Rate limiting, the acquired permit is dropped after the task completes.
```
hoytak, line 86: Maybe a comment here about fair scheduling.

data/src/parallel_xorb_uploader.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    // Utils
+    shard_manager: Arc<ShardFileManager>,
+    cas: Arc<dyn Client + Send + Sync>,
+
+    // Internal worker
+    upload_tasks: Mutex<JoinSet<Result<()>>>,
+
+    // Rate limiter
+    concurrency: Arc<Semaphore>,
```
hoytak: With the semaphore here, concurrent calls to upload_files will create multiple instances of this semaphore which will then mean the actual number of concurrent uploads is MAX_CONCURRENT_UPLOADS  times the number of concurrent calls to upload files.  Should we push this lower, maybe into the cas.put (so we don't have to worry about it elsewhere?)

seanses: I think the problem there is memory usage: each task owns a xorb data (up to 64 MiB), Cleaner (chunking & hashing & deduping) will run fast enough to spawn a ton of upload tasks, and that soon adds up to GiB+, and up to file size, memory usage.

data/src/data_client.rs
```diff
@@ -18,8 +18,8 @@ use crate::errors::DataProcessingError;
 use crate::{errors, PointerFile, PointerFileTranslator};
 
 // Concurrency in number of files
-pub const MAX_CONCURRENT_UPLOADS: usize = 8; // TODO
-pub const MAX_CONCURRENT_DOWNLOADS: usize = 8; // TODO
+const MAX_CONCURRENT_UPLOADS: usize = 8; // TODO
```
assafvayner, line 21: why TODO?

data/src/data_processing.rs
```diff
@@ -125,14 +127,30 @@ impl PointerFileTranslator {
...
+                    &self.config.cas_storage_config.prefix,
+                    self.shard_manager.clone(),
+                    self.cas.clone(),
+                    *MAX_CONCURRENT_XORB_UPLOADS,
+                    self.threadpool.clone(),
+                )
+                .await
+            },
+        };
+        *xorb_uploader = Some(uploader.clone());
```
assafvayner: can we not create this on initialization of the PFT?

data/src/parallel_xorb_uploader.rs
seanses: Switched to use a global rate limiter to avoid this problem.

data/src/parallel_xorb_uploader.rs
seanses: Done

data/src/data_client.rs
seanses: There is no specific reason to choose 8, just a random pick, will revisit this.

data/src/data_processing.rs
seanses: In the previous design I didn't want to create the upload queue and upload worker task when there's no clean task, but now that we switched to use JoinSet  that concern no longer holds. Will change

PR #100, opened 2024-11-27
xorb stream validation

assafvayner:
This PR introduces the tooling in (mostly) cas_object package to validate a xorb passed in as an AsyncReader. The consumer of these utilities is cas_server, not various client pieces.

Also creates a CopyReader util that wraps an async reader that enables retrieving all the data that was read before. In this use case, cas_server will consume the reader to push the data to s3 after validation. This can be used for shard validation as well.

This PR is tied with (and should be merged before) https://github.com/huggingface-internal/xetcas/pull/131

### DIFF
--- a/cas_object/src/cas_object_format.rs
+++ b/cas_object/src/cas_object_format.rs
@@ -15,7 +15,8 @@ use crate::cas_chunk_format::{deserialize_chunk, serialize_chunk};
 use crate::error::{CasObjectError, Validate};
 use crate::{range_hash_from_chunks, CompressionScheme};
 
-pub(crate) const CAS_OBJECT_FORMAT_IDENT: [u8; 7] = [b'X', b'E', b'T', b'B', b'L', b'O', b'B'];
+pub type CasObjectIdent = [u8; 7];
+pub(crate) const CAS_OBJECT_FORMAT_IDENT: CasObjectIdent = [b'X', b'E', b'T', b'B', b'L', b'O', b'B'];
 pub(crate) const CAS_OBJECT_FORMAT_VERSION: u8 = 0;
 const CAS_OBJECT_INFO_DEFAULT_LENGTH: u32 = 60;
 
@@ -23,7 +24,7 @@ const CAS_OBJECT_INFO_DEFAULT_LENGTH: u32 = 60;
 /// Info struct for [CasObject]. This is stored at the end of the XORB.
 pub struct CasObjectInfo {
     /// CAS identifier: "XETBLOB"
-    pub ident: [u8; 7],
+    pub ident: CasObjectIdent,
 
     /// Format version, expected to be 0 right now.
     pub version: u8,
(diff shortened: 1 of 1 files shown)


cas_object/src/cas_chunk_format.rs
```diff
@@ -111,16 +122,18 @@ pub fn serialize_chunk<W: Write>(
...
 
-pub fn deserialize_chunk_header<R: Read>(reader: &mut R) -> Result<CASChunkHeader, CasObjectError> {
-    let mut result = CASChunkHeader::default();
-    unsafe {
-        let buf = slice::from_raw_parts_mut(&mut result as *mut _ as *mut u8, size_of::<CASChunkHeader>());
-        reader.read_exact(buf)?;
-    }
-
+pub fn parse_chunk_header(chunk_header_bytes: [u8; CAS_CHUNK_HEADER_LENGTH]) -> Result<CASChunkHeader, CasObjectError> {
+    let result: CASChunkHeader = unsafe { std::mem::transmute_copy(&chunk_header_bytes) };
```
assafvayner, line 126: note this unsafe is by design, we are transmuting an 8 byte slice to a packed struct intentionally, see removed unsafe from previous version.

cas_object/src/cas_object_format.rs
```diff
@@ -163,6 +165,87 @@ impl CasObjectInfo {
...
+    /// assumes that the ident and version have already been read and verified.
+    ///
+    /// verifies that the length of the footer data matches the length field at the very end of the buffer
+    #[cfg(feature = "stream_xorb")]
+    pub async fn deserialize_async<R: futures::io::AsyncRead + Unpin>(
+        reader: &mut R,
+        version: u8,
+    ) -> Result<(Self, u32), CasObjectError> {
+        // already read 8 bytes (ident + version)
+        let mut total_bytes_read: u32 = (size_of::<[u8; 7]>() + size_of::<u8>()) as u32;
```
sirahd: [nit] declare `[u8; 7]` as its own type and use it here and in `CAS_OBJECT_FORMAT_IDENT` const declaration to avoid hard-coding constant

cas_object/src/validate_xorb_stream.rs
```diff
@@ -0,0 +1,119 @@
...
+    reader: &mut R,
+    hash: &MerkleHash,
+) -> Result<bool> {
+    _validate_cas_object_from_async_read(reader, hash)
+        .await
+        .ok_for_format_error()
+        .map(|o| o.is_some())
+}
+
+async fn _validate_cas_object_from_async_read<R: AsyncRead + Unpin>(reader: &mut R, hash: &MerkleHash) -> Result<()> {
```
sirahd, line 27: [q] just for my context, why is validation code the async version drastically different from the current non-async version in `validate_cas_object`? 

I thought the main difference would be that anywhere `read` is called will now needs to be awaited but that seems to not be the case.

assafvayner: The key is that there is no \*"seek" capability on the streaming interface.

So we essentially need to read/parse/validate the xorb serially from the first chunk onwards until we finally receive the footer. The sync version assumes all data is present and starts validating from the footer first and then the individual chunks.

\* if we were to try and seek to the end, then we would have to wait for all the data anyway which would not be more performant than the current sync version.

PR #101, opened 2024-11-27
Check local shards to see if file is already registered

seanses:
Always check local shards to see if a file is already registered so we don't create a shard when uploading the same file twice.

### DIFF
--- a/data/src/clean.rs
+++ b/data/src/clean.rs
@@ -26,7 +26,6 @@ use utils::ThreadPool;
 
 use crate::cas_interface::Client;
 use crate::chunking::{chunk_target_default, ChunkYieldType};
-use crate::configurations::FileQueryPolicy;
 use crate::constants::MIN_SPACING_BETWEEN_GLOBAL_DEDUP_QUERIES;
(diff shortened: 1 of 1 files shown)


data/src/clean.rs
```diff
@@ -594,13 +593,8 @@ impl Cleaner {
 
         let file_hash = file_node_hash(&tracking_info.file_hashes, &self.repo_salt.unwrap_or_default())?;
 
-        // Is the file registered already?  If so, nothing needs to be added now.
-        let file_already_registered = match self.remote_shards.file_query_policy {
-            FileQueryPolicy::LocalFirst | FileQueryPolicy::LocalOnly => {
-                self.shard_manager.get_file_reconstruction_info(&file_hash).await?.is_some()
-            },
-            FileQueryPolicy::ServerOnly => false,
```
hoytak, line 602: Any idea what the original logic was here?  My guess was that there wouldn't be any shards for the server only policy anyway, so defer early to the server.

seanses: Yeah that's probably the reason before when we didn't intend to run the cleaning file feature on server only policy.

PR #102, opened 2024-12-02
feat: set up infra for xet telemetry

port8080:


port8080:
The high level idea is described in the following data flow
```mermaid
graph LR
    subgraph client
        A[(CAS)] --> |Telemetry| B[Circular Buffer]
        B --> |Telemetry| C[Consumer]
    end
    
    subgraph HF
        C --> |Telemetry Data| D[huggingface.co/api/telemetry/xet/cli]
        D --> |Processing| E[EKS]
        E --> |Visual Data| F[Dashboards]
    end
```

Unlike git-lfs, this code is likely sitting in a long running process and having a periodic timer to optimistically fire telemetry should work. I have been rather conservative with the rate (once every 100ms) given that some [context](https://huggingface.slack.com/archives/C07F4C7F7A8/p1722928095308369?thread_ts=1722866161.025209&cid=C07F4C7F7A8) in slack states that "client side telemetry data is not extensively used". 

Also, take a look at https://www.codeproject.com/articles/3479/the-bip-buffer-the-circular-buffer-with-a-twist

hf_xet/src/log.rs
```diff
@@ -1,10 +1,177 @@
...
+///   start_time = "2024-12-01T12:00:00Z",
+///   metadata_timestamp_offset = 500,
+///   metadata_bytes = 1024,
+///   data_bytes = 1048576,
+///   data_deduped_bytes = 524288,
+///   data_compressed_bytes = 262144,
+///   end_time = "2024-12-01T12:01:00Z"
+/// );
+/// ```
+struct LogBufferLayer {
```
jgodlew: nit: can we move the LogBufferLayer and associated structs/impls/task to a separate file (e.g. `log_buffer.rs`)?

hf_xet/src/log.rs
```diff
@@ -17,5 +184,133 @@ pub fn initialize_logging() {
...
+        };
+        let telemetry_filter_layer =
+            telemetry_buffer_layer.with_filter(FilterFn::new(|meta| meta.target() == "client_telemetry"));
+        tracing_subscriber::registry()
+            .with(fmt_layer)
+            .with(filter_layer)
+            .with(telemetry_filter_layer)
+            .init();
+
+        let _telemetry_task = tokio::task::spawn({
```
jgodlew: This requires that we are running in an async runtime and since we are not (as `initialize_logging` is run on module init), this line will panic. We will need to change `lib.rs` to have this call run while inside of a task spawned by the `ThreadPool` (or on creation of the ThreadPool).

hf_xet/src/log.rs
```diff
@@ -17,5 +184,133 @@ pub fn initialize_logging() {
...
+                        }
+                    }
+                    if read_len > 0 {
+                        let mut buffer = log_buffer.lock().unwrap();
+                        buffer.decommit(read_len);
+                    }
+                    if !http_header_map.is_empty() {
+                        if let Some(endpoint) = get_telemetry_endpoint() {
+                            let telemetry_url = format!("https://{endpoint}/{TELEMETRY_SUFFIX}");
+                            let client = reqwest::Client::new();
```
jgodlew: IIRC, we shouldn't need to construct a new Client each iteration of the loop and should share it across calls.

hf_xet/src/log.rs
jgodlew: See: https://github.com/huggingface-internal/xet-core/pull/82/files#diff-a61f2215d9dcf14b2d9e3480cf05a41176c3e552af529d02376e8c5decb4b21aR19-R30 for how you might be able to fix this.

port8080:
Adding @XciD and @Wauplin with the following requests:

- Please lmk the best json format for sending data to `api/telemetry` for later conversion of that into useful dashboards. 
- I am concerned about [client side telemetry data is not extensively used](https://huggingface.slack.com/archives/C07F4C7F7A8/p1722928095308369?thread_ts=1722866161.025209&cid=C07F4C7F7A8) and have put in a timer to keep the traffic at low volume. However, the number will need tweaking based on our estimate of the number of concurrent connections for lfs-files. I would love to get input on how I can go about estimating that.

hf_xet/Cargo.toml
```diff
@@ -21,3 +21,7 @@ tracing-subscriber = { version = "0.3", features = [
     "tracing-log",
     "env-filter",
 ] }
+bipbuffer = "0.1.2"
```
XciD: nit: `"0.1"` allow you to have all fix without breakings

hf_xet/src/log.rs
```diff
@@ -1,10 +1,177 @@
...
+const TELEMETRY_SUFFIX: &str = "api/telemetry/xet/cli";
+
+fn is_telemetry_disabled() -> bool {
+    matches!(env::var("HF_HUB_DISABLE_TELEMETRY").as_deref(), Ok("1"))
+}
+fn is_staging_mode() -> bool {
+    matches!(env::var("HUGGINGFACE_CO_STAGING").as_deref(), Ok("1"))
+}
+fn get_telemetry_endpoint() -> Option<String> {
+    if is_telemetry_disabled() {
```
XciD: why not rely on huggingface.co and if HF_ENDPOINT is define on it ?

XciD:
Why not using structured logs in kibana ?

port8080:
I am not 100% sure that I understand "Why not using structured logs in kibana ?", adding the following points for context:

- This code will be run in the hub client. So, user laptops etc. where we do not have access to filebeat. We are shipping the critical logs (i.e. those marked with the `target: "client_telemetry"`) over to the telemetry endpoint for later processing.
- @jgodlew is working on the following enhancement "Currently our log all the json payload is bundled up inside of the “message” field, which isn’t what we want. We should flatten this field so that the span fields (e.g. path, status, repo_id, …) are queriable by Kibana. "

Hope that clarifies the usage.

hf_xet/src/log.rs
port8080: I am attempting to match the python logic at 
- https://github.com/huggingface-internal/xetpoc_huggingface_hub/blob/main/src/huggingface_hub/utils/_telemetry.py#L27-L33
- https://github.com/huggingface-internal/xetpoc_huggingface_hub/blob/main/src/huggingface_hub/constants.py#L64-L70

with this rust code. Hope that clarifies your q.

XciD:
> I am not 100% sure that I understand "Why not using structured logs in kibana ?", adding the following points for context:

🤦 I'm not used to have rust in frontend. Discard my comment

hf_xet/src/log.rs
```diff
@@ -17,5 +184,135 @@ pub fn initialize_logging() {
...
+            .with(filter_layer)
+            .with(telemetry_filter_layer)
+            .init();
+
+        if let Some(endpoint) = get_telemetry_endpoint() {
+            let _telemetry_task = tokio::task::spawn({
+                let log_buffer = Arc::clone(&telemetry_log_buffer);
+                let log_stats = Arc::clone(&telemetry_log_stats);
+                let client = reqwest::Client::new();
+                let telemetry_url = format!("https://{endpoint}/{TELEMETRY_SUFFIX}");
```
XciD: `get_telemetry_endpoint` already have https

hf_xet/src/log.rs
```diff
@@ -1,10 +1,177 @@
...
+            };
+            event.record(&mut visitor);
+        }
+        let serializable: SerializableHeaders = (&http_headers).into();
+        if let Ok(serialized_headers) = serde_json::to_string(&serializable) {
+            if let Ok(reserved) = buffer.reserve(serialized_headers.len()) {
+                if reserved.len() < serialized_headers.len() {
+                    // log goes to /dev/null if not enough free space
+                    self.stats.records_refused.fetch_add(1, Ordering::Relaxed);
+                    self.stats
```
XciD: can be done before the if ? as it's the same in the else

hf_xet/src/log.rs
```diff
@@ -1,10 +1,177 @@
+use std::collections::HashMap;
+use std::env;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::{Arc, Mutex};
```
XciD: might need to use `use tokio::sync::Mutex;` to avoid blocking in the code ?

hf_xet/src/log.rs
```diff
@@ -17,5 +184,135 @@ pub fn initialize_logging() {
...
+                                }
+                            }
+                        }
+                        if read_len > 0 {
+                            let mut buffer = log_buffer.lock().unwrap();
+                            buffer.decommit(read_len);
+                        }
+                        if !http_header_map.is_empty() {
+                            if let Ok(response) =
+                                client.head(telemetry_url.clone()).headers(http_header_map).send().await
```
XciD: might be good to log the error in a file or somewhere for debug purpose ?

hf_xet/src/log.rs
```diff
@@ -1,10 +1,177 @@
...
+        let mut http_headers = HeaderMap::new();
+
+        {
+            let mut visitor = |field: &tracing::field::Field, value: &dyn std::fmt::Debug| {
+                http_headers.insert(field.name(), HeaderValue::from_str(&format!("{:?}", value)).unwrap());
+            };
+            event.record(&mut visitor);
+        }
+        let serializable: SerializableHeaders = (&http_headers).into();
+        if let Ok(serialized_headers) = serde_json::to_string(&serializable) {
```
XciD: just checking, you are serializing in mememory, then deserializing when sending the request ?
Why not keeping the full struct in memory ?

XciD:
> Please lmk the best json format for sending data to api/telemetry for later conversion of that into useful dashboards.

Telemetry endpoint, for convert the user-agent header from this: 

```
gradio_client/0.10.1; hf_hub/0.24.6; python/3.10.12; torch/2.4.0; src/http://0.0.0.0:8000/Mixtral-8x7B-Instruct-v0.1/
```

to this
Screenshot 2024-12-03 at 20 37 27

You can check here: https://kibana.elastic.huggingface.tech/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:60000),time:(from:now-15m,to:now))&_a=(columns:!(),filters:!(),hideChart:!f,index:'60cb2590-eb2b-11ec-abc6-8728463c304a',interval:auto,query:(language:kuery,query:''),sort:!(!('@timestamp',desc)))

If the format don't suit your needs, we can modify following your spec the endpoint.

hf_xet/src/log.rs
port8080: subtle diff - the stats suffix are `_refused` and `_written` in the two branches

hf_xet/src/log.rs
port8080: [Some](https://github.com/tokio-rs/tracing/issues/738) [good reasons](https://github.com/tokio-rs/tracing/issues/668#issuecomment-609945964) for `tracing_subscriber::layer::Layer::on_event` to not support async means that we _do_ need to use the blocking `std::sync::Mutex`. I am hoping the time spent inside the lock is tiny enough that the performance impact is not measurable and it will be ok.

hf_xet/src/log.rs
port8080: Ack, this is a perf impact. But the design constraint of "using a pre-allocated memory buffer" makes the option of storing a certain count of (e.g. 100) of log records a non-starter because these records can be variable sized. 

This channel will be used for high value metrics, and I do not expect the serde cost to impact the CPU. I am more concerned that the fixed memory budget that we allocate to this is strictly respected.

port8080:
I have changed to the code into the format specified in https://github.com/huggingface-internal/xet-core/pull/102#issuecomment-2515445168 but want to make sure I ask one more q:

All the values are categorical in your example (e.g. versions). I am planning to send numbers and counters (e.g. `metadata_bytes`, `data_bytes`, `retry_counts`, `error_counts` etc.). And then, I would love to be able to create a dashboard on these logs that add up those numbers to give us a sense of the client experience of using xet for upload and download. 

Is my goal valid, reasonable and achievable with the current ES setup?

XciD:
> All the values are categorical in your example (e.g. versions). I am planning to send numbers and counters (e.g. `metadata_bytes`, `data_bytes`, `retry_counts`, `error_counts` etc.). And then, I would love to be able to create a dashboard on these logs that add up those numbers to give us a sense of the client experience of using xet for upload and download.

Current set-up is only setup of string. Number are not parsed. We might have to change the code to make it work

hf_xet/src/log_buffer.rs
```diff
@@ -146,14 +147,42 @@ impl TryFrom<SerializableHeaders> for HeaderMap {
...
     pub stats: Arc<LogBufferStats>,
+    telemetry_versions: String,
 }
 
 impl LogBufferLayer {
     pub fn new(size: usize) -> Self {
         let buffer = Arc::new(Mutex::new(BipBuffer::new(size)));
         let stats = Arc::new(LogBufferStats::default());
+        // populate remote telemetry calls with versions for python and hf_hub if possible
+        let mut telemetry_versions = String::new();
```
jgodlew, line 158: Should we also add the hf_xet version to the version as well?

hf_xet/src/log.rs
```diff
@@ -17,5 +19,23 @@ pub fn initialize_logging() {
...
+        let telemetry_filter_layer =
+            telemetry_buffer_layer.with_filter(FilterFn::new(|meta| meta.target() == "client_telemetry"));
+
+        tracing_subscriber::registry()
+            .with(fmt_layer)
+            .with(filter_layer)
+            .with(telemetry_filter_layer)
+            .init();
+
+        let _telemetry_task = tokio::task::spawn(telemetry_task);
```
jgodlew: Although this should work, we might want to spawn using the ThreadPool for consistency in how tasks are spawned in our library.

PR #103, opened 2024-12-02
async xorb validation returns CasObject

assafvayner:
fix needed to return CasObject from validation so that it can be cached later on by the server without re-parsing the footer.

### DIFF
--- a/cas_object/src/validate_xorb_stream.rs
+++ b/cas_object/src/validate_xorb_stream.rs
@@ -12,7 +12,7 @@ use crate::cas_object_format::CAS_OBJECT_FORMAT_IDENT;
 use crate::error::{CasObjectError, Result, Validate};
 use crate::{parse_chunk_header, CASChunkHeader, CasObject};
 
-/// takes an async reader to the entire xorb data and validated that the xorb is correctly formatted
+/// takes an async reader to the entire xorb data and validates that the xorb is correctly formatted
 /// and returns the deserialized CasObject (metadata)
 ///
 /// if either the hash stored in the metadata section of the validated xorb or the computed hash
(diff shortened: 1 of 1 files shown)


cas_object/src/cas_object_format.rs
```diff
@@ -318,6 +318,17 @@ impl CasObject {
         Ok(Self { info, info_length })
     }
 
+    /// Construct CasObject object from AsyncRead.
+    /// assumes that the ident and version have already been read and verified.
+    #[cfg(feature = "stream_xorb")]
```
rajatarya: I would drop the feature flag since we don't have a use case where we would want to build without async support today.

cas_object/src/validate_xorb_stream.rs
```diff
@@ -10,24 +10,24 @@ use merklehash::MerkleHash;
 use crate::cas_chunk_format::decompress_chunk_to_writer;
 use crate::cas_object_format::CAS_OBJECT_FORMAT_IDENT;
 use crate::error::{CasObjectError, Result, Validate};
-use crate::{parse_chunk_header, CASChunkHeader, CasObjectInfo};
+use crate::{parse_chunk_header, CASChunkHeader, CasObject};
 
 // returns Ok(false) on a validation error, returns Err() on a real error
```
rajatarya: These comments are incorrect now since return signature for this function is `Option<CasObject>` and not `Option<bool>`. I would also like these comments to be docstring-style comments (`/// `) explaining the `pub fn`.

cas_object/src/validate_xorb_stream.rs
```diff
@@ -10,24 +10,24 @@ use merklehash::MerkleHash;
...
-    _validate_cas_object_from_async_read(reader, hash)
-        .await
-        .ok_for_format_error()
-        .map(|o| o.is_some())
+) -> Result<Option<CasObject>> {
+    _validate_cas_object_from_async_read(reader, hash).await.ok_for_format_error()
 }
 
-async fn _validate_cas_object_from_async_read<R: AsyncRead + Unpin>(reader: &mut R, hash: &MerkleHash) -> Result<()> {
+async fn _validate_cas_object_from_async_read<R: AsyncRead + Unpin>(
```
rajatarya, line 32: The assumption that the provided Reader is at the beginning of a CasObject should be documented in the doc string here (or in the public fn).

cas_object/src/validate_xorb_stream.rs
```diff
@@ -36,8 +36,8 @@ async fn _validate_cas_object_from_async_read<R: AsyncRead + Unpin>(reader: &mut
                 return Err(CasObjectError::FormatError(anyhow!("Xorb Invalid Format Version")));
             }
             // try to parse footer
-            let (cas_object_info, _) = CasObjectInfo::deserialize_async(reader, version).await?;
-            break cas_object_info;
+            let cas_object = CasObject::deserialize_async(reader, version).await?;
+            break cas_object;
         }
 
         // parse the chunk header, decompress the data, compute the hash
```
rajatarya, line 54: I think there should be additional logging here for 1st time in this loop - because right now the logs for parsing/validating will occur for 1st chunk vs subsequent chunks - so if the Reader is not where it should be - the logs will make it really hard to diagnose/debug the situation.

PR #104, opened 2024-12-03
export_as_keyed_shard now works as streaming function

hoytak:
With this PR, export_as_keyed_shard_streaming now allows generation of a protected shard using only a reader and writer, without the footer needing to be loaded.


mdb_shard/src/shard_format.rs
```diff
@@ -599,11 +599,15 @@ impl MDBShardInfo {
     pub fn read_all_cas_blocks_full<R: Read + Seek>(&self, reader: &mut R) -> Result<Vec<MDBCASInfo>> {
         let mut ret = Vec::with_capacity(self.num_cas_entries());
 
-        reader.seek(SeekFrom::Start(self.metadata.cas_info_offset))?;
+        let (cas_info_start, _cas_info_end) = self.cas_info_byte_range();
+
+        reader.seek(SeekFrom::Start(cas_info_start)).unwrap();
 
-        while let Some(cas_info) = MDBCASInfo::deserialize(reader)? {
+        while let Some(cas_info) = MDBCASInfo::deserialize(reader).unwrap() {
```
seanses: Why `unwrap` instead of `?`

mdb_shard/src/shard_format.rs
```diff
@@ -599,11 +599,15 @@ impl MDBShardInfo {
     pub fn read_all_cas_blocks_full<R: Read + Seek>(&self, reader: &mut R) -> Result<Vec<MDBCASInfo>> {
         let mut ret = Vec::with_capacity(self.num_cas_entries());
 
-        reader.seek(SeekFrom::Start(self.metadata.cas_info_offset))?;
+        let (cas_info_start, _cas_info_end) = self.cas_info_byte_range();
+
+        reader.seek(SeekFrom::Start(cas_info_start)).unwrap();
```
assafvayner: should we be unwrapping here?

mdb_shard/src/shard_format.rs
```diff
@@ -967,52 +970,151 @@ impl MDBShardInfo {
...
+            include_file_info,
+            include_cas_lookup_table,
+            include_chunk_lookup_table,
+            Some(self),
+        )
+    }
+
+    /// Export the current shard as an hmac keyed shard,
+    #[allow(clippy::too_many_arguments)]
+    pub fn export_as_keyed_shard_streaming<R: Read + Seek, W: Write>(
```
assafvayner, line 987: how are we streaming the shard data with a sync `Read` that would entail that all the data must be present for this reader IIUC (pls correct me if I'm incorrect). That or this reader will be blocking a thread waiting for another async io to complete.

mdb_shard/src/shard_format.rs
```diff
@@ -967,52 +970,151 @@ impl MDBShardInfo {
...
+        reader: &mut R,
+        writer: &mut W,
+        hmac_key: HMACKey,
+        key_valid_for: std::time::Duration,
+        include_file_info: bool,
+        include_cas_lookup_table: bool,
+        include_chunk_lookup_table: bool,
```
assafvayner, line 994: should we wrap these in some struct that will be less total arguments, I think that would also slightly optimize space for the 3 bools, maybe just put those in a struct? This is a teensy nit, don't have to act on it

mdb_shard/src/shard_format.rs
hoytak: Debug code trying to track down something wrong.  Will revert!

mdb_shard/src/shard_format.rs
hoytak: Debug code to trace an issue, will revert.

mdb_shard/src/shard_format.rs
hoytak: Good question.  Currently none of these functions are async so we can't call async functions here...  Let me look at the interface on the cas side and see what needs to happen.

mdb_shard/src/shard_format.rs
hoytak: This keeps the existing API except for removing the &self; but yeah, it could be cleaned up.

PR #106, opened 2024-12-04
progress bar for download

assafvayner:


hf_xet/src/progress_update.rs
```diff
@@ -0,0 +1,58 @@
...
+        trace!("updating progress bar");
+        Python::with_gil(|py| {
+            let f = self.py_func.bind(py);
+            if !f.is_callable() {
+                error!("ProgressUpdater func: {} is not callable", self.name);
+                return;
+            }
+            let _ = f
+                .call1((total,))
+                .inspect_err(|e| error!("python exception trying to update progress bar {e:?}"));
```
hoytak: This is where CTRL-C will show up -- calling back into python after a ctrl-c will cause that python function to raise a KeyboardInterrupt exception, so that should be handled instead of dropped...

rajatarya:
Is this WIP? Are you looking for a quick pass through it or are you ready to merge and looking for a full review?

data/src/data_client.rs
```diff
@@ -56,19 +57,36 @@ pub async fn download_async(
...
-    .await
-    .map_err(|e| match e {
-        ParallelError::JoinError => DataProcessingError::InternalError("Join error".to_string()),
-        ParallelError::TaskError(e) => e,
-    })?;
+    let paths =
+        tokio_par_for_each(pointer_files_plus, MAX_CONCURRENT_DOWNLOADS, |(pointer_file, updater), _| async move {
+            let proc = processor.clone();
+            let res = smudge_file(&proc, &pointer_file, updater).await;
+            tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
rajatarya: Why sleep?

rajatarya: Is this a typo that should be removed?

hf_xet/src/progress_update.rs
```diff
@@ -0,0 +1,58 @@
+use std::fmt::{Debug, Formatter};
+
+use pyo3::exceptions::PyTypeError;
+use pyo3::prelude::PyAnyMethods;
+use pyo3::{Py, PyAny, PyErr, PyResult, Python};
+use tracing::{error, trace};
+use utils::progress::ProgressUpdater;
+
+/// A wrapper struct of a python function to update a progress bar
+pub struct WrappedProgressUpdater {
```
rajatarya, line 15: I'd like to see more docs describing how to use this / where it is used / with python example code as well. Otherwise this is very easy to completely forget in 6mo.

data/src/data_client.rs
assafvayner: yes, that was a left by mistake from testing, good catch.

assafvayner:
Screenshot 2024-12-10 at 10 00 26 AM
@rajatarya

assafvayner:
https://github.com/huggingface-internal/xetpoc_huggingface_hub/pull/11

PR #113, opened 2024-12-11
Each PFT uses isolated shard session directory

seanses:
Fix STO-190
- Each PFT uses isolated shard session directory so they don't conflict on shard upload stage. The implication is that concurrent PFTs don't dedupe with each other. We may consider to push down parallelism logic from Python library to xet-core library in the future.
- Move `fn default_config` back from configurations.rs to data_client.rs: configurations.rs contains general definitions while `fn default_config` contains more business logic configs.
- Fix a critical bug introduced in this refactor [PR](https://github.com/huggingface-internal/xet-core/pull/90): we are supposed to use a custom defined [SMALL_FILE_THRESHOLD](https://github.com/huggingface-internal/xet-core/blob/6825cc91a3810576377787b1de8015e4d8881548/hf_xet/src/config.rs#L10) that equals `1` because we now process every file delegated from the Python library, but that PR mistakenly used https://github.com/huggingface-internal/xet-core/blob/ebee1d8e49e28d8f15c28cd647159a4b794ea999/data/src/constants.rs#L29.

### DIFF
--- a/data/src/data_client.rs
+++ b/data/src/data_client.rs
@@ -8,7 +8,7 @@ use std::sync::Arc;
 use cas_client::CacheConfig;
 use dirs::home_dir;
 use parutils::{tokio_par_for_each, ParallelError};
-use tempfile::{tempdir, tempdir_in, TempDir};
+use tempfile::{tempdir_in, TempDir};
 use utils::auth::{AuthConfig, TokenRefresher};
 use utils::progress::ProgressUpdater;
 use utils::ThreadPool;


data/src/data_client.rs
```diff
@@ -1,25 +1,86 @@
...
+pub fn default_config(
+    endpoint: String,
+    token_info: Option<(String, u64)>,
+    token_refresher: Option<Arc<dyn TokenRefresher>>,
+) -> errors::Result<(TranslatorConfig, TempDir)> {
+    let home = home_dir().unwrap_or(current_dir()?);
+    let xet_path = home.join(".xet");
+    std::fs::create_dir_all(&xet_path)?;
+
+    let cache_path = home.join(".cache").join("huggingface").join("xet");
```
ylow, line 39: std::fs::create_dir_all(&cache_path)?;

data/src/data_client.rs
```diff
@@ -1,25 +1,86 @@
...
+
+    let (token, token_expiration) = token_info.unzip();
+    let auth_cfg = AuthConfig::maybe_new(token, token_expiration, token_refresher);
+
+    let shard_staging_directory = tempdir_in(xet_path.join("shard-session"))?;
```
rajatarya, line 44: std::fs::create_dir_all(xet_path.join("shard-session")?;

PR #114, opened 2024-12-11
Create .xet/shard-session directory before tempdir_in

seanses:
tempdir_in will panic if the root directory doesn't exist already

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,17 +20,18 @@ permissions:
 
 jobs:
   linux:
-    runs-on: ${{ matrix.platform.runner }}
+    runs-on:
+      group: ${{ matrix.platform.runner }}
     strategy:
       matrix:
         platform:
-          - runner: ubuntu-latest
+          - runner: cpu-high
             target: x86_64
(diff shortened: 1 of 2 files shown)


data/src/data_client.rs
```diff
@@ -41,7 +41,9 @@ pub fn default_config(
     let (token, token_expiration) = token_info.unzip();
     let auth_cfg = AuthConfig::maybe_new(token, token_expiration, token_refresher);
 
-    let shard_staging_directory = tempdir_in(xet_path.join("shard-session"))?;
+    let shard_staging_root = xet_path.join("shard-session");
```
assafvayner, line 44: do we need to store this content under the `xet_path`, (presumably `$HOME/.xet`). Can we instead define these as temp files under the OS tmp directory?

seanses: No specific reason, open to change it

rajatarya: Let's commit as-is for now and we can revisit later. I think it is good for this to be rooted a specific/known place so user has a place to go. The user already knows about ~/.cache/huggingface so part of me thinks it should go in there.

PR #119, opened 2024-12-12
Python CTRL-C cancellation

hoytak:
This PR implements CTRL-C cancellation for tasks by using a non-tokio background thread that periodically checks the python signal handler for a keyboard interrupt signal, then issues a shutdown to the current runtime.  After the runtime is shut down, the execution returns the appropriate python exception.

### DIFF
--- a/utils/src/threadpool.rs
+++ b/utils/src/threadpool.rs
@@ -2,7 +2,6 @@ use std::fmt::Display;
 use std::future::Future;
 use std::sync::atomic::Ordering::SeqCst;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
-use std::time::Duration;
 
 /// This module provides a simple wrapper around Tokio's runtime to create a thread pool
 /// with some default settings. It is intended to be used as a singleton thread pool for
@@ -52,14 +51,13 @@ use std::time::Duration;
 /// - `new_threadpool`: Creates a new Tokio runtime with the specified settings.
 use tokio;
 use tokio::task::{JoinError, JoinHandle};
-use tracing::debug;
+use tracing::{debug, error};
 use xet_error::Error;
 
 const THREADPOOL_NUM_WORKER_THREADS: usize = 4; // 4 active threads
 const THREADPOOL_THREAD_ID_PREFIX: &str = "hf-xet"; // thread names will be hf-xet-0, hf-xet-1, etc.
 const THREADPOOL_STACK_SIZE: usize = 8_000_000; // 8MB stack size
 const THREADPOOL_MAX_BLOCKING_THREADS: usize = 100; // max 100 threads can block IO
-const RUNTIME_SHUTDOWN_WINDOW_MS: u64 = 5 * 1000; // The runtime shuts down in 5 seconds when Ctrl-C is pressed.
 
 /// Define an error time for spawning external threads.
 #[derive(Debug, Error)]
@@ -71,7 +69,7 @@ pub enum MultithreadedRuntimeError {
     #[error("Task Panic: {0:?}.")]
     TaskPanic(JoinError),
 
-    #[error("Task cancelled; possible runtime shutdown in progress {0}.")]
+    #[error("Task cancelled; possible runtime shutdown in progress ({0}).")]
     TaskCanceled(String),
 
     #[error("Unknown task runtime error: {0}")]
@@ -83,8 +81,9 @@ pub struct ThreadPool {
     // This has to allow for exclusive access to enable shutdown when
     runtime: std::sync::RwLock<Option<tokio::runtime::Runtime>>,
 
-    // We should use this handle when we actually enter the runtime, which means that during shutdown, spawned tasks
-    // and the like are actually
+    // We use this handle when we actually enter the runtime to avoid the lock.  It is
+    // the same as using the runtime, with the exception that it does not block a shutdown
+    // while holding a reference to the runtime does.
     handle: tokio::runtime::Handle,
 
     // The number of external threads calling into this threadpool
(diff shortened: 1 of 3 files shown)


hf_xet/src/log.rs
```diff
@@ -16,7 +17,7 @@ const DEFAULT_LOG_LEVEL: &str = "warn";
 #[cfg(debug_assertions)]
 const DEFAULT_LOG_LEVEL: &str = "info";
 
-pub fn initialize_logging(threadpool: Arc<ThreadPool>) {
+pub fn initialize_logging(py: Python, runtime: Arc<ThreadPool>) {
```
ylow, line 20: py doesn't seem to be used (Also in logbufferlayer)

utils/src/threadpool.rs
```diff
@@ -108,24 +112,35 @@ impl ThreadPool {
     }
 
     /// Cancels and shuts down the runtime.  All tasks currently running will be aborted.
-    pub fn cancel_all_and_shutdown(&self) {
+    pub fn perform_sigint_shutdown(&self) {
         // Shutdown with a timeout.  This waits up to 5 seconds for all the running tasks to complete or yield at an
         // await statement, then it drops the worker threads to force shutdown.  This may cause resource
         // leaks, but this is usually used in the context where the larger process needs to be shut down
         // anyway and thus that is okay.
```
ylow: This is not generally true. For instance running in jupyter notebook.

ylow:
To take as an example, the difficulty of forced cancellation, there is an inherent race in tokio shutdown. For instance if due to timing of ctrl-c we happen to start a `block_on` while awaiting shutdown. Tokio can panic (and so take the whole process down). https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.block_on 
Forced task cancellation falls into the domain of "this is stupid hard and you should never do it" category.

hf_xet/src/log.rs
hoytak: It's used in LogBufferLayer::new to query for the python version and system information, which I guess we could query beforehand and pass in, but I don't see a reason to not have it here, given that this is the python specific code for initializing the logging. 

The reason for this change was that now calling get_threadpool() is done before releasing the GIL, as it's used after the execution call completes.  The with_gil call in LogBufferLayer::new was blocking.

utils/src/threadpool.rs
hoytak: Fair.  In testing, the task cancellation mechanism works well, so I can switch it to just shutdown and not use the one with the hard timeout.

hoytak:
> To take as an example, the difficulty of forced cancellation, there is an inherent race in tokio shutdown. For instance if due to timing of ctrl-c we happen to start a `block_on` while awaiting shutdown. Tokio can panic (and so take the whole process down). https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.block_on Forced task cancellation falls into the domain of "this is stupid hard and you should never do it" category.

Yes, this is exactly why I wrote the external_run_async call to do nothing but call tokio spawn, which will return a TaskCanceled error on the join handle.  Long running block_on processes without an await will block the runtime and cause these sorts of panics.  

I agree that forced task cancellation is hard, but also it seems to be the rust async way.  I see all this more as an argument to isolate the async parts to the IO and rework the rest of the computation using traditional threading with periodic checks for a shutdown.

utils/src/threadpool.rs
hoytak: Fixed and updated the comment.

hoytak:
> Structurally ok, seemingly some additional unrelated changes?
> 
> As mentioned in huddle, I am broadly uncomfortable with forced runtime shutdown as forced task cancellation is generally recognized as unsafe. People do not generally write code expecting await calls to also be cancellation points.
> 
> I would prefer a simple polling based approach where tasks can "opt-in" to being cancellation points. It will generally require a bit more work especially in "upload" code as we currently just give the whole body contents to the http library and let it deal with it. Instead we will need do something like attach it to a stream instead (https://stackoverflow.com/questions/70252995/how-to-monitor-reqwest-client-upload-progress) which will also help with progress bars.
> 
> Ok with the approach as is for now as a quick solution, but I do want to switch over a explicit cancellation approach soon.

Agreed, that it is not ideal, but it's also a built in aspect of a lot of the rust async design.  Note that it also will occur currently; e.g. when using a joinset, an error propegating back up may cause the destructor of the joinset to run while there are still tasks in it, which causes all these to be cancelled. 

I see this as an argument to isolate async calls...

ylow:
> > Structurally ok, seemingly some additional unrelated changes?
> > As mentioned in huddle, I am broadly uncomfortable with forced runtime shutdown as forced task cancellation is generally recognized as unsafe. People do not generally write code expecting await calls to also be cancellation points.
> > I would prefer a simple polling based approach where tasks can "opt-in" to being cancellation points. It will generally require a bit more work especially in "upload" code as we currently just give the whole body contents to the http library and let it deal with it. Instead we will need do something like attach it to a stream instead (https://stackoverflow.com/questions/70252995/how-to-monitor-reqwest-client-upload-progress) which will also help with progress bars.
> > Ok with the approach as is for now as a quick solution, but I do want to switch over a explicit cancellation approach soon.
> 
> Agreed, that it is not ideal, but it's also a built in aspect of a lot of the rust async design. Note that it also will occur currently; e.g. when using a joinset, an error propegating back up may cause the destructor of the joinset to run while there are still tasks in it, which causes all these to be cancelled.
> 
> I see this as an argument to isolate async calls...

It is a builtin, but we need not use it. We can treat async tasks exactly as threads and cancel them normally.

hf_xet/src/log.rs
ylow: Can the LogBufferLayer::new be done without gil?

rajatarya:
Created https://linear.app/xethub/issue/STO-226/refactor-ctrlc-cancellation to track the refactoring of this so we don't forget it.

STO-226

hf_xet/src/log.rs
hoytak: It could, but that would just mean copying that code to the a place higher up and passing all that information down to this function explicitly.  Which could make sense, but doesn't seem to give us much for that refactoring.  

More detailed, we should probably rename LogBufferLayer as it's not really that...

PR #120, opened 2024-12-13
upload progress bar work

assafvayner:
fix STO-155

Adds progress tracking on uploads. requires https://github.com/huggingface-internal/xetpoc_huggingface_hub/pull/12

upload progress is reported on a xorb upload or a chunk having been detected as a duplicate.

### DIFF
--- a/data/src/data_processing.rs
+++ b/data/src/data_processing.rs
@@ -77,7 +77,11 @@ pub struct PointerFileTranslator {
 
 // Constructors
 impl PointerFileTranslator {
-    pub async fn new(config: TranslatorConfig, threadpool: Arc<ThreadPool>, upload_progress_updater: Option<Arc<dyn ProgressUpdater>>) -> Result<PointerFileTranslator> {
+    pub async fn new(
+        config: TranslatorConfig,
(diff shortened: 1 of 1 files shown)


data/src/bin/example.rs
```diff
@@ -162,7 +162,7 @@ async fn clean(mut reader: impl Read, mut writer: impl Write) -> Result<()> {
 
     let mut read_buf = vec![0u8; READ_BLOCK_SIZE];
 
-    let translator = PointerFileTranslator::new(default_clean_config()?, get_threadpool()).await?;
+    let translator = PointerFileTranslator::new(default_clean_config()?, get_threadpool(), None).await?;
```
rajatarya, line 165: Not necessary, but would be nice to have progress bar in Rust as well. This would likely be used as the huggingface_hub Python client is rewritten in Rust. And would be a simpler debugging loop for verifying progress bar bookkeeping maintenance.

assafvayner: Let's add the progress bar from the rust context when we're removing the progress bar from the python context

PR #122, opened 2024-12-16
Update threadpool

hoytak:
This PR updates the current threadpool implementation, anticipating changes needed for the CTRL-C cancelation issues.  
Calls from the external environment into the runtime will record how many execution threads are running in the threadpool and ensure that the code actually gets run in one of the worker threads; panics or other runtime issues are propagated as errors.

### DIFF
--- a/cas_types/src/key.rs
+++ b/cas_types/src/key.rs
@@ -83,7 +83,7 @@ pub mod hex {
 #[derive(Debug, Clone, Copy, Serialize, Deserialize, Default, PartialEq, Eq, Hash)]
 pub struct HexMerkleHash(#[serde(with = "hex::serde")] pub MerkleHash);
 
-impl std::fmt::Display for HexMerkleHash {
+impl Display for HexMerkleHash {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         write!(f, "{}", self.0.hex())
     }
@@ -113,7 +113,7 @@ impl From<&MerkleHash> for HexMerkleHash {
     }
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+#[derive(Debug, Clone, Serialize, Deserialize, Default, Hash, PartialEq, Eq)]
 pub struct HexKey {
     pub prefix: String,
     #[serde(with = "hex::serde")]
--- a/cas_types/src/lib.rs
+++ b/cas_types/src/lib.rs
@@ -1,5 +1,5 @@
 use core::fmt;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use merklehash::MerkleHash;
 use serde::{Deserialize, Serialize};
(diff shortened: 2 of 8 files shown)


utils/src/threadpool.rs
```diff
@@ -51,54 +50,151 @@ use std::sync::atomic::Ordering::SeqCst;
...
     }
 
-    pub fn block_on<F: std::future::Future>(&self, future: F) -> F::Output {
-        debug!("threadpool: block_on called, {}", self);
-        self.inner.block_on(future)
+    /// Gives the number of concurrent calls to external_run_async_task.
+    #[inline]
+    pub fn external_executor_count(&self) -> usize {
+        self.external_executor_count.load(Ordering::SeqCst)
     }
```
assafvayner, line 108: make this not pub?

utils/src/threadpool.rs
```diff
@@ -51,54 +50,151 @@ use std::sync::atomic::Ordering::SeqCst;
...
+        ret
+    }
+
+    /// Call this function from within the tokio runtime to call an async function from a
+    /// non-async context.  However, it is expensive and use should be minimized.
+    pub fn internal_block_on<F: std::future::Future>(&self, future: F) -> F::Output {
+        debug!("threadpool: internal_block_on called, {}", self);
+        let handle = self.handle();
+        tokio::task::block_in_place(move || handle.block_on(future))
+    }
```
assafvayner: this isn't used anywhere (right?), why do we need it?

utils/src/threadpool.rs
hoytak: It will be used in hf_xet

utils/src/threadpool.rs
hoytak: It was going to be used from implementing the Drop trait, but that got refactored.   Can drop.

utils/src/threadpool.rs
```diff
@@ -51,54 +50,143 @@ use std::sync::atomic::Ordering::SeqCst;
...
+
+    #[error("Unknown task runtime error: {0}")]
+    Other(String),
+}
 
 #[derive(Debug)]
 pub struct ThreadPool {
-    inner: tokio::runtime::Runtime,
-}
+    // This has to allow for exclusive access to enable shutdown when
```
seanses, line 83: Some comments here in this struct seemed to be cut off?

utils/src/threadpool.rs
```diff
@@ -51,54 +50,143 @@ use std::sync::atomic::Ordering::SeqCst;
...
 
 impl ThreadPool {
-    pub fn new() -> Self {
-        Self {
-            inner: new_threadpool(),
-        }
+    pub fn new() -> Result<Self, MultithreadedRuntimeError> {
+        let runtime = new_threadpool()?;
+        Ok(Self {
+            handle: runtime.handle().clone(),
```
seanses, line 98: This `handle` is cached but not invalidated in `cancel_all_and_shutdown`, feels like some weird concurrency related issues can arise. Getting the handle is unlikely to be a bottleneck, maybe we don't cache it?

utils/src/threadpool.rs
```diff
@@ -15,16 +16,14 @@ use std::sync::atomic::Ordering::SeqCst;
...
-///     // Your async code here
-///     42
-/// });
+/// let result = pool
+///     .external_run_async_task(async {
+///         // Your async code here
+///         42
+///     })
+///     .expect("Task Error.");
 ///
```
rajatarya, line 27: Pretty sure the docstring thing will be unhappy because this assert doesn't work anymore, need to add back the 42 to the docstring example.

utils/src/threadpool.rs
hoytak: It's actually internally reference counted, so it doesn't require us to hold onto it.  running tasks on the handle will themselves just abort if it's in shutdown.

PR #125, opened 2024-12-18
POC: bridge dl with end-to-end async streaming

co42:
Testing if it's possible to completely stream the download via the bridge without storing chunk in RAM.
And is it fast enough ?


co42:
Wrong project

PR #130, opened 2024-12-20
Compress models better

seanses:
Implements ZipNN byte grouping compression based on lz4 (bg4-lz4).
For a comparison of compression ratio between bg4-lz4 and lz4, see the [test](https://github.com/huggingface/xet-core/blob/8eeb1bdcfb8b8441c34cb4e89f4500813b599b4a/cas_object/src/compression_scheme.rs#L227), also copied below. 

In summary, bg4-lz4 provides compression ratio no worse than lz4. bg4-lz4 achieves 7.4% reduction for **random** f32 values in [-1, 1], 13% for those in [0, 2], 15% for **random** bf16 values in [-1,1] and 26% for those in [0,2], while lz4 yields none. bg4-lz4 doesn't compress for random f64 values, I think we need 8 byte groups to make the compression happen. It doesn't compress for f16 values either, we need to extract and shuffle the sign bit to make the compression happen.

405360161-0a9de0e0-3fe6-44e2-aaea-a5ac751789af

See benchmark results on Apple M2 Max [here](https://github.com/huggingface/xet-core/blob/8eeb1bdcfb8b8441c34cb4e89f4500813b599b4a/cas_object/benches/compression_bench.rs#L10), also copied below.

405359672-99d8e290-308d-4734-ac0c-d42e2e2c454e

Byte splits occupies 2% - 34% out of total compression time, while for data types that bg4-lz4 yields additional compression to lz4, split occupies 2% - 7% total time. Byte regrouping occupies 17% - 57% out of total decompression time. This likely can be optimized more with a group of "*gather*" SIMD instructions.

Will merge into xetcas before changing the default compression config in the cas client.


cas_object/src/compression_scheme.rs
```diff
@@ -12,75 +14,267 @@ pub enum CompressionScheme {
...
-    fn from(value: CompressionScheme) -> Self {
-        From::from(&value)
+fn lz4_compress_from_slice(data: &[u8]) -> Result<Vec<u8>> {
+    let mut enc = FrameEncoder::new(Vec::new());
+    enc.write_all(data)?;
+    Ok(enc.finish()?)
+}
+
+fn lz4_decompress_from_slice(data: &[u8]) -> Result<Vec<u8>> {
+    let mut dest = vec![];
```
assafvayner, line 94: style nit, Vec::new()

cas_object/src/compression_scheme.rs
```diff
@@ -12,75 +14,267 @@ pub enum CompressionScheme {
...
     }
+
+    let regrouped = bg4_regroup(&groups);
+
+    writer.write_all(&regrouped)?;
+
+    Ok(regrouped.len() as u64)
+}
+
+fn bg4_split(data: &[u8]) -> [Vec<u8>; 4] {
```
assafvayner: please leave a comment explaining this protocol/format, if there is a design doc then a link here as well would be nice.

cas_object/src/compression_scheme.rs
```diff
@@ -12,75 +14,267 @@ pub enum CompressionScheme {
...
+fn lz4_decompress_from_reader<R: Read, W: Write>(reader: &mut R, writer: &mut W) -> Result<u64> {
+    let mut dec = FrameDecoder::new(reader);
+    Ok(copy(&mut dec, writer)?)
+}
+
+fn bg4_lz4_compress_from_slice(data: &[u8]) -> Result<Vec<u8>> {
+    let groups = bg4_split(data);
+    let mut dest = vec![];
+    for g in groups {
+        let mut enc = FrameEncoder::new(&mut dest);
```
assafvayner: do you need to recreate the encoder for each group?

cas_object/src/compression_scheme.rs
```diff
@@ -12,75 +14,267 @@ pub enum CompressionScheme {
...
+}
+
+fn lz4_decompress_from_reader<R: Read, W: Write>(reader: &mut R, writer: &mut W) -> Result<u64> {
+    let mut dec = FrameDecoder::new(reader);
+    Ok(copy(&mut dec, writer)?)
+}
+
+fn bg4_lz4_compress_from_slice(data: &[u8]) -> Result<Vec<u8>> {
+    let groups = bg4_split(data);
+    let mut dest = vec![];
```
assafvayner, line 112: nit vec![] -> Vec::new()

Maybe it's just me, but I recall the linter used to disallow an empty `vec![]`, maybe if the linter doesn't care I shouldn't either 🤷 ?

cas_object/src/compression_scheme.rs
port8080: https://github.com/huggingface/dedupe_estimator/blob/main/dedupe_estimator.cpp#L287C8-L314

But, this code does seem different from the zipnn paper - https://arxiv.org/pdf/2411.05239 - where the sign bit is moved over. See image below
Screenshot 2024-12-20 at 2 58 20 PM

@seanses I think your code matches @ylow dedup_estimater, but it is worth checking if the original zipnn idea will give even better results

seanses: @port8080 Indeed, that's actually where I doubt in the first place if ZipNN is applicable to our architecture. Moving the sign bit requires knowing where the sign bit is, that means 1). our chunker produces a clean cut at f32/f64/f16/bf16 boundaries; 2). uses different algorithm for f32 and f16/bf16. Both of them require deep plumbing to learn model file format from the python library. Thus I think bit moving is more suitable for ad-hoc solutions instead of serving as a general approach. With that being said, I still like to try it and see how it performs, so I don't plan to merge this in before a complete benchmark.

cas_object/src/compression_scheme.rs
seanses: I actually like the macro a lot, less typing, linters never bark at me for using these macros.

seanses:
> Benchmarking request, I would love to see data for compression & decompression speeds - similar to Table 3 in the [paper](https://arxiv.org/pdf/2411.05239)
> 
> ```
> 5.2 Compression and Decompression speed
> We ran our tests on an Apple M1 Max machine with 10
> cores and 64GB of RAM running macOS Sonoma 14.3.
> The tests run in a single process and on a single core. Table 3 shows the speed benefits of our method vs. vanilla compressors on 3 representative models. 
> ```

Yeah, all coming along.

cas_object/src/compression_scheme.rs
seanses: The main idea is not to have the compression of a group interfering with each other. It's a bit difficult to control so that a group maps to a single block within a lz4 frame ([context](https://github.com/lz4/lz4/blob/dev/doc/lz4_Frame_format.md)), but I won't say that's impossible to achieve. There are still a lot of details to revisit.

cas_object/src/compression_scheme.rs
port8080: As we discussed this morning, safetensors is ~30% of uploaded bytes, and we can dig into that format in particular to understand the layout. Of course, let us first measure to see if the advantage is worth it.

ylow: I strongly prefer *not* building something format specific. This leaves us highly format agnostic and this will work automatically for dduf, gguf, Tensorflow files, pytorch etc. Which in aggregate is a lot more than just safetensors.

seanses:
Improved compression ratio and speed. Added more benchmarks.

port8080 (approved):
🚢 
This is a wonderful addition to our toolbox, I will keep asking for more benchmarking numbers but this code is good to go

seanses:
Results on real models (dedup against dev):
compression

seanses:
All numbers available in table at https://docs.google.com/spreadsheets/d/1SJ8Dv3EcNTuA41JXT-ggL4SzMnqs84hn/edit?usp=sharing&ouid=108235600614994105911&rtpof=true&sd=true

seanses:
Results on real datasets (dedup against dev):
compression2



This makes a clear distinction that this algorithm should only be applied to models.

PR #131, opened 2024-12-20
XetRuntime: new concurrent runtime for isolating async. 

hoytak:
This PR introduces a new Xet Runtime class with both synchronous and async compute runtimes.  There are several design objectives:

1. Allow us to isolate the async part of our codebase to only the portions that really need it.  
2. Abstract away the specific async runtime (now tokio) to allow us to swap it out for lighter weight runtime down the road if desired.
3. Expose compute focused non-async primitives (e.g. par_for, JoinSet, JoinHandle) that we use frequently in the async world, but that could easily be done in the multithreaded sync world instead. 
4. Expose reliable ways to call into async routines from the non-async runtime. 
5. Lay the groundwork for better CTRL-C cancellations. 

Internally, the runtime maintains both a small tokio runtime and a small rayon threadpool for the parallel compute part. 

### Cancellation check
```rust
xet_runtime().check_runtime_cancellation()?; // Returns an error if cancellation is requested
```

### Background compute
```rust
let handle = xet_runtime().spawn_compute_task(|| ...)?;

let result = handle.join()?; // Wait for the result
```

### Run an async task from non-async code runtime.
```rust
let result = xet_runtime().run_async_task(async { ... })?;
```

### Run a long-running task on the compute threadpool from an async context
```rust
let result = xet_runtime().run_compute_task_from_async(|| ...).await?;
```

### Spawn a background async task
```rust
let handle = xet_runtime().spawn_async_task(async { ... });  // async or non-async
let result = handle.await?; // In async code 
let result = handle.join()?; // In non-async code
```

### par_for
```rust
let results: Vec<_> = xet_runtime().par_for(inputs, |x| { let v = ... ; Ok(v) })?; 
```

###  `par_for_each`
```rust
xet_runtime().par_for_each(inputs, |x| { do_something(); Ok(()) })?; 
```

### Compute JoinSet
```rust
let mut join_set = xet_runtime().compute_joinset();
join_set.spawn(|| 42);
join_set.spawn(|| 99);
while let Some(result) = join_set.join_next()? { ... } 
```

### Async JoinSet
```rust
let mut join_set = xet_runtime().async_joinset();
join_set.spawn(async { 42 });
join_set.spawn(async { 99 });
while let Some(result) = join_set.join_next().await? { ... }
```

### External Entry Points

To call in to the runtime from python or other places use enter_runtime or enter_runtime_async.  These are similar to the above but have specific error reporting and cancelation bookkeeping. 

```rust
let result = xet_runtime().enter_runtime(|| {...})?;
let result = xet_runtime().enter_runtime_async(async { ... } )?; 
```

### DIFF
--- a/xet_runtime/src/errors.rs
+++ b/xet_runtime/src/errors.rs
@@ -1,0 +1,53 @@
+use tracing::error;
+use xet_error::Error;
+
+/// An error class for everyone to wrap that indicates cancellation.
+#[derive(Debug, Error)]
+pub struct RuntimeCancellation {}
+
+impl std::fmt::Display for RuntimeCancellation {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "RuntimeCancellation")
+    }
+}
+
+/// Define an error time for spawning external threads.
+#[derive(Debug, Error)]
+#[non_exhaustive]
+pub enum XetRuntimeError {
+    #[error("Error Initializing Async Runtime: {0:?}")]
+    AsyncRuntimeInitializationError(#[from] std::io::Error),
+
+    #[error("Error Initializing Compute Runtime: {0:?}")]
+    ComputeRuntimeInitializationError(String),
+
+    #[error("Task Panic: {0:?}.")]
+    TaskPanic(String),
+
+    #[error("Task cancelled; possible runtime shutdown in progress ({0}).")]
+    TaskCanceled(String),
+
+    #[error("Async Task Join Error: {0}")]
+    OtherJoinError(tokio::task::JoinError),
+
+    #[error("RuntimeCancellation")]
+    RuntimeCancellation(#[from] RuntimeCancellation),
+
+    #[error("ask runtime error: {0}")]
+    Other(String),
+}
+
+pub type Result<T> = std::result::Result<T, XetRuntimeError>;
+
+pub(crate) fn map_join_error(e: tokio::task::JoinError) -> XetRuntimeError {
+    if e.is_panic() {
+        // The task panic'd.  Pass this exception on.
+        error!("Panic reported on xet worker task: {e:?}");
+        XetRuntimeError::TaskPanic(format!("{e:?}"))
+    } else if e.is_cancelled() {
+        // Likely caused by the runtime shutting down (e.g. with a keyboard CTRL-C).
+        XetRuntimeError::TaskCanceled(format!("{e}"))
+    } else {
+        XetRuntimeError::OtherJoinError(e)
+    }
+}
--- a/xet_runtime/src/primatives.rs
+++ b/xet_runtime/src/primatives.rs
@@ -49,16 +49,18 @@ impl<T: Send + Sync> Future for ComputeJoinHandle<T> {
 pub struct AsyncJoinHandle<T: Send + Sync> {
     // Just wrap the inner struct and translates any errors to XetRuntimeErrors.
     inner: tokio::task::JoinHandle<T>,
+    runtime_handle: tokio::runtime::Handle,
 }
 
 impl<T: Send + Sync> AsyncJoinHandle<T> {
-    pub(crate) fn new(inner: tokio::task::JoinHandle<T>) -> Self {
-        Self { inner }
+    pub(crate) fn new(runtime_handle: tokio::runtime::Handle, inner: tokio::task::JoinHandle<T>) -> Self {
+        Self { inner, runtime_handle }
     }
 
-    /// Awaits the task, for API consistency with the ComputeJoinHandle
-    pub async fn join(self) -> Result<T> {
-        self.inner.await.map_err(map_join_error)
+    /// Join on an async task from the sync runtime.  Use join_handle.await it the
+    /// async context.  
+    pub fn join(self) -> Result<T> {
+        self.runtime_handle.block_on(self.inner).map_err(map_join_error)
     }
 }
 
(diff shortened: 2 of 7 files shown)


xet_runtime/src/errors.rs
```diff
@@ -0,0 +1,53 @@
...
+    #[error("RuntimeCancellation")]
+    RuntimeCancellation(#[from] RuntimeCancellation),
+
+    #[error("ask runtime error: {0}")]
+    Other(String),
+}
+
+pub type Result<T> = std::result::Result<T, XetRuntimeError>;
+
+pub(crate) fn map_join_error(e: tokio::task::JoinError) -> XetRuntimeError {
```
assafvayner, line 42: Should this be a `From<tokio::task::JoinError>` impl?

xet_runtime/src/primatives.rs
```diff
@@ -0,0 +1,401 @@
...
+impl<T: Send + Sync> Future for AsyncJoinHandle<T> {
+    type Output = Result<T>;
+    fn poll(mut self: std::pin::Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> std::task::Poll<Self::Output> {
+        self.inner.poll_unpin(cx).map_err(map_join_error)
+    }
+}
+
+// Join Sets
+
+struct _ComputeJoinSetSharedData<T: Send + Sync + 'static> {
```
assafvayner: renname to `ComputeJoinSetSharedDataInner or just `ComputeJoinSetSharedData`? It's unexported anyway, and I'm not sure the `_` is very rusty

xet_runtime/src/primatives.rs
```diff
@@ -0,0 +1,401 @@
...
+    /// Returns the number of tasks currently running or completed with results waiting.
+    pub fn len(&self) -> usize {
+        self.shared_data.incoming_result_count.load(Ordering::Acquire)
+    }
+}
+
+/// An async wrapper class around the tokio JoinSet that provides two additional things:
+/// - tasks are spawned explicitly on the runtime handle.
+/// - This will allow us to swap out tokio perhaps down the road if we want to...
+pub struct AsyncJoinSet<T> {
```
assafvayner, line 429: I think we need a comment somewhere explaining why there are 2 kinds of join sets and the scenario of which case to use either.

xet_runtime/src/runtime_executor.rs
```diff
@@ -0,0 +1,554 @@
...
+
+/// Call this function within long-running loops or tasks to check whether a cancellation request
+/// (e.g. a CTRL-C press) has happened.  If so, it returns an error that can be handled outside.  
+#[inline]
+pub fn check_runtime_cancellation() -> std::result::Result<(), RuntimeCancellation> {
+    xet_runtime().check_for_cancellation()
+}
+
+#[derive(Debug)]
+pub struct XetRuntime {
```
assafvayner, line 123: I think this will need a big doc comment explaining everything.

xet_runtime/src/runtime_executor.rs
```diff
@@ -0,0 +1,554 @@
...
+    //
+    async_runtime: tokio::runtime::Runtime,
```
assafvayner, line 125: can we in any way make this not a hard requirement on tokio here? Fine for v1, but we will at some point want to see if we can use a non-tokio runtime.

xet_runtime/src/runtime_executor.rs
```diff
@@ -0,0 +1,554 @@
...
+
+    // The number of external threads calling into this threadpool
+    external_executor_count: AtomicUsize,
+
+    // Are we in the middle of a sigint shutdown?
+    cancelation_requested: AtomicBool,
```
assafvayner: missing an l in cancellation

xet_runtime/src/runtime_executor.rs
```diff
@@ -0,0 +1,554 @@
...
+        self.check_for_cancellation()?;
+
+        let (jh, tx) = ComputeJoinHandle::create();
+
+        self.compute_threadpool.spawn(move || {
+            let result = task();
+            let _ = tx.send(result).map_err(|e| {
+                debug!("Return result on join handle encountered error: {e:?}");
+                e
+            });
```
assafvayner: this map_err should be a `debug_error` from error printer maybe

xet_runtime/src/primatives.rs
hoytak: One is for async tasks and the other is for multithreaded synchronous tasks.  

Also, these are never created explicitly; they are returned by spawn_compute_task and spawn_async_task, these are the only way for the user to create these objects.

xet_runtime/src/runtime_executor.rs
hoytak: The main idea here is that all uses of tokio can be pushed behind this runtime API.  So making tokio not a hard requirement in our code is definitely the goal.   Once we build around this class, we can swap out the tokio runtime inside this class without having to change other parts of the code.

xet_runtime/src/errors.rs
hoytak: The derive(Error) proc macro already creates an implementation of that so we have to be explicit here.

xet_runtime/src/primatives.rs
hoytak: Updated documentation.

xet_runtime/src/runtime_executor.rs
```diff
@@ -0,0 +1,754 @@
...
+        let async_runtime = tokio::runtime::Builder::new_multi_thread()
+            .worker_threads(ASYNC_THREADPOOL_NUM_WORKER_THREADS) // 4 active threads
+            .thread_name_fn(get_async_thread_name) // thread names will be hf-xet-0, hf-xet-1, etc.
+            .thread_stack_size(THREADPOOL_STACK_SIZE) // 8MB stack size, default is 2MB
+            .enable_all() // enable all features, including IO/Timer/Signal/Reactor
+            .build()
+            .map_err(XetRuntimeError::AsyncRuntimeInitializationError)?;
+
+        let compute_threadpool = rayon::ThreadPoolBuilder::new()
+            .num_threads(COMPUTE_THREADPOOL_NUM_WORKER_THREADS.min(num_cpus::get() - 2).max(1)) // Use the number of CPUs by default
```
seanses, line 150: Subtraction may panic here: usize  - 2

hoytak: Good catch, though in release mode it would underflow...

ylow:
How much of this class is actually of use today? Like all the compute_task stuff. Will they actually be used by any of the rest of the xet-core?

hoytak:
After some discussion on this and getting some good reviews and feedback, I'm closing this PR and instead separating out the specific components needed to achieve the same goals.

PR #134, opened 2025-01-07
Drop xet_error.

hoytak:
xet_error was a source drop of thiserror with integrated logging and the ability to put in a breakpoint to note exception propagation.   However, the benefits of maintaining a source drop with changes, at least on client side, are less clear in the current ecosystem and with upgrades to the thiserror package (e.g. better integrated stacktracing).  

This PR removes xet_error and reverts everything back to the original thiserror implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,6 @@ Xet-core is the repo responsible for Rust-based code running on the client machi
 * [parutils](src/parutils): Provides parallel execution utilities relying on Tokio (ex. parallel foreach).
 * [progress_reporting](src/progress_reporting): offers ReportedWriter so progress for Writer operations can be displayed.
 * [utils](src/utils): general utilities - **unclear how much is currently in use**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,7 +21,7 @@ Xet-core is the repo responsible for Rust-based code running on the client machi
 * [parutils](src/parutils): Provides parallel execution utilities relying on Tokio (ex. parallel foreach).
 * [progress_reporting](src/progress_reporting): offers ReportedWriter so progress for Writer operations can be displayed.
 * [utils](src/utils): general utilities - **unclear how much is currently in use**
-* [xet_error](src/xet_error): Error utility crate, widely used for anyhow! logging in other crates.
+* [thiserror](src/thiserror): Error utility crate, widely used for anyhow! logging in other crates.
```
assafvayner: we should get rid of this line

PR #135, opened 2025-01-10
CI & Release GH Action updates

rajatarya:
- To support sha2 builds on windows needed to change dependency for sha2 crate to not use asm feature on Windows.


rajatarya:
Verified all the builds are working now.

jgodlew (approved):
:shipit:

PR #140, opened 2025-01-13
Deadlock fix.

hoytak:
Fix for issue where: 
- multiple python threads call into download_files.  
- Thread 1 begins to initializes the runtime, including logging. 
- Thread 2 acquires the GIL but then blocks on the init lock on the runtime. 
- Thread 1 hits the logging section, which requires the GIL, and is blocked. 

Fix is to: 
- Initialize logging and runtime on module load.  
- Delay initializing logging until after runtime write lock is released for the case where CTRL-C forces the runtime to be re-initialized.

### DIFF
--- a/hf_xet/Cargo.lock
+++ b/hf_xet/Cargo.lock
@@ -225,11 +225,11 @@ dependencies = [
  "reqwest-middleware",
  "reqwest-retry",
  "tempfile",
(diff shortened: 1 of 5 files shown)


ylow (approved):

jgodlew (approved):
:shipit:

PR #143, opened 2025-01-16
Profiler 

hoytak:
This PR adds in an easy way to profile xet-core operations.   Internally, the code uses pprof-rs, which links in the GPerfTools suite. 

To use, compile with --features=profiling; with maturin, install with `maturin develop --release --features=profiling`

When running, it saves the profile to profilers/<timestamp>, writing both a gperf protobuf file and a flamegraph svg file.   The protobuf file can be read using the pprof tool: 

```bash
go install github.com/google/pprof@latest
# Replace this with the correct path to the hf_xet library file.
~/go/bin/pprof  ./venv/lib/python3.10/site-packages/hf_xet/hf_xet.abi3.so  profiles/<timestamp>/pprof.pb
```

On OSX, I like using qcachegrind as the interactive visualizer of the results, which requires the profile data to be exported to the callgrind format: 

```bash
brew install qcachegrind     # If needed
~/go/bin/pprof  -callgrind ./venv/lib/python3.10/site-packages/hf_xet/hf_xet.abi3.so  profiles/<timestamp>/pprof.pb
qcachegrind <reported callgrind output file> 
```


hf_xet/Cargo.toml
```diff
@@ -32,6 +32,14 @@ reqwest = "0.11.27"
 serde = { version = "1.0.215", features = ["derive"] }
 serde_json = "1.0.133"
 lazy_static = "1.5"
+once_cell = "1.17"
```
assafvayner: I don't see once_cell used, remove?

hf_xet/src/profiling.rs
```diff
@@ -0,0 +1,100 @@
+use std::fs;
+use std::path::PathBuf;
+
+use chrono::Local;
+use pprof::protos::Message;
+use pprof::{ProfilerGuard, ProfilerGuardBuilder};
+
+/// A simple pprof-rs integration assuming only one function type is active at once.
```
assafvayner: hf hub calls download_files many times potentially for a directory download and it does so concurrently, but given this comment I assume this is fine right?

hf_xet/src/profiling.rs
```diff
@@ -0,0 +1,100 @@
...
+    fn save_report(&self) {
+        let Some(guard) = &self.guard else {
+            return;
+        };
+
+        // Build the report from the guard
+        let report = match guard.report().build() {
+            Ok(r) => r,
+            Err(e) => {
+                eprintln!("Error building profiler report: {e:?}");
```
assafvayner: From my understanding any place where we may log errors happens after the log subscriber is initialized (in runtime::init_threadpool). Can we (or is there a good reason not to) replace all these eprintln's with log_error? ErrorPrinter trait is available, it will clean up some code later, fairly straight forward in all the inspect_err cases.

this snippet will change to `let Ok(report) = guard.report().build().log_error(...) else { return; };`

Mostly regarding code consistency with the rest of the codebase, functionally the only difference is we would get different log formatting

hf_xet/src/profiling.rs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            if let Ok(profile) = report
+                .pprof()
+                .inspect_err(|e| eprintln!("Error creating pprof profile report: {e:?}"))
+            {
+                let _ = profile.write_to_writer(&mut pb_file).inspect_err(|e| {
+                    eprintln!("Failed writing protobuf perf out file: {e:?}");
+                });
+            }
+        }
```
assafvayner: This looks good, I don't have an opinion on which to choose, but we can also replace with an `and_then` pattern, I've seen both around xet-core.

feel free to resolve immediately w/o an answer if you prefer the current version.

assafvayner (approved):
Left some comments, mostly minor nits. Thanks a lot for this, I would like to use it as well.

hf_xet/Cargo.toml
hoytak: Good catch.

hf_xet/src/profiling.rs
hoytak: I should update the comment.  I had a previous version that separated out the profiles by either download or upload, but that had other issues.  Will clean up the comment there.

hf_xet/src/profiling.rs
hoytak: The result error types are different so there's a lot of extra code to use the and_then pattern, which is why I did it this way.

hf_xet/src/profiling.rs
hoytak: I thought about that, but decided to go with eprintln for all the profiling stuff as the way we would use this code would be with the user sitting down and compiling and running manually.  In this case, eprintln would be better than using log messages, which may be filtered out or json formatted or dumped to a log file.   Also, it's not about the functionality or operation of the code.

PR #147, opened 2025-01-23
Fragmentation Prevention

ylow:
For https://linear.app/xet/issue/XET-246/fragmentation-prevention We use average chunks / range as a fragmentation estimator, targetting an average of 16 chunks per range which roughly equates to 1MB per range. This is computed over the last window of 32 ranges. If the average drops below the target, dedupe is disabled until the average is above the target again.

Running on first 1GB of a *highly* fragmented file (comprising of a few hundred KB of an existing file, followed by a hundred KB of zeros, repeat) we see the following:
 - Baseline: 1000000001 bytes -> 726845953 bytes, 2975 ranges, 336134 average bytes per range
 - 512KB target (anti-fragmentation goal of 8 chunk per range): 1000000001 bytes -> 873515521 bytes, 1465 ranges, 682594 average bytes per range
 - 1MB target (anti-fragmentation goal of 16 chunks per range): 1000000001 bytes -> 932235777 bytes, 829 ranges, 1206273 average bytes per range

This also includes a hysteresis implementation:
 - 512KB target (anti-fragmentation goal of 8 chunk per range): 1000000001 bytes -> 873515521 bytes, 1657 ranges, 603500 average bytes per range.

The hysteresis turned out to be pretty important for deduping a content defined chunked variant of Parquet:
Without hysteresis (only concern is how v2 dedupes against v1):
``` 
parquet file v1: 5728317968 bytes -> 5728137283 bytes
parquet file v2: 5726717793 bytes -> 4544391399 bytes (11.14 chunks per range)
```
With hysteresis
``` 
parquet file v1: 5728317968 bytes -> 5728137283 bytes
parquet file v2: 5726717793 bytes -> 3568275084 bytes (8.11 chunks per range)
```
So with the hysteresis implementation we are closer to the target chunk per range and we are able to still dedupe pretty well. As comparison, *without* any fragmentation prevention:
```
parquet file v1: 5728317968 bytes -> 5728137283 bytes
parquet file v2: 5726717793 bytes -> 3402767500 bytes (6.89 chunks per segment)
```

### DIFF
--- a/data/src/clean.rs
+++ b/data/src/clean.rs
@@ -550,7 +550,7 @@ impl Cleaner {
                 // This is new data.
                 let add_new_data;
 
-                if tracking_info.current_cas_block_hashes.get(&chunk.hash).is_some() && !forced_nodedupe {
+                if tracking_info.current_cas_block_hashes.contains_key(&chunk.hash) && !forced_nodedupe {
                     let idx = tracking_info.current_cas_block_hashes.get(&chunk.hash).unwrap();
                     let idx = *idx;
                     // This chunk will get the CAS hash updated when the local CAS block
--- a/data/src/clean.rs
+++ b/data/src/clean.rs
@@ -497,6 +497,7 @@ impl Cleaner {
 
             // check the fragmentation state and if it is pretty fragmented
             // we skip dedupe
+            let mut forced_nodedupe = false;
             if let Some((n_deduped, _)) = dedupe_query {
                 if let Some(chunks_per_range) = tracking_info.rolling_chunks_per_range() {
                     if chunks_per_range < MIN_N_CHUNKS_PER_RANGE {
@@ -507,6 +508,7 @@ impl Cleaner {
                         // the chunks per range and so we skip it.
                         if (n_deduped as f32) < chunks_per_range {
                             dedupe_query = None;
+                            forced_nodedupe = true;
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


ylow:
Other ideas which may or may not improve things are:
 - have a high-water and low-water mark so if we are near the 1MB boundary we don't keep jumping back and forth between dedupe and no-dedupe.
 - look ahead a bunch of chunks (this is somewhat complicated)

ylow:
Having high and low watermark doesn't seem to do much and hurts readability.

data/src/clean.rs
```diff
@@ -462,8 +497,41 @@ impl Cleaner {
...
+                        // once I start deduping again, we lower CPR
+                        // to the low threshold so we allow for more small
+                        // fragments.
+                        tracking_info.defrag_at_low_threshold = true;
+                    }
+                }
+            }
 
-            if let Some((n_deduped, fse)) = deduped_blocks[cur_idx].take() {
+            if let Some((n_deduped, fse)) = dedupe_query {
```
hoytak, line 542: FYI: in the current configuration, n_deduped is always one due to https://github.com/huggingface/xet-core/blob/ebb9ff517f1d0caeca6ad9277ce56398302062fb/data/src/clean.rs#L44  (I discovered that just a bit ago).

ylow: Yeah, that's fine. If n_deduped is bigger it provides some lookahead. But even if not it will still work as designed --> turn off dedupe when we are too fragmented.

data/src/constants.rs
```diff
@@ -42,3 +42,15 @@ pub const CURRENT_VERSION: &str = env!("CARGO_PKG_VERSION");
...
 /// which stores File Hash -> reconstruction instructions
 pub const FILE_RECONSTRUCTION_CACHE_SIZE: usize = 65536;
+
+/// Number of ranges to use when estimating fragmentation
+pub const NRANGES_IN_STREAMING_FRAGMENTATION_ESTIMATOR: usize = 128;
+
+/// Minimum number of chunks per range. Used to control fragmentation
+/// This targets an average of 1MB per range.
+/// The hysteresis factor multiplied by the target Chunks Per Range (CPR) controls
+/// the low end of of the hysteresis range. Basically, dedupe will stop
```
seanses: Typo: "of of"

seanses (approved):
LGTM!

PR #148, opened 2025-01-23
Reduce memory usage

seanses:
Client memory footprint on file upload has been high due to several locations of inefficiency. This fixes some of those:
- Reduce disk read queue size.
- Limit upload concurrency to hardware parallelism (# of cores).
- Drop large buffer preemptively before "slow" network transmission.

Context:
Before the above changes the theoretical memory usage is:

405850025-bdfc5fe0-f891-4fba-9b39-44be6aa3b6f6

This PR reduces the memory usage to the below without compromising speed (benchmarked with same settings as in https://www.notion.so/huggingface2/Xorb-upload-speed-1531384ebcac8012bf74fa08ed822f67):

405850334-3de3648b-b30d-41a9-a0f7-9088ee6f7644

There is still ~650 MiB extra memory footprint demanding further investigation. The 512 MiB global Xorb uploader memory footprint can also be significantly reduced, but that requires more work in the future.

### DIFF
--- a/cas_client/src/remote_client.rs
+++ b/cas_client/src/remote_client.rs
@@ -89,7 +89,7 @@ impl UploadClient for RemoteClient {
             hash: *hash,
         };
 
-        let was_uploaded = self.upload(&key, &data, chunk_and_boundaries).await?;
+        let was_uploaded = self.upload(&key, data, chunk_and_boundaries).await?;
 
         if !was_uploaded {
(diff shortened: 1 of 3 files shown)


port8080 (approved):
Nice! :shipit:

code looks good, the tables in the PR are hard to read in 'dark mode'

seanses:
Edited the images background color. Now good for "dark mode".

PR #149, opened 2025-01-23
Remove small file passthrough

ylow:
Removing all the old code that passes through small text file through to git. Since we are not using that anymore.


data/src/clean.rs
```diff
@@ -213,14 +206,8 @@ impl Cleaner {
         let file_size = self.metrics.file_size.load(Ordering::Relaxed);
 
         // File is small, all data kept in the small file buffer.
```
seanses: rm this comment

seanses (approved):
LGTM!

PR #159, opened 2025-01-30
Caching system for shard file manager

hoytak:
This PR enables a caching system for the shard file manager by: 
- shard file manager is associated with a single directory.  All registered shards must be in that directory.  
- Within-process caching is optionally done by directory.   
- The shard file manager can optionally reference a secondary manager for chunk and file lookup.  In our case, the main cache directory is cached and global, but each session directory has its own shard file manager where lookup is attempted first, falling back to the global cache directory manager if that fails.

### DIFF
--- a/data/src/shard_interface.rs
+++ b/data/src/shard_interface.rs
@@ -26,14 +26,14 @@ pub async fn create_shard_manager(
     let cache_shard_manager = ShardFileManager::builder(shard_cache_directory)
         .with_chunk_dedup(!download_only_mode)
         .with_expired_shard_cleanup(true)
-        .with_caching(true)
+        .from_global_manager_cache(true)
         .build()
         .await?;
 
     let session_shard_manager = ShardFileManager::builder(shard_session_directory)
         .with_chunk_dedup(!download_only_mode)
         .with_expired_shard_cleanup(false)
-        .with_caching(false)
+        .from_global_manager_cache(false)
         .with_upstream_manager(cache_shard_manager)
         .build()
         .await?;
--- a/data/src/shard_interface.rs
+++ b/data/src/shard_interface.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use cas_client::{HttpShardClient, LocalShardClient, ShardClientInterface};
 use mdb_shard::ShardFileManager;
-use tracing::{debug, warn};
+use tracing::debug;
 
 use super::configurations::Endpoint::*;
 use super::configurations::StorageConfig;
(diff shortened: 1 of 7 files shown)


mdb_shard/src/shard_file_handle.rs
```diff
@@ -128,37 +129,69 @@ impl MDBShardFile {
         }
     }
 
-    pub fn load_all(path: &Path) -> Result<Vec<Arc<Self>>> {
+    pub fn load_all_valid(path: impl AsRef<Path>) -> Result<Vec<Arc<Self>>> {
+        Self::load_all(path, false)
+    }
+
+    pub fn load_all(path: impl AsRef<Path>, load_expired: bool) -> Result<Vec<Arc<Self>>> {
```
seanses, line 136: Is `load_all(.., true)` used anywhere? If not we don't need to make this function `pub`.

mdb_shard/src/shard_file_manager.rs
```diff
@@ -103,25 +111,39 @@ impl SFMBuildParameters {
...
-    pub fn with_expired_shard_cleanup(mut self, cleanup: bool) -> Self {
+    /// Makes a fallback manager that unresolved queries are queried against.  
+    ///
+    /// To be used to combine a session shard manager with a cache shard manager.
+    pub fn with_upstream_manager(mut self, upstream_manager: Arc<ShardFileManager>) -> Self {
+        self.upstream_manager = Some(upstream_manager);
+        self
+    }
+
+    pub fn with_custom_expired_shard_cleanup(mut self, cleanup: bool, deletion_buffer_window: u64) -> Self {
```
seanses: I think `deletion_buffer_window` is ignored if `cleanup` is false. That makes this function a bit confusing. Maybe supplying a custom "deletion_buffer_window" means `cleanup` is true.

mdb_shard/src/shard_file_manager.rs
```diff
@@ -103,25 +111,39 @@ impl SFMBuildParameters {
...
-        Ok(ShardFileManager::new_from_builder(self).await?.0)
+        ShardFileManager::new_from_builder(self).await
     }
 }
 
 pub struct ShardFileManager {
     shard_bookkeeper: RwLock<ShardBookkeeper>,
     current_state: RwLock<MDBInMemoryShard>,
 
+    upstream_manager: Option<Arc<ShardFileManager>>,
```
seanses, line 147: Is there a plan to separate shards and group those with the same expiry into a common subdirectory? If so I wonder how this is going to evolve. It seems inefficient to have multiple upstream managers.

data/src/shard_interface.rs
```diff
@@ -16,21 +16,29 @@ pub async fn create_shard_manager(
...
-        shard_manager.register_shards_by_path(&[shard_cache_directory]).await?;
-    } else {
-        warn!("Merkle DB Cache path {:?} does not exist, skipping registration.", shard_cache_directory);
-    }
-
-    Ok(shard_manager)
+    let cache_shard_manager = ShardFileManager::builder(shard_cache_directory)
+        .with_chunk_dedup(!download_only_mode)
+        .with_expired_shard_cleanup(true)
+        .with_caching(true)
```
seanses: At first glance I thought this `with_caching` is to designate whether this holds "cached" shards. Now that it means whether to add this to the global shard manager cache, do we need to make the distinction here? I don't think it's problematic to do for both managers?

mdb_shard/src/shard_file_manager.rs
```diff
@@ -103,25 +111,39 @@ impl SFMBuildParameters {
...
 
     pub async fn build(self) -> Result<Arc<ShardFileManager>> {
-        Ok(ShardFileManager::new_from_builder(self).await?.0)
+        ShardFileManager::new_from_builder(self).await
     }
 }
 
 pub struct ShardFileManager {
     shard_bookkeeper: RwLock<ShardBookkeeper>,
     current_state: RwLock<MDBInMemoryShard>,
```
seanses, line 145: So now both cache_shard_manager and session_shard_manager holds a current_state, this structure is bit unintuitive. Would it be better if there is a read-only "ShardFileManager" struct and another with add-on write ability.

mdb_shard/src/shard_file_manager.rs
hoytak: Currently, this recursively loads all the shards under a directory too, so that's still compatible... 

I think there are multiple ways of handling this, including just writing out a flat index file that gets deserialized, so I'm not too worried about it now....

data/src/shard_interface.rs
hoytak: Good call! will rename.

mdb_shard/src/shard_file_manager.rs
hoytak: Yes, I could agree that separating out the functionality would make it easier, but that seemed too invasive of a change...  This seemed like a small switch.

mdb_shard/src/shard_file_handle.rs
hoytak: I think it was used for testing, but then changed that.

mdb_shard/src/shard_file_manager.rs
seanses: Fair

seanses (approved):
LGTM!

PR #161, opened 2025-01-31
Avoid uploading empty xorbs.

hoytak:
Avoids uploading empty xorbs.

### DIFF
--- a/data/src/parallel_xorb_uploader.rs
+++ b/data/src/parallel_xorb_uploader.rs
@@ -89,51 +89,59 @@ impl XorbUpload for ParallelXorbUploader {
 
         // Only upload a new xorb if there is new data; it may be that an existing new file is formed only
         // from existing chunks.
-        let xorb_data_len = cas_data.data.len();
-
-        let cas_hash = cas_node_hash(&cas_data.chunks[..]);
-
-        // Rate limiting, the acquired permit is dropped after the task completes.
-        // The chosen Semaphore is fair, meaning xorbs added first will be scheduled to upload first.
-        let permit = self
-            .rate_limiter
-            .clone()
-            .acquire_owned()
-            .await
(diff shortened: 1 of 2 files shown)


data/src/parallel_xorb_uploader.rs
```diff
@@ -89,51 +89,59 @@ impl XorbUpload for ParallelXorbUploader {
...
-
-        let mut upload_tasks = self.upload_tasks.lock().await;
-        let upload_progress_updater = self.upload_progress_updater.clone();
-        upload_tasks.spawn_on(
-            async move {
-                let ret = upload_and_register_xorb(item, shard_manager, cas, cas_prefix).await;
-                if ret.is_ok() {
-                    if let Some(updater) = upload_progress_updater {
-                        updater.update(xorb_data_len as u64);
+        if cas_data.data.is_empty() {
```
seanses, line 92: Is this the real cause? I examine all references and we always check if `cas_data.data` is empty before calling into this function.

seanses: But this indeed helps me identify a bug: [here](https://github.com/huggingface/xet-core/blob/5cf29c1be2a3092fb81e9c4d87880b20228797a9/data/src/data_processing.rs#L199) we check `new_cas_data.data.is_empty()` but I don't think we should do that because `new_cas_data` may have non-empty `pending_file_info` but empty `data`?

hoytak: The is_empty() check returns false if there is pending file info but empty CAS data, so that is correct with this PR.

hoytak: UGH, this bug was introduced accidentally with b015f31c; before it tested new_cas_data.is_empty(), so it would upload empty xorbs but not fail to register new files made up of mixed but existing chunks.

seanses (approved):
LGTM!

PR #163, opened 2025-02-04
Up the chunk deduplication size.

hoytak:
Changed the number of chunks processed at once to 256 (increased from 1).


seanses (approved):
LGTM!

PR #165, opened 2025-02-04
add clean_file_from_url

assafvayner:
Adds a utility function to clean a file given a url to use to "GET" the file, i.e. an S3 presigned url.

The contents are then read incrementally and passed into the cleaner handle to chunk, dedup and upload.

Intended potential use for streaming files to migrate them.

### DIFF
--- a/data/src/data_client.rs
+++ b/data/src/data_client.rs
@@ -6,7 +6,7 @@ use std::num::NonZero;
 use std::path::PathBuf;
 use std::sync::Arc;
 
-use cas_client::CacheConfig;
+use cas_client::{build_http_client, CacheConfig};
 use dirs::home_dir;
 use futures::{AsyncReadExt, TryStreamExt};
 use lazy_static::lazy_static;
@@ -209,7 +209,9 @@ async fn clean_file_from_url(
     let path = PathBuf::from(path);
     let mut read_buf = vec![0u8; READ_BLOCK_SIZE];
 
-    let response = reqwest::get(file_url).await?;
+    let client = build_http_client(&None)?;
+
+    let response = client.get(file_url).await?;
 
     let mut reader = response.bytes_stream().map_err(std::io::Error::other).into_async_read();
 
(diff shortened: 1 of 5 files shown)


data/src/data_client.rs
```diff
@@ -200,6 +201,39 @@ async fn clean_file(processor: &PointerFileTranslator, f: String) -> errors::Res
...
 
+async fn clean_file_from_url(
+    processor: &PointerFileTranslator,
+    file_url: String,
+    path: String,
+) -> errors::Result<PointerFile> {
+    let path = PathBuf::from(path);
+    let mut read_buf = vec![0u8; READ_BLOCK_SIZE];
+
+    let response = reqwest::get(file_url).await?;
```
seanses: Our [build_http_client](https://github.com/huggingface/xet-core/blob/0bc9a42bbb0c35cdfa44c18e6d610987cf9e5eae/cas_client/src/http_client.rs#L63) can build a client with retry capability.

data/src/data_client.rs
```diff
@@ -200,6 +201,39 @@ async fn clean_file(processor: &PointerFileTranslator, f: String) -> errors::Res
...
+
+    let handle = processor
+        .start_clean(
+            IDEAL_CAS_BLOCK_SIZE / READ_BLOCK_SIZE, // enough to fill one CAS block
+            Some(&path),                            // for logging & telemetry
+        )
+        .await?;
+
+    loop {
+        let bytes = reader.read(&mut read_buf).await?;
```
seanses, line 226: But I'm not sure how does the RetryMiddleware works with stream reading. If an error occurs within this `read()` does the RetryMiddleware retries this read or the entire GET? If it retries the entire GET what does the next successful read return? If it returns bytes from the beginning of the file then that's problematic.

assafvayner: I believe at this point it will not retry, though I also think we should not retry plainly at that moment since if we could have submitted some content to the cleaner already. The behavior for an error from the stream (e.g. in the highlighted line) should be the same behavior as if we had a file read return an error, which I think we just bubble it up in this case.

seanses: Let's test to confirm the behavior of RetryMiddleware interacting with error reading from a stream (the error may be internally handled by retry and not propagated to this level).

assafvayner:
closing in favor of #288

PR #168, opened 2025-02-07
Include s in repo type arg of xtool

sirahd:
Use the plural as default for `repo_type` arg for the xtool so that it is consistent with how the caller (repo scanner) parses the repo type as plural


seanses (commented):
But huggingface_hub uses the singular form. I'd prefer we stay consistent with that.

PR #169, opened 2025-02-10
singleflight remove explicit threadpool

assafvayner:
Removing threadpool references in the singleflight crate. The crate uses Handle::current().spawn() to spawn tasks. If a `work` call is issued within a runtime within a xet_threadpool::ThreadPool runtime it will use that tokio runtime handle to spawn tasks.


hoytak (approved):
LGTM!

PR #170, opened 2025-02-14
Fix calling async from sync inside async runtime

seanses:
The "external_run_async_task" function, as documented in the comment, should only be used by threads outside of tokio, otherwise it can panic if `block_on` is called on a thread that is to drive async tasks. The `HubClientTokenRefresher::refresh` function, however, needs to call an async fn inside of tokio. This PR implements the correct routine to do that.


seanses:
cc @sirahd We should get this into the repo scanner.

sirahd (approved):
LGTM! I'll send a PR to repo-scanner once this is merged.

PR #173, opened 2025-02-14
Adding request_id to logs from CAS requests

jgodlew:
* Add debug log to http client whenever we receive a response from CAS, logging the request_id and status code.
* Add request_id to error logs when a response from CAS fails
* Refactor logic to process http response result into a helper trait/fn for consistency across CAS apis.


jgodlew:
Example error message:
```
{"timestamp":"2025-02-14T20:08:00.057728Z","level":"ERROR","fields":{"message":"upload_xorb api failed: request id: 01JM316G9MD5PVP61YX80YZHRK, error: reqwest::Error { kind: Status(400), url: \"http://localhost:4884/xorb/default/73287bdf1a5ff5e17d61cee9302adc33a38b10d2344ba65443e50a3ef4db5148\" }","caller":"/Users/jgodlew/git/huggingface/xet-core/cas_client/src/http_client.rs:198"},"filename":"/Users/jgodlew/git/huggingface/xet-core/error_printer/src/lib.rs","line_number":28}
```

cas_client/src/http_client.rs
```diff
@@ -114,20 +115,21 @@ impl Middleware for LoggingMiddleware {
...
+            .await
+            .inspect(|res| {
+                let status_code = res.status().as_u16();
+                let request_id = request_id_from_response(res);
+                debug!(request_id, status_code, "Received CAS response");
+                if Some(Retryable::Transient) == default_on_request_success(res) {
+                    warn!(request_id, "Status Code: {status_code:?}. Retrying...");
+                }
+            })
+            .inspect_err(|err| {
```
assafvayner: personal preference combine the 2 inspect_errs, and also add a comment on what each section is meant to do

assafvayner (approved):
small nits, otherwise LGTM

cas_types/src/lib.rs
```diff
@@ -9,6 +9,12 @@ mod error;
 mod key;
 pub use key::*;
 
+/// Indicates a "session id" that clients can use to group together related requests
+/// (e.g. all requests made to CAS to support a user-triggered upload (xorbs + shards)).
+pub const SESSION_ID_HEADER: &str = "X-Xet-Session-Id";
+/// Request id generated by CAS for a request.
+pub const REQUEST_ID_HEADER: &str = "X-Xet-Request-Id";
```
XciD: maybe use the same as hub, and all infra `X-Request-Id` ?

seanses (approved):
LGTM! Can you also help refactor similar places in HttpShardClient?

jgodlew:
> Can you also help refactor similar places in HttpShardClient?

`HttpShardClient` should also have the refactors (uses `process_error` instead of multiple `log_errors`). The only place I didn't update was `get_dedup_shards`, as it is a "best-effort" API and we swallow unsuccessful requests.

cas_client/src/http_client.rs
jgodlew: I can't combine these as the first is an `inspect` (ok) and the second is `inspect_err`. I can switch this to a match instead if you prefer, but, IMO, I find `inspect/inspect_err` pattern to be more readable as we aren't modifying the Result.

Regardless, I can add comments on what each part does.

assafvayner: oh, I didn't notice the inspect ok and inspect err, I thought both were ok 🤦

PR #174, opened 2025-02-14
Shard query resilient to shard file deletion

seanses:
Since [add a caching layer for shard file manager](https://github.com/huggingface/xet-core/pull/159), a shard file manager along with shards registered to it are kept in memory through out the entire life time of a running process. If the shard files on disk are (manually) deleted after one operation (e.g. upload_folder in a iPython session), the next operation will fail due to file not found I/O error. This PR turns this I/O error into a soft failure such that a query into a deleted shard returns `Ok(None)`. I don't think there's need to remove them from the in memory registry for the numbers being small.

### DIFF
--- a/mdb_shard/src/error.rs
+++ b/mdb_shard/src/error.rs
@@ -50,25 +50,18 @@ impl PartialEq for MDBShardError {
     }
 }
 
-// Helper trait to swallow io::ErrorKind::NotFound errors. In the case of
-// a cache shard was registered but later deleted during the lifetime
-// of a shard file manager, dedup queries to this shard should not fail hard.
-pub trait CacheDeletionResilience<T> {
-    fn ok_for_io_error_not_found(self) -> Result<Option<T>>;
-}
-
-impl<T> CacheDeletionResilience<T> for Result<T> {
-    fn ok_for_io_error_not_found(self) -> Result<Option<T>> {
-        match self {
-            Ok(v) => Ok(Some(v)),
-            Err(MDBShardError::IOError(e)) => {
-                if e.kind() == io::ErrorKind::NotFound {
-                    Ok(None)
-                } else {
-                    Err(MDBShardError::IOError(e))
-                }
-            },
-            Err(other_err) => Err(other_err),
-        }
+// Helper function to swallow io::ErrorKind::NotFound errors. In the case of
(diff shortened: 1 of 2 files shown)


mdb_shard/src/error.rs
```diff
@@ -49,3 +49,26 @@ impl PartialEq for MDBShardError {
         }
     }
 }
+
+// Helper trait to swallow io::ErrorKind::NotFound errors. In the case of
+// a cache shard was registered but later deleted during the lifetime
+// of a shard file manager, dedup queries to this shard should not fail hard.
+pub trait CacheDeletionResilience<T> {
```
hoytak: I feel like this is a bit too specific of a case to justify introducing a whole new trait to handle just this since the query functions already return Option<T>.  Maybe we can just do this: 

```
pub fn not_found_as_none(e : E) -> Result<T, E> { 
 /// code below okay
}
```

Then just use `let res = get_reader().or_else(not_found_as_none)?`

seanses: Ah yes you're right!

mdb_shard/src/shard_file_handle.rs
```diff
@@ -234,15 +234,25 @@ impl MDBShardFile {
 
     #[inline]
     pub fn get_file_reconstruction_info(&self, file_hash: &MerkleHash) -> Result<Option<MDBFileInfo>> {
-        self.shard.get_file_reconstruction_info(&mut self.get_reader()?, file_hash)
```
hoytak, line 237: Urg.  Sorry to flipflop again....

Let's just consolidate and add in the function:

```
fn get_reader_if_present(&self) -> Result<Option<File>> {
   match self.get_reader() {
      Ok(r) => Ok(Some(r)),
      Err(io::error) => // ...
      Err(e) => Err(e)
    }
 }       
 ```
 
 And then replace all these with 
 
 ```
 let Some(mut reader) = self.get_reader_if_present()? else { return Ok(None); }; 
 ```

hoytak (approved):
One thought, take it or leave it.... 

But LGTM!

mdb_shard/src/shard_file_handle.rs
seanses: sounds good!

PR #175, opened 2025-02-18
Removed unused parts of the remote shard interface.

hoytak:
These functions are used anymore; removing for clarity.


seanses (approved):
Looks good! Can we also remove FileQueryPolicy from the configs? I don't think it's related anymore

hoytak:
> Looks good! Can we also remove FileQueryPolicy from the configs? I don't think it's related anymore

Yes, but that has some larger implications IMO so let's do that in a separate PR.

PR #177, opened 2025-02-18
only store compressed chunk if the compressed chunk is smaller

assafvayner:
Can close this PR if we don't want this.

Changes the chunk serialization functionality to only use the compression scheme if it is actually better than storing the data raw. In the case where the uncompressed data takes less space, the chunk is serialized with no compression.


cas_object/src/cas_chunk_format.rs
```diff
@@ -116,8 +117,14 @@ pub fn serialize_chunk<W: Write>(
 
     let compressed = compression_scheme.compress_from_slice(chunk)?;
     let compressed_len = compressed.len();
+    if compressed_len > uncompressed_len {
```
hoytak: pedantic, could be >=

hoytak: Could just change the base info instead of copy/pasting the writing code: 

```
let (write_data, compression_scheme) = {
if compressed.len() > chunk.len() {
   (chunk, CompressionScheme::None)
} else {
   (&compressed, compression_scheme)
}
};
```

Also, from my tests this is really common so no need to log this...

hf_xet/Cargo.lock
seanses, line 1: LGTM but I don't think this will change the lock file? May want to double check if this lock file was updated from changes in another commit

seanses (approved):

hf_xet/Cargo.lock
assafvayner: I'm pretty sure the lock file was not up to date, my IDE + other cargo commands are forcing the lock file to sync

assafvayner: actually looks like a rustc/cargo version issue.

ylow: Can you revert the lockfile?

PR #178, opened 2025-02-20
Consolidate local client testing code into local_client.rs

hoytak:
Currently, the functionality for the local CAS client testing is spread between local_shard_client.rs, global_dedup_table.rs, local_client.rs, and local_test_client.rs, with functionality spread between these and unneeded boilerplate code to interface between them. 

This PR consolidates all of these into just LocalClient, as defined in local_client.rs.


cas_client/src/local_client.rs
```diff
@@ -193,99 +304,135 @@ impl UploadClient for LocalClient {
...
+            let Some(one_range) = self
+                .get_object_range(&entry.cas_hash, vec![(entry.chunk_index_start, entry.chunk_index_end)])?
+                .pop()
+            else {
+                return Err(CasClientError::InvalidRange);
+            };
+            let start = byte_range.as_ref().map(|range| range.start as usize).unwrap_or(0);
+            let end = byte_range.as_ref().map(|range| range.end as usize).unwrap_or(one_range.len());
+
+            writer.write_all(&one_range[start..end])?;
```
seanses: BUG: `byte_range` is defined on the entire file not on a single segment

hoytak: Hmm.  This part of the code wasn't changed at all; it was moved over as is.  But I can definitely fix that.

hoytak: Fixed.

seanses (approved):

PR #179, opened 2025-02-24
Use KL divergence on bg4 groups to choose compression scheme 

hoytak:
This PR switches the determination of whether or not to use BG4 grouping for compression from manual (by repository type) to using the KL divergence of the bit counts of bytes across each of the 4 elements in the bytegroups against the distribution of the combined bitcounts. 

Specifically, we build 9-element histograms of the number of 1-bits in the u8 elements of each of the 4 slices that we would rearrange under byte grouping.   Then, we measure the maximum KL-divergence between the combined histogram of these counts and each of the 4 histograms of counts in the slices.   The intuition here is that this is a first-order information-theoretic approximation of how different the slice distributions are from the base distribution.   Then, if max_kl >= 0.02, we do byte grouping, otherwise not.  This value was chosen using logistic regression, and the code to reproduce the results is included in this PR. 

I tested this out on random blocks from a bunch of different models, datasets, and other files from public hub repos, and found that this rule predicts which compression method is more efficient by around 92%.  In addition, the rate at which it gets it wrong when the method not predicted would be more than a 5% improvement is less than 0.2%:

```
Accuracy on test set: 0.917349629923716
              precision    recall  f1-score   support
           0       0.59      0.95      0.73     32850
           1       0.99      0.91      0.95    249254

Incorrectly predicted 41 / 52900 of cases where bg4_lz4_size < 0.95 * lz4_size
Incorrectly predicted 307 / 229204 of cases where bg4_lz4_size > 1.05 * lz4_size
```


sirahd (approved):
LGTM!

seanses (changes requested):
Looks like something is off: all auto selection decisions lead towards "lz4". See below for a list of benchmarks on models and datasets. (The red font column is the this using 
this predictor). For ref see https://github.com/huggingface/xet-core/pull/130#issuecomment-2618125466 and https://github.com/huggingface/xet-core/pull/130#issuecomment-2625419389

auto-compression

seanses (commented):
LGTM! But need to fix the pyfunction signature for the updated function to let hf_xet compile.

hoytak:
> LGTM! But need to fix the pyfunction signature for the updated function to let hf_xet compile.

Good catch!  I thought the CI would catch that; filed an issue to add that in the CI checks.

seanses (approved):

seanses:
Benchmark numbers here: https://docs.google.com/spreadsheets/d/1SJ8Dv3EcNTuA41JXT-ggL4SzMnqs84hn/edit?gid=1247117275#gid=1247117275

PR #180, opened 2025-02-26
Clean up global dedup query API 

hoytak:
Changed the trait ShardDedupProber to return a PathBuf instead of just the shard hash for the global dedup, as the shard hash was reformed to get the path but with unclear location.  Also, as this is used only in the cas_client, moved the trait there. 

Furthermore, corrected an issue with the local testing client and added test coverage for that. 

This fixes the issue where global dedup shards were downloaded to the cache but attempted to be registered by the session shard manager.


cas_client/src/http_shard_client.rs
```diff
@@ -152,40 +148,35 @@ impl ShardDedupProber<CasClientError> for HttpShardClient {
...
     use mdb_shard::{MDBShardFile, MDBShardInfo};
     use merklehash::MerkleHash;
 
     use super::HttpShardClient;
+    use crate::interface::ShardDedupProber;
     use crate::RegistrationClient;
 
     #[tokio::test]
     #[ignore = "need a local cas_server running"]
     async fn test_local() -> anyhow::Result<()> {
```
seanses: We can drop this manual test entirely because the CAS server implementation has changed and no longer follows the logic defined here, can be misleading.

data/src/file_cleaner.rs
```diff
@@ -426,24 +426,20 @@ impl SingleFileCleaner {
...
+                                    else { return Ok(false); };
 
-                                let Some(shard_hash) = query_result else {
+                                let Some(new_shard_file) = query_result else {
                                     debug!("Queried shard for global dedup with hash {query_chunk:?}; nothing found.");
-                                    return false;
+                                    return Ok(false);
                                 };
 
                                 // Okay, we have something, so go ahead and download it in the background.
```
seanses: Update this comment

data/src/remote_shard_interface.rs
```diff
@@ -131,42 +131,19 @@ impl RemoteShardInterface {
...
-    pub async fn register_local_shard(&self, shard_hash: &MerkleHash) -> Result<()> {
-        let shard_manager = self.shard_manager()?;
-        let cache_dir = self.shard_cache_directory()?;
-
-        if shard_manager.shard_is_registered(shard_hash).await {
-            info!("register_local_shard: shard {shard_hash:?} already registered, ignore.");
-            return Ok(());
+    ) -> Result<Option<PathBuf>> {
+        if let Some(shard_client) = self.shard_client.as_ref() {
+            debug!("get_dedup_shards: querying for shards with chunk {:?}", chunk_hash[0]);
```
seanses: Probably want to print the entire chunk_hash instead of chunk_hash[0]

seanses (approved):
LGTM with minor suggestions.

data/src/remote_shard_interface.rs
hoytak: Yes, used to be an array; now just a single chunk.  nice.

PR #182, opened 2025-02-26
Xorb Format Upgrade

hoytak:
Upgrade to support uncompressed lengths stored in the xorb footer; other rearrangements as needed.

In detail,

- Rename `CasObjectInfo` to `CasObjectInfoV0`
- Add `CasObjectInfoV1` that implements the new footer format.
- Change `CasObject` to contain the new footer. A new object will be serialized to use the new footer. When deserializing a xorb, it will check the footer version and dispatch the corresponding footer deserialization logic.
- Augments xorb validation from async stream to support three cases:  xorb with V1 footer, xorb with V0 footer, and xorb with no footer. For the last two cases, the footer is disregarded and a new V1 footer is generated using information gained from chunk list validation and returned to the caller.


cas_object/src/cas_object_format.rs
```diff
@@ -249,6 +257,507 @@ impl CasObjectInfo {
...
+
+        // Write this out here, though it's written out multiple places.  Here as it applies all over
+        // the place.
+        write_u32(w, self.num_chunks)?;
+
+        write_u32(w, self.hashes_section_offset_from_end)?;
+        write_u32(w, self.boundary_section_offset_from_end)?;
+
+        // write closing metadata
+        write_bytes(w, &self._buffer)?;
```
assafvayner: Add a comment to specify this as a section. I read this as a separate bit after the chunk lengths. Maybe call it the toes

cas_object/src/validate_xorb_stream.rs
```diff
@@ -58,104 +93,137 @@ async fn _validate_cas_object_from_async_read<R: AsyncRead + Unpin>(
...
+        cas_info.num_chunks = chunk_hash_and_size.len() as u32;
+        cas_info.fill_in_boundary_offsets();
+
+        CasObject {
+            info: cas_info,
+            info_length: 0,
+        }
+    });
+
+    Ok((cas_object, go_back_bytes))
```
assafvayner: make into a separate fn

assafvayner (approved):
Looks good. I'd like some more comments in cas_object_format.rs but also we're going to change this again to remove V0 pretty soon so not too important.

ylow (approved):
Not trivial to review :-) Since its pretty much all parsing code. But pretty much if the tests pass it is probably fine.

PR #183, opened 2025-03-03
Merge RemoteClient and HttpShardClient.

hoytak:
Since these have the same the same functionality and the remotes are the same, this simplifies a lot of the code paths.


cas_client/src/remote_client.rs
```diff
@@ -171,7 +176,7 @@ impl ReconstructionClient for RemoteClient {
         // TODO: spawn threads to reconstruct each file
         for (hash, terms) in manifest.files {
             let w = files.get_mut(&(hash.into())).unwrap();
-            self.reconstruct_file_to_writer(http_client.clone(), terms, fetch_info.clone(), 0, None, w, None)
+            self.reconstruct_file_to_writer(terms, fetch_info.clone(), 0, None, w, None)
```
seanses, line 181: The deleted "http_client" is used to issue requests to S3 using pre-signed S3 URL with no auth header. But `self.http_client` is constructed with an AuthMiddleware which injects a bearer token (granted to access the CAS server) to the request header. I think AWS S3 is going to reject this request for invalid auth. @jgodlew and @bpronan has more knowledge on this than me.

cas_client/src/interface.rs
```diff
@@ -56,28 +55,21 @@ pub trait ReconstructionClient {
     /// with the backing store (S3) to retrieve xorbs.
     async fn get_file(
         &self,
-        http_client: Arc<ClientWithMiddleware>,
```
seanses, line 59: The comment above says the "http_client" is intentionally passed in for better customization. I'm not sure if that's eventually where we will end up at but better to confirm with @rajatarya.

cas_client/src/remote_client.rs
```diff
@@ -114,7 +125,7 @@ impl UploadClient for RemoteClient {
         };
 
         let url = Url::parse(&format!("{}/xorb/{key}", self.endpoint))?;
-        let response = self.http_auth_client.head(url).send().await?;
+        let response = self.http_client.head(url).send().await?;
```
rajatarya: There were two http_clients because auth_client had additional headers added / middleware. Are those differences captured somewhere?

cas_client/src/remote_client.rs
rajatarya: This is why we had two http_client instances before. One was called http_auth_client and one was http_client -> to handle this specific case.

cas_client/src/interface.rs
rajatarya: The idea is cas_client should be a very simple crate that knows how to talk with CAS backend. It shouldn't have smarts around parallelism and concurrency. So having the `get_file` method take in a http_client parameter is one way to ensure that this crate isn't responsible for controlling its execution. It is given the implementation to use  instead.

rajatarya (commented):
My biggest concern is the reconciliation of the http_auth_client and the http_client - I do not expect these to work if combined. Because any time the comms are going to S3 directly - the AuthMiddleware cannot be included.

cas_client/src/remote_client.rs
hoytak: Okay, gotcha.  I will look a bit closer to make sure those cases are separate.

cas_client/src/interface.rs
hoytak: So in the code for data layer, what I saw is that that the http client was created and passed in to the cas client, and it was only used in exactly that way; what I did is simply move that into the class itself as it was simply an extra thing that the class using the cas client held only to pass in to these functions.  It seemed confusing to not have it work with the code that uses it.  I'm not sure how having it be a parameter to the function vs. held in the class affects concurrency, at least with how it is written now?

I think it would make sense to take this as a parameter to the constructor, which would be very similar. 

The authenticated client vs. the non-authenticated distinction definitely makes sense; I will definitely make sure that path is clear.

cas_client/src/remote_client.rs
hoytak: This was just a rename as http_auth_client sounds like a client for authentication, not an authenticated client.  I can clear up the name properly.

cas_client/src/interface.rs
hoytak: So, to clarify: 

Currently, we have http_auth_client stored in the class and built on construction, and this is built using 

```
let http_auth_client = cas_local::build_auth_http_client(auth_config, &None).unwrap();  
```

(nvm the unwrap here; that should get cleaned up).  

We then have http_client built in the file download class, and this is built using 

```
let http_client = cas_client::build_http_client(&None)?;
```

This is stored in the FileDownloader class and only passed in to the get_file function.   It's not used for anything else, and it's all built and configured in the cas_client module already. 

The main reason I see to keep it separated from the full remote client is so that the paths that don't do the download don't construct it.  Am I missing something?  

It seems to me that we should perhaps separate out the file download stuff from the full Client trait, since it's used uniquely for this part.  But otherwise, it was just confusing to me...

seanses (approved):
LGTM!

PR #192, opened 2025-03-06
Enable unpacked_chunk_offsets check

seanses:
We use `boundaries_version` to distinguish a V1 xorb from V1 view of a V0 xorb.


assafvayner (approved):
I like that we have a light-load check and it is used often, I do think it could be better enforcement i.e. ensure cumulative sums are growing for example, I'm also thinking to possibly memoize the validation status.

These are ideas that don't need implementing right now, I think this is just good for now.

hoytak (approved):
LGTM!

PR #196, opened 2025-03-07
Make Disk Cache Configurable by environment variable

ylow:
Make the chunk cache configurable through environment variable, specifically HF_XET_CACHE_SIZE_GB. Defaults to 10GB. Setting it to 0 disables the chunk cache.

### DIFF
--- a/data/src/data_client.rs
+++ b/data/src/data_client.rs
@@ -251,10 +251,10 @@ mod tests {
     fn test_cache_size_env_var() {
         // Test with different cache sizes
         let test_cases = vec![
-            (Some("5"), 5 << 30),    // 5GB
-            (Some("20"), 20 << 30),  // 20GB
-            (Some("1"), 1 << 30),    // 1GB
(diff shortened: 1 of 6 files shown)


rajatarya:
My only concern with this PR is that the unit-tests don't actually confirm that chunk cache behaves as expected when capacity set. If there was an integration test that would cover that case I would feel better.

data/src/data_client.rs
```diff
@@ -32,6 +33,14 @@ const MAX_CONCURRENT_DOWNLOADS: usize = 8; // Download is not CPU-bound
 const DEFAULT_CAS_ENDPOINT: &str = "http://localhost:8080";
 const READ_BLOCK_SIZE: usize = 1024 * 1024;
 
+fn get_configured_cache_size() -> u64 {
+    env::var("HF_XET_CACHE_SIZE_GB")
```
assafvayner, line 37: should use new configurable_constants! macro for this here, if needs to rebase off main

assafvayner:
Done in another already merged PR

PR #197, opened 2025-03-10
Restrict memory allocation size on Xorb deserialization

seanses:
Limit the initial variable length array allocation size and increment the array size as needed. This is to prevent allocating giant array based on the size declared in an uploaded xorb. This will pair up with [another PR](https://github.com/huggingface-internal/xetcas/pull/268) that limits the total read size of a reader to effectively limit total memory usage.

### DIFF
--- a/cas_object/src/cas_object_format.rs
+++ b/cas_object/src/cas_object_format.rs
@@ -5,6 +5,7 @@ use std::mem::{size_of, size_of_val};
 use anyhow::anyhow;
 use bytes::Buf;
 use futures::AsyncReadExt;
+use merkledb::constants::{IDEAL_CAS_BLOCK_SIZE, TARGET_CDC_CHUNK_SIZE};
 use merkledb::prelude::MerkleDBHighLevelMethodsV1;
 use merkledb::{Chunk, MerkleMemDB};
 use merklehash::{DataHash, MerkleHash};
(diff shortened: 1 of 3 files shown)


cas_object/src/cas_object_format.rs
```diff
@@ -163,13 +166,13 @@ impl CasObjectInfoV0 {
         read_bytes(&mut num_chunks)?;
         let num_chunks = u32::from_le_bytes(num_chunks);
 
-        let mut chunk_boundary_offsets = Vec::with_capacity(num_chunks as usize);
+        let mut chunk_boundary_offsets = Vec::with_capacity((num_chunks as usize).min(AVERAGE_NUM_CHUNKS_PER_XORB));
```
hoytak: I'm confused what we're doing here?  The next loop puts exactly num_chunks in the vector, so all the min would do is force a later reallocation.

hoytak: If this is to prevent an invalid xorb from triggering an OOM error here, wouldn't we want to test num_chunks and then reject the xorb if it's too large?

seanses: This is exactly to prevent an invalid xorb with giant num_chunks. The difficulty with the approach that you suggested is that, there's not a reasonable limit we can reference, because chunk size can be as small as it can be, e.g. created from a single file with 100 bytes. Even if there's a hard limit `64 MiB / 1 byte` num of chunks in a xorb, that's too large.

What I'm doing here is that, if `num_chunks` is giant, we don't allocate that much memory ahead of time, instead we increment the array size as needed. The reader is limited by certain custom value, as I describe above and implemented in [this PR](https://github.com/huggingface-internal/xetcas/pull/268), so deserializing `num_chunks` entries will naturally fail when hitting the end of the reader.

hoytak: Ah, I see.  That makes more sense.  I would perhaps add 10% on it to reduce allocations, but I'm fine with this.  

As a stylistic nit: maybe define 

```
let prealloc_num_chunks = ( (num_chunks as usize).min(AVERAGE_NUM_CHUNKS_PER_XORB) * 9 / 8);
```

 to make it a bit cleaner... (with a small buffer above the average).

assafvayner: If realloc is pretty cheap... then it doesn't really matter what we put here, maybe just hardcode AVG_CHUNKS_PER_XORB and then let it be that way until we get a deserialization error, (which I need to check, but will we even get it before we fill up num_chunks entries at all? we could still be filling num_chunks entries of bogus data)

edit/tldr, don't fret too much on this, just preset something, I think I agree with Di's code at its current state.

seanses: Added this suggestion as an inline fn and added comments.

PR #199, opened 2025-03-11
Data Processing Refactor / Simplification towards WASM compatibility

hoytak:
This PR refactors the data/ package part of xet-core with the following goals: 
- Simplification/readability.  The previous version had a number of design decisions made when we were still figuring out the backend and the different use cases.  Now that we have finalized the overall data flow, we can redo these with simpler and more readable designs.
- WASM compatibility.  This PR abstracts away the chunking and deduplication flow into a standalone package that will allow a different data flow not relying on disk-based shards to still take advantage of the deduplication logic.

The PR: 
- Adds a deduplication package that contains and isolates the chunking  and deduplication logic without the use of tokio channels in the API.
- Simplifies the logic in the file cleaning process.  Now, a FileUploadSession instance creates SingleFileCleaner instances that use &mut self to avoid the frequent use of locks.


data/src/shard_interface.rs
```diff
@@ -1,39 +1,142 @@
...
+                debug!("Error encountered attempting to query global dedup table: {e:?}; ignoring.");
+                e
+            })
+        else {
+            return Ok(false);
+        };
+
+        let Some(new_shard_file) = query_result else {
+            return Ok(false);
+        };
```
assafvayner:
```suggestion
        let Ok(Some(query_result)) = self
            .client
            .query_dedup_shard_by_chunk(&chunk_hash, repo_salt)
            .await
            .debug_error("Error encountered attempting to query global dedup table; ignoring")
        else {
            return Ok(false);
        };
```

use the error_printer trait for debug_error and also we're returning Ok(false) in both these branches.

xet_threadpool/src/threadpool.rs
```diff
@@ -88,6 +89,13 @@ impl ThreadPool {
         })
     }
 
+    /// Use this when already running within a tokio runtime; this simply
+    /// pulls in the handle from the current runtime.  Can be used in testing functions
+    /// when the runtime needs to be passed in as an argument to current functions.
+    pub fn from_current_runtime() -> Arc<Self> {
+        Arc::new(Self::from_external(tokio::runtime::Handle::current()))
```
assafvayner: use named import `TokioRuntimeHandle::current()`?

deduplication/src/data_aggregator.rs
```diff
@@ -22,51 +26,63 @@ pub(crate) struct CASDataAggregator {
...
-        let raw_bytes_len = self.data.len();
-        let cas_hash = cas_node_hash(&self.chunks[..]);
+        let xorb_data = RawXorbData::from_chunks(&self.chunks);
+        let xorb_hash = xorb_data.hash();
 
         // Now that we have the CAS hash, fill in any blocks with the referencing xorb
         // hash as needed.
         for (fi, chunk_hash_indices) in self.pending_file_info.iter_mut() {
-            for &i in chunk_hash_indices.iter() {
+            for i in std::mem::take(chunk_hash_indices) {
```
assafvayner: this to me seems slightly awkward. possible suggestion:
- create an empty vec of MDBFileInfo, make the outer loop an `into_iter()` to take ownership of both fi and chunk_hash_indices
- after modifying `fi` push in into the result vec
- modify return statement to return that vec as the second return value

Another thing would be to make this a 
```
self.pending_file_info.into_iter().map(|(fi, chunk_hash_indices)| {
  // fill in xorb_hash and map to fi
  fi
})
```

deduplication/src/interfaces.rs
```diff
@@ -0,0 +1,34 @@
+use std::result::Result;
+
+use async_trait::async_trait;
+use mdb_shard::file_structs::FileDataSequenceEntry;
+use merklehash::MerkleHash;
+
+use crate::raw_xorb_data::RawXorbData;
+
+#[async_trait]
+pub trait DataInterface: Send + Sync + 'static {
```
assafvayner: doc comment on purpose of this interface, also DataInterface could mean many things, is there a possibly better name?

deduplication/src/interfaces.rs
```diff
@@ -0,0 +1,34 @@
...
+    /// global deduplication.  This is expected to run in the background.
+    async fn register_global_dedup_query(&mut self, _chunk_hash: MerkleHash) -> Result<(), Self::ErrorType> {
+        Ok(())
+    }
+
+    /// Waits for all the current queries to complete, then returns true if there is
+    /// new deduplication information available.
+    async fn complete_global_dedup_queries(&mut self) -> Result<bool, Self::ErrorType> {
+        Ok(false)
+    }
```
assafvayner: are these default returns intentional?

deduplication/src/defrag_prevention.rs
```diff
@@ -0,0 +1,107 @@
...
+        if self.rolling_last_nranges.len() < NRANGES_IN_STREAMING_FRAGMENTATION_ESTIMATOR {
+            None
+        } else {
+            Some(self.rolling_nranges_chunks as f32 / self.rolling_last_nranges.len() as f32)
+        }
+    }
+
+    /// Check to see if we should update against this entry or continue from the previous one?
+    pub(crate) fn allow_dedup_on_next_range(&mut self, dedup_range_size: usize) -> bool {
+        if let Some(chunks_per_range) = self.rolling_chunks_per_range() {
```
assafvayner: might be nice to have de-indented block and early return here i.e.
```
let Some(chunks_per_range) = self.rolling_chunks_per_range() else {
  return true;
}
// de-indent the rest
```

assafvayner (commented):
some initial comments before walkthrough, I think mostly syntax changes on old code, can ignore for now

data/src/data_interface.rs
```diff
@@ -0,0 +1,94 @@
...
+    async fn register_global_dedup_query(&mut self, chunk_hash: MerkleHash) -> Result<()> {
+        if !self.global_dedup_queries_enabled() {
+            return Ok(());
+        }
+
+        // Now, query for a global dedup shard in the background to make sure that all the rest of this
+        // can continue.
+        let session: Arc<FileUploadSession> = self.session.clone();
+
+        self.active_global_dedup_queries.spawn(async move {
```
seanses: Use spawn_on to the self-managed threadpool

data/src/data_interface.rs
```diff
@@ -0,0 +1,94 @@
...
+use deduplication::{DataInterface, RawXorbData};
+use mdb_shard::file_structs::FileDataSequenceEntry;
+use merklehash::MerkleHash;
+use tokio::task::JoinSet;
+
+use crate::configurations::GlobalDedupPolicy;
+use crate::errors::Result;
+use crate::file_upload_session::FileUploadSession;
+
+pub struct UploadSessionDataManager {
```
seanses: This struct only holds queries handles in addition to a FileUploadSession, sounds like it can be merged into FileUploadSession so not to have many layers of wraps.

data/src/parallel_xorb_uploader.rs
```diff
@@ -52,105 +42,81 @@ pub(crate) struct ParallelXorbUploader {
...
+            .map_err(|e| UploadTaskError(e.to_string()))?;
+
+        let client = self.client.clone();
+        let cas_prefix = self.cas_prefix.clone();
+        let upload_progress_updater = self.upload_progress_updater.clone();
+
+        self.upload_tasks.lock().await.spawn_on(
+            async move {
+                let n_bytes_transmitted = client
+                    .put(&cas_prefix, &xorb.hash(), xorb.to_vec(), xorb.cas_info.chunks_and_boundaries())
```
seanses: The `if cas_data.data.is_empty()` check is removed, we may upload empty xorbs again?

data/src/shard_interface.rs
```diff
@@ -1,39 +1,176 @@
...
+        self.session_shard_manager.flush().await?;
+
+        // First, scan, merge, and fill out any shards in the session directory
+        let shard_list =
+            consolidate_shards_in_directory(self.session_shard_manager.shard_directory(), MDB_SHARD_MIN_TARGET_SIZE)?;
+
+        // Upload all the shards and move each to the common directory.
+        let mut shard_uploads = JoinSet::<Result<()>>::new();
+
+        let xorb_bytes_uploaded = Arc::new(AtomicUsize::new(0));
```
seanses:
```suggestion
        let shard_bytes_uploaded = ...
```

data/src/data_client.rs
```diff
@@ -187,29 +178,25 @@ pub async fn download_async(
...
+    let mut handle = processor.start_clean(filename.as_ref().to_string_lossy().into());
 
     loop {
         let bytes = reader.read(&mut read_buf)?;
         if bytes == 0 {
             break;
         }
 
-        handle.add_bytes(read_buf[0..bytes].to_vec()).await?;
+        handle.add_data(&read_buf[0..bytes]).await?;
```
seanses: My major concern it that it is going to be slow if we don't buffer I/O

data/src/shard_interface.rs
```diff
@@ -1,39 +1,176 @@
...
+        Ok(self.cache_shard_manager.chunk_hash_dedup_query(query_hashes).await?)
+    }
+
+    // Add everything to the session shard manager
+    pub async fn add_cas_block(&self, cas_block_contents: MDBCASInfo) -> Result<()> {
+        Ok(self.session_shard_manager.add_cas_block(cas_block_contents).await?)
+    }
+
+    pub async fn add_file_reconstruction_info(&self, file_info: MDBFileInfo) -> Result<()> {
+        // Only add the file reconstruction info if it's not already in the cache.  In this
```
seanses: 🙅‍♂️ We always register a new file info to be uploaded. This is because each file is associated with a repo and the client doesn't know if this file already exists in that repo given local information: the shard cache is repo ignorant.

deduplication/src/interfaces.rs
hoytak: They are, as the global dedup is optional, but in our case I can drop them.  They were more there as an intermediate step.

data/src/data_interface.rs
hoytak: So I read deeper into the documentation.  When spawn is called on a worker thread, it will always spawn the task on the same runtime that the worker thread is running on.  spawn_on is used to spawn onto a different runtime, and we don't need to do that here.

data/src/parallel_xorb_uploader.rs
hoytak: Ah, good check.  I will add it again.

data/src/data_client.rs
hoytak: So the BufReader reads in 32K blocks at a time, and it's meant for reading many small calls (e.g. like our read_u32).  What we're doing here is buffering by always reading in a large block, which is what the BufReader does.   So adding another layer of buffering won't help anything.

data/src/shard_interface.rs
hoytak: Good catch!

data/src/data_interface.rs
hoytak: In this case, it uses a &mut self to avoid layers of mutexes, and as a result it's given by value to the FileDeduper.  Also, there is one instance per file, as we want to complete only the current global dedup queries that are specific to that file during that part of the querying.  So it does need to be its own struct.

deduplication/src/data_aggregator.rs
hoytak: Updated to make it more explicit.

data/src/shard_interface.rs
hoytak: Consolidated and updated.

data/src/data_client.rs
seanses: I don't think so. BufReader merges small reads into one big syscall read, but it is a sync read so we will eventually wait for reading in the one large block, and while we wait for it CPU has nothing to work on and no xorb gets produced. Typically we want to buffer the next block of bytes in memory to keep the CPU always busy, using OS native async I/O or disk threads.

hoytak: The OS already does that level of in-memory buffering, so the syscall will be quite fast; it's also the cost we pay with BufReader as there's not really any async context switching here.     From the documentation on BufReader, 

```
BufReader<R> can improve the speed of programs that make small and repeated read calls to the same file or network socket. It does not help when reading very large amounts at once, or reading just one or a few times. It also provides no advantage when reading from a source that is already in memory, like a [Vec](https://doc.rust-lang.org/std/vec/struct.Vec.html)<u8>.
```

Also, it calls in 8KiB blocks by default (not 32KiB, I stand corrected), so while a single read into a 10MB buffer will be a single syscall, BufReader will sequentially load 8KiB blocks and copy the data into the big buffer, with each one requiring a syscall.

mdb_shard/src/shard_file_manager.rs
```diff
@@ -1,24 +1,35 @@
...
 use crate::file_structs::*;
 use crate::shard_file_handle::MDBShardFile;
 use crate::shard_file_reconstructor::FileReconstructor;
 use crate::shard_in_memory::MDBInMemoryShard;
 use crate::utils::truncate_hash;
 
+// Store a maximum of this many indices in memory
+const CHUNK_INDEX_TABLE_DEFAULT_MAX_SIZE: usize = 64 * 1024 * 1024;
+lazy_static! {
+    static ref CHUNK_INDEX_TABLE_MAX_SIZE: usize = std::env::var("XET_CHUNK_INDEX_TABLE_MAX_SIZE")
```
assafvayner: should use configurable_constants! in this file

deduplication/src/raw_xorb_data.rs
```diff
@@ -0,0 +1,77 @@
...
+    // avoid this copy / memory overhead
+    pub fn to_vec(&self) -> Vec<u8> {
+        let mut new_vec = Vec::with_capacity(self.num_bytes());
+
+        for ch in self.data.iter() {
+            new_vec.extend_from_slice(ch);
+        }
+
+        new_vec
+    }
```
assafvayner, line 76: this almost makes me think this struct should be defined in either the cas_object crate or its own crate to also be referenced by the cas_object crate.

I don't think cas_object, which should still handle serializing xorbs, should use the deduplication crate

deduplication/src/defrag_prevention.rs
```diff
@@ -0,0 +1,110 @@
...
+/// This targets an average of 1MB per range.
+/// The hysteresis factor multiplied by the target Chunks Per Range (CPR) controls
+/// the low end of the hysteresis range. Basically, dedupe will stop
+/// when CPR drops below hysteresis * target_cpr, and will start again when
+/// CPR increases above target CPR.
+const MIN_N_CHUNKS_PER_RANGE_HYSTERESIS_FACTOR: f32 = 0.5;
+const DEFAULT_MIN_N_CHUNKS_PER_RANGE: f32 = 8.0;
+
+lazy_static::lazy_static! {
+    static ref MIN_N_CHUNKS_PER_RANGE: f32 = std::env::var("XET_MIN_N_CHUNKS_PER_RANGE")
```
assafvayner: use configurable_constants! macro here

data/src/file_upload_session.rs
```diff
@@ -119,92 +104,128 @@ impl FileUploadSession {
...
+        metrics.xorb_bytes_uploaded = self.xorb_uploader.finalize().await?;
 
-        Ok(())
+        let all_file_info = {
+            if return_files {
+                self.shard_interface.session_file_info_list().await?
+            } else {
+                Vec::new()
+            }
+        };
```
assafvayner: syntax option here, I personally prefer the suggested change, but go with your preference/what is consistent here
```suggestion
        let all_file_info = if return_files {
            self.shard_interface.session_file_info_list().await?
        } else {
            Vec::new()
        };
```

data/src/sha256.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /// Update the generator with some bytes.
+    pub fn update(&mut self, new_chunks: Arc<[Chunk]>) {
+        let current_state = self.hasher.take();
+
+        // The previous task returns the hasher; we consume that and pass it on.
+        self.hasher = Some(tokio::spawn(async move {
+            let mut hasher = match current_state {
+                Some(jh) => jh.await.unwrap(),
```
assafvayner: did we always have this unwrap here? do we want to mitigate it?

data/src/shard_interface.rs
```diff
@@ -1,39 +1,161 @@
...
+        let xorb_bytes_uploaded = Arc::new(AtomicUsize::new(0));
+
+        for si in shard_list {
+            let salt = self.config.shard_config.repo_salt;
+            let shard_client = self.client.clone();
+            let shard_prefix = self.config.shard_config.prefix.clone();
+            let cache_shard_manager = self.cache_shard_manager.clone();
+            let xorb_bytes_uploaded = xorb_bytes_uploaded.clone();
+
+            shard_uploads.spawn(async move {
```
assafvayner, line 150: may want to use the limiting join set here if we don't have a wait to control shard upload rate, of course we do want a lot of parallelism, but maybe cap it at 100 or 1000 concurrent shard uploads.

data/src/sha256.rs
```diff
@@ -0,0 +1,119 @@
...
+        let current_state = self.hasher.take();
+
+        let hasher = match current_state {
+            Some(jh) => jh.await.unwrap(),
+            None => return MerkleHash::default(),
+        };
+
+        let sha256 = hasher.finalize();
+        let hex_str = format!("{sha256:x}");
+        MerkleHash::from_hex(&hex_str).unwrap()
```
assafvayner: I think this is essentially guaranteed to succeed, but maybe we change this to `.expect()` for cleanliness

assafvayner (commented):
getting better, at some point we want to flip this PR to ready, final reviews and merge

deduplication/src/raw_xorb_data.rs
hoytak: I agree, but I think that's something for a later PR.

data/src/sha256.rs
hoytak: Good catch -- it only triggers if the task has been canceled, i.e. during shutdown.   But I can still propagate it.

data/src/shard_interface.rs
hoytak: We're already doing that using the parallel_upload_limiter semaphore, so there are at most MAX_CONCURRENT_XORB_UPLOADS going on at a time.

data/src/data_client.rs
seanses: Probably I didn't convey my thought correctly. I didn't mean that we need to wrap the reader inside a BuReader, which is indeed unnecessary here. What I meant is that we need to read ahead and buffer: while CPU is processing bytes `0..10`, someone should already initiate a read call to fetch bytes `10..20` into memory, so this loop will never wait for I/O from disk. Whether OS brings more data from disk into memory in a background kernel thread, if any OS does, depends on file opening flag hints and its implementation, and thus it is more reliable to ensure from user level.

Actually, since we always read fixed size bytes from disk and the size is align to sector size, it's more beneficial to disable OS buffering to reduce data copying overhead.

deduplication/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+name = "deduplication"
+version = "0.14.5"
+edition = "2021"
+
+[dependencies]
+gearhash = "0.1.3"
+merklehash = { path = "../merklehash" }
+mdb_shard = { path = "../mdb_shard" }
+merkledb = { path = "../merkledb" }
+cas_object = { path = "../cas_object" }
```
seanses: Request to drop the cas_object dep. It is only needed for 1. two constants that limit xorb size, which should be passed into deduplication as a config for generality; 2. compute verification range hash, which I think should be moved out of cas_object into mdb_shard, because it doesn't need any information about a cas object layout, but only needs a slice of MerkleHash to compute.

hoytak: If something is a value that's tied to the larger system -- I.e. settings on the server -- then I think it's confusing to make it a parameter that can be passed in as our system doesn't allow for generality.  However, I agree that it can be moved here.  Will do.

data/src/data_client.rs
hoytak: BufReader doesn't do any read-ahead like this, so I really don't think this is a scenario worth optimizing for.  Plus, the OS does pretty well at paging data ahead.  Usually on file open the whole file is loaded into memory cache if it's small enough -- this is usually what all the memory unused by programs goes to -- so I don't think we have to worry about this, especially when multiple files are being accessed in parallel and there's a lot of other work going on.

seanses: I don't have a clear number in mind that how well OS read ahead, but can do benchmarks later to confirm. I'm good with this for now.

data/src/file_cleaner.rs
```diff
@@ -1,273 +1,111 @@
...
+        self.sha_generator.update(chunks.clone());
 
-        Self::run(cleaner.clone(), chunk_c).await;
+        // Run the deduplication interface here.
+        self.dedup_manager.process_chunks(&chunks).await?;
 
-        Ok(cleaner)
-    }
+        // Update the progress bar
+        if let Some(updater) = self.session.upload_progress_updater.as_ref() {
```
seanses, line 80: I think the progress is double updated w.r.t to the same byte, the other place that this is updated is in parallel_xorb_uploader.rs and that's after xorb upload succeeds. Between these two I prefer to keep the other one that updates after the bytes leave the client.

deduplication/src/dedup_metrics.rs
```diff
@@ -0,0 +1,40 @@
+#[derive(Default, Debug, Clone, Copy)]
+pub struct DeduplicationMetrics {
+    pub total_bytes: usize,
+    pub deduped_bytes: usize,
+    pub new_bytes: usize,
```
seanses, line 5: This metric `new_bytes` is never effectively updated in dedup, but this number is actually needed for xtool reporting.

data/src/file_upload_session.rs
```diff
@@ -119,92 +104,126 @@ impl FileUploadSession {
...
-        self.shard_manager
-            .all_file_info_of_session()
-            .await
-            .map_err(DataProcessingError::from)
+    pub async fn finalize(self: Arc<Self>) -> Result<DeduplicationMetrics> {
+        Ok(self.finalize_impl(false).await?.0)
+    }
+
+    pub async fn finalize_with_file_info(self: Arc<Self>) -> Result<(DeduplicationMetrics, Vec<MDBFileInfo>)> {
+        self.finalize_impl(false).await
```
seanses: The parameter should be `true` according to the function name.

data/src/file_upload_session.rs
```diff
@@ -119,92 +104,126 @@ impl FileUploadSession {
...
-        Ok(())
+        let all_file_info = if return_files {
+            self.shard_interface.session_file_info_list().await?
+        } else {
+            Vec::new()
+        };
+
+        // Upload and register the current shards in the session, moving them
+        // to the cache.
+        metrics.shard_bytes_uploaded = self.shard_interface.upload_and_register_session_shards().await?;
```
seanses, line 216: In `dry_run` mode we don't want to upload and register session shards.

hoytak: That is intercepted by the dryrun client, which swallows that call.

data/src/file_cleaner.rs
hoytak: According to @assafvayner, that is how it should be done; this was implemented this way as people wanted to show progress quickly.  I agree with you, but I actually think we are now set up to do this properly where we actually track which files depend on what % of which xorbs, so that when xorbs get uploaded we report the proper file % for the depending files.  But that would also be another PR.

data/src/file_upload_session.rs
hoytak: OK, I added that in with https://github.com/huggingface/xet-core/pull/199/commits/331738848fe558f9dd17575250fed252e06f8cdf; it seemed more appropriate to put that in the logic for the remote_client since it's already there for the rest of the functions.

hoytak: I believe with this, we can just drop https://github.com/huggingface/xet-core/pull/213, is that correct?

data/src/file_upload_session.rs
hoytak: Yes, good catch!

data/src/file_cleaner.rs
assafvayner: The initial progress updating logic increments progress when a xorb is uploaded, and when a chunk deduped against an existing chunk regardless of it's uploaded yet or not.

As far as changing the progress logic to be more accurate, we can do that now if it's not a really big task, ideally in a new PR. If it's a lot of work, let's keep the old logic, get this PR merged and work on supporting huggingface.js (higher priority).

As far as I understood, the more accurate tracking will require a lot of work but you'll know better. If it'll take you just 1 day or less then just go for it, but I'd rather it be in a new PR into main after this is merged.

hoytak: I don't think it will be a lot of work -- I've actually prototyped it out already -- but I may need your help interfacing it to the python world.  I think let's just get this in and we can count on that landing.

data/src/shard_interface.rs
assafvayner: Are you sure, this is for shard uploads? I don't see how in this code path we call into the parallel_upload_limiter code

data/src/file_upload_session.rs
seanses: Yes we can drop #213

seanses: But we should also entirely skip this statement, because moving a session shard to cache without uploading the corresponding xorbs is fatal.

hoytak: At this point, all of the xorbs for the session have have been successfully uploaded, so this should be correct.

data/src/shard_interface.rs
hoytak: OH, that's a typo.  Will fix.

hoytak: Ah, you're right.  That's not correct.  Fixed to also use the max parallel uploads.

I'm hesitant to use the limited join set here as there can be many multiple sessions running, and so a global limiting semaphore is needed to actually have the effect we want.

hoytak: Fixed.  Now uses the same semaphore.

data/src/file_cleaner.rs
hoytak: Fixed now too.

data/src/file_upload_session.rs
seanses: But in dry run mode we didn't upload xorbs so we should skip this entirely.

cas_object/src/constants.rs
assafvayner, line 1: delete?

assafvayner (approved):
LGTM after addressing @seanses's last comment

data/src/file_upload_session.rs
hoytak: Ah, I see.   Fixed.

seanses (approved):
LGTM!

PR #203, opened 2025-03-11
optionally deserialize only boundaries section of xorb metadata

assafvayner:
This PR will enable the CAS server to only download the boundaries section of xorb footers to reduce data required to be transmitted and cached.

### DIFF
--- a/cas_object/src/cas_object_format.rs
+++ b/cas_object/src/cas_object_format.rs
@@ -594,7 +594,10 @@ impl CasObjectInfoV1 {
         let offset_to_boundary_section_offset =
             size_of::<u32>() + size_of_val(&s._buffer) + size_of_val(&s.boundary_section_offset_from_end);
         reader.seek(SeekFrom::End(-(offset_to_boundary_section_offset as i64)))?;
-        let boundary_section_offset_from_end = read_u32(reader)?;
+        let mut boundary_section_offset_from_end = read_u32(reader)?;
+
+        // add 4 bytes to offset from info_length at the end
+        boundary_section_offset_from_end += size_of::<u32>() as u32;
         reader.seek(SeekFrom::End(-(boundary_section_offset_from_end as i64)))?;
 
         let mut counting_reader = countio::Counter::new(reader);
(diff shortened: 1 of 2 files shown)


cas_object/src/cas_object_format.rs
```diff
@@ -1942,4 +2023,48 @@ mod tests {
...
+            assert_eq!(boundaries_footer.num_chunks, num_chunks);
+            assert_eq!(boundaries_footer.num_chunks, original.num_chunks);
+            assert_eq!(boundaries_footer.boundary_section_offset_from_end, original.boundary_section_offset_from_end);
+            assert_eq!(boundaries_footer.hashes_section_offset_from_end, original.hashes_section_offset_from_end);
+            assert_eq!(boundaries_footer.chunk_boundary_offsets, original.chunk_boundary_offsets);
+            assert_eq!(boundaries_footer.unpacked_chunk_offsets, original.unpacked_chunk_offsets);
+
+            // check that the hashes are not filled in
+            assert!(boundaries_footer.chunk_hashes.is_empty());
+            assert!(!boundaries_footer.has_chunk_hashes());
```
seanses, line 2063: I think this assert is doing the same thing as the above one?

cas_object/src/cas_object_format.rs
```diff
@@ -579,6 +582,79 @@ impl CasObjectInfoV1 {
...
+        let end_byte_offset = r.reader_bytes();
+
+        if end_byte_offset != s.boundary_section_offset_from_end as usize {
+            return Err(CasObjectError::FormatError(anyhow!(
+                "Xorb Invalid: incorrect boundary_section_offset_from_end."
+            )));
+        }
+
+        // set the version and ident field since this is a valid CasObjectInfoV1
+        s.hashes_version = 0;
```
seanses: use the const value

seanses (approved):

cas_object/src/cas_object_format.rs
assafvayner: removing this line, it's preset in the Default impl

assafvayner:
Going to proceed with merging this, but hold off on using server-side, we do use the hashes section on the server side for shard validation.

PR #204, opened 2025-03-11
xet_threadpool is wasm compilable

assafvayner:
Small step towards wasm-uploads, makes xet_threadpool work in singlethreaded wasm context (alledgedly, more testing to come)

- no impact on threadpool outside wasm/e.g. hf-xet
- required that I remove the use of `tokio::task::block_in_place` which is not available in tokio single thread/current_thread mode which is required in the WASM usage of tokio
  - this was used downstream of token refreshes, i.e. call async token refreshing function in hub_client from a sync `refresh` trait function. Ended up making the `TokenRefresher` trait an `async_trait` hence the various extra changes
- formatting fixes renamed imports for tokio::runtime::* imports.

### DIFF
--- a/cas_client/src/http_client.rs
+++ b/cas_client/src/http_client.rs
@@ -1,4 +1,4 @@
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
 use std::time::Duration;
 
 use anyhow::anyhow;
@@ -9,6 +9,7 @@ use reqwest::{Request, Response};
 use reqwest_middleware::{ClientBuilder, ClientWithMiddleware, Middleware, Next};
 use reqwest_retry::policies::ExponentialBackoff;
 use reqwest_retry::{default_on_request_failure, default_on_request_success, RetryTransientMiddleware, Retryable};
+use tokio::sync::Mutex;
 use tracing::{debug, warn};
 use utils::auth::{AuthConfig, TokenProvider};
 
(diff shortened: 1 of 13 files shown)


cas_client/src/http_client.rs
```diff
@@ -9,6 +9,7 @@ use reqwest::{Request, Response};
 use reqwest_middleware::{ClientBuilder, ClientWithMiddleware, Middleware, Next};
 use reqwest_retry::policies::ExponentialBackoff;
 use reqwest_retry::{default_on_request_failure, default_on_request_success, RetryTransientMiddleware, Retryable};
+use tokio::sync::Mutex;
```
assafvayner, line 12: need a tokio Mutex in order to be Sync

xet_threadpool/src/threadpool.rs
```diff
@@ -151,55 +150,21 @@ impl ThreadPool {
...
-                    error!("Panic reported on xet worker task: {e:?}");
-                    MultithreadedRuntimeError::TaskPanic(e)
-                } else if e.is_cancelled() {
-                    // Likely caused by the runtime shutting down (e.g. with a keyboard CTRL-C).
-                    MultithreadedRuntimeError::TaskCanceled(format!("{e}"))
-                } else {
-                    MultithreadedRuntimeError::Other(format!("task join error: {e}"))
-                }
-            })
+            self.handle.spawn(future).await.map_err(MultithreadedRuntimeError::from)
```
hoytak, line 161: The default from method doesn't distinguish between cancelled and panic, so without the above code CTRL-C will show up as a panic runtime error in python instead of a TaskCancellation exception, which is handled differently.  If this code isn't part of the key path, let's not change it.

xet_threadpool/src/errors.rs
```diff
@@ -16,3 +16,18 @@ pub enum MultithreadedRuntimeError {
     #[error("Unknown task runtime error: {0}")]
     Other(String),
 }
+
+impl From<tokio::task::JoinError> for MultithreadedRuntimeError {
+    fn from(err: tokio::task::JoinError) -> Self {
```
hoytak, line 21: This doesn't cause a compilation issue with the From trait implemented in the #[derive(Error)]...?

assafvayner: for the `TaskPanic` variant, it only holds inside a `JoinError` they would conflict if it was `TaskPanic(#[from] JoinError)`

hoytak (approved):
LGTM

PR #205, opened 2025-03-12
update merklehash to wasm compat

assafvayner:
Makes the merklehash crate be compilable with wasm i.e. `--target wasm32-unknown-unknown`

this requires a specific feature on `getrandom` which is a dep of `rand` (and in particular this feature name changed in the version used by v0.9.0 of `rand`, opting to bump up `rand` as we go through all the crates.

Also requires setting a rust flag option, rather than forcing everyone to set RUSTFLAGS I've instead set it in a .cargo/config.toml in the workspace root. Only applies for the wasm target.


hoytak (approved):
LGTM

PR #206, opened 2025-03-12
WIP: Separate out chunking and file dedup into standalone package

hoytak:
This draft PR creates a new package, deduplication/, isolating file deduplication into a class that handles that.  This class is meant to be used and consumed upon completion and wraps a simple interface that gets called to actually do the deduplication.

This is a WIP, but the aim is to streamline the SingleFileCleaner class by separating out the detailed implementation of the deduplication into isolated functionality and make the dedup accessible for wasm as well as.


merkledb/src/aggregate_hashes.rs
```diff
@@ -17,7 +20,10 @@ pub fn cas_node_hash(chunks: &[(MerkleHash, usize)]) -> MerkleHash {
 
     let nodes: Vec<MerkleNode> = chunks
         .iter()
-        .map(|(h, len)| mdb.maybe_add_node(h, *len, Vec::default()).0)
+        .map(|t| {
+            let t: (MerkleHash, usize) = t.into();
```
assafvayner, line 24: maybe change this to `let (t_hash, t_usize): (MerkleHash, usize) = t.into();` with better variable names, makes the next line more readable. I'm not exactly sure what this is doing yet

deduplication/src/errors.rs
assafvayner, line 1: delete?

hoytak:
Closing WIP progress after incorporating discussion; see https://github.com/huggingface/xet-core/pull/199

PR #208, opened 2025-03-12
make utils wasm compat

assafvayner:
This PR ensures that the utils crate (singleflight, serialization_utils) compiles for WASM, should be rebased/merge-main after the xet_threadpool change #204  is in main.

### DIFF
--- a/utils/src/output_bytes.rs
+++ b/utils/src/output_bytes.rs
@@ -4,6 +4,7 @@
 /// * `v` - the size in bytes
 pub fn output_bytes(v: usize) -> String {
     let map = vec![
+        #[cfg(not(target_family = "wasm"))]
         (1_099_511_627_776, "TiB"),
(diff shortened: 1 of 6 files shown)


utils/src/output_bytes.rs
```diff
@@ -4,6 +4,7 @@
 /// * `v` - the size in bytes
 pub fn output_bytes(v: usize) -> String {
     let map = vec![
+        #[cfg(not(target_family = "wasm"))]
```
seanses, line 7: Is there a way to tell if it's wasm32 or wasm64? I think this is only invalid in wasm32

seanses: Actually.. maybe a better fix is to change the parameter type to u64.

assafvayner:
closing in favor of #257

PR #209, opened 2025-03-13
concurrent limited join set

assafvayner:
An idea to be primarily used on the cas server.

A join set that only has at most X tasks running at once. In CAS server (and possibly in the client) we use a standard tokio JoinSet when spawning a number of tasks for 1 operation e.g. shard verification, fetch_info generation. In order to not cause too large a cpu spike/steal work from other requests being handled, we will use this limiting join set to ensure only some tasks are ever actually running at once. (tasks will be in the task pool but will only have work on them after the semaphore is acquired).

### DIFF
--- a/utils/src/lib.rs
+++ b/utils/src/lib.rs
@@ -8,6 +8,7 @@ pub mod singleflight;
 mod async_read;
 mod output_bytes;
 pub mod progress;
+pub mod limited_joinset;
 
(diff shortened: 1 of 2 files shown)


hoytak:
I think this would be easier using just a JoinSet + Semaphore.   Each task first acquires a permit before running the rest of the task.

jgodlew (approved):

PR #211, opened 2025-03-18
Macro for declaring configurable constants. 

hoytak:
Currently, we use a lot of const expressions in which a corresponding environment variable can be used to override the default value.  This PR introduces a macro to standardize this usage.

Example:
```
use utils::configurable_constants;

configurable_constants! {

    /// The target shard size; shards.
    MDB_SHARD_TARGET_SIZE : u64 = 64 * 1024 * 1024;

    /// Minimum shard size; shards are aggregated until they are at least this.
    MDB_SHARD_MIN_TARGET_SIZE: u64 = 64 * 1024 * 1024;

    /// The global dedup chunk modulus; a chunk is considered global dedup
    /// eligible if the hash modulus this value is zero.  Can only be changed for testing.
    MDB_SHARD_GLOBAL_DEDUP_CHUNK_MODULUS : u64 = release_fixed(1024);
}
```

In this example, on initialization, the code would attempt to read from XET_ MDB_SHARD_TARGET_SIZE, using the value given above if that doesn't work.   

For MDB_SHARD_GLOBAL_DEDUP_CHUNK_MODULUS, which is denoted by release_fixed, it would attempt to load the value  from the corresponding environment variable only when debug_assertions are enabled, allowing that value to be changed in tests but not in release mode.  Similarly, fixed(...) denotes that a value is never changed. 

Additionally, the macro `test_set_global!(VAR, VALUE);` can be called at the beginning of a test to ensure that the global referenced by that value is set to a specific value for that test.

### DIFF
--- a/mdb_shard/src/constants.rs
+++ b/mdb_shard/src/constants.rs
@@ -1,6 +1,6 @@
-use utils::{configurable_globals, release_fixed};
+use utils::configurable_constants;
 
-configurable_globals! {
+configurable_constants! {
 
     /// The target shard size; shards.
     MDB_SHARD_TARGET_SIZE : u64 = 64 * 1024 * 1024;
--- a/mdb_shard/src/constants.rs
+++ b/mdb_shard/src/constants.rs
@@ -1,13 +1,27 @@
-pub const MDB_SHARD_TARGET_SIZE: u64 = 64 * 1024 * 1024;
-pub const MDB_SHARD_MIN_TARGET_SIZE: u64 = 48 * 1024 * 1024;
+use utils::{configurable_globals, release_fixed};
 
-pub const MDB_SHARD_GLOBAL_DEDUP_CHUNK_MODULUS: u64 = 1024;
+configurable_globals! {
 
-/// The amount of time a shard should be expired by before it's deleted, in seconds.  
-/// By default set to 7 days.
-pub const MDB_SHARD_EXPIRATION_BUFFER_SECS: u64 = 7 * 24 * 3600;
+    /// The target shard size; shards.
+    MDB_SHARD_TARGET_SIZE : u64 = 64 * 1024 * 1024;
+
+    /// Minimum shard size; shards are aggregated until they are at least this.
+    MDB_SHARD_MIN_TARGET_SIZE: u64 = 64 * 1024 * 1024;
+
+    // The global dedup chunk modulus; a chunk is considered global dedup
+    // eligible if the hash modulus this value is zero.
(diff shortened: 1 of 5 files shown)


hoytak:
Note: I only converted the mdb_shard crate with this PR, I'll convert the constants in more crates if this gets approved.

mdb_shard/src/constants.rs
```diff
@@ -1,13 +1,27 @@
...
+    // eligible if the hash modulus this value is zero.
+    MDB_SHARD_GLOBAL_DEDUP_CHUNK_MODULUS : u64 = release_fixed(1024);
+
+    /// The amount of time a shard should be expired by before it's deleted, in seconds.
+    /// By default set to 7 days.
+    MDB_SHARD_EXPIRATION_BUFFER_SECS : u64 = 7 * 24 * 3600;
+
+    /// The maximum size of the chunk index table that's stored in memory.  After this,
+    /// no new chunks are loaded for deduplication.
+    CHUNK_INDEX_TABLE_MAX_SIZE : usize = 64 * 1024 * 1024;
```
assafvayner: minor style nit to fix the colons after the constant names in some of these

utils/src/constant_declarations.rs
```diff
@@ -0,0 +1,107 @@
...
+pub fn env_configurable<T>(t: T) -> GlobalConfigMode<T> {
+    GlobalConfigMode::EnvConfigurable(t)
+}
+
+// Make env_configurable the default
+impl<T> From<T> for GlobalConfigMode<T> {
+    fn from(value: T) -> Self {
+        GlobalConfigMode::EnvConfigurable(value)
+    }
+}
```
assafvayner, line 18: how do you ever get any of the other variants?

do we need a separate macro for those?
do we even care about having any other variants realistically, I can get around release fixed by building from source and regular users are unlikely to accidentally set such env vars. Fixed is just a regular const, so we don't need a special value for it.

assafvayner (approved):
I like this idea, we do this very often and it will make it easier to set defaults and configure env vars.

Approved, but can you address my comment in constant_declarations.rs? (respond or change, not sure if there was a relevant purpose I'm missing).

utils/src/constant_declarations.rs
hoytak: The idea is that some constants should never be changed in release mode, but could be helpful to change in local unit tests -- e.g. things dealing with the xorb format or chunking.  So using `= release_fixed(value);` means that value can only be changed in debug mode.  

Yes, if it's truly fixed, you probably would use a full const instead of a ref like this, so I can drop that part.

PR #212, opened 2025-03-20
Implement memory control for upload pipeline.

hoytak:
In the large update PR, the memory usage was unbounded as it was possible to load everything into chunks before the xorb upload begins.  This implements a simple strategy to ensure that more data is not ingested while the upload is unable to keep up.   By default, it limits memory usage by chunks to 512 MB per upload session, which does not include the additional potential global usage of MAX_PARALLEL_XORB_UPLOAD * MAX_XORB_BYTES or the local buffer used to read in the files.  

Memory usage is limited using a global semaphore combined with acquire_many_permits, where each permit effectively represents a byte.  Permits for a block of chunks that run through the processing end-to-end are acquired at once to ensure that parallel calls to add_data do not cause a deadlock.

### DIFF
--- a/data/src/constants.rs
+++ b/data/src/constants.rs
@@ -1,5 +1,6 @@
 utils::configurable_constants! {
 
+
     // Approximately 4 MB min spacing between global dedup queries.  Calculated by 4MB / TARGET_CHUNK_SIZE
     ref MIN_SPACING_BETWEEN_GLOBAL_DEDUP_QUERIES: usize = 256;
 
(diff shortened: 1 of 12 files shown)


seanses (commented):
Is the memory permit necessary? The chunker will consume as much data as possible to cut  out chunks, the deduper will consume all chunks to form a xorb, they buffer limit data (chunker buffers number of bytes < `MAX_CHUNK_SIZE`; deduper buffers number of bytes < `MAX_XORB_SIZE`). Now if the upload is slow, `register_new_xorb` will just block because it can't acquire a permit beyond `MAX_CONCURRENT_XORB_UPLOADS`, so `dedup_manager.process_chunks` will block, `SingleFileCleaner::add_data` will block, and so no more byte can enter the system.

The only concerning case is when data.len() is huge. This issue is beautifully handled by breaking it up into smaller blocks in this PR.

hoytak:
Good point.   

The one place where we can build up a huge amount of data in memory is in the sha256 hasher, which doesn't force completion of the pipeline until the end of the file.  However, I can change that, which means your analysis is correct, and we don't need this complexity.  I'll change that in the main PR.

PR #213, opened 2025-03-20
Added dry run client for running uploads in dry run mode.

hoytak:
Adds a DryRunClient that wraps a given client, but swallows all uploads.  This enables dry run mode in the cleaning path simplification PR.

### DIFF
--- a/cas_client/src/dry_run_client.rs
+++ b/cas_client/src/dry_run_client.rs
@@ -1,0 +1,99 @@
+use std::collections::HashMap;
+use std::io::Write;
+use std::path::PathBuf;
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use cas_types::FileRange;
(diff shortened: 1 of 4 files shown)


assafvayner (approved):

seanses (changes requested):
The xtool actually wants to obtain the precise number of bytes a client needs to transfer, this includes the additional bytes to represent the xorb format and counting the effect of compression. In old `remote_client.rs` this is implemented by just skipping the `reqwest_client.put`. The `put()` in DryRunClient returns `Ok(0)` and is not suffice.

hoytak:
Given that comment, I think we can actually just drop this pr in favor of https://github.com/huggingface/xet-core/pull/199/commits/331738848fe558f9dd17575250fed252e06f8cdf; is that correct?

hoytak:
Dropping; incorporated changes needed into PR 199.

PR #217, opened 2025-03-28
Fold parallel xorb uploader into FileUploadSession.

hoytak:
This PR folds the parallel xorb uploader into the FileUploadSession class.   It should not add or subtract any functionality.

The main reason for this is to make the bookkeeping around the upload tracking easier, as after this only the session manager has to record and track completions and dependencies.


data/src/file_upload_session.rs
```diff
@@ -130,9 +141,52 @@ impl FileUploadSession {
...
+                .put(&cas_prefix, &xorb_hash, xorb_data, chunks_and_boundaries)
+                .await?;
+
+            drop(upload_permit);
+
+            if let Some(updater) = session.upload_progress_updater.as_ref() {
+                updater.update(n_bytes_transmitted as u64);
+            }
+
+            session.deduplication_metrics.lock().await.xorb_bytes_uploaded += n_bytes_transmitted;
```
hoytak, line 180: Tracking completed bytes is now done here instead of as a returned value.

data/src/file_upload_session.rs
```diff
@@ -130,9 +141,52 @@ impl FileUploadSession {
...
+        }
+
+        let xorb_hash = xorb.hash();
+        let xorb_data = xorb.to_vec();
+        let chunks_and_boundaries = xorb.cas_info.chunks_and_boundaries();
+
+        drop(xorb);
+
+        let session = self.clone();
+        let upload_permit = acquire_upload_permit().await?;
```
assafvayner, line 165: This will block the current function until there is a permit available to start the upload, is this intentional?

We can return early to the caller if we acquire the permit inside the future that gets put into the join set

data/src/shard_interface.rs
```diff
@@ -141,11 +141,7 @@ impl SessionShardInterface {
             // It's also important to acquire the permit before the task is launched; otherwise, we may spawn an
             // unlimited number of tasks that end up using up a ton of memory; this forces the pipeline to
             // block here while the upload is happening.
-            let upload_permit = UPLOAD_CONCURRENCY_LIMITER
-                .clone()
-                .acquire_owned()
-                .await
-                .map_err(|e| DataProcessingError::UploadTaskError(e.to_string()))?;
+            let upload_permit = acquire_upload_permit().await?;
```
assafvayner, line 144: the code to upload the shard makes more sense for me to be part of the FileUploadSession, particularly I don't think acquire_upload_permit needs to be exported out of the file/module.

Not a required change, just a nit.

data/src/file_upload_session.rs
```diff
@@ -130,9 +141,52 @@ impl FileUploadSession {
         SingleFileCleaner::new(file_name, self.clone())
     }
 
+    pub(crate) async fn register_new_xorb_for_upload(self: &Arc<Self>, xorb: RawXorbData) -> Result<()> {
+        // First check the current xorb upload tasks to see if they
```
seanses: incomplete comment?

seanses (approved):

data/src/file_upload_session.rs
hoytak: This is intentional.  If we return to the caller early, then the future waiting on the permit holds a lot of data in memory while the caller builds more.  We end up running loading everything into memory while waiting for the uploads to progress (also why we can't use the limited join set here).

data/src/shard_interface.rs
hoytak: I considered that, but I kind of like to isolate the shard handling interface (or fold it completely into file_upload_session).

PR #218, opened 2025-03-30
Conservative retry strategy for global dedup query

seanses:
Global dedup queries don't retry on 429. Fix XET-441, XET-448.

### DIFF
--- a/cas_client/src/http_client.rs
+++ b/cas_client/src/http_client.rs
@@ -258,73 +258,184 @@ mod tests {
     #[tokio::test]
     #[traced_test]
     async fn test_retry_policy_500() {
-        // Arrange
-        let server = MockServer::start();
-        let mock_500 = server.mock(|when, then| {
-            when.method(GET).path("/data");
-            then.status(500).body("500: Internal Server Error");
-        });
+        {
+            // Arrange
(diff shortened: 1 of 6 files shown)


rajatarya:
Adding @jgodlew to review as well.

assafvayner (approved):
LGTM

assafvayner:
only question is why separate out the 2 clients, should not all clients stop sending requests on 429? Is there a way instead of not retrying to explicitly delay how long until the next retry? (I don't think it's important, ok to fail here)

jgodlew (approved):
LGTM! :shipit:

seanses:
> only question is why separate out the 2 clients, should not all clients stop sending requests on 429? 

For certain requests we want them to finally succeed, e.g. put xorb and put shard.

> Is there a way instead of not retrying to explicitly delay how long until the next retry? (I don't think it's important, ok to fail here)

Depends on how we want to do dedup. Our current setup does dedup as we stream in and chunk the file and the file reconstruction info is built based on the dedup result. If we choose to explicitly wait (for a long time), we have to pause the entire cleaning, otherwise we need to go back and rebuild the file reconstruction info based on the new dedup info fetched from server.

PR #219, opened 2025-03-31
Completion tracking for accurate upload and download progress reporting. 

hoytak:
This PR adds detailed completion tracking to the upload and download paths that reports accurate per-file metrics to the progress updating interface.  The dependencies between xorbs and files are tracked, so that when a xorb is uploaded, the progress of all the files containing data in that xorb is incremented correctly.  And, when a file is deduped against any xorb, that progress is also reported correctly.  

Updates are passed to a struct implementing the TrackingProgressUpdater trait, which reports progress as batches of 
per-file updates. 

```
pub trait TrackingProgressUpdater: Debug + Send + Sync {
    /// Register a set of updates as a list of ProgressUpdate instances, which 
    /// contain the name and progress information.
    fn register_updates(&self, updates: &[ProgressUpdate]);
}
```

where ProgressUpdate is defined as

```
pub struct ProgressUpdate {
    pub item_name: Arc<str>,
    pub total_count: u64,
    pub completed_count: u64,
    pub update_increment: u64,
}
```

In addition to the above functionality, numerous test cases were added to ensure accuracy and correctness across a lot of different cases.


assafvayner:
Need to resolve TODO's and ensure backwards compatibility while hf_hub keeps providing only 1 progress bar, then make a change to hf_hub to add progress bars for each file.

hoytak:
Updated the PR to the current main branch, so everything should work now. 

On further thought, I changed the trait for the progress updating to be async to avoid the use of the std::sync::Mutex; this should now be WASM compatible. 

The current implementation should be backwards compatible with the previous progress updating as reported to python.   

I propose we do this in stages (the TODO's reflect this):

1. Get this PR in. This is backwards compatible; all the per-file updates are just folded into the global updating counter. 
2. Update hub to expose a progress reporting function that utilizes with the new detailed per-file reporting.
3. A final and small PR to connect the two.
4. Remove the old global update API in hub.

merklehash/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-rand = "0.9.0"
+rand = { version = "0.8.5", features = ["small_rng"] }
```
assafvayner, line 9: upgrade rand version

hf_xet/src/progress_update.rs
```diff
@@ -49,18 +49,27 @@ impl WrappedProgressUpdater {
...
             let f = self.py_func.bind(py);
             if !f.is_callable() {
                 error!("ProgressUpdater func: {} is not callable", self.name);
                 return;
             }
-            let _ = f
-                .call1((increment,))
-                .log_error("python exception trying to update progress bar");
+            for update in updates {
+                // TODO: Connect the full information in the ProgressUpdate struct to python.
```
assafvayner, line 65: this is a TODO in the huggingface_hub library, we should remove this comment before merging or remove this comment after merging the PR into huggingface_hub. we have to release hf_xet before the huggingface_hub PR gets merged

assafvayner (approved):
I think this is all good, would urge to get a DX or YL stamp too!

data/src/progress_tracking.rs
```diff
@@ -0,0 +1,497 @@
...
+    file_indices: Vec<usize>,
+
+    /// True if the xorb has already been updated successfully.
+    is_completed: bool,
+}
+
+/// Represents a file that depends on one or more xorbs.
+struct FileDependency {
+    /// Human-readable name of the file.
+    name: Arc<str>,
```
seanses, line 28: Maybe we don't need this field. There's a trend for xet-core to support upload from any `Read` data source, e.g. https://github.com/huggingface/xet-core/pull/273/files, and "file name" doesn't matter. The caller will be responsible for mapping file names to each progress bar.

data/src/file_cleaner.rs
```diff
@@ -39,7 +40,10 @@ enum DedupManagerBackgrounder {
 /// A class that encapsulates the clean and data task around a single file.
 pub struct SingleFileCleaner {
     // Auxiliary info
-    file_name: String,
+    file_name: Arc<str>,
```
seanses: We can delete this field.

seanses:
Looks good in general! Will take another look when the conflicts are resolved

hf_xet/src/progress_update.rs
hoytak: The huggingface_hub would just expose a python API that should be called something different than this one, so that can get released before the connecting PR.   That connecting PR can clean up all these comments.

data/src/progress_tracking.rs
hoytak: Since the progress bar functions as a callback, there needs to be some way of tracking which update is for which file.  This can just be a unique tag that the python end simply translates into the filename.

data/src/file_cleaner.rs
hoytak: Again, the main issue here is that the callback gives per-file updates, and this field is needed to specify which updated applies to which file.  This can be a tag instead of a name that the end point translates, but needs to be something propagated through the whole process.

seanses (approved):

data/src/progress_tracking.rs
seanses: A tag sounds good!

PR #220, opened 2025-03-31
Fix hf xet download speed

Hugoch:
Check #221 for implem.

Fix `hf_xet` download logic to write file chunk in parallel rather than iterating.
While this fix improves performances on SSD, it will degrade performances on HDD.

For reviewers:
- Please check `remote_client.rs:317` as I am unsure of Xet logic at this point. Commented code is what I would have expected but it was not implemented previously and breaks the tests. I you can enlighten me...
- I increased `NUM_CONCURRENT_RANGE_GETS`, but it should maybe depend on the underlying hardware not to hog desktop machines.
- Seek in file is slow, we can probably use a `write_at` from `std::os::unix::fs::FileExt` to get a bit more perf, but that would be a unix-only option

## Performance comparison
`hf_xet_fixed` is current PR.

xet_gcp_aws_tools
xet_tool_boxplot

### DIFF
--- a/cas_client/src/remote_client.rs
+++ b/cas_client/src/remote_client.rs
@@ -314,12 +314,6 @@ impl RemoteClient {
         if let Some(byte_range) = &byte_range {
             let mut current_offset = 0;
             for (idx, (term_range, &length)) in file_range_per_term.iter_mut().zip(term_lengths.iter()).enumerate() {
-                // FIXME I don't understand why the current logic does not take the start range into account
-                // term_range.start = if byte_range.start < current_offset + length && byte_range.start >=
-                // current_offset {     byte_range.start - current_offset
-                // } else {
-                //     0
-                // };
                 // Adjust the end based on whether the byte_range.end is within the current term.
                 term_range.end = if byte_range.end > current_offset && byte_range.end <= current_offset + length {
                     byte_range.end - current_offset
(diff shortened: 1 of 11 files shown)


cas_client/src/remote_client.rs
```diff
@@ -300,45 +302,103 @@ impl RemoteClient {
...
+        let term_lengths = terms.iter().map(|x| x.unpacked_length as u64).collect::<Vec<_>>();
+        let file_range_per_term = &mut vec![FileRange::default(); terms.len()];
+        for item in file_range_per_term.iter_mut() {
+            item.start = 0;
+            item.end = 0;
+        }
+        if let Some(byte_range) = &byte_range {
+            let mut current_offset = 0;
+            for (idx, (term_range, &length)) in file_range_per_term.iter_mut().zip(term_lengths.iter()).enumerate() {
+                // FIXME I don't understand why the current logic does not take the start range into account
```
jgodlew: IIUC, the start range is unneeded since the terms will be received from CAS pruned to the range already, meaning that we don't know how many terms / bytes were skipped to get to the first term in `byte_range.start` is the offset globally across the complete file, however, the terms we are given may not start with the first term in `terms`. This is why we have the `offset_into_first_range` instead of relying on `byte_range.start`.

Hugoch: Super clear! Thanks for the explanation 🙏 !

Hugoch:
Closing to consolidate in #221

PR #221, opened 2025-03-31
Env to write reconstruction terms in parallel

jgodlew:
When investigating slowness between hf_xet download and hf_transfer, we noticed that one source of slowness was due to the fact that although terms were downloaded concurrently, they were being written sequentially to disk. In [hf_transfer](https://github.com/huggingface/hf_transfer/blob/main/src/lib.rs#L334-L361), each chunk downloaded is written to the output file directly.

This PR updates the way the cas_client writes reconstructed terms to the output file to behave similarly to hf_transfer:
* Refactors the reconstruction path to take a `OutputProvider` enum instead of a `&Box<dyn Write + Send>`.
* `OutputProvider` enum can be called to provide a `Write` implementation for a particular location.
* New method: `RemoteClient::reconstruct_file_to_writer_parallel` to write terms to the output file in parallel.
* Env variable: `HF_XET_RECONSTRUCT_WRITE_SEQUENTIALLY` that, when set by the user, will enable writing terms to the output sequentially instead of in parallel.

TODO:

- [x] Clean up
- [x] More Testing

### DIFF
--- a/cas_client/src/interface.rs
+++ b/cas_client/src/interface.rs
@@ -176,7 +176,7 @@ pub mod buffer {
 
     impl BufferProvider {
         pub fn get_writer_at(&self, _start: u64) -> Result<Box<dyn Write + Send>> {
-            Ok(Box::new(self.buf.clone())) // TODO: fix tests once we start writing in parallel
+            Ok(Box::new(self.buf.clone()))
         }
     }
 
--- a/cas_client/src/interface.rs
+++ b/cas_client/src/interface.rs
@@ -175,18 +175,22 @@ pub mod buffer {
     }
 
     impl BufferProvider {
-        pub fn get_writer_at(&self, _start: u64) -> Result<Box<dyn Write + Send>> {
-            Ok(Box::new(self.buf.clone()))
+        pub fn get_writer_at(&self, start: u64) -> Result<Box<dyn Write + Send>> {
+            let mut buffer = self.buf.clone();
+            buffer.idx = start;
+            Ok(Box::new(buffer))
         }
     }
 
     #[derive(Debug, Default, Clone)]
+    /// Thread-safe in-memory buffer that implements [Write](Write) trait at some position
+    /// within an underlying buffer and allows access to inner buffer.
     /// Thread-safe in-memory buffer that implements [Write](Write) trait and allows
     /// access to inner buffer
     pub struct ThreadSafeBuffer {
+        idx: u64,
         inner: Arc<Mutex<Cursor<Vec<u8>>>>,
     }
-
     impl ThreadSafeBuffer {
         pub fn value(&self) -> Vec<u8> {
             self.inner.lock().unwrap().get_ref().clone()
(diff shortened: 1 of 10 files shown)


cas_client/src/interface.rs
```diff
@@ -71,6 +72,47 @@ pub trait ReconstructionClient {
...
+#[derive(Debug, Clone)]
+pub enum OutputProvider {
+    File(FileProvider),
+    #[cfg(test)]
+    Buffer(buffer::BufferProvider),
+}
+
+impl OutputProvider {
+    /// Create a new writer to start writing at the indicated start location.
+    pub(crate) fn get_writer_at(&self, start: u64) -> Result<Box<dyn Write + Send>> {
```
jgodlew, line 85: I tried to make this a trait that could be implemented by `FileProvider`/`BufferProvider`, but ran into Rust trait object and generics issues with the `Client` trait that would use this (i.e. `ReconstructionClient::get_file`). We might be able to possibly do this with some bigger refactors that are outside the scope of this PR. For now, the enum approach works.

cas_client/src/remote_client.rs
```diff
@@ -341,6 +366,120 @@ impl RemoteClient {
...
+            range_download_single_flight: self.range_download_single_flight.clone(),
+            fetch_info,
+            semaphore: Arc::new(Semaphore::new(NUM_CONCURRENT_RANGE_GETS)),
+            output: output_provider.clone(),
+        };
+        // Build term tasks, computing the offsets needed for the downloaded term and
+        // offset for the output.
+        let mut bytes_written = 0;
+        let mut remaining = total_len;
+        let terms = terms.into_iter().enumerate().map(|(idx, term)| {
```
port8080:
```suggestion
        let term_tasks = terms.into_iter().enumerate().map(|(idx, term)| {
```

Hugoch:
Benchmark of latest version:

xet_gcp_aws_tools
xet_tool_boxplot

Hugoch (approved):
LGTM, nice to have the fallback to sequential writes!  🚀

cas_client/src/interface.rs
assafvayner: We may need to revisit a lot of this code in prep for hf.js since we are adding  File ops, something to address later

cas_client/src/remote_client.rs
```diff
@@ -142,26 +152,40 @@ impl ReconstructionClient for RemoteClient {
...
-            manifest.offset_into_first_range,
-            byte_range,
-            writer,
-            progress_updater,
-        )
-        .await
+
+        // If the user has set the `ENV_RECONSTRUCT_WRITE_SEQUENTIALLY` env variable, then we should
+        // write the file to the output sequentially instead of in parallel.
+        if env::var(ENV_RECONSTRUCT_WRITE_SEQUENTIALLY).is_ok() {
```
assafvayner, line 165: we should be using the configurable_constants! macro from now on, will submit a PR later to change this.

cas_client/src/remote_client.rs
```diff
@@ -341,6 +370,120 @@ impl RemoteClient {
...
+            let task = task_info.clone();
+            task.write_term(term, start..end, file_offset)
+        });
+
+        // Spawn the tasks
+        let mut handles = FuturesUnordered::new();
+        term_tasks.for_each(|task| {
+            let handle = self.threadpool.spawn(task);
+            handles.push(handle);
+        });
```
assafvayner, line 423: this will cause all tasks to run at the same time, and then each task is limited explicitly by the semaphore acquire, we could fold away some of the code by using the LimitedJoinSet from the utils crate.

cas_client/src/remote_client.rs
```diff
@@ -341,6 +370,120 @@ impl RemoteClient {
...
+
+        // Join the tasks as they come in.
+        let mut total_written = 0;
+        while let Some(result) = handles.next().await {
+            match result {
+                Ok(Ok(len_written)) => {
+                    progress_updater.as_ref().inspect(|updater| updater.update(len_written));
```
assafvayner, line 430: need to verify how this tracks with https://github.com/huggingface/xet-core/pull/219

PR #222, opened 2025-04-01
Update hf_xet to 1.0.0

jgodlew:
Now that we've released the integration with huggingface_hub and users are expected to use hf_xet in production, we should bump up the version of hf_xet to reflect this.

This does imply that we will not be breaking the interface to hf_xet (i.e. the pyo3 interface), however, I think that that is something to adhere to as breaking that interface will break current huggingface_hub versions.


rajatarya:
Love doing this with this upcoming release! Let's try to get the conservative dedup change in as well and snap that as 1.0

port8080 (approved):
https://semver.org/#spec-item-4
Inflection point indeed

rajatarya (approved):
Let's go!

PR #223, opened 2025-04-01
use configurable_constants macro for remote_client

assafvayner:
re comment on #221 . uses configurable_constants! macro to set boolean to download sequentially.


jgodlew (approved):
Thanks for making this change!

PR #226, opened 2025-04-02
Adding windows testing for CI

jgodlew:
* Adds a job to run tests on windows in addition to linux.
* Fixes linting for windows builds
* Fixes the LocalClient that tries to set files as read-only, which, during tests, will break on windows due to file deletion behaving differently than on Linux.
* Identified an issue with the disk cache deleting items if there are simultaneous `put`s to the cache for the same key, range. This is fine on unix, but on windows, causes errors (again, due to file deletion behavior differences). This change mitigates the issue, allowing huggingface_hub tests to pass on windows, but opens up another issue of us needing to vet our filesystem deletes (e.g. cache eviction) for correctness on windows.


chunk_cache/src/disk.rs
```diff
@@ -372,8 +372,13 @@ impl DiskCache {
         // removing by index in reverse to guarantee lower-index items aren't shifted/moved
         for item_idx in to_remove.into_iter().rev() {
             let item = items.swap_remove(item_idx);
-            overlapping_item_paths.insert(self.item_path(key, &item)?);
-            total_bytes_rm += item.len;
+            // We only remove from the disk if the item found is not equal to the cache_item
```
seanses, line 375: But this already effectively drops the item from the state by the above line `items.swap_remove(item_idx)`? Probably we shouldn't enqueue this item in the first place if an identical one is found in the state: line #363 `if item.range.start >= cache_item.range.start && item.range.end <= cache_item.range.end` should change to exclude this case

.github/workflows/ci.yml
```diff
@@ -40,3 +40,18 @@ jobs:
       - name: Build and Test
         run: |
           cargo test --verbose --no-fail-fast --features "strict"
+  build_and_test-win:
+    runs-on: windows-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Install Rust 1.86
```
rajatarya, line 48: Any concerns with using 1.86 for Windows when we're using 1.79 for rest of xet-core?

chunk_cache/src/disk.rs
jgodlew: Yeah, I had originally thought to change the conditional in line 363 to check for equality, however, we still want to remove the duplicate item from `items` (just not from disk). Unfortunately, mutating `items` isn't possible while it is being iterated over, so I went with this approach as we are looping over the duplicates anyways. 

We could make this better by integrating a singleflight of some sort over the cache put (then there shouldn't be duplicates at all), but that would require a larger refactor of the cache.

seanses: I see that makes sense.

seanses (approved):
LGTM!

rajatarya (approved):

.github/workflows/ci.yml
jgodlew: I merged in Assaf's PR (https://github.com/huggingface/xet-core/pull/230) to upgrade to 1.86 for the unix builds, so all builds should be using rust 1.86 now.

PR #229, opened 2025-04-02
configure cache size and default endpoint

assafvayner:
allows cache size and default endpoint to be configured with env vars:

XET_CHUNK_CACHE_SIZE_BYTES
XET_DEFAULT_CAS_ENDPOINT

Fix XET-276, concurrency controls already configurable from previous PR.

### DIFF
--- a/cas_client/src/remote_client.rs
+++ b/cas_client/src/remote_client.rs
@@ -80,13 +80,17 @@ impl RemoteClient {
     ) -> Self {
         // use disk cache if cache_config provided.
         let chunk_cache = if let Some(cache_config) = cache_config {
-            debug!(
-                "Using disk cache directory: {:?}, size: {}.",
-                cache_config.cache_directory, cache_config.cache_size
-            );
-            chunk_cache::get_cache(cache_config)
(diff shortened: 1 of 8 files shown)


cas_client/src/remote_client.rs
```diff
@@ -80,13 +80,17 @@ impl RemoteClient {
...
         let chunk_cache = if let Some(cache_config) = cache_config {
-            debug!(
-                "Using disk cache directory: {:?}, size: {}.",
-                cache_config.cache_directory, cache_config.cache_size
-            );
-            chunk_cache::get_cache(cache_config)
-                .log_error("failed to initialize cache, not using cache")
-                .ok()
+            if cache_config.cache_size == 0 {
+                None
```
rajatarya, line 85: Please add an info! indicating no chunk cache is used because cache size is set to 0.

rajatarya (approved):
Think we should emit an `info!` since setting size to 0 is unexpected.

Also, can you take the env variables in this PR and put them into a Linear issue for updating docs - so we can document these?

PR #230, opened 2025-04-03
bump xet-core to rust 1.86

assafvayner:


cas_client/Dockerfile
assafvayner, line 1: this file is not used anywhere.

jgodlew (approved):

PR #236, opened 2025-04-08
rm tmpdir use in safe file creater

assafvayner:
Fix XET-477.

Removes all (non-test) uses of temp directories from SafeFileCreator, changes functionality of SafeFileCreator::new_unnamed to take a directory in which to operate. Changes usage of SafeFileCreator::new_unnamed to use the shard cache directory.


seanses (approved):
LGTM!

PR #239, opened 2025-04-09
Avoid a potential divide by zero in the cache

ylow:
Should fix https://linear.app/xet/issue/XET-489/bug-divide-by-zero-on-download-with-llama4-model

Now, I am not *entirely* sure why this happens. It could happen if the target num_items drop to 0. Possibly because cache is disabled?


rajatarya (approved):
This looks good to me. I'd prefer to see a unit-test that covers this case so we don't regress this in the future, but since this is urgent we can file an issue to add the test later on.

ylow:
Oh... clippy...

PR #241, opened 2025-04-09
high performance mode

assafvayner:
fix XET-481

Adds high performance mode configurations.


ylow (approved):

data/src/constants.rs
```diff
@@ -15,13 +15,22 @@ utils::configurable_constants! {
 
     /// The maximum number of simultaneous xorb upload streams.
     /// The default value is 8 and can be overwritten by environment variable "XET_CONCURRENT_XORB_UPLOADS".
```
rajatarya, line 17: This comment seems stale now.

rajatarya: Rather: rename the comment to match current environment variable.

xet_threadpool/src/threadpool.rs
```diff
@@ -215,14 +204,12 @@ impl Display for ThreadPool {
...
 /// This is a simple wrapper around tokio's runtime, with some default settings.
 /// Intentionally unwrap this because if it fails, the application should not continue.
-fn new_threadpool(maximum_worker_threads: bool) -> Result<TokioRuntime, MultithreadedRuntimeError> {
+fn new_threadpool() -> Result<TokioRuntime, MultithreadedRuntimeError> {
     #[cfg(not(target_family = "wasm"))]
     let mut builder = TokioRuntimeBuilder::new_multi_thread();
     #[cfg(target_family = "wasm")]
     let mut builder = TokioRuntimeBuilder::new_current_thread();
-    if !maximum_worker_threads {
-        builder.worker_threads(THREADPOOL_NUM_WORKER_THREADS); // 4 active threads
```
rajatarya, line 224: Help me understand, how many threadpool threads exist now? With this line removed, what happens?

assafvayner: Without this condition tokio will default to have the number of threads be the number of cores on the machine.

data/src/constants.rs
assafvayner: https://github.com/huggingface/xet-core/pull/247

utils/src/constant_declarations.rs
```diff
@@ -112,3 +116,32 @@ macro_rules! test_set_globals {
...
+        val == 1
+    } else {
+        false
+    }
+}
+
+lazy_static! {
+    /// To set the high performance mode to true, set either of the following environment variables to 1 or true:
+    ///  - HF_XET_HIGH_PERFORMANCE
+    ///  - HF_XET_HP
```
jgodlew, line 140: nit: should we also check for: `HF_HUB_ENABLE_HF_TRANSFER` to set the high performance mode too? If someone was already using hf_transfer, then they are likely more of a power user, so we should default their perf to higher.

jgodlew (commented):
LGTM! Just a question on whether we should check for hf_transfer env as well.

utils/src/constant_declarations.rs
ylow: Mildly agnostic about this though as it can be confusing. If we do we might need to print a nice message like "turning on HF_XET_HIGH_PERFORMANCE" since we detected "HF_HUB_ENABLE_HF_TRANSFER" or something like that... Not confident about the right approach here.

assafvayner: I am not 100% sure what will happen with the hub library, but if you have HF transfer downloaded and enabled it might be used over hf-xet? if not then this is still an odd behavior IMO

PR #242, opened 2025-04-09
Adding a pre-release workflow to test huggingface_hub

bpronan:
Since changes to xet-core and hf-xet can break huggingface_hub, we're adding automation to test the whole cross-platform huggingface_hub test suite against our RC builds.

This uses the same trick that huggingface_hub uses with the transformers, datasets, and diffusers libraries. It creates a ci branch on huggingface_hub which triggers the cross-platform huggingface_hub test suite.


.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,80 @@
+name: hf-xet prerelease
+
+on:
+  push:
+    tags:
+      - v*rc*
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Tag to test (e.g., 1.0.3rc2)"
```
assafvayner: does the tag need to start with a "v"? then the example in description should have the v

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,80 @@
...
+          git config user.email "bot+xet@huggingface.co"
+
+      - name: Wait for prerelease to be out on PyPI
+        run: |
+          VERSION=${{ steps.get-version.outputs.VERSION }}
+          echo "Waiting for hf_xet==${VERSION} to be available on PyPI"
+          while ! pip install hf_xet==${VERSION}; do
+            echo "hf_xet==${VERSION} not available yet, retrying in 15s"
+            sleep 15
+          done
```
assafvayner: non-blocker (I see this is what hf_hub does too), is it at all possible to make this job run strictly after the pypi upload job? Is that a part of a different workflow?

don't spend time on this.

assafvayner (approved):
LGTM assuming this is tested.

seanses (approved):
LGTM!

.github/workflows/pre-release.yml
bpronan: The pypi upload is done by the release workflow. The basic release steps are:

1. cut a RC build with the release workflow which pushes to pypi
2. this will run and block on the pypi build being available
3. once it is it will create a test branch on huggingface_hub to run the full suite of python tests

bpronan: Renamed files and added a comment to hopefully clarify this.

PR #245, opened 2025-04-10
Running huggingface_hub xet tests on PRs

bpronan:
Adding huggingface_hub xet tests to run on every PR and checkin. 

We build the hf_xet wheel, clone huggingface_hub, and run the xet tests in huggingface_hub agains the new build.

### DIFF
--- a/.github/workflows/hf-xet-tests.yml
+++ b/.github/workflows/hf-xet-tests.yml
@@ -16,7 +16,9 @@ jobs:
   hub-python-tests:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      # checkout out xet-core
+      - uses: actions/checkout@v4 
+      # checkout out huggingface_hub
       - uses: actions/checkout@v4
         with:
           repository: huggingface/huggingface_hub
--- a/.github/workflows/hf-xet-tests.yml
+++ b/.github/workflows/hf-xet-tests.yml
@@ -1,0 +1,42 @@
+name: Test huggingface_hub xet tests
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - '*'
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  hub-python-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/hf-xet-tests.yml
```diff
@@ -0,0 +1,42 @@
+name: Test
```
assafvayner: rename to Test with huggingface_hub?

assafvayner (approved):
LGTM

.github/workflows/hf-xet-tests.yml
```diff
@@ -0,0 +1,42 @@
...
+      - '*'
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  hub-python-tests:
+    runs-on: ubuntu-latest
```
jgodlew, line 17: should we also run on windows too?

jgodlew (approved):
LGTM! We should probably check windows integration too.

.github/workflows/hf-xet-tests.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/checkout@v4
```
seanses, line 22: I'm curious why this `- uses: actions/checkout@v4` is repeated here?

.github/workflows/hf-xet-tests.yml
bpronan: @rajatarya was concerned about cost/build times, so we decided to only run the xet specific tests on linux for now.

The full platform suite will be [run in the pre-release/RC workflow](https://github.com/huggingface/xet-core/pull/242) before we push any new releases, so that should cover us there.

.github/workflows/hf-xet-tests.yml
bpronan: The first one is to checkout `xet-core`. The second one is to checkout `huggingface_hub`.

This is a good question. I've added comments to explain this.

.github/workflows/hf-xet-tests.yml
bpronan: If we find Windows tests breaks are common enough, we can always add them in here.

seanses (approved):

PR #247, opened 2025-04-11
fix comment on max concurrent downloads

assafvayner:
Addressing final comments on #241

### DIFF
--- a/data/src/constants.rs
+++ b/data/src/constants.rs
@@ -14,7 +14,8 @@ utils::configurable_constants! {
     ref MDB_SHARD_LOCAL_CACHE_EXPIRATION_SECS: u64 = 3 * 7 * 24 * 3600;
 
     /// The maximum number of simultaneous xorb upload streams.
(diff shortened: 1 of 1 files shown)


data/src/constants.rs
```diff
@@ -14,7 +14,8 @@ utils::configurable_constants! {
     ref MDB_SHARD_LOCAL_CACHE_EXPIRATION_SECS: u64 = 3 * 7 * 24 * 3600;
 
     /// The maximum number of simultaneous xorb upload streams.
-    /// The default value is 8 and can be overwritten by environment variable "XET_CONCURRENT_XORB_UPLOADS".
+    /// can be overwritten by environment variable "HF_XET_CONCURRENT_XORB_UPLOADS".
+    /// The default value changes from 8 to 100 when "High Performance Mode" is enabled
```
rajatarya, line 18: This still seems incorrect - HF_XET_CONCURRENT_UPLOADS seems to be the new name. XORB isn't there, right?

assafvayner: great catch!

assafvayner: fixed, resolving

PR #250, opened 2025-04-11
use safe file writer in local client

assafvayner:
Attempted to identify the error we keep getting on windows tests. Tracked it down to persisting the tempfile in LocalClient. While I'm not 100% confident using SafeFileCreator will fix this, I expect it will make it a bit more consistent.


seanses (approved):
LGTM!

PR #252, opened 2025-04-14
Segmented download and refresh fetch info on 403

seanses:
This PR implements segmented download such that the file reconstruction and fetch info is not queried in its entirety at the beginning of the download, but queried in segments. This
- reduces start delay due to fetching file info: it takes 8s to get info for a 100 GB file, and an estimate of 40s for a 500 GB file;
- parallelizes file info fetching and range downloading;
- reduces probability of pre-signed URL expiration;
- auto refreshes a segment of file fetch info on pre-signed URL expiration;
- dynamically tunes range download degree of concurrency

Fix XET-270


cas_types/src/lib.rs
```diff
@@ -27,8 +30,41 @@ pub type FileRange = Range<u64>;
...
+    }
+
+    pub fn len(&self) -> u64 {
+        self.end - self.start
+    }
+}
+
 // note that the standard PartialOrd/Ord impls will first check `start` then `end`
-#[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq, PartialOrd, Ord, Default, Hash)]
+#[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq, PartialOrd, Ord, Default, Hash, Copy)]
```
assafvayner: Can we make this impl below

`impl <Idx: Copy> Copy for Range<Idx>`

seanses: Done

cas_client/src/remote_client.rs
```diff
@@ -813,193 +787,4 @@ mod tests {
         // Assert
         assert!(result.is_ok());
     }
-
-    #[test]
-    fn test_reconstruct_file_to_writer() {
```
seanses, line 818: I'm rewriting this test due to API change.

cas_client/src/remote_client.rs
```diff
@@ -324,314 +311,298 @@ impl RemoteClient {
...
-    /// outputting the file content via indicated OutputProvider.
-    ///
-    /// Unlike `reconstruct_file_to_writer`, this function will spawn a task for writing each
-    /// term to the output, with each task writing to its part of the file.
-    pub async fn reconstruct_file_to_writer_parallel(
+    // Segmented download such that the file reconstruction and fetch info is not queried in its entirety
+    // at the beginning of the download, but queried in segments. Range downloadeds are executed with
+    // a certain degree of parallilism, and so does writing out to storage. Ideal when the external
+    // storage is fast at seeks, e.g. RAM or SSDs.
+    async fn reconstruct_file_to_writer_segmented_parallel_write(
```
seanses, line 458: This function is 90+% identical to the `reconstruct_file_to_writer_segmented`, with the only difference being
1. using `JoinSet` to store download futures
2. each download task is `TermDownloadAndWrite` instead of `TermDownload`
3. writing out to file is done inside `TermDownloadAndWrite::run()`.

I'm keeping these two separated for now so @assafvayner can make optimizations to this parallel write situation easily. Will consider merging these two later.

cas_client/src/error.rs
```diff
@@ -73,3 +82,33 @@ impl From<utils::errors::SingleflightError<CasClientError>> for CasClientError {
         }
     }
 }
+
+impl<T> From<std::sync::PoisonError<T>> for CasClientError {
+    fn from(value: std::sync::PoisonError<T>) -> Self {
+        CasClientError::InternalError(anyhow!("{value:?}"))
```
assafvayner: Can you define and call (optionally `#[inline]`) something like:

```
impl CasClientError {
  pub fn internal(value: impl std::fmt::Debug) -> Self {
    CasClientError::InternalError(anyhow!("{value:?}"))
  }
}
```

we repeat that line often enough, plus it let's us clean up other places where we return an InternalError variant.

cas_client/src/remote_client.rs
```diff
@@ -324,314 +311,298 @@ impl RemoteClient {
...
 
-    /// Uses the reconstruction response and optionally requested FileRange to re-create a file,
-    /// outputting the file content via indicated OutputProvider.
-    ///
-    /// Unlike `reconstruct_file_to_writer`, this function will spawn a task for writing each
-    /// term to the output, with each task writing to its part of the file.
-    pub async fn reconstruct_file_to_writer_parallel(
+    // Segmented download such that the file reconstruction and fetch info is not queried in its entirety
+    // at the beginning of the download, but queried in segments. Range downloadeds are executed with
+    // a certain degree of parallilism, and so does writing out to storage. Ideal when the external
```
assafvayner: typos, downloads and parallelism

cas_client/src/remote_client.rs
```diff
@@ -324,314 +311,298 @@ impl RemoteClient {
...
+                            let data = term_download.run().await?;
+                            drop(permit);
+                            Ok(data)
+                        });
+                        running_downloads_clone.lock().await.push(future);
+                    },
+                    DownloadQueueItem::Metadata(fetch_info) => {
+                        // query for the file info of the first segment
+                        let segment_size = download_scheduler_clone.next_segment_size()?;
+                        info!("querying file info of size {segment_size}");
```
assafvayner: a lot of info!'s could be `debug!'s`

cas_client/src/remote_client.rs
```diff
@@ -324,314 +311,298 @@ impl RemoteClient {
...
-                // use offset_into_first_range for the first term if needed
-                max(0, offset_into_first_range as usize)
-            } else {
-                0
+
+        let mut writer = writer.get_writer_at(0)?;
+        let mut total_written = 0;
+        loop {
+            // polling the futures queue will return None when there's no future in it yet.
+            let Some(result) = running_downloads.lock().await.next().await else {
```
assafvayner: we are holding the lock to running_downloads while waiting for the next task to be done, why?

I'm not sure anymore that FuturesOrdered is the right thing to use here (might not have been before either?), since we are spawning on the threadpool we can just use a joinset

seanses: We can't use "JoinSet" in sequential writing mode because JoinSet is not ordered, and the tasks will be returned in the order they complete. But good point we should not hold the lock as it reduces parallelism.

cas_types/src/lib.rs
```diff
@@ -20,26 +22,106 @@ pub struct UploadXorbResponse {
...
 /// Start and exclusive-end range for file content
-pub type FileRange = Range<u64>;
+pub type FileRange = Range<u64, _F>;
 /// Start and inclusive-end range for HTTP range content
-pub type HttpRange = Range<u32>;
+pub type HttpRange = Range<u64, _H>;
+
+impl FileRange {
+    pub fn full() -> Self {
+        Self {
```
jgodlew: nit: should we use `Self::new(0, u64::MAX)` instead to simplify the construction here and in `take_segment`?

cas_client/src/download_utils.rs
```diff
@@ -0,0 +1,400 @@
...
+        if metrics.n_retries_on_403 > 0 {
+            info!("detected retries on 403, shrinking segment size by one range");
+            let mut num_range_in_segment = self.n_range_in_segment.lock()?;
+            if *num_range_in_segment > 1 {
+                *num_range_in_segment -= 1;
+            }
+            self.n_concurrent_download_task.forget_permits(1);
+        } else {
+            // TODO: check download speed and consider if we should increase or decrease
+            info!("expanding segment size by one range");
```
seanses, line 271: Also increase segment size

cas_client/src/download_utils.rs
```diff
@@ -0,0 +1,400 @@
+use std::collections::HashMap;
```
jgodlew, line 1: It would be good to have some unit tests for these structs / methods to verify that they're storing and fetching the right things.

cas_client/src/remote_client.rs
```diff
@@ -324,314 +309,280 @@ impl RemoteClient {
...
-        while let Some(result) = handles.next().await {
-            match result {
-                Ok(Ok(len_written)) => {
-                    progress_updater.as_ref().inspect(|updater| updater.update(len_written));
-                    total_written += len_written;
-                },
-                Ok(Err(e)) => Err(e)?,
-                Err(e) => Err(CasClientError::Other(format!("Error joining download task {e:?}")))?,
+        let mut remaining_total_len = total_len;
+        while let Some(item) = task_rx.recv().await {
```
jgodlew, line 513: it seems that the same task is both the producer and consumer of this queue. We probably don't need to have it and can replace it with a `VecDeque`.

Although, as discussed offline, we might want to instead have separate producer/consumer tasks to avoid potential hiccups in processing where we are stuck waiting for the next segment's metadata to be fetched from CAS (can be changed as a separate PR).

jgodlew (approved):
Overall, the code looks good.

As discussed in the meeting, we might want to look into having the producer and consumer of the queue be different tasks: the producer fetches the next segment's metadata from CAS and pushes Term tasks into the (bounded) queue. The consumer task pulls off of the queue and spawns the Term fetching tasks, adding them to the JoinSet/running_download channel. This would reduce the risk of "hiccups" in downloads where we've downloaded a segments terms and are stuck waiting for CAS to provide the next set of segment metadata.

Also, some more tests would be good.

cas_client/src/remote_client.rs
```diff
@@ -324,314 +309,280 @@ impl RemoteClient {
...
+        //
+        // If the queue item is `DownloadQueueItem::Metadata`, it fetches the file reconstruction info
+        // of the first segment, whose size is linear to `num_concurrent_range_gets`. Once fetched, term
+        // download tasks are enqueued and spawned with the degree of concurrency equal to `num_concurrent_range_gets`.
+        // After the above, a task that defines fetching the remainder of the file reconstruction info is enqueued,
+        // which will execute after the first of the above term download tasks finishes.
+        let threadpool = self.threadpool.clone();
+        let chunk_cache = self.chunk_cache.clone();
+        let range_download_single_flight = self.range_download_single_flight.clone();
+        let download_scheduler = DownloadScheduler::new(*NUM_CONCURRENT_RANGE_GETS);
```
assafvayner, line 351: The download scheduler being created within the context of a single download op means that if we're downloading multiple files in parallel, they can each have X concurrent downloads going where X is the configured value. I think this semaphore should be some kind of global singleton.

cas_client/src/download_utils.rs
```diff
@@ -0,0 +1,400 @@
...
+            .log_error("invalid response from CAS server: failed to match hash in fetch_info")?
+            .clone();
+
+        Ok((fetch_term, v))
+    }
+
+    pub async fn query(&self) -> Result<Option<(u64, Vec<CASReconstructionTerm>)>> {
+        let Some(manifest) = get_reconstruction_with_endpoint_and_client(
+            &self.endpoint,
+            &self.client,
```
assafvayner, line 97: this reqwest client is used both to get the reconstruction info, and later is cloned to use to communicate with S3/CF. This means that we are sending our CAS token to S3/CF which is considered bad for security reasons. We have 2 separate http clients for this reason in the remote client, one which has auth configured for CAS, and one without.

cas_client/src/download_utils.rs
seanses: Added

cas_client/src/remote_client.rs
seanses: Will change to separate producer/consumer in a later PR

cas_client/src/remote_client.rs
seanses: I'm not sure. This follows our current setting that each file has its own range fetch concurrency of `NUM_CONCURRENT_RANGE_GETS` and then the global fetch concurrency is `NUM_CONCURRENT_RANGE_GETS * MAX_CONCURRENT_DOWNLOADS`. This is in our published tuning doc so I'm hesitated to change.

cas_client/src/remote_client.rs
seanses: Done

cas_client/src/error.rs
seanses: Done

cas_client/src/remote_client.rs
assafvayner: In the docs it's per file? I reckon we will want to change both then, can wait for later

seanses: Don't remember where the doc is, but definitely we'll come back to tune all these parameters.

PR #253, opened 2025-04-15
ChunkCache interface returns chunk indices

assafvayner:
Needed step for: https://linear.app/xet/issue/XET-493/download-each-fetch-info-once-to-fulfill-all-repeat-uses

When using the cache to fulfill multiple ranges at once, we will need to be able to separate out chunks from the cache hit range. This is necessary to avoid multiple cache get requests.


jgodlew (approved):
LGTM!

PR #256, opened 2025-04-17
download and write by fetch info instead of by term

assafvayner:
Fix XET-493.

This change changes the standard parallelized download process to run with respect to fetch info's within a segment rather than on a term by term basis.

Why? terms which are adjacent or overlapping in a xorb utilize a single fetch info instance or a presigned url to CDN/S3. Downloading term by term *can* cause repeat downloads of the same fetch info url. We do not often see this happen because the cache & single flight are used to combat this happening. However the cache does not help if it is disabled/slow, and in general we can get rid of the latency even for a cache hit (file read into memory) with this change. Notably local cache hit is still better than a CDN download so we still check the cache first.


cas_client/src/remote_client.rs
```diff
@@ -451,32 +452,27 @@ impl RemoteClient {
         Ok(total_written)
     }
 
-    // Segmented download such that the file reconstruction and fetch info is not queried in its entirety
-    // at the beginning of the download, but queried in segments. Range downloads are executed with
-    // a certain degree of parallelism, and so does writing out to storage. Ideal when the external
-    // storage is fast at seeks, e.g. RAM or SSDs.
-    async fn reconstruct_file_to_writer_segmented_parallel_write(
+    async fn reconstruct_file_to_writer_segmented_parallel_write_by_fetch_info(
```
seanses, line 457: Let's add (back) comments.

cas_client/src/remote_client.rs
```diff
@@ -541,58 +528,99 @@ impl RemoteClient {
...
         while let Some(result) = running_downloads.join_next().await {
-            process_result(result, &mut total_written, &download_scheduler)?;
+            process_result(result)?;
         }
 
         Ok(total_written)
     }
 }
 
+async fn make_tasks(
```
seanses: A short summary of how this function re-arranges terms?

cas_client/src/remote_client.rs
```diff
@@ -743,6 +771,16 @@ mod tests {
     use super::*;
     use crate::interface::buffer::BufferProvider;
 
+    // test reconstruction contains 20 chunks, where each chunk size is 10 bytes
```
seanses: These constants are no longer used, probably added back due to merge error

cas_client/src/download_utils.rs
```diff
@@ -331,6 +501,8 @@ pub(crate) async fn get_one_term(
         cache.put(&key, &fetch_term.range, &chunk_byte_indices, &data)?;
     }
 
+    let mut data = data.to_vec();
```
seanses: Declaring `DownloadRangeResult::Data` as `(Arc<[u8]>,...` and calling `to_vec()` here introduced unnecessary clone. If declared as `Data(Vec<u8>,...)` then this is just transferring ownership. Why do we change in this way?

seanses: I noticed we are doing the same thing above `cached.data.to_vec()`. If `CacheRange` is `{ .. data: Vec<u8>}` we don't need this unnecessary clone.

cas_client/src/download_utils.rs
```diff
@@ -345,7 +517,8 @@ pub(crate) async fn get_one_term(
         // [0, len] -> [0, end_byte_index)
         data.truncate(end_byte_index);
```
seanses: These two operations can be combined into one:
`data[start_byte_index..end_byte_index].to_vec()`, which only copies `(end_byte_index - start_byte_index)` bytes, while `split_at` copies `end_byte_index` bytes into two vecs.

But if `start_byte_index == 0`, this can just be truncate.

For reference: https://github.com/huggingface/xet-core/blob/dc02eba2dac127614cd043b3a3b806f0c490ae3e/cas_client/src/download_utils.rs#L189

cas_client/src/download_utils.rs
```diff
@@ -398,7 +612,7 @@ async fn download_range(
         response.bytes_stream().map_err(std::io::Error::other),
     )
     .await?;
-    Ok(DownloadRangeResult::Data(data, chunk_byte_indices))
+    Ok(DownloadRangeResult::Data(data.into(), chunk_byte_indices.into()))
```
seanses: Reading the impl of `into()` here, looks like it's actually another unnecessary clone/memcpy: https://doc.rust-lang.org/src/alloc/sync.rs.html#3746

cas_client/src/remote_client.rs
```diff
@@ -1165,8 +1214,16 @@ mod tests {
...
+            assert_eq!(test.expected_data.len() as u64, resp.unwrap(), "response len mismatch {test}");
+            assert_eq!(test.expected_data.len(), buf.len(), "written len mismatch {test}");
+            assert_eq!(test.expected_data[..100], buf.value()[..100], "response data head mismatch {test}");
+            let tail_start = test.expected_data.len() - 100;
+            assert_eq!(
+                test.expected_data[tail_start..],
+                buf.value()[tail_start..],
+                "response data tail mismatch {test}"
+            );
+            assert_eq!(test.expected_data, buf.value(), "response data mismatch {test}");
```
seanses, line 1233: I think this assert covers the above two, are they for debugging purpose?

cas_client/src/remote_client.rs
```diff
@@ -485,35 +481,26 @@ impl RemoteClient {
...
 
-        let mut total_written = 0;
-        let mut remaining_total_len = total_len;
-        while let Some(item) = task_rx.recv().await {
+        while let Some(item) = task_queue.pop_front() {
             // first try to join some tasks
             while let Some(result) = running_downloads.try_join_next() {
-                process_result(result, &mut total_written, &download_scheduler)?;
+                process_result(result)?;
+                // process_result(result, &mut total_written, &download_scheduler)?;
```
seanses: can delete this

cas_client/src/remote_client.rs
```diff
@@ -541,58 +528,99 @@ impl RemoteClient {
...
-                        remaining_total_len -= take;
-                        remaining_segment_len -= take;
-                        debug!("enqueueing {download_and_write_task:?}");
-                        task_tx.send(DownloadQueueItem::Term(download_and_write_task))?;
+                    // define the term download tasks
+                    debug!("enqueueing {} download tasks", tasks.len());
+                    for task_def in tasks {
+                        debug!("enqueueing {task_def:?}");
+                        task_queue.push_back(DownloadQueueItem::DownloadTask(task_def));
+                        // task_tx.send(DownloadQueueItem::DownloadTask(task_def))?;
```
seanses: can delete this

cas_client/src/remote_client.rs
```diff
@@ -541,58 +528,99 @@ impl RemoteClient {
...
-                            },
-                            write_offset: total_len - remaining_total_len,
-                            output: writer.clone(),
-                        };
+                    let tasks = make_tasks(
+                        &segment,
+                        terms,
+                        offset_into_first_range,
+                        segment_size,
+                        total_len,
```
seanses, line 543: BUG: this parameter is called "remaining_total_len" but this never gets updated, i.e. ranges will be written at incorrect offset for 2+ segments.

cas_client/src/download_utils.rs
```diff
@@ -211,25 +211,195 @@ pub(crate) struct TermDownloadAndWrite {
...
+        http_client: Arc<ClientWithMiddleware>,
+    ) -> Self {
+        Self {
+            fetch_info,
+            chunk_cache,
+            http_client,
+        }
+    }
+
+    pub fn generate(&self, hash: MerkleHash, range: ChunkRange) -> XorbRangeDownload {
```
seanses: This generator logic is very simple that just copies over utilities, I feel it's not necessary to abstract this.

seanses (changes requested):
There's a bug that write offset is not updated correctly. Some memcpy can also be avoided.

cas_client/src/download_utils.rs
assafvayner: The cache returns an Arc<[u8]> right now so a cache hit is a clone, I tried deferring the clone down as much as possible. In this function I will try to see how I can get around it if possible. Note that this function is also only used by the sequential download (I'll check if I can merge the one used for the non-sequential one somehow).

cas_client/src/remote_client.rs
assafvayner: yes, I'd like to leave them here to make it easier to diagnose, if you just compare the 2 values altogether you get a massive printed output that doesn't help debug, it is easier to debug first checking subsections.

cas_client/src/download_utils.rs
assafvayner: I agree, but then we get into a clippy too many parameters lint issue, I can either disable the lint or keep this, I don't really care which to do, I usually try to avoid disabling lints but for no particularly good reason.

cas_client/src/remote_client.rs
assafvayner: added a comment

cas_client/src/remote_client.rs
assafvayner: this turns out to need to be passed in total_len in this case, the calculation ensures not writing past the end of the requested range.

There was an additional bug here though that each segment began writing at byte 0.

seanses: If you derive the remaining_total_len correctly, then the start offset for each segment is total_len - remaining_total_len

assafvayner:
closing in favor of #320

PR #257, opened 2025-04-18
make utils wasm-compat

assafvayner:
Progress for wasm, this is required before any other crate that uses utils can even attempt to export anything that is wasm compat.

This doesn't make all of utils wasm compat, it just excludes things we don't care about for now e.g. singleflight.


seanses (approved):
lgtm!

PR #263, opened 2025-04-23
Updating the release jobs to write the new version

bpronan:
Setting up the release script to write the Cargo.toml version before uploading to PyPi.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,6 +38,11 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: 3.x
+      - name: Update version in toml
+        run: |
+          TAG=${{ github.event.inputs.tag }}
+          VERSION=${TAG#v}
+          sed -i '/^version /s/=.*$/= '"$VERSION"'/' hf_xet/Cargo.toml
       - name: Build wheels
         uses: PyO3/maturin-action@v1
         with:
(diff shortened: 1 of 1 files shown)


rajatarya (approved):
With ChatGPT help verified that: `VERSION=${TAG#v}` does string manipulation and removes the prefix "v" from the tag name.

LGTM!

.github/workflows/pre-release-testing.yml
```diff
@@ -20,14 +20,17 @@ jobs:
...
         run: |
           if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
-            echo "VERSION=${{ inputs.tag }}" >> $GITHUB_OUTPUT
+            TAG=${{ inputs.tag }}
           else
-            echo "VERSION=${GITHUB_REF#refs/tags/}" >> $GITHUB_OUTPUT
+            TAG=${GITHUB_REF#refs/tags/}
           fi
+          SEM_VERSION=${TAG#v}
+          TRIMMED_VERSION=${SEM_VERSION//-}
```
rajatarya, line 32: Why is this line needed?

From what I understand, this would be:
TAG="v1.0.4-rc0"
SEM_VERSION="1.0.4-rc0"
TRIMMED_VERSION="1.0.4rc0"

Why add the - to the SEM version if we need to remove it later?

rajatarya (approved):
I'm assuming you're testing this as you go..

.github/workflows/pre-release-testing.yml
rajatarya: Good discussion, inconsistency between Rust semantic versioning and Python semantic versioning.

rajatarya (approved):

PR #264, opened 2025-04-23
Skipping the version update on empty tag

bpronan:


rajatarya (approved):
This way whatever version string is in the repo will be the one used for Release workflow (when not specifying the tag/version).

PR #266, opened 2025-04-24
Fixing hub test branch creation

bpronan:
Fixing the pre-release testing workflow. There were two issues which this resolves:

1. test branches on huggingface_hub need to start with `ci_`.
2. our sed command wasn't catching any of the cases we were using in setup.py on huggingface_hub:
```
hf_xet>=1.0.0,<2.0.0
hf-xet>=1.0.0,<2.0.0; platform...
```
We should be handling these cases with some basic future proofing now.


rajatarya (approved):
Good talk through it!

PR #267, opened 2025-04-24
Reduce Usage of Compression Format Detection

ylow:
We use a KL-divergence based predictor to select between the LZ4 and BG-LZ4 schemes. This is done for every chunk and seems to consume about 20% CPU time. Instead, we only run this for the first Xorb, lock the scheme and use it from then on.

 - Removes compression scheme from RemoteClient construction and instead as an argument to put (xorb)
 - FileUploadSession is used to track compression scheme
 - Other changes demanded by clippy.

### DIFF
--- a/cas_client/src/interface.rs
+++ b/cas_client/src/interface.rs
@@ -36,7 +36,7 @@ pub trait UploadClient {
         hash: &MerkleHash,
         data: Vec<u8>,
         chunk_and_boundaries: Vec<(MerkleHash, u32)>,
-        compression: Option<CompressionScheme>
+        compression: Option<CompressionScheme>,
     ) -> Result<usize>;
 
     /// Check if a XORB already exists.
--- a/cas_client/src/interface.rs
+++ b/cas_client/src/interface.rs
@@ -5,6 +5,7 @@ use std::path::PathBuf;
 use std::sync::Arc;
 
 use async_trait::async_trait;
+use cas_object::CompressionScheme;
 use cas_types::{FileRange, QueryReconstructionResponse};
 use mdb_shard::shard_file_reconstructor::FileReconstructor;
 use merklehash::MerkleHash;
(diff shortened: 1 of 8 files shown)


seanses (approved):

seanses:
Need to be careful when this is adapted to different usage patterns (e.g. web workers), that a file upload session should not be responsible for uploading files to different repos (especially a mix of models and datasets). cc @assafvayner

assafvayner:
Is it possible to tie the compression scheme to a `Cleaner`?

ylow:
Plausible but more annoying as the Cleaner plumbs it at a different level (chunk production) which makes the tracking more intrusive. Here it is really using the first Xorb produced to set the compression for the remaining Xorbs. 
Will a FileUploadSession ever be used for more than one repo? Given that it has a `repo_id: Option<String>`, it should not be used for than one repo at a time anyway.

assafvayner:
@seanses I think what this tells me is that in the XetSession implementation we should hold a map of repo_id -> FileUploadSession, this should not be heavy, I expect that there will usually be only 1 repo being referenced, but we should be defensive against the case of 2. For all the heavy objects such as ShardManager we can pass/clone them with Arc's right? Unless we need the shard manager to be separated?

seanses:
> @seanses I think what this tells me is that in the XetSession implementation we should hold a map of repo_id -> FileUploadSession, this should not be heavy, I expect that there will usually be only 1 repo being referenced, but we should be defensive against the case of 2. For all the heavy objects such as ShardManager we can pass/clone them with Arc's right? Unless we need the shard manager to be separated?

ShardManagers are separated by the directory they manage. https://github.com/huggingface/xet-core/blob/2fe54ea94d8321769e5e6d9345192e1e5f63270a/mdb_shard/src/shard_file_manager.rs#L22

seanses:
> Plausible but more annoying as the Cleaner plumbs it at a different level (chunk production) which makes the tracking more intrusive. Here it is really using the first Xorb produced to set the compression for the remaining Xorbs. Will a FileUploadSession ever be used for more than one repo? Given that it has a `repo_id: Option<String>`, it should not be used for than one repo at a time anyway.

Ah yes right, the AuthConfig should be unique for each "operation", so maybe a map of "AuthConfig" -> FileUploadSession @assafvayner

PR #271, opened 2025-04-25
Make dedup critical crates compilation-compat with wasm

seanses:
1. Upgrade `rand` so we can use the new "wasm_js" feature for the underlying `getrandom` dependency.
2. Change deprecated rand `gen`, `gen_range` functions.
3. Restrict tokio features.
4. Change `tempdir` to `tempfile` because `tempdir` uses a very old non-wasm compat version of `rand`, and is merged into the latter and archived.
5. Clean out some unnecessary dependencies.
6. Moved a file from parutils to utils.


parutils/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-tokio = { version = "1.36", features = ["full"] }
+tokio = { version = "1.44", features = ["sync"] }
 async-scoped = { version = "0.7", features = ["use-tokio"] }
```
assafvayner, line 10: I think async-scoped with use-tokio feature is likely still going to cause issues?

mdb_shard/Cargo.toml
```diff
@@ -26,6 +25,11 @@ futures-util = "0.3.31"
 serde = { version = "1.0.215", features = ["derive"] }
 blake3 = "1.5.4"
 
+[target.'cfg(target_family = "wasm")'.dependencies]
+uuid = { version = "1", features = ["v4", "js"] }
```
assafvayner, line 29: if you include the "js" feature for non-wasm build is that an error, I think the exports will just get ignored?

mdb_shard/Cargo.toml
```diff
@@ -9,15 +9,14 @@ edition = "2021"
 merklehash = { path = "../merklehash" }
 thiserror = "2.0"
 utils = { path = "../utils" }
-tempdir = "0.3.7"
-tokio = { version = "1.36", features = ["full"] }
+tempfile = "3.19.1"
```
assafvayner, line 12: is this wasm compat??

Cargo.toml
```diff
@@ -19,7 +19,7 @@ members = [
     "xet_threadpool", 
     "deduplication"]
 
-exclude = ["hf_xet", "chunk_cache_bench"]
+exclude = ["hf_xet", "chunk_cache_bench", "wasm_xet"]
```
assafvayner, line 22: I don't see the wasm_xet files in this PR, can we also rename to hf_xet_wasm?

assafvayner (approved):

mdb_shard/Cargo.toml
seanses: This is compilation-compat but won't work in web workers so we won't use this. But tempdir is very old and uses a very old version of `rand`, which uses a very old version of `getrandom`, which doesn't support the new `wasm_js` feature.

Cargo.toml
seanses: OK. Done in #272

mdb_shard/Cargo.toml
seanses: I'm not sure about that

parutils/Cargo.toml
seanses: Just wanted to make it compilation-compat. We are not going to use in JS/web workers anyway.

PR #272, opened 2025-04-25
wasm poc

seanses:
This implements uploading through Xet protocol in WASM environment, and makes necessary changes to make dependent crates WASM compatible.
1. Uploading through Xet protocol is done in hf_xet_wasm crate;
2. Separate Cas Client trait definitions into upload and download functionality groups and disable download for WASM;
3. Disable Cas Client request retry in WASM environment, which isn't critical for a POC (until we have a retry strategy that doesn't depends on time);
4. Disable async CasObject deserialization;
5. Enable in-memory global dedup;


wasm_xet/web/wasm_xet_bg.wasm.d.ts
assafvayner, line 1: pls don't check the generated files in. Using wasm-pack should git ignore them!

cas_client/src/error.rs
```diff
@@ -95,6 +95,7 @@ impl PartialEq for CasClientError {
     }
 }
 
+#[cfg(not(target_family = "wasm"))]
```
rajatarya, line 98: Is this because concurrency doesn't exist in WASM, so no need for single-flight?

cas_client/src/http_client.rs
```diff
@@ -78,6 +78,7 @@ impl RetryConfig<No429RetryStrategy> {
 
 /// Builds authenticated HTTP Client to talk to CAS.
 /// Includes retry middleware with exponential backoff.
+#[allow(unused_variables)]
```
rajatarya: What is considered unused? Can this warning be removed?

cas_client/src/http_client.rs
```diff
@@ -86,14 +87,26 @@ pub fn build_auth_http_client<R: RetryableStrategy + Send + Sync + 'static>(
...
-    Ok(ClientBuilder::new(reqwest_client)
-        .maybe_with(auth_middleware)
-        .with(retry_middleware)
-        .maybe_with(logging_middleware)
-        .maybe_with(session_middleware)
-        .build())
+    #[cfg(not(target_family = "wasm"))]
+    {
+        let retry_middleware = get_retry_middleware(retry_config);
+        Ok(ClientBuilder::new(reqwest_client)
```
rajatarya: Why does this need the compile-flag over the entire `ClientBuilder::new`, can't the compile flag simply be on whever retry_middleware is None or Some(...)?

cas_client/src/interface.rs
```diff
@@ -1,211 +1,156 @@
...
-            buffer.idx = start;
-            Ok(Box::new(buffer))
+    /// A Client to the CAS (Content Addressed Storage) service to allow reconstructing a
+    /// pointer file based on FileID (MerkleHash).
+    ///
+    /// To simplify this crate, it is intentional that the client does not create its own http_client or
+    /// spawn its own threads. Instead, it is expected to be given the parallelism harness/threadpool/queue
+    /// on which it is expected to run. This allows the caller to better optimize overall system utilization
+    /// by controlling the number of concurrent requests.
+    #[cfg_attr(not(target_family = "wasm"), async_trait::async_trait)]
```
rajatarya, line 51: What do these two lines mean? Can you add a comment to explain how the config attribute for async_trait is different across wasm or not.

rajatarya (commented):
(Partially completed, will continue with latest.)

cas_client/src/http_client.rs
```diff
@@ -154,21 +133,34 @@ pub fn build_auth_http_client_no_retry(
 
 /// Builds HTTP Client to talk to CAS.
 /// Includes retry middleware with exponential backoff.
+#[allow(unused_variables)]
```
rajatarya: Why are there unused variables here?

cas_client/src/http_client.rs
```diff
@@ -296,7 +290,8 @@ impl Middleware for AuthMiddleware {
 
 pub struct SessionMiddleware(String);
 
-#[async_trait::async_trait]
+#[cfg_attr(not(target_family = "wasm"), async_trait::async_trait)]
```
rajatarya, line 273: A comment explaining the compilation flag changes would be helpful here - they both look so similiar but I imagine have very different behaviors.

cas_client/src/error.rs
assafvayner: A) yes
B) we can't compile the dependencies of the singleflight portion of the utils crate to WebAssembly so it is not exported.

cas_client/src/http_client.rs
assafvayner: I'll remove these. It only gives this error when linting in webassembly where the retry config arg is not used.

cas_client/src/interface.rs
assafvayner: Added to the hf_xet_wasm README.md, this change happens across many parts of the code so I think it would be too much to include it in all places where this is used.

Open to other places to put the commentary about it.

tldr for wasm env, the async_trait should not enforce `Send`'ness

cas_client/src/http_client.rs
assafvayner: changed this to be more reflective of the pattern you may be seeking.
I readded the retry middleware. I think we changed the type of body used such that it is now fine to use.

cas_client/src/http_client.rs
```diff
@@ -124,16 +89,22 @@ pub fn build_auth_http_client<R: RetryableStrategy + Send + Sync + 'static>(
...
     let logging_middleware = Some(LoggingMiddleware);
     let session_middleware = (!session_id.is_empty()).then(|| SessionMiddleware(session_id.to_owned()));
-    let retry_middleware = get_retry_middleware(retry_config);
-    let reqwest_client = reqwest::Client::builder()
-        .dns_resolver(Arc::from(GaiResolverWithAbsolute::default()))
-        .build()?;
-    Ok(ClientBuilder::new(reqwest_client)
+
+    let mut reqwest_client_builder = reqwest::Client::builder();
+    #[cfg(not(target_family = "wasm"))]
```
rajatarya: I think a comment here indicating why this is compiled away for WASM is a good idea. Future maintainers of this package might not understand this and know how to make changes here.

cas_client/src/http_client.rs
```diff
@@ -214,7 +175,8 @@ pub struct Api(pub &'static str);
 /// Adds logging middleware that will trace::warn! on retryable errors.
 pub struct LoggingMiddleware;
 
-#[async_trait::async_trait]
+#[cfg_attr(not(target_family = "wasm"), async_trait::async_trait)]
```
rajatarya, line 181: What does this pattern mean?

hf_xet_wasm/README.md
```diff
@@ -0,0 +1,82 @@
+# hf_xet_wasm: xet-core for WebAssembly
+
+This crate enables functionality to use the xet upload protocol from the browser with the use of a wasm based binary replicating the functionality of the `hf_xet` python library.
+Functionality included but not limited to chunking, global deduplication, xorb formation, xorb upload, shard formation, shard upload.
```
rajatarya, line 4: I would rephrase this to be explicit about what is included and what is missing - so future maintainers can understand what to expect from the rest of the repo. Something like: 

hf_xet_wasm has: chunking, global deduplication, xorb formation, xorb upload, shard formation, shard upload

hf_xet_wasm is missing: complete download support (xorbs, shards, chunk caching)

hf_xet_wasm/README.md
```diff
@@ -0,0 +1,82 @@
...
+- Install sfz:
+```bash
+cargo install sfz
+```
+
+- Serve the web
+```bash
+sfz --coi -r examples
+```
+
```
rajatarya, line 62: I would mention the examples directory in this README.

hf_xet_wasm/src/auth.rs
rajatarya, line 1: This file could use a comment block describing what it is - I believe this is the Rust side of the token refreshing code - maybe an example of how a JS application would use this would help explain it. Or point to an example using this code.

hf_xet_wasm/src/blob_reader.rs
rajatarya, line 1: This file needs some comments. Something that connects this to being necessary because huggingface.js exposes a BlobReader interface to read LFS files (or something like that).

hf_xet_wasm/src/configurations.rs
```diff
@@ -0,0 +1,28 @@
+use cas_object::CompressionScheme;
+use utils::auth::AuthConfig;
+
+#[derive(Debug)]
+pub struct DataConfig {
```
rajatarya, line 7: Why is the DataConfig struct defined here? Is this defined in `hf-xet` and so needs to be redefined here for webassembly?

rajatarya: Same for ShardConfig and TranslatorConfig - why are these structs defined in the wasm crate?

hf_xet_wasm/src/errors.rs
rajatarya, line 1: Why can't this use the standard errors crate? Why redefine these?

hf_xet_wasm/src/session.rs
```diff
@@ -0,0 +1,104 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct JsPointerFile {
+    pub file_size: f64,
+    pub file_hash: String,
+    pub sha256: String,
+}
+
+#[wasm_bindgen(js_name = "XetSession")]
+pub struct XetSession {
```
rajatarya, line 38: public structs should have _some_ comments.

hf_xet_wasm/src/sha256.rs
```diff
@@ -0,0 +1,68 @@
+use deduplication::Chunk;
+use merklehash::MerkleHash;
+use sha2::{Digest, Sha256};
+
+use super::errors::*;
+
+pub enum ShaGeneration {
```
rajatarya, line 10: comments

hf_xet_wasm/src/sha256.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    pub fn finalize(self) -> Result<MerkleHash> {
+        match self {
+            ShaGeneration::Value(hash) => Ok(hash),
+            ShaGeneration::Action(sha_generator) => sha_generator.finalize(),
+        }
+    }
+}
+
+pub struct ShaGenerator {
```
rajatarya, line 47: comments

hf_xet_wasm/src/wasm_file_cleaner.rs
```diff
@@ -0,0 +1,213 @@
...
+use tokio::sync::mpsc;
+use tokio_with_wasm::alias as wasmtokio;
+
+use super::errors::*;
+use super::wasm_deduplication_interface::UploadSessionDataManager;
+use super::wasm_file_upload_session::FileUploadSession;
+use crate::sha256::ShaGeneration;
+use crate::wasm_timer::Timer;
+
+enum CPUTask {
```
rajatarya, line 21: Can you add a comment here educating maintainers what the difference is between CurrentThread and WorkerThread? Why have this enum, what behavior does it drive?

hf_xet_wasm/src/wasm_file_cleaner.rs
```diff
@@ -0,0 +1,213 @@
...
+            dedup_manager: FileDeduper::new(UploadSessionDataManager::new(session), file_id),
+        }
+    }
+
+    pub async fn add_data(&mut self, data: &[u8]) -> Result<()> {
+        let chunks = match self.cpu_task {
+            CPUTask::CurrentThread((ref mut chunker, ref mut sha_generation)) => {
+                Self::add_data_to_cpu_task_in_current_thread(chunker, sha_generation, data).await
+            },
+            CPUTask::WorkerThread(_) => todo!(),
```
rajatarya: nit: might want the `todo!()` to have a string describing the current limitation, like: `todo!("no background support for cleaning data")`

hf_xet_wasm/src/wasm_file_cleaner.rs
```diff
@@ -0,0 +1,213 @@
...
+        }?;
+
+        // Run the deduplication interface here.
+        let msg = format!("dedupping {} chunks", chunks.len());
```
rajatarya: typo: dedupping -> deduping

hf_xet_wasm/src/wasm_file_cleaner.rs
```diff
@@ -0,0 +1,213 @@
...
+        // Update the sha256 generator
+        let _timer = Timer::new(format!("computing sha256 over {} chunks", chunks.len()));
+        sha_generation.update(&chunks);
+        drop(_timer);
+
+        Ok(chunks)
+    }
+
+    async fn add_data_to_cpu_task_in_worker_thread() -> Result<Vec<Chunk>> {
+        todo!()
```
rajatarya: explain limitation in `todo!()`

hf_xet_wasm/src/wasm_file_upload_session.rs
```diff
@@ -0,0 +1,179 @@
...
+            return Err(DataProcessingError::internal("register xorb after drop"));
+        };
+        xorb_uploader.upload_xorb(cas_object).await?;
+
+        Ok(())
+    }
+
+    pub async fn finalize(self: Arc<Self>) -> Result<()> {
+        // Register the remaining xorbs for upload.
+        let data_agg = take(&mut *self.current_session_data.lock().await);
```
rajatarya, line 133: How is `take()` different from just `self.current_session_data.lock().await;`? In other places in this impl the lock is received with a `.lock().await` - how come this code does a `take()` on top of that?

hf_xet_wasm/src/wasm_file_upload_session.rs
```diff
@@ -0,0 +1,179 @@
...
+
+        // Finalize the xorb uploads.
+        if let Some(mut xorb_uploader) = self.xorb_uploader.lock().await.take() {
+            xorb_uploader.finalize().await?;
+        }
+
+        // Now that all the tasks there are completed, there shouldn't be any other references to this session
+        // hanging around; i.e. the self in this shession should be used as if it's consuming the class, as it
+        // effectively empties all the states.
+        debug_assert_eq!(Arc::strong_count(&self), 1);
```
rajatarya, line 144: Should this be in the config compilation block for `debug_assertions`?

hf_xet_wasm/src/wasm_timer.rs
```diff
@@ -0,0 +1,31 @@
+use web_sys::console;
+
+static PROFILING: bool = false;
+
+pub struct Timer {
```
rajatarya: nit: I'd rename this to `LoggingTimer` or `ConsoleLogTimer` because nothing in the implementation (and no comments) indicate that it is a helper object to time operations in JS and report them to the console.

hf_xet_wasm/src/xorb_uploader.rs
```diff
@@ -0,0 +1,99 @@
...
+    async fn finalize(&mut self) -> Result<()>;
+}
+
+pub struct XorbUploaderLocalSequential {
+    client: Arc<dyn Client + Send + Sync>,
+    cas_prefix: String,
+}
+
+impl XorbUploaderLocalSequential {
+    #[allow(dead_code)]
```
rajatarya: why need this?

rajatarya (approved):
Overall this is great, thank you @seanses and @assafvayner for putting this together!

The vast majority of my comments are asking for more documentation/comments/explanation. It will be hard for _future you_ or anyone new to the repo to remember the context under which all the WASM changes were made, and that makes maintaining them without documentation risky. The more comments now the better.

I think with ~60m of effort to add comments to this PR it will be ready to merge.

Again, great work on this - really appreciate the effort.

hf_xet_wasm/src/wasm_file_upload_session.rs
assafvayner: https://doc.rust-lang.org/std/mem/fn.take.html

take replaces the value being referenced with a default, more importantly moving out the value referenced to use in the current scope.

hf_xet_wasm/src/wasm_file_upload_session.rs
assafvayner: actually, the #[cfg(debug_assertions)] block is not needed as long as we use the `debug_assert.*!` macros, the debug_assertions check is included in the macro.

hf_xet_wasm/src/xorb_uploader.rs
assafvayner: was used for linting, removing the macro it and letting the warning stick

cas_client/src/http_client.rs
assafvayner: it removed a lint warning in wasm-mode, removed it, and allowing warnings.

cas_client/src/http_client.rs
assafvayner: this is explained in hf_xet_wasm README.md; there's too many occurrences in other places to state it everywhere.

hf_xet_wasm/src/errors.rs
assafvayner: each crate has a wrapper over error types. In this case, we can't use the data crate, so we have this file.

hf_xet_wasm/src/configurations.rs
assafvayner: these are redefined because we need less configuration values and also no access to where the structs with those same names are defined in the data crate.

cas_client/src/http_client.rs
rajatarya: My main concern is that the wasm crate isn't something most xet-core maintainers will need to read/understand. So it might make sense to repeat or reference the message in wasm readme in a comment, something like: "# see hf_xet_wasm README.md for context

rajatarya (approved):
Love the changes, let's merge!

PR #273, opened 2025-04-25
`upload_bytes` and pointer file "removal"

bpronan:
We need to support the ability for clients to upload binary data directly to xet, so we are adding an `upload_bytes` function to `hf_xet`. 

This required some refactoring since files were baked into the upload process. Instead of lightly changing `PointerFile` to support this, we're taking the opportunity to refactor it out since it doesn't conceptually make sense in the workflow we need to support.

I took a light hand to the test refactoring. IMO, the bash integration tests should be turned into python/JS integration tests, but that seemed too invasive a change for a small feature like this.

I'm also taking the opportunity to clean up the `hf_xet` interface a bit. The whole pointer file concept leaks across the API, so we're moving towards a more conceptually accurate set of data structures. We have a `PyXetDownloadInfo` struct to support specifying the information we need to download a file, and we have a `PyXetUploadInfo` struct to report information on the file that was uploaded.

Note: since PyO3 doesn't have support for deprecation of the Python->Rust interface, we need to support both the old and new types until the next major version (IE, breaking) update to `hf_xet`. 

@jgodlew, I would love your opinion on the `hf_xet` interface. Could you take a very critical eye towards the struct and field naming for the python data structures? Since we renaming is a hassle, we want names that we can live with for a while.

Partially resolves XET-379.


data/src/xet_file.rs
```diff
@@ -0,0 +1,43 @@
...
+    pub fn hash(&self) -> &str {
+        &self.hash
+    }
+
+    /// Returns the parsed merkle hash of the file.
+    pub fn merkle_hash(&self) -> std::result::Result<MerkleHash, DataHashHexParseError> {
+        MerkleHash::from_hex(&self.hash).map_err(|e| {
+            error!("Error parsing hash value for file info {e:?}");
+            e
+        })
```
sirahd: [nit] use `log_error` for this
```suggestion
        MerkleHash::from_hex(&self.hash).log_error("Error parsing hash value for file info")
```

ylow:
This all seems reasonable and looks good!

hf_xet/src/lib.rs
```diff
@@ -106,53 +141,109 @@ fn try_parse_progress_updaters(funcs: Vec<Py<PyAny>>) -> PyResult<Vec<Arc<dyn Pr
...
+// This is used to support backward compatibility for PyPointerFile with old versions of huggingface_hub
+type PyPointerFile = PyXetDownloadInfo;
+
 #[pyclass]
 #[derive(Clone, Debug)]
-pub struct PyPointerFile {
+pub struct PyXetDownloadInfo {
     #[pyo3(get, set)]
-    path: String,
+    destination_path: String,
```
sirahd, line 148: Wouldn't renaming the field in the struct/object break the backward compatibility too?

sirahd (commented):
Please don't mind my unsolicited review, just popped in to take a look and left some comments. There's a mention of `PyXetUploadInfo` in the description but didn't see the implementation. Is that for later down the road?

hf_xet/src/lib.rs
bpronan: This was handled by adding an extra getter/setter to the trait implemenation.

We've had to change this around to handle a backwards compatibility story through PyO3.

hf_xet/src/lib.rs
```diff
@@ -32,7 +32,42 @@ fn convert_data_processing_error(e: DataProcessingError) -> PyErr {
 }
 
 #[pyfunction]
-#[pyo3(signature = (file_paths, endpoint, token_info, token_refresher, progress_updater, _repo_type), text_signature = "(file_paths: List[str], endpoint: Optional[str], token_info: Optional[(str, int)], token_refresher: Optional[Callable[[], (str, int)]], progress_updater: Optional[Callable[[int], None]], _repo_type: Optional[str]) -> List[PyPointerFile]")]
+#[pyo3(signature = (file_contents, endpoint, token_info, token_refresher, progress_updater, _repo_type), text_signature = "(file_contents: List[bytes], endpoint: Optional[str], token_info: Optional[(str, int)], token_refresher: Optional[Callable[[], (str, int)]], progress_updater: Optional[Callable[[int], None]], _repo_type: Optional[str]) -> List[PyXetUploadInfo]")]
+pub fn upload_bytes(
+    py: Python,
+    file_contents: Vec<Vec<u8>>,
```
jgodlew, line 38: Q: since huggingface_hub can support a `BinaryIO` for the file content (e.g. from some python I/O stream), do we expect to create a new method to support that type of input (provided we can find a way to translate python `BinaryIO` to a rust `Read`)?

data/src/data_client.rs
```diff
@@ -95,26 +95,55 @@ pub fn default_config(
     Ok(Arc::new(translator_config))
 }
 
+pub async fn upload_bytes_async(
+    thread_pool: Arc<ThreadPool>,
+    file_contents: Vec<Vec<u8>>,
```
jgodlew, line 100: nit: for extensibility, does it make sense to have this accept a `Vec<T>` where `T: Read`? Then we could merge the lower level `clean_bytes`/`clean_file` functions into a single `clean_reader`.

hf_xet/src/lib.rs
```diff
@@ -106,53 +140,129 @@ fn try_parse_progress_updaters(funcs: Vec<Py<PyAny>>) -> PyResult<Vec<Arc<dyn Pr
...
+    fn __str__(&self) -> String {
+        format!("{self:?}")
+    }
+
+    fn __repr__(self_: PyRef<'_, Self>) -> String {
+        let super_ = self_.as_super();
+        format!("PyPointerFile({}, {}, {})", super_.destination_path, super_.hash, super_.file_size)
+    }
+
+    fn get_path(self_: PyRef<'_, Self>) -> String {
```
jgodlew, line 198: You need to annotate with `#[getter]` for the getter to be correctly produced in the python class
```suggestion
    #[getter]
    fn get_path(self_: PyRef<'_, Self>) -> String {
```

hf_xet/src/lib.rs
```diff
@@ -106,53 +140,129 @@ fn try_parse_progress_updaters(funcs: Vec<Py<PyAny>>) -> PyResult<Vec<Arc<dyn Pr
...
+        self_.as_super().destination_path.clone()
+    }
+
+    fn get_hash(self_: PyRef<'_, Self>) -> String {
```
jgodlew: This isn't needed since hash is unchanged from the parent class.

However, you do need a `getter` for `filesize` to keep this backwards-compatible.

hf_xet/src/lib.rs
```diff
@@ -106,53 +140,129 @@ fn try_parse_progress_updaters(funcs: Vec<Py<PyAny>>) -> PyResult<Vec<Arc<dyn Pr
...
+        self_.as_super().hash.clone()
+    }
+
+    fn set_path(mut self_: PyRefMut<'_, Self>, path: String) {
+        self_.as_super().destination_path = path;
+    }
+
+    fn set_filesize(mut self_: PyRefMut<'_, Self>, filesize: u64) {
+        self_.as_super().file_size = filesize;
+    }
```
jgodlew: nit: do we want `filesize` to be settable? We don't allow that in the super-class or in the previous version.

hf_xet/src/lib.rs
```diff
@@ -41,15 +76,15 @@ pub fn upload_files(
     token_refresher: Option<Py<PyAny>>,
     progress_updater: Option<Py<PyAny>>,
     _repo_type: Option<String>,
-) -> PyResult<Vec<PyPointerFile>> {
+) -> PyResult<Vec<PyXetUploadInfo>> {
```
jgodlew, line 79: nit: `PyXetUploadInfo` doesn't have the same fields as `PyPointerFile` so this is technically breaking the interface as someone could have depended on the `filesize` or `path` field (or that the returned type is a `PyPointerFile`)...

jgodlew (changes requested):
Overall the changes look good. There are some changes needed for the PyPointerFile to be backwards-compatible and I think that `upload_files` isn't fully compatible.

hf_xet/src/lib.rs
bpronan: I understand. I'm not too concerned about this change. None of our python code is using the return values at the moment. 

While there could be a client that's using hf-xet directly, I'm very doubtful that they are using the return values either.

hf_xet/src/lib.rs
bpronan: `BinaryIO` will require quite a bit of work to get working in a secure way with PyO3. According to the huggingface_hub folks, this is a very small use case. 

If we end up supporting it, we will probably need a decent amount of code to support it. In short, it's extremely likely that we'll need an additional routine anyway.

data/src/data_client.rs
bpronan: This is a good point. There are a few subtle differences between the routines, so it may be difficult.

I'll give it a try in the morning, but it may not be feasible without some serious Rust hacks.

jgodlew (approved):
LGTM! :shipit:

data/src/data_client.rs
bpronan: I’ve looked into refactoring out the `clean_bytes` and `clean_file` into the same function, and it looks like it’s non-trivial to do. There are a few potential perf concerns too. The biggest is having all these open file descriptors up front for the file case. Additionally, it would add an unnecessary buffer copy for the bytes case. I fully acknowledge that these could be premature optimizations, but I would rather not risk it until we can get some cross-platform benchmarking in place to test whether having 1000s of open file descriptors is a concern.

If we end up needing to handle BinaryIO, it will be worth it to revisit this refactor. I'll add it to the issue tracking that work.

PR #277, opened 2025-04-25
Release commit from event rather than latest on main

bpronan:
The release should tag the sha at the time of the workflow dispatch time instead of when the release gets executed


jgodlew (approved):
LGTM!

PR #283, opened 2025-04-29
sanitize url in reqwest error

assafvayner:
When we have a reqwest::Error in hf_xet it can get logged at different places and it currently logs presigned urls that include data to reach S3/CF, this PR attempts to sanitize those urls when the error is logged/returned

### DIFF
--- a/cas_client/src/error.rs
+++ b/cas_client/src/error.rs
@@ -50,8 +50,8 @@ pub enum CasClientError {
     #[error("ReqwestMiddleware Error: {0}")]
     ReqwestMiddlewareError(#[from] reqwest_middleware::Error),
 
-    #[error("Reqwest Error: {0}")]
-    ReqwestError(#[from] reqwest::Error),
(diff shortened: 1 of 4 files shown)


cas_client/src/error.rs
```diff
@@ -63,6 +63,15 @@ pub enum CasClientError {
     PresignedUrlExpirationError,
 }
 
+impl From<reqwest::Error> for CasClientError {
+    fn from(value: reqwest::Error) -> Self {
+        // strip non-domain info from url, primarily query parameters which can be sensitive information
+        let domain = value.url().and_then(|url| url.domain()).unwrap_or_default().to_string();
```
jgodlew: I would rather we strip out query parameters instead of keeping only the domain. Not having the path will make it difficult to consult CDN logs for problematic requests.

jgodlew (approved):
Overall, this is ok, however, I am concerned about us stripping off too much info from the URL (namely the path) as it will hinder our ability to help debug issues.

PR #284, opened 2025-04-29
cache put

assafvayner:
FIX XET-531

There are issues in the cache code with state interior mutability such that we attempt to "smart evict" some items that overlap with the new items to be inserted, the file in the issue triggered this path in particular and exposed a bug causing the the total_bytes and num_items tracked by the cache to be incorrect that then caused a lot of other issues, namely evictions stopped working and cache puts were failing as a result.

### DIFF
--- a/chunk_cache/src/disk.rs
+++ b/chunk_cache/src/disk.rs
@@ -12,7 +12,7 @@ use cas_types::{ChunkRange, Key};
 use error_printer::ErrorPrinter;
 use file_utils::SafeFileCreator;
 use merklehash::MerkleHash;
-use tracing::{debug, error, warn};
+use tracing::{debug, error};
 use utils::output_bytes;
 
 use crate::disk::cache_file_header::CacheFileHeader;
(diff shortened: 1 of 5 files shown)


chunk_cache/src/disk.rs
```diff
@@ -485,46 +476,57 @@ impl DiskCache {
...
-        let num_items = state.num_items;
-        if num_items == 0 {
+        debug_assert_eq!(
+            state.inner.values().map(|v| v.len()).sum::<usize>(),
+            state.num_items,
+            "real num items != stored num items"
+        );
+
+        if state.num_items == 0 {
+            error!("cache random_item for eviction: no items in cache");
```
jgodlew, line 517: Should we use error or warn here? It seems that this isn't fatal, right?

jgodlew (approved):
Overall, looks good. We may want to change some of the `error` logs to `warn` if they are not actually issues the user should address.

chunk_cache/src/disk.rs
assafvayner: this is an intentional error, this is a rather bad state invariant violation where we are trying to evict items from an empty cache.

No error in the cache code is supposed to be fatal, some are such as real FS errors, but mostly the cache is intended as an optimization so any internal errors are set to not disrupt the rest of the operations.

PR #286, opened 2025-04-29
Initial warning huggingface-hub < 0.30.0, no tests

rajatarya:
@jsulz: I'm not sure emitting this warning will ever be useful - because in order for this warning to happen the user has to do the following:

```
# pip install huggingface_hub==0.29.0
# pip install hf-xet

ipython
import hf_xet
```

Since users don't actually interact with `hf-xet` directly they won't run the import themselves, and no version of `huggingface-hub` < 0.30.0 will know to try importing `hf-xet`, so the user won't see the warning.

I am more and more inclined to close XET-461 as wontfix for these reasons.

Fixes [XET-461](https://linear.app/xet/issue/XET-461/hf-xet-should-emit-a-warning-if-in-a-python-environment-where)

I have tested this builds into the .whl from maturin

### DIFF
--- a/hf_xet/.gitignore
+++ b/hf_xet/.gitignore
@@ -1,0 +1,2 @@
+python/**/*.so
+**/__pycache__*
--- a/hf_xet/python/hf_xet/__init__.py
+++ b/hf_xet/python/hf_xet/__init__.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from importlib.metadata import version as _version, PackageNotFoundError
+from packaging.version import Version as _Version
+import warnings
+
+def get_hfhub_version() -> str:
+    try:
+        return _version("huggingface_hub")
+    except importlib.metadata.PackageNotFoundError:
+        # package is not installed
+        return "0.0.0"
+
+if _Version(get_hfhub_version()) < _Version("0.30.0"):
+    warnings.warn(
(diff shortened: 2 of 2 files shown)


jsulz:
@rajatarya yup, it's an edge case. 

What I was thinking was if someone has a pinned dependency (e.g., an older version of `transformers`) that requires `huggingface_hub<0.30.0`, and then tries to follow the documentation here: https://huggingface.co/docs/hub/en/storage-backends#using-xet-storage, which currently recommends installing `hf-xet` via `pip install hf-xet`

That said, no strong opinions here either way.

jsulz (approved):

rajatarya:
Yeah, my thinking in that case is that the user still won't see this message (because they will never type in 'import hf-xet'.

I think probably better to add more explicit documentation in the docs that hf-hub needs to be > 0.30.0.

I'm inclined to close this PR and make the doc change instead. @jsulz you cool with that?

jsulz:
Ahhh, gotcha. Yeah, if we can't emit this when someone is using the `huggingface_hub` APIs w/out `hf-xet` being explicitly imported, then we can drop this. 

FWIW, I think the https://huggingface.co/docs/hub/en/storage-backends#using-xet-storage are clear enough on this front.

rajatarya:
Closing, as users don't interact with `hf-xet` directly.

PR #288, opened 2025-04-30
clean file from url

assafvayner:
Adds functionality to read a file directly from a URL and clean/upload it. This can enable/speed up migration jobs where we don't want to download the whole file to disk first.

Particular attention has been added to retry requests on failures reading from the bytes stream (with no backoff). Retrying reading from the stream requires that we restart the cleaning process for the file. This retry is a particular ask from @seanses on the previous attempt.


sirahd:
I'm not sure if the streaming download approach in repo-scanner will speed up the migration jobs. The current [download implementation](https://github.com/huggingface-internal/repository-scanner/blob/5ac8a12a32659984ae7a9e3d0b0696611e1a33c5/apps/shared/src/hub/api.rs#L522) splits up the download into 10MB chunk and parallelizes them over multiple connections where each can be retried independently. The streaming approach as implemented in this PR uses a single connection, which means significantly less download throughput.

If that's not enough to convince you, the median file download time (on all migrations we've run so far) is 0.18s, compared to 6s for the migration time (which includes cleaning and uploading to CAS). The juice is probably not worth the squeeze in this particular scenario.

assafvayner:
The main reason to keep this idea isn't for most migrations. There was a note from @jgodlew that there were some large files that exceed the migrate worker's node's whole EBS volume. Streaming the file in is relevant in those cases, not as relevant in the common case as you mentioned.

You bring up a possible idea to download the file in chunks rather than streaming, that allows us to control retries at a more granular level.

This PR is optional, I'd like to get JG's take before closing. I'd also maybe like to keep feature around for some other side-explorations.

jgodlew:
Yeah, having the ability to download from a URL would be good. I agree with Sam that it would be good to have a chunk-level download to better stream the content into the FileCleaner as that would be faster. 

This work isn't as important now for the migration service since we've upgraded to better EBS volumes with higher provisioned throughput, which has mitigated the issue of reading/writing the LFS files to the disk. As for files being too large, we are ok on that front and are not encountering files > disk capacity.

However, I would suggest that this sort of change (upload from a url) is better supported either as a separate library or even just being done by the client. The interface for uploading using xet-core is currently sufficient (add bytes to a FileCleaner, regardless of the source of that data). If this becomes a user ask, then I would suggest we incorporate it / provide a library to help with this.

PR #82, opened 2024-11-02
add tracing spans

jgodlew:

PR #291, opened 2025-04-30
Adding session_id to requests and spans

jgodlew:
* Creates a session_id whenever the `data_client` is used or a FileUploadSession/FileDownloader is created to be propagated to the new remote clients.
* Adds a middleware to http clients to push the session_id into the `X-Xet-Session-Id` header for outgoing requests (CAS is already configured to accept this header).
* Adds info-level spans to the key parts of xet-core for cases where xet-core is used as a library by long-running systems (e.g. migration service or internal systems) for aid in debugging / tracing (essentially bringing: https://github.com/huggingface/xet-core/pull/82 up to current minus the hf_xet logging changes).

### DIFF
--- a/cas_client/src/http_client.rs
+++ b/cas_client/src/http_client.rs
@@ -2,9 +2,9 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use anyhow::anyhow;
-use cas_types::REQUEST_ID_HEADER;
+use cas_types::{REQUEST_ID_HEADER, SESSION_ID_HEADER};
 use error_printer::{ErrorPrinter, OptionPrinter};
-use http::StatusCode;
+use http::{Extensions, StatusCode};
 use reqwest::header::{HeaderValue, AUTHORIZATION};
 use reqwest::{Request, Response};
 use reqwest_middleware::{ClientBuilder, ClientWithMiddleware, Middleware, Next};
@@ -14,7 +14,7 @@ use reqwest_retry::{
     Retryable, RetryableStrategy,
 };
 use tokio::sync::Mutex;
-use tracing::{debug, warn};
+use tracing::{debug, info_span, warn, Instrument};
 use utils::auth::{AuthConfig, TokenProvider};
 
 use crate::{error, CasClientError};
@@ -81,15 +81,18 @@ impl RetryConfig<No429RetryStratey> {
 pub fn build_auth_http_client<R: RetryableStrategy + Send + Sync + 'static>(
     auth_config: &Option<AuthConfig>,
     retry_config: RetryConfig<R>,
-) -> std::result::Result<ClientWithMiddleware, CasClientError> {
+    session_id: &str,
+) -> Result<ClientWithMiddleware, CasClientError> {
     let auth_middleware = auth_config.as_ref().map(AuthMiddleware::from).info_none("CAS auth disabled");
     let logging_middleware = Some(LoggingMiddleware);
+    let session_middleware = (!session_id.is_empty()).then(|| SessionMiddleware(session_id.to_owned()));
     let retry_middleware = get_retry_middleware(retry_config);
     let reqwest_client = reqwest::Client::builder().build()?;
     Ok(ClientBuilder::new(reqwest_client)
         .maybe_with(auth_middleware)
         .maybe_with(Some(retry_middleware))
         .maybe_with(logging_middleware)
+        .maybe_with(session_middleware)
         .build())
 }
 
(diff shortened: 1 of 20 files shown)


cas_object/src/cas_object_format.rs
```diff
@@ -272,6 +272,7 @@ impl CasObjectInfoV0 {
     }
 }
 
+#[allow(clippy::empty_line_after_doc_comments)]
```
assafvayner, line 275: what is this? Is this lint fixed by moving the comment about the `#[derive(Debug...)]`?

cas_client/src/http_client.rs
```diff
@@ -146,7 +155,9 @@ impl Middleware for LoggingMiddleware {
         extensions: &mut http::Extensions,
         next: Next<'_>,
     ) -> reqwest_middleware::Result<Response> {
+        let api = extensions.get::<Api>().map(|a| a.0);
         next.run(req, extensions)
+            .instrument(info_span!("client::request", api))
```
assafvayner, line 160: If api is None, what is the behavior here?
very minor nit, to have an `unwrap_or("unset")`/default value
I assume these instrumentations are just for local debugging, not going to the server, so it doesn't matter too much.

data/src/migration_tool/migrate.rs
```diff
@@ -69,10 +75,15 @@ pub async fn migrate_files_impl(
         FileUploadSession::new(config, threadpool, None).await?
     };
 
-    let clean_ret = tokio_par_for_each(file_paths, num_workers, |f, _| async {
-        let proc = processor.clone();
-        let (pf, metrics) = clean_file(proc, f).await?;
-        Ok((pf, metrics.new_bytes as u64))
+    let file_paths_plus = add_spans(file_paths, || info_span!("migration::clean_file"));
```
assafvayner: rename to file_paths_and_spans maybe? other places that have "_plus" too?

data/src/data_client.rs
```diff
@@ -250,12 +276,23 @@ async fn smudge_file(
     Ok(file_path.to_string())
 }
 
+/// Adds spans to the indicated list for each element.
+///
+/// Although a span will be added for each element, we need an Option<Span> since
+/// tokio_par_for_each requires the input list be Default, which Span isn't.
+pub fn add_spans<I, F: Fn() -> Span>(v: Vec<I>, create_span: F) -> Vec<(I, Option<Span>)> {
```
assafvayner, line 283: this function is good and all, gives me an idea for the future to make tokio_par_for_each to take an iterator rather than a vec, then you can just chain the iterators here and create the spans when you need them, i.e.
`v.map(|item| (item, Some(create_span()))` and you wouldn't actually have to iterate and collect at all

assafvayner (approved):
most comments are nits, not too important.

For the python client, is each download/upload batch 1 session?

sirahd (approved):
LGTM. Are the spans only logged for the upload path since the tracing subscriber is filtering on [`client_telemetry` target](https://github.com/huggingface/xet-core/blob/5f1176d22cbb904a06a157af66b70f41883aee82/hf_xet/src/log.rs#L42), which is only set within [`SingleFileCleaner`](https://github.com/huggingface/xet-core/blob/47175d447573ae6935096338555055d844b06cc2/data/src/file_cleaner.rs#L131)?

cas_object/src/cas_object_format.rs
jgodlew: This is because inside the struct, there are many doc-comments that are for fields that are commented out (since they only exist in the physical structure like lines 296-298) and section comments (e.g. line 288). I would rather not do some hacky commenting of those docs to satisfy clippy and instead add the `#[allow]` clause to the struct.

cas_client/src/http_client.rs
jgodlew: If api is None, then the field is not printed, which IMO is fine as an `unwrap_or("unset")` doesn't give us any more info about the API the request is for. Yeah, these spans don't go to a server unless the user sets up the config (e.g. in repo_scanner or some other service directly depending on xet-core).

data/src/data_client.rs
jgodlew: Yeah, I agree that having `tokio_par_for_each` take an iterator would be better. It may also help with removing the `I: Default` requirement for the input list. I also think we may want to revisit why we need this field over some combination of futures/Joinset/ThreadPool...

I can look into refactoring/changing that in a separate PR.

jgodlew:
> Are the spans only logged for the upload path since the tracing subscriber is filtering on [`client_telemetry` target](https://github.com/huggingface/xet-core/blob/5f1176d22cbb904a06a157af66b70f41883aee82/hf_xet/src/log.rs#L42), which is only set within [`SingleFileCleaner`](https://github.com/huggingface/xet-core/blob/47175d447573ae6935096338555055d844b06cc2/data/src/file_cleaner.rs#L131)?

No, that filter is only filtering logs sent to the telemetry service (if set by the user). It shouldn't affect the other logs being printed or spans being added.

jgodlew:
> For the python client, is each download/upload batch 1 session?

There is 1 session_id per call to hf_xet's download_files or upload_files/upload_bytes. Since the python client may call these methods multiple times in a single python-level upload/download, there might be multiple batches (e.g. `huggingface_hub.snapshot_download` might call `hf_xet::download_files` multiple times, creating multiple sessions).

Ideally, once we have the `hf_xet.XetSession` implemented/integrated, we might be able to change this to 1 session_id per XetSession.

PR #295, opened 2025-05-03
Fix some CAS info missing from shard bug

seanses:
CAS info left in the DataAggregator are not registered to the session shard, and thus will not be added to global dedup. This PR fixes this bug.

### DIFF
--- a/data/src/file_upload_session.rs
+++ b/data/src/file_upload_session.rs
@@ -304,9 +304,9 @@ impl FileUploadSession {
         if xorb.num_bytes() > 0 {
             // Add the xorb info to the current shard.
             self.shard_interface.add_cas_block(xorb.cas_info.clone()).await?;
-
-            self.register_new_xorb_for_upload(xorb).await?;
         }
+
+        self.register_new_xorb_for_upload(xorb).await?;
 
         // Now, we need to scan all the file dependencies for dependencies on this xorb, as
         // these would not have been registered yet.
@@ -315,7 +315,7 @@ impl FileUploadSession {
 
         {
             for (file_id, fi, bytes_in_xorb) in new_files {
-                if xorb_hash != MerkleHash::default() && bytes_in_xorb != 0 {
+                if xorb_hash == MerkleHash::default() || bytes_in_xorb != 0 {
                     new_dependencies.push((file_id, xorb_hash, bytes_in_xorb, false));
                 }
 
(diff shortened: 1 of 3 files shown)


data/src/file_upload_session.rs
```diff
@@ -301,15 +301,21 @@ impl FileUploadSession {
...
         // Now, we need to scan all the file dependencies for dependencies on this xorb, as
         // these would not have been registered yet.
-        self.register_new_xorb_for_upload(xorb).await?;
 
         let mut new_dependencies = Vec::with_capacity(new_files.len());
 
         {
             for (file_id, fi, bytes_in_xorb) in new_files {
-                if xorb_hash == MerkleHash::default() || bytes_in_xorb != 0 {
+                if xorb_hash != MerkleHash::default() && bytes_in_xorb != 0 {
```
hoytak: Is there a reason for this change here?  I remember the || being needed to properly bookkeep file progress when there was an empty xorb...

data/src/file_upload_session.rs
```diff
@@ -301,15 +301,21 @@ impl FileUploadSession {
         debug_assert_le!(xorb.num_bytes(), *MAX_XORB_BYTES);
         debug_assert_le!(xorb.data.len(), *MAX_XORB_CHUNKS);
 
+        if xorb.num_bytes() > 0 {
+            // Add the xorb info to the current shard.
+            self.shard_interface.add_cas_block(xorb.cas_info.clone()).await?;
+
+            self.register_new_xorb_for_upload(xorb).await?;
```
hoytak: Let's drop the change below and move this call out of the if statement -- inside this function we already do:

```
         if xorb.num_bytes() == 0 {
            self.completion_tracker.register_xorb_upload_completion(xorb_hash).await;
            return Ok(true);
        }
```

And I'd like to keep the completion_tracker logic minimal

data/src/file_upload_session.rs
```diff
@@ -301,15 +301,21 @@ impl FileUploadSession {
         debug_assert_le!(xorb.num_bytes(), *MAX_XORB_BYTES);
         debug_assert_le!(xorb.data.len(), *MAX_XORB_CHUNKS);
 
+        if xorb.num_bytes() > 0 {
+            // Add the xorb info to the current shard.
+            self.shard_interface.add_cas_block(xorb.cas_info.clone()).await?;
```
hoytak: good catch

data/src/file_upload_session.rs
seanses: Will double check the logic

hoytak:
This exposed two much deeper bugs than I suspected.  Put the full fix up at https://github.com/huggingface/xet-core/pull/302 (needs more review).

hoytak:
Let's close this; fix incorporated into https://github.com/huggingface/xet-core/pull/302.

seanses:
Close in favor of #302

hoytak:
fix incorporated into https://github.com/huggingface/xet-core/pull/302.

PR #297, opened 2025-05-05
Replace passed-around threadpool refs with thread local variable

hoytak:
Currently, we pass references to the threadpool around the code in order to use it.  However, all of this code is currently on a worker thread of the tokio runtime used to create the threadpool.  

This PR simplifies this by using thread local storage; each worker thread sets a reference to the runtime on start that can be accessed at any time using ThreadPool::current().  

Fallback for running within an existing tokio runtime (E.g. with tokio::test) is also handled using the from_external() mechanism.

There should be no functionality change, just code simplification.


seanses (approved):
LGTM!

PR #299, opened 2025-05-06
Chunk Partitioning

ylow:
Implements a procedure that identifies partitioning points in a file where chunking can be occur in parallel while *very likely* aligning with the regular sequential chunking procedure.

See the parallel chunking.pdf file for a proof as to why this can be done. However, due to an implementation error in the chunker, the proof does not actually apply. However, it is still quite likely to behave well.

Even if the partitioning does not align, the worst case is a reduction in deduplication quality so its not really a significant issue.


hoytak (approved):
All looks good to me; definitely optimizations as you mentioned but that's stuff for later.

seanses (approved):

PR #301, opened 2025-05-07
Connect detailed upload progress to hub

hoytak:
This PR connects detailed upload progress to the hub in a backwards compatible way. 

It works by testing the number of arguments and argument names on the progress updating function.  If the progress reporting function takes a single argument, this function calls it using the old method; if it has the appropriate arguments for detailed reporting -- `item_id, completed_bytes, total_bytes, update_increment` -- then it calls it using the new method.   Additionally, if None is passed in, the progress reporting is disabled.

### DIFF
--- a/hf_xet/Cargo.lock
+++ b/hf_xet/Cargo.lock
@@ -1194,6 +1194,7 @@ dependencies = [
  "ctrlc",
  "data",
  "error_printer",
+ "itertools 0.14.0",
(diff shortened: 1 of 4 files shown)


hoytak:
This PR has been tested and verified on a draft implementation in huggingface_hub; I want to get it in now before I put up a PR there.

seanses (approved):

PR #302, opened 2025-05-08
Fixed race condition in dependency tracking.

hoytak:
Testing discovered two dependency tracking issues: 

The first is when a file references a new xorb multiple times in non-contiguous locations.  In this case, the logic will cause an assertion failure when debug_assertions are enabled, though likely the underlying logic is actually correct.

The second is that tracking which xorbs are part of a given session and which xorbs have been uploaded previously was done by recording the xorb hashes when they are registered for upload.  However, it turns out this needs to happen before registering the xorb as available for dedup; otherwise a race condition could cause a xorb to be incorrectly registered as already uploaded when in reality the xorb hash was in the queue and not actually added to the session registry yet.  This causes the progress to be incorrectly counted as already completed. 

A clean fix for this is to actually get that information directly from the shard_manager instead of tracking it separately by returning whether it's deduped against the session data or the cache.  This fix allows newly cut xorbs to immediately be used for dedup by other threads and correctly tracks all the information across threads.

### DIFF
--- a/data/src/progress_tracking.rs
+++ b/data/src/progress_tracking.rs
@@ -118,7 +118,6 @@ impl CompletionTrackerImpl {
                 } else {
                     // Insert a new xorb entry if needed.
                     entry.file_indices.insert(dep.file_id as usize);
-                    eprintln!("Registering dep: {} -> {}", dep.file_id, dep.xorb_hash);
                     *file_entry.remaining_xorbs_parts.entry(dep.xorb_hash).or_default() += dep.n_bytes;
                 }
             }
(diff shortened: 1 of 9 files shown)


data/src/progress_tracking.rs
```diff
@@ -76,50 +77,49 @@ impl CompletionTrackerImpl {
...
-                        update_increment: n_bytes,
+                        update_increment: dep.n_bytes,
                     };
                     ret.push(progress_update);
                 } else {
                     // Insert a new xorb entry if needed.
-                    entry.file_indices.push(file_id as usize);
-                    *file_entry.remaining_xorbs_parts.entry(xorb_hash).or_default() += n_bytes;
+                    entry.file_indices.insert(dep.file_id as usize);
+                    eprintln!("Registering dep: {} -> {}", dep.file_id, dep.xorb_hash);
```
seanses: Don't print to stderr

seanses (approved):
LGTM except need to change the logging mentioned above.

PR #306, opened 2025-05-09
Switch chunk cache to use async RWlock instead of std::sync mutex.

hoytak:
Currently, the chunk cache holds a std::sync Mutex around the internal state.  This can block the tokio worker threads from processing other tasks, and in most cases a read lock is sufficient for the operations.

The PR does some mild rearranging of the functions to accommodate the different read/write locks and makes the primary trait async, but should currently involve no other functionality change.

### DIFF
--- a/chunk_cache/src/disk.rs
+++ b/chunk_cache/src/disk.rs
@@ -3,8 +3,9 @@ use std::fs::{DirEntry, File};
 use std::io::{self, Cursor, ErrorKind, Read, Seek, SeekFrom, Write};
 use std::mem::size_of;
 use std::path::{Path, PathBuf};
-use std::sync::{Arc, Mutex, MutexGuard};
+use std::sync::Arc;
 
+use async_trait::async_trait;
 use base64::engine::general_purpose::URL_SAFE;
 use base64::engine::GeneralPurpose;
 use base64::Engine;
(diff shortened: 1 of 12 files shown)


chunk_cache/src/lib.rs
```diff
@@ -39,6 +40,7 @@ pub struct CacheRange {
 /// implementors are allowed to evict data, a get after a put is not required to
 /// be a cache hit.
 #[automock]
+#[async_trait]
```
assafvayner, line 43: making this interface async requires refactoring the chunk_cache_bench crate a lot, can you give it a shot in this PR. If it's too old/not-worth-our-time, we can delete that crate.

chunk_cache/src/disk.rs
```diff
@@ -136,11 +213,12 @@ impl DiskCache {
         let capacity = config.cache_size;
         let cache_root = config.cache_directory.clone();
 
+        // May take a while; don't block the runtime for this.
```
assafvayner, line 216: I think this does block whichever thread is initializing the cache. not sure if there is a task spawned to get/scan the cache. Can look into/file an issue

assafvayner (commented):
LGTM except we need to make a decision on the benchmark crate, will approve after that's resolved.

chunk_cache/src/lib.rs
hoytak: Pushed an update!  Not the worst. 

Since it's being used heavily in production, I suspect iteratively reducing bottlenecks and addressing user issues is the majority of what we'll be doing going forward.

assafvayner (approved):

PR #308, opened 2025-05-10
Optimize bg4 prediction 

hoytak:
This PR optimizes the bg4 prediction used in the compression scheme analysis by bypassing the per-byte popcnt call, which can be quite slow.  This was shown to be a bottleneck; this PR speeds this operation up.

On Arm/neon, this uses an intrinsic vector popcnt instruction, and on intel it uses a bit twiddling method to calculate the per-byte popcount on u128s that is much more efficient than subsequent popcnt calls.

### DIFF
--- a/cas_object/Cargo.toml
+++ b/cas_object/Cargo.toml
@@ -14,6 +14,8 @@ harness = false
 bench = true
 
 [dependencies]
+tokio = { version = "1.44", features = ["full"] }
+half = "2.4.1"
 thiserror = "2.0"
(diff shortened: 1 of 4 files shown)


hoytak:
Method gives a 2x speedup.   Benchmark: 

Reference: 1608.03 MB/s
V1: 1663.79 MB/s
SWAR: 2392.19 MB/s   (Fallback method on intel)
Optimized: 3222.63 MB/s  (simd method with neon)

PR #315, opened 2025-05-13
Move CAS object serialization to before parallel upload gate.

hoytak:
This PR moves the CAS object serialization outside of the semaphore controlling the number of parallel uploads.  This serialization can be expensive relative to the upload itself on fast connections, so gating only the upload itself ensures the upload link stays saturated.


danielhanchen:
Would this help make `upload_large_folder` work?

hoytak:
> Would this help make `upload_large_folder` work?

It's a step in that direction; this PR is part of optimizing the upload pipeline to make sure bandwidth is used efficiently and should improve all the upload paths.  There are a few more optimizations coming.

cas_object/src/cas_object_format.rs
```diff
@@ -1303,6 +1304,130 @@ impl CasObject {
...
+        Ok(Self {
+            serialized_data,
+            hash: xorb_hash,
+            raw_num_bytes,
+            num_chunks,
+        })
+    }
+
+    // Used in testing
+    #[allow(dead_code)]
```
jgodlew: I think if this is just needed for testing, you can probably just put it under the `test_utils` module below (also, we should probably tag that module with a `#[cfg(test)]`)?

deduplication/src/raw_xorb_data.rs
```diff
@@ -63,8 +63,8 @@ impl RawXorbData {
         n
     }
 
-    // Todo: Push this Xorb data format all the way down to the compression levels to
-    // avoid this copy / memory overhead
+    // Used largely for testing to test newer serialization method against known older one.
+    #[allow(dead_code)]
```
jgodlew: Again, probably just want to have this tagged with `#[cfg(test)]` (or in a testing module) instead?

cas_client/src/interface.rs
```diff
@@ -27,17 +27,7 @@ pub trait UploadClient {
...
-    /// will be needed.
-    async fn put(
-        &self,
-        prefix: &str,
-        hash: &MerkleHash,
-        data: Vec<u8>,
-        chunk_and_boundaries: Vec<(MerkleHash, u32)>,
-        compression: Option<CompressionScheme>,
-    ) -> Result<usize>;
+    async fn upload_xorb(&self, prefix: &str, serialized_cas_object: SerializedCasObject) -> Result<u64>;
```
seanses, line 25: Also update the comments

cas_object/src/cas_object_format.rs
```diff
@@ -1303,6 +1304,130 @@ impl CasObject {
...
+}
+
+impl SerializedCasObject {
+    pub fn from_xorb(xorb: RawXorbData, compression_scheme: Option<CompressionScheme>) -> Result<Self, CasObjectError> {
+        Self::from_xorb_impl(xorb, compression_scheme, false)
+    }
+
+    // Used in testing
+    #[allow(dead_code)]
+    pub fn from_xorb_with_verification(
```
jgodlew: Also move this to `#[cfg(test)]`

jgodlew (approved):
Overall looks good! Just have a few comments about tagging some of the "Used in testing" pieces with the `#[cfg(test}]` or move to a test module.

cas_object/src/cas_object_format.rs
```diff
@@ -1303,6 +1304,130 @@ impl CasObject {
...
+    ) -> Result<Self, CasObjectError> {
+        Self::from_xorb_impl(xorb, compression_scheme, true)
+    }
+
+    /// Builds the xorb from
+    fn from_xorb_impl(
+        xorb: RawXorbData,
+        compression_scheme: Option<CompressionScheme>,
+        verify_against_previous: bool,
+    ) -> Result<Self, CasObjectError> {
```
seanses: Let's not rewrite the serialization code again, preferably call `CasObject::serialize()`

hf_xet/src/lib.rs
```diff
@@ -240,7 +240,7 @@ impl From<XetFileInfo> for PyXetUploadInfo {
 
 impl From<PyXetDownloadInfo> for (XetFileInfo, DestinationPath) {
     fn from(pf: PyXetDownloadInfo) -> Self {
-        (XetFileInfo::new(pf.hash, pf.file_size as usize), pf.destination_path)
+        (XetFileInfo::new(pf.hash, pf.file_size as u64), pf.destination_path)
```
seanses: I think this field is already u64

cas_client/src/remote_client.rs
```diff
@@ -116,28 +116,43 @@ impl RemoteClient {
...
+            hash: serialized_cas_object.hash,
         };
 
-        let (was_uploaded, nbytes_trans) = self.upload(&key, data, chunk_and_boundaries, compression).await?;
+        let url = Url::parse(&format!("{}/xorb/{key}", self.endpoint))?;
+
+        let n_bytes = serialized_cas_object.serialized_data.len();
+
+        let xorb_uploaded = {
+            if !self.dry_run {
```
seanses, line 131: Now that we do serialization outside we can actually separate the dry_run mode into a dummy client implementation.

cas_object/src/cas_object_format.rs
hoytak: This is meant to replace that code; to use that function requires an intermediate copy which I want to avoid. Following Joe's suggestion, I moved CasObject::serialize to test_utils as serialize_xorb_to_stream_reference; now this code is the only path for the serialization.

cas_client/src/remote_client.rs
hoytak: Nice!  Though that should be a separate PR.

seanses (approved):

PR #319, opened 2025-05-13
Fixing the sdist install and adding better project information

bpronan:
Attempting to address the `sdist` problems. For an sdist build, the pyproject.toml gets copied to the root directory of the repo and edited to include the manifest location. We need to refer to the `python` directory from the context of the root directory of the repo.

Note: this doesn't appear to have any affect on the wheel behavior or setup.

Local testing with the following now works:
```
cd hf_xet
maturin sdist

cd <anywhere>
cp <xet-core dir>/hf_xet/target/wheels/hf_xet-1.1.1.tar.gz .
tar -xvzf hf_xet-1.1.1.tar.gz
cd hf_xet-1.1.1
maturin build
```

### DIFF
--- a/hf_xet/pyproject.toml
+++ b/hf_xet/pyproject.toml
@@ -5,16 +5,26 @@ build-backend = "maturin"
 [project]
 name = "hf-xet"
 requires-python = ">=3.8"
+description = "Fast transfer of large files with the Hugging Face Hub."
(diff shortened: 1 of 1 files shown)


jgodlew (approved):

rajatarya:
Love to see this, yay!

PR #320, opened 2025-05-13
download terms once and write where needed

assafvayner:
Fix XET-493

Changes the parallel download process to download each fetch info term only once ever and write all terms fulfilled by it at once.


jgodlew (approved):
Overall, looks good!

PR #323, opened 2025-05-14
Add incremental upload progress and total bytes to progress tracking.

hoytak:
This PR adds the ability for the completion tracker to report xorb uploading incrementally, with file progress within the xorb being updated in proportion to how many bytes of that file are in the xorb. 

In addition, it also now tracks total byte completion, so we can also report on new bytes uploaded instead of using the aggregated per-file upload speed, which can be very confusing to interpret given that deduplication is treated the same as uploading.

This PR contains the underlying machinery and tests; a future PR will connect this to hf_xet.


progress_tracking/src/lib.rs
```diff
@@ -1,7 +1,16 @@
...
 pub mod verification_wrapper;
 
-pub use update_interfaces::{
-    ItemProgressUpdater, NoOpProgressUpdater, ProgressUpdate, SimpleProgressUpdater, TrackingProgressUpdater,
-};
+pub use no_op_tracker::NoOpProgressUpdater;
+pub use progress_info::{ItemProgressUpdate, ProgressUpdate};
+
+/// The trait that a progress updater that reports per-item progress completion.
+#[async_trait::async_trait]
```
assafvayner: not a comment on the code, a reminder for when we integrate this package into the WASM code; we need to use the arch gating async_trait macro here.

progress_tracking/src/upload_tracking.rs
```diff
@@ -34,14 +41,17 @@ struct XorbDependency {
...
+
     /// Total bytes already uploaded for this file (across its xorbs).
     completed_bytes: u64,
-    /// Mapping of xorb_hash -> number of bytes of the file contained in that xorb.  Only
+
+    /// Mapping of xorb_hash -> (number of completed bytes / number of bytes of the file contained in that xorb).  Only
     /// xorbs that are not uploaded yet are tracked here.
     /// Once an xorb is uploaded, we remove it from here (and add to `completed_bytes`).
-    remaining_xorbs_parts: HashMap<MerkleHash, u64>,
+    remaining_xorbs_parts: HashMap<MerkleHash, (u64, u64)>,
```
assafvayner: instead of (u64, u64) would prefer a struct with named fields for readability. Also the comment may need to be updated, could just be replace `/` (looks like dividing to me) with `,` or better yet named fields

assafvayner (approved):

progress_tracking/src/upload_tracking.rs
hoytak: Changed to named fields.

PR #325, opened 2025-05-14
SDist fix v2

bpronan:
The original approach for building our `sdist` package [didn't work](https://github.com/huggingface/xet-core/actions/runs/15026118241/job/42227177543). Shifting the workaround to a post-process step in the sdist package build step.

Note: maturin caches pretty aggressively, so it matters whether it's tested on a clean platform.


.github/workflows/release.yml
```diff
@@ -184,15 +184,23 @@ jobs:
...
       - name: Build sdist
         uses: PyO3/maturin-action@v1
         with:
           command: sdist
           args: --out dist
           working-directory: hf_xet
+      - name: Patch sdist
+        run: |
+          pushd hf_xet/dist
+          DISTFILE=$(ls | grep hf_xet-)
```
rajatarya: I am always nervous with this type of shell command getting assigned to a variable.

Instead, can this be `hf_xet-$VERSION.tar.gz` or something like that? That feels more deterministic to other files being listed in the dist directory.

bpronan: Using `$VERSION` won't work here for two reasons:
1. it's in cargo format, eg `1.1.1-rc0` while the tar file is in python format `1.1.1rc0`.
2. we don't always have a version since this action can runs on checkin to the main branch. 

So, we would have to extract the version from `Cargo.toml` and mutate it to be sure this will work for all cases. The extraction is tricky in just bash. It can be done, but I may not get to it this afternoon.

rajatarya: OK how about a compromise of making sure only one value, something like:

```
DISTFILE=$(ls -tr | grep hf_xet- | head -n 1) 
```

or use `tail -n 1` if you want the final value if the list contains multiple values.

rajatarya (approved):

rajatarya (approved):

Issue #311, opened 2025-05-12
Constant `CAS service error` on XET enabled repos


danielhanchen:
People who download from our repo https://huggingface.co/unsloth unfortunately keep getting `RuntimeError: Data processing error: CAS service error : IO Error: Too many open files (os error 24)`

Longer error:
```
"timestamp":"2025-05-12T02:12:04.887262Z","level":"WARN","fields":{"message":"Reqwest(reqwest::Error { kind:
 Request, url: \"https://transfer.xethub.hf.co/xorbs/default/34b2f024c1843d3c470c2fab0c55ccfbc86fc8f207ae08e
cc1070d1080734b5e?X-Xet-Signed-Range=bytes%3D0-57223773&Expires=1747019506&Policy=eyJTdGF0ZW1lbnQiOlt7IlJlc2
91cmNlIjoiaHR0cHM6Ly90cmFuc2Zlci54ZXRodWIuaGYuY28veG9yYnMvZGVmYXVsdC8zNGIyZjAyNGMxODQzZDNjNDcwYzJmYWIwYzU1Y2
NmYmM4NmZjOGYyMDdhZTA4ZWNjMTA3MGQxMDgwNzM0YjVlP1gtWGV0LVNpZ25lZC1SYW5nZT1ieXRlcyUzRDAtNTcyMjM3NzMiLCJDb25ka$
Rpb24iOnsiRGF0ZUxlc3NUaGFuIjp7IkFXUzpFcG9jaFRpbWUiOjE3NDcwMTk1MDZ9fX1dfQ__&Signature=vpEdXcP0~0xiOxt-l8DuG-k
buNPbA72k5DPR~2t8NOrytM~f77IqEiHuSFDHIYQIlcxkrICvkcjBujiCygzUNkFHXG35Y7i5wdsQP6imOKJtAxKNyGd1kGuvQfW2SEYl9o5
~6nRtx47apnjBTC1TMzApBle-wZIk8iSC7U-CGBmEmto3cZcbmKZR-QOHr8WMtgVqvsESf0e9FmjmubVGNkxifEuwhFkgUZnM1G2NReut5Ly
BFe~VZBOh~c-08is87Gkgm5rawut93q8zRrPsJfL6Dhh~F6XGTV-muliFcFAmphYeN9359~nutzvV7Q4z80Xw3m8SzOm6pOTwlBfucg__&Ke
[...]

PR #326, opened 2025-05-14
don't bump num concurrent downloads live

assafvayner:
Removing the ability of the number of concurrent downloads to grow fixes the error for "Too many open files" see #311 .

Tracking how many download tasks are running concurrently over the download of a long and fragmented file showed a steady growth in the number of allowed concurrent downloads (we grow by 1 for every successful download without 403's). This causes the total number of allowed concurrent downloads to reach ~1K and then we hit "Too many open files" in my repro. Notably, my soft ulimit is 1024, and at ~1K (could be a bit more or less, my tracking is not exactly accurate), for each task we have 1 fd for a connection CF, so this is roughly when we reach the ulimit.

We leave the constant number of concurrent downloads user configurable, but constant. We can add a different live updating mechanism later.

### DIFF
--- a/cas_client/src/download_utils.rs
+++ b/cas_client/src/download_utils.rs
@@ -266,12 +266,10 @@ impl DownloadScheduler {
             if *num_range_in_segment > 1 {
                 *num_range_in_segment -= 1;
             }
-            self.n_concurrent_download_task.forget_permits(1);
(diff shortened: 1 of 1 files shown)


seanses (approved):
LGTM!

PR #327, opened 2025-05-15
Back out streaming upload of xorb.

hoytak:
Currently, the streaming upload of the xorb breaks the upload path due to attempts to clone the stream for retries.  

Backing this change out until we find a different solution for incremental progress reporting.


assafvayner (approved):
Do we still have a progress update on the full xorb upload finished?

hoytak:
Yes, xorb completion is still reported.   This happens at a higher level.

PR #329, opened 2025-05-15
add retry on range download

assafvayner:
fix XET-567

This PR adds the retry mechanism on downloading a range from CF/S3 for when an error occurs reading or parsing the body.
The upload is handled by reverting a change to reintroduce retries on a clonable body PR #328.


assafvayner:
closing in favor of #331

PR #330, opened 2025-05-15
Extracting debug symbols from wheels

bpronan:
To reduce the binary size, we are extracting the debug symbols from our wheel packages. For windows and mac, we can use cargo to output the debug symbols. For Linux, we are stripping them out as a post processing step. This can reduce our installed binary size by 20-80MB depending on the platform.


hoytak (approved):
Looks fine to me....  If it works, it works.

PR #331, opened 2025-05-15
add retries on body read errors during download

assafvayner:
fix XET-567

This PR adds the retry mechanism on downloading a range from CF/S3 for when an error occurs reading or parsing the body.
The upload is handled by reverting a change to reintroduce retries on a clonable body PR https://github.com/huggingface/xet-core/pull/328.

### DIFF
--- a/cas_client/src/download_utils.rs
+++ b/cas_client/src/download_utils.rs
@@ -2,6 +2,11 @@ use std::collections::HashMap;
 use std::sync::{Arc, Mutex, RwLock};
 use std::time::{Duration, Instant};
 
+use crate::error::{CasClientError, Result};
+use crate::http_client::{Api, BASE_RETRY_DELAY_MS, BASE_RETRY_MAX_DURATION_MS, NUM_RETRIES};
+use crate::remote_client::{get_reconstruction_with_endpoint_and_client, PREFIX_DEFAULT};
+use crate::OutputProvider;
+use cas_object::error::CasObjectError;
 use cas_types::{CASReconstructionFetchInfo, CASReconstructionTerm, ChunkRange, FileRange, HexMerkleHash, Key};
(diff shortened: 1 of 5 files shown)


cas_client/src/download_utils.rs
```diff
@@ -360,6 +362,25 @@ pub(crate) async fn get_one_term(
     Ok(data)
 }
 
+struct ChunkRangeDeserializeFromBytesStreamRetryCondition;
+
+impl tokio_retry::Condition<CasClientError> for ChunkRangeDeserializeFromBytesStreamRetryCondition {
+    fn should_retry(&mut self, err: &CasClientError) -> bool {
+        // we only care about retrying some error yielded by trying to deserialize the stream
```
seanses, line 369: This can't handle this case: https://github.com/huggingface/huggingface_hub/issues/3036#issuecomment-2837721585, which comes from `send()`

jgodlew (approved):
LGTM!

cas_client/src/download_utils.rs
assafvayner: that should be handled by the retry middleware on the client

seanses: Hmm.. The message says `kind: Request` and the [implementation](https://github.com/TrueLayer/reqwest-middleware/blob/591ab444ec58b0e241724fef8a76a655dcaafdbc/reqwest-retry/src/retryable_strategy.rs#L151) on Request kind is only to retry on `hyper_error.is_incomplete_message() || hyper_error.is_canceled()` or if the underlying I/O error is `std::io::ErrorKind::ConnectionReset | std::io::ErrorKind::ConnectionAborted`, but not including `hyper::Error(ChannelClosed)`

seanses (approved):

PR #332, opened 2025-05-15
Fix release build issues

bpronan:
There were a number of issues with #330 that didn't show up in local dev testing that showed up in github actions:

1. Mac, Win: building with a target on mac and windows placed the debug symbols in a different place (rust quirk)
2. Linux: the output folder by maturin didn't have write permissions for the user (maturin runs as sudo)
3. Linux: objcopy does not work on aarch64 (you need to install an aarch64 version of binutils) 
4. sdist: the path was broken (the path for `ls` was wrong)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -258,7 +258,7 @@ jobs:
       - name: Patch sdist
         run: |
           pushd hf_xet/dist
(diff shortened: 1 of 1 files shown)


jgodlew (approved):
:shipit:

rajatarya (approved):

PR #333, opened 2025-05-16
Incremental progress on upload_xorb with retry_wrapper

hoytak:
This PR implements incremental progress reporting on the upload_xorb function, reporting progress every 512KB of data uploaded.  

In addition, errors are retried using the same retry policy as the other clients.  To get around Body::wrap_stream preventing retries due to cloning failing, this PR adds a simple retry wrapper utility that allows the entire request to be retried instead of doing it as part of the middleware layer.

### DIFF
--- a/cas_client/Cargo.toml
+++ b/cas_client/Cargo.toml
@@ -38,6 +38,7 @@ heed = "0.11"
 futures = "0.3.31"
 derivative = "2.2.0"
 more-asserts = "0.3.1"
+http-body = "1"
 
 [dev-dependencies]
 rand = "0.8.5"
--- a/cas_client/src/http_client.rs
+++ b/cas_client/src/http_client.rs
@@ -96,6 +96,22 @@ pub fn build_auth_http_client<R: RetryableStrategy + Send + Sync + 'static>(
         .build())
 }
 
+/// Builds authenticated HTTP Client to talk to CAS.
+pub fn build_auth_http_client_no_retry(
+    auth_config: &Option<AuthConfig>,
+    session_id: &str,
+) -> Result<ClientWithMiddleware, CasClientError> {
+    let auth_middleware = auth_config.as_ref().map(AuthMiddleware::from).info_none("CAS auth disabled");
+    let logging_middleware = Some(LoggingMiddleware);
+    let session_middleware = (!session_id.is_empty()).then(|| SessionMiddleware(session_id.to_owned()));
+    let reqwest_client = reqwest::Client::builder().build()?;
+    Ok(ClientBuilder::new(reqwest_client)
+        .maybe_with(auth_middleware)
+        .maybe_with(logging_middleware)
+        .maybe_with(session_middleware)
+        .build())
+}
+
 /// Builds HTTP Client to talk to CAS.
 /// Includes retry middleware with exponential backoff.
 pub fn build_http_client<R: RetryableStrategy + Send + Sync + 'static>(
--- a/cas_client/src/http_client.rs
+++ b/cas_client/src/http_client.rs
@@ -129,18 +129,27 @@ pub fn build_http_client<R: RetryableStrategy + Send + Sync + 'static>(
         .build())
 }
 
+/// RetryStrategy
+pub fn get_retry_policy_and_strategy<R: RetryableStrategy + Send + Sync>(
+    config: RetryConfig<R>,
+) -> (ExponentialBackoff, R) {
+    (
+        ExponentialBackoff::builder()
+            .retry_bounds(
+                Duration::from_millis(config.min_retry_interval_ms),
+                Duration::from_millis(config.max_retry_interval_ms),
+            )
+            .build_with_max_retries(config.num_retries),
+        config.strategy,
+    )
+}
+
 /// Configurable Retry middleware with exponential backoff and configurable number of retries using reqwest-retry
 fn get_retry_middleware<R: RetryableStrategy + Send + Sync>(
     config: RetryConfig<R>,
 ) -> RetryTransientMiddleware<ExponentialBackoff, R> {
-    let retry_policy = ExponentialBackoff::builder()
-        .retry_bounds(
-            Duration::from_millis(config.min_retry_interval_ms),
-            Duration::from_millis(config.max_retry_interval_ms),
-        )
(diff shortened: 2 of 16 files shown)


cas_client/src/http_client.rs
```diff
@@ -96,6 +96,22 @@ pub fn build_auth_http_client<R: RetryableStrategy + Send + Sync + 'static>(
         .build())
 }
 
+/// Builds authenticated HTTP Client to talk to CAS.
+pub fn build_auth_http_client_no_retry(
```
assafvayner, line 100: I don't think this is really necessary to use, it's fine to use the retry-middleware enabled client on all requests, it just won't actually retry when it's not able to due to a single use provided request body.

cas_client/src/remote_client.rs
```diff
@@ -158,19 +162,27 @@ impl UploadClient for RemoteClient {
...
+
+                let response = retry_wrapper(
+                    move || {
+                        let upload_stream = upload_stream.clone_with_reset();
+                        let url = url.clone();
+
+                        client
+                            .post(url)
+                            .with_extension(Api("cas::upload_xorb"))
+                            .header(CONTENT_LENGTH, HeaderValue::from(n_upload_bytes)) // this line took WAY too long to figure out
```
assafvayner: remove the comment?

Also do we need to set this header explicitly, presumably it was previously set implicitly by reqwest when we were uploading a full byte slice as the body

cas_client/src/retry_utils.rs
assafvayner, line 1: In #331  I already added the use of tokio_retry, I would prefer if we could just use it here again rather than write our own version of this to re-use reqwest-retry, find to just reuse the constants IMO than match the exact code path to provide the config IMO. Arguably we could set different constants too.

cas_client/Cargo.toml
```diff
@@ -38,6 +38,7 @@ heed = "0.11"
 futures = "0.3.31"
 derivative = "2.2.0"
 more-asserts = "0.3.1"
+http-body = "1"
```
assafvayner: is this actually used anywhere? may need to remove

cas_client/src/remote_client.rs
hoytak: It was set then, but because we are using a stream it must be set explicitly.

cas_client/src/retry_utils.rs
hoytak: Unfortunately, with the use of the the stream upload, we can't implement retry in the middleware layer. 

One thought would be to convert everything to this method and not have the retry happen in the middleware layer, which also make it easy to also retry on json parsing errors...

cas_client/src/http_client.rs
hoytak: Unfortunately, that's a no-go.  If retry middleware is enabled, then the body value must be cloneable and streams are not.

cas_client/src/retry_utils.rs
assafvayner: I would still slightly prefer to use tokio_retry rather than maintain our own utility. I don't feel all that strongly about it, but you essentially re-implemented pieces of it but instead using the reqwest-retry strategy.

@jgodlew I think you authored the original (pre-HF) retry mechanic, do you have any opinions?

hoytak: I'm open to converting it to use tokio_retry, but doing it that way requires a different policy type than how we specify things now.  It's doable to roughly convert it over, but I did it this way for consistency with the current approach. 

In general, I would be quite open to using tokio_retry here and only having the tokio retry policy here, dropping the reqwuest_middleware layer from everything and just using this function.   (We could easily add a version that also tried to parse the contents and retry on EOF there too).

cas_client/src/retry_utils.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    for attempt in 0.. {
+        let result = create_request().await;
+
+        // Do we retry?
+        if result.is_err() && matches!(strategy.handle(&result), Some(Retryable::Transient)) {
+            // Does our retry count / timing policy allow us to retry, and when?
+            if let RetryDecision::Retry { execute_after } = retry_policy.should_retry(start_time, attempt) {
+                // Retry after system time is a specific value.
+                if let Ok(wait_dur) = execute_after.duration_since(SystemTime::now()) {
```
assafvayner, line 53: is it a real error if this returns an error?

assafvayner (approved):
Generally looks good, I'm a bit cautious about having our own retry utility, but we can go back and modify it over time, also to use it in more places

cas_client/src/retry_utils.rs
jgodlew: I don't have a strong preference here. I think it is fine to build our own for now to keep the policies consistent with the retry middleware used within a request instead of maintaining some adapter between the reqwest retry config and the tokio retry config. 

However, I do think that eventually, we may want to move the retry out of the middleware and move to tokio_retry (seems that we used that in xetdata's xet-core). I am fine with that being a separate PR though.

cas_client/src/retry_utils.rs
```diff
@@ -0,0 +1,64 @@
...
+                }
+                continue;
+            }
+        }
+
+        return result;
+    }
+
+    unreachable!("Retry loop should exit via return on success or final failure");
+}
```
jgodlew, line 72: I would like to see some unit tests for the `retry_wrapper` to make sure that it retries as expected.

jgodlew (approved):
Overall looks good. We should probably have some more unit testing on the retry_wrapper though.

cas_client/src/retry_utils.rs
hoytak: It means that now is after the execute_after time, which means we should just go ahead and execute it.

cas_client/src/retry_utils.rs
hoytak: I tried to create unit tests for it using a mock, but there isn't a straightforward way to create a Response object outside of reqwest creating it so I gave up.  I can look deeper. .

assafvayner: rather than make the retry specific to reqwest's Request + Response, can you make it generic? Pass in something that generates a future that returns a response after it has been validated for the upload retry case

PR #335, opened 2025-05-16
MacOS dbg symbols zipped in release workflow

rajatarya:
- needed because `gh release` command does not support uploading directories

cc: fyi @bpronan


assafvayner (approved):

jgodlew (approved):

rajatarya:
Tested locally as much as possible with Joe and Assaf looking over my shoulders.

PR #339, opened 2025-05-16
Merging Cargo.toml dependencies into workspace Cargo.toml

jgodlew:
Merges the dependencies of crate Cargo.toml into the workspace dependencies, allowing for consistent versions of our dependent packages across our crates. Note that since hf_xet is not part of the workspace, it is not able to merge dependencies in there.

Also does the following as a result of this change
* Upgrades the clap version used by utility binaries to clap v4 (many were already there, upgraded the others)
* Upgrades reqwest to `0.12` across all packages (including hf_xet)
* Sorts the dependencies in our Cargo.toml files
* Removes a few dependencies that are no longer used / needed (e.g. toml)


Cargo.toml
```diff
@@ -5,20 +5,19 @@ members = [
...
+    "merklehash",
     "parutils",
+    "progress_tracking",
     "utils",
-    "cas_object",
-    "cas_types",
-    "chunk_cache",
-    "xet_threadpool", 
-    "deduplication", "progress_tracking"]
+    "xet_threadpool"]
```
assafvayner: newline after xet_threadpool

sirahd (approved):
LGTM. Just for my context, why isn't `hf_xet` included in the workspace?

jgodlew:
> Just for my context, why isn't `hf_xet` included in the workspace?

hf_xet wasn't included since it needs to be built with `maturin`, not `cargo`. Adding hf_xet to the workspace will cause certain cargo commands to fail when used at the workspace-level (e.g. `cargo build`), which can make it a bit more difficult to work with the repo.

However, we could revisit this decision as adding hf_xet to the workspace would help keep the dependencies sync'ed. It seems that our CI "should" continue to work (cargo test / fmt / clippy all seem to work with hf_xet in the workspace).

Alternatively, we could try and reduce the hf_xet dependencies by refactoring utilities out of hf_xet and into other/new xet-core libraries.

bpronan (approved):

assafvayner (approved):

PR #340, opened 2025-05-16
Streamline and aggregate file updates for reporting to python

hoytak:
With the current incremental progress updates, the amount of updates going to python is substantial, and each has to acquire a global GIL lock.  This negatively affects the upload speed on fast connections. 

This PR introduces an intermediate aggregation class that quickly aggregates all the incoming progress updates, then sends the aggregated update list to hf_xet once every 200 ms.   With this, the thread contention experienced by the frequent incremental updates is eliminated while still reporting accurate progress to the user.


hoytak:
Note that this depends on https://github.com/huggingface/xet-core/pull/328; Will update this PR once that is merged.

hf_xet/src/progress_update.rs
```diff
@@ -1,20 +1,59 @@
...
+    pub total_bytes: u64,
+    #[pyo3(get)]
+    pub bytes_completed: u64,
+    #[pyo3(get)]
+    pub bytes_completion_increment: u64,
+}
+
+/// Update class for total updates
+#[pyclass]
+pub struct PyTotalProgressUpdate {
```
assafvayner, line 74: Are we passing these back to python? Do you have a sample of how we're going to interface with this in python? (let's add that to the doc comments ideally)

assafvayner (approved):
Can you provide a sample python snippet of how to use the new constructs?

hf_xet/src/progress_update.rs
hoytak: I can add a sample of this.

PR #345, opened 2025-05-20
upload xorb head call

assafvayner:
fix XET-583

Adds a head call on a xorb before uploading a xorb to save time of transmitting a whole xorb. Only enabled when high performance mode is not enabled, users of high performance mode may suffer worse performance with an additional RTT per xorb than by transmitting 64MB.

### DIFF
--- a/data/src/file_upload_session.rs
+++ b/data/src/file_upload_session.rs
@@ -278,7 +278,9 @@ impl FileUploadSession {
 
         // Check if the xorb already exists in the CAS.  If it does, we don't need to upload it again.
         // only check if configured to check
-        if *PREUPLOAD_CHECK_XORB_EXISTS && session.client.exists(&cas_prefix, &xorb_hash).await? {
+        if *PREUPLOAD_CHECK_XORB_EXISTS
+            && session.client.exists(&cas_prefix, &xorb_hash).await.is_ok_and(|exists| exists)
+        {
             // update tracker and return early
             debug!("Xorb {cas_prefix}/{xorb_hash} already exists in CAS, skipping upload");
             session.completion_tracker.register_xorb_upload_completion(xorb_hash).await;
--- a/data/src/file_upload_session.rs
+++ b/data/src/file_upload_session.rs
@@ -17,7 +17,7 @@ use progress_tracking::verification_wrapper::ProgressUpdaterVerificationWrapper;
 use progress_tracking::{NoOpProgressUpdater, TrackingProgressUpdater};
 use tokio::sync::{Mutex, OwnedSemaphorePermit, Semaphore};
 use tokio::task::JoinSet;
-use tracing::{info_span, instrument, Instrument};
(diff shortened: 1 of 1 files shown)


data/src/file_upload_session.rs
```diff
@@ -264,10 +273,19 @@ impl FileUploadSession {
             tokio::task::spawn_blocking(move || SerializedCasObject::from_xorb(xorb, compression_scheme)).await??;
 
         let session = self.clone();
-        let upload_permit = acquire_upload_permit().await?;
         let cas_prefix = session.config.data_config.prefix.clone();
         let completion_tracker = self.completion_tracker.clone();
 
+        // Check if the xorb already exists in the CAS.  If it does, we don't need to upload it again.
+        // only check if configured to check
+        if *PREUPLOAD_CHECK_XORB_EXISTS && session.client.exists(&cas_prefix, &xorb_hash).await? {
```
rajatarya: A thought: what about checking the size of xorb in deciding? Meaning, if xorb is > 10MB then do additional HEAD. Something like that?

rajatarya: Does client.exists() do retries? Will this fail out the registration entirely if await? doesn't return Ok()

rajatarya (commented):
Can some unit-tests be added for this case?

And I am concerned around retries/error handling. Any time we are adding a round-trip to server we need to consider all the error conditions that can happen and make sure we do the right thing.

data/src/file_upload_session.rs
assafvayner: yes to retries, yes on failing the registration on an Err() notably, only an actual irrecoverable error will return an Err() after retries

rajatarya: Good that retries are baked in and only irrecoverable errors will return Err().

However, shouldn't we swallow that irrecoverable error (log it) and continue to try to upload this xorb? Feels strange that we'd bail out from the upload because our call to check if it exists fails.

assafvayner: I will add this, though any irrecoverable error that would be returned by the head call will just be reached again by the post call, these are errors such as not being able to connect to CAS.

assafvayner: Is there a particular reason for a 10MB threshold? Is there some logic with which to determine this threshold?
This factor is relevant

rajatarya: I made up the 10MB number. 

If we "knew" the RTT and current network throughput then we could model that time * 1500-byte TCP packets (assuming not running on BGP networks and routing across older connections), and then assume 200 bytes overhead (gzip, headers, blah) and then see 1300-bytes*\<throughput\> to develop the threshold.

Put another way, we should only consider the HEAD call on cases when pushing 64MB is slower than the RTT for HEAD. But since we probably don't know that (and xorbs aren't always 64MB) we can guess on a lower threshold.

assafvayner:
not planned

PR #346, opened 2025-05-21
Resume capability for interrupted upload file sessions.

hoytak:
Currently, the metadata for uploaded data is only registered at the end of an upload session, and it's discarded if the session is interrupted before it completes.  This means that if an upload session is interrupted, a session resuming that won't take advantage of the previous data. 

With this PR, we cache the metadata for all of the successfully uploaded Xorbs in a persistent directory separate from the cache.  All the metadata in this directory gets pulled in to the temporary session directory at the start, allowing it to be used for deduplication.  Upon successful upload of all the data, it's registered on the server side along with the rest of the session metadata as before.  If that upload is successful, the intermediate xorb metadata along with the rest of the session metadata is merged into the persistent cache.

The user experience now is that when an upload session resumes from a previously interrupted session, all the data is still processed but the data that was uploaded previously is rapidly deduped against the previously uploaded xorbs, effectively meaning the session quickly resumes where it left off.

### DIFF
--- a/cas_client/src/local_client.rs
+++ b/cas_client/src/local_client.rs
@@ -643,7 +643,7 @@ mod tests {
         )
         .unwrap();
 
-        let new_shard_path = shard_in.write_to_directory(&shard_dir_1).unwrap();
+        let new_shard_path = shard_in.write_to_directory(&shard_dir_1, None).unwrap();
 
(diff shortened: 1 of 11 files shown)


assafvayner (approved):
LGTM

PR #348, opened 2025-05-21
Debug Symbol cleanup and instructions

bpronan:
A few adjustments to the debug symbol build and release process:

1. Consolidating the debug symbols into a single zip file to clean up the release asset list and the contents of each zip. 
2. Baking the platforms into the names, so we don't have heterogenous layers of zip files. 
3. Added instructions for how to use the debug symbols to the top-level readme.

Note: for Linux, since we use the fully qualified platform name in the symbol linking phase, this will work. Mac doesn't do any name matching for dSYM files, it only matters that the relocation file is correct. Windows is unchanged.

Resolves XET-582 and XET-571.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,10 +27,12 @@ jobs:
         platform:
           - runner: ubuntu-22.04
             target: x86_64
+            python-target: x86_64
             manylinux: auto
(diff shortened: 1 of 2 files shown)


jgodlew (approved):
LGTM! :shipit:

PR #352, opened 2025-05-27
Added tests for session resuming +  fixed reporting issues.

hoytak:
This PR adds in a test suite to verify that interrupted session indeed resume correctly, with previously uploaded xorbs correctly tracked and reported.  Along with this test suite, it fixes a reporting issue that causes such xorbs to incorrectly be labeled in the progress tracking as still needing to be uploaded, which results is these files appearing to only partially complete.


assafvayner (approved):
:shipit:

PR #353, opened 2025-05-28
Limit shard cache directory

hoytak:
Currently, the shard cache has no upper limit by size; shards are evicted when they expire, which is typically a few weeks.  In large repositories, this still allows it to grow substantially.  

This PR puts a soft upper limit on the number of shards in the shard cache.  When the shard cache is loaded, shards are deleted to bring the total size down to HF_XET_SHARD_CACHE_SIZE_LIMIT in order of oldest to newest.


hoytak:
Note that this PR depends on https://github.com/huggingface/xet-core/pull/352, which should be merged first.

PR #355, opened 2025-05-28
Improved XORB compression method determination

hoytak:
Currently, to save time, we determine the optimal compression scheme by running a detection heuristic on the first xorb encountered during an upload session and using that to set the compression scheme for the entire file.  However, this can leave a lot on the table as it effectively just uses the first file -- often chosen at random due to the parallelism -- to determine the compression for the entire batch.  

This PR strikes a compromise position, where the compression is determined periodically -- specifically at the start of every contiguous run of chunks that come from the same file or every 32 chunks (configurable).  This effectively reduces the time spent determining the optimal compression scheme to a negligible computation burden while still adapting to different file types and file sections within the same repo.

### DIFF
--- a/cas_object/src/lib.rs
+++ b/cas_object/src/lib.rs
@@ -2,9 +2,9 @@ pub mod byte_grouping;
 mod cas_chunk_format;
 mod cas_object_format;
 mod compression_scheme;
+pub mod constants;
 pub mod error;
 mod validate_xorb_stream;
-pub mod constants;
 
 pub use cas_chunk_format::*;
 pub use cas_object_format::*;


assafvayner (approved):
LGTM

cas_object/src/cas_object_format.rs
```diff
@@ -1303,10 +1305,42 @@ impl SerializedCasObject {
...
+            } else {
+                num_chunks
+            };
+
+            if compression_scheme.is_none() && num_chunks != 0 {
+                debug_assert!(xorb.file_boundaries.is_sorted());
+                debug_assert_ge!(xorb.file_boundaries.len(), 0);
+                debug_assert_lt!(*xorb.file_boundaries.last().unwrap(), xorb.data.len());
+
+                for (f_idx, &start_idx) in xorb.file_boundaries.iter().enumerate() {
```
seanses, line 1320: When processing a pile of small files each of one chunk size, this adds the determination overhead to each chunk

cas_object/src/cas_object_format.rs
```diff
@@ -1303,10 +1305,42 @@ impl SerializedCasObject {
         {
             let mut writer = Cursor::new(&mut serialized_data);
 
-            for chunk in xorb.data {
-                // now serialize chunk directly to writer (since chunks come first!)
-                serialize_chunk(&chunk, &mut writer, compression_scheme)?;
-                cas.info.chunk_boundary_offsets.push(writer.stream_position()? as u32);
+            // Set the periodic retesting interval
+            let retest_interval = if *CAS_OBJECT_COMPRESSION_SCHEME_RETEST_INTERVAL > 0 {
```
seanses, line 1309: If we sample I'd rather we sample with Poisson arrivals (exponential intervals) instead of fixed intervals for [PASTA property](https://en.wikipedia.org/wiki/Arrival_theorem). Fixed intervals don't see average when a file has some specific (adversarial) structures.

cas_object/src/cas_object_format.rs
assafvayner: I think this is fine. If this is not fine, then we need a special case for small/1-chunk files to use a the same (ideally-preset LZ4?) compression scheme.

cas_object/src/cas_object_format.rs
hoytak: That's possible, but I don't really want to introduce randomness into how we construct the xorb.   There's already enough effective randomness in the chunk boundaries that correlations with file structure should be broken.   I'm also not really worried about adversarial agents trying to construct something that doesn't compress as well...  it seems like a lot of work to ... increase your own storage costs?

cas_object/src/cas_object_format.rs
hoytak: This is just to reduce the overall overhead; it is still cheaper than running the compression, so a lot of small files is  just slightly slower on that specific xorb, and likely negligible (I think the size requirement for a file to go down this pathway is much larger than a chunk).

cas_object/src/cas_object_format.rs
seanses: sounds good

PR #356, opened 2025-05-28
CI builds with dev, alpha, or beta tags have debug symbols and debug_assertions enabled 

hoytak:
With this PR, the build profile and debug symbol stripping for python wheels is set based on the tag.   If the tag contains *dev*, *alpha*, or *beta* then it builds the wheel using the opt-test profile, which enables the debug assertions within the code and without stripping out the debug symbols.  All other tags keep the current release path.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1415,6 +1415,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "84b26c544d002229e640969970a2e74021aadf6e2f96372b9c58eff97de08eb3"
 
 [[package]]
+name = "heapify"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0049b265b7f201ca9ab25475b22b47fe444060126a51abe00f77d986fc5cc52e"
+
+[[package]]
 name = "heck"
 version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


.github/workflows/release.yml
```diff
@@ -197,11 +207,14 @@ jobs:
...
           fi
+      - uses: ./.github/actions/set-build-profile
+        with:
+          tag: ${{ github.event.inputs.tag }}
       - name: Build wheels
         uses: PyO3/maturin-action@v1
         with:
           target: ${{ matrix.platform.target }}
-          args: --release --out dist
+          args: --profile ${{ env.BUILD_PROFILE }} --out dist
```
bpronan: Unfortunately, we do want to build with `--profile release` for windows. The way that the debug symbols are created are different for the different platforms.

You'll likely have to update the `set-build-profile` action to account for this.

.github/actions/set-build-profile/action.yml
```diff
@@ -0,0 +1,22 @@
...
+  steps:
+    - shell: bash
+      run: |
+        TAG=${{ inputs.tag }}
+        LOWERTAG=$(echo "$TAG" | tr '[:upper:]' '[:lower:]')
+        if [[ "$LOWERTAG" == *dev* || "$LOWERTAG" == *alpha* || "$LOWERTAG" == *beta* ]]; then
+          BUILD_PROFILE=opt-test
+          IS_RELEASE=false
+        else
+          BUILD_PROFILE=release-dbgsymbols
```
bpronan, line 14: See below, this will have to branch based on whether it's windows or not. If windows, it needs to be `release`.

.github/workflows/release.yml
hoytak: I'll just have windows not call the set-build-profile action.

bpronan (approved):

PR #357, opened 2025-05-28
DO NOT MERGE: abort investigation

hoytak:


hoytak:
Found and fixed with https://github.com/huggingface/xet-core/pull/359

PR #360, opened 2025-05-29
limit concurrent gets and update segment size tuner config

assafvayner:
fixes XET-605

Updates the semaphore used to limit concurrent downloads to be used across all uses of a remote client which is just 1 for a session.

As a side effect renames the DownloadScheduler to DownloadSegmentLengthTuner since that's all it does now and adds configuration values for it.

### DIFF
--- a/cas_client/src/download_utils.rs
+++ b/cas_client/src/download_utils.rs
@@ -14,9 +14,8 @@ use http::header::RANGE;
 use http::StatusCode;
 use merklehash::MerkleHash;
 use reqwest_middleware::ClientWithMiddleware;
-use tokio::sync::{OwnedSemaphorePermit, Semaphore};
 use tokio_retry::strategy::ExponentialBackoff;
-use tracing::{debug, error, info, trace};
+use tracing::{debug, error, info, trace, warn};
 use url::Url;
 use utils::singleflight::Group;
 
@@ -24,6 +23,31 @@ use crate::error::{CasClientError, Result};
 use crate::http_client::{Api, BASE_RETRY_DELAY_MS, BASE_RETRY_MAX_DURATION_MS, NUM_RETRIES};
 use crate::remote_client::{get_reconstruction_with_endpoint_and_client, PREFIX_DEFAULT};
 use crate::OutputProvider;
+
+utils::configurable_constants! {
+    // Env (HF_XET_NUM_RANGE_IN_SEGMENT_BASE) base value for the approx number of ranges in the initial
+    // segment size used to download, where a segment is a range of a file that is downloaded
+    // setting this value to 0 causes no segments to be downloaded, this will cause downloads to fail/hang
+    ref NUM_RANGE_IN_SEGMENT_BASE: usize = GlobalConfigMode::HighPerformanceOption {
+        standard: 16, // 16 * ~64MB -> ~1GB initial segment size
+        high_performance: 256, // 128 * ~64MB -> ~16GB initial segment size
+    };
+
+    // Env (HF_XET_NUM_RANGE_IN_SEGMENT_DELTA) delta value for the approx number of ranges in a segment size
+    // used to increase/decrease the segment size by this many approximate ranges
+    // setting this value to 0 causes no segment size change, i.e. will remain constant
+    ref NUM_RANGE_IN_SEGMENT_DELTA: usize = GlobalConfigMode::HighPerformanceOption {
+        standard: 8, // increase/decrease segment size by 8 range ~512MB
+        high_performance: 64, // increase/decrease segment size by 8 ranges ~4GB
+    };
+
+    // Env (HF_XET_NUM_RANGE_IN_SEGMENT_MAX) max value for the approx number of ranges in a segment size
+    // setting this value to 0 will be ignored and the max size will be set to usize::MAX
+    ref NUM_RANGE_IN_SEGMENT_MAX: usize = GlobalConfigMode::HighPerformanceOption {
+        standard: 1024 * 2, // * ~64MB -> max at 128GB segment
+        high_performance: 1024 * 16, // * ~64MB -> max at 1TB segment
+    };
+}
 
 #[derive(Clone, Debug)]
 pub(crate) enum DownloadRangeResult {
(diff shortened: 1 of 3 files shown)


cas_client/src/download_utils.rs
```diff
@@ -25,6 +24,31 @@ use crate::http_client::{Api, BASE_RETRY_DELAY_MS, BASE_RETRY_MAX_DURATION_MS, N
...
+    // setting this value to 0 causes no segment size change, i.e. will remain constant
+    ref NUM_RANGE_IN_SEGMENT_DELTA: usize = GlobalConfigMode::HighPerformanceOption {
+        standard: 8, // increase/decrease segment size by 8 range ~512MB
+        high_performance: 64, // increase/decrease segment size by 8 ranges ~4GB
+    };
+
+    // Env (HF_XET_NUM_RANGE_IN_SEGMENT_MAX) max value for the approx number of ranges in a segment size
+    // setting this value to 0 will be ignored and the max size will be set to usize::MAX
+    ref NUM_RANGE_IN_SEGMENT_MAX: usize = GlobalConfigMode::HighPerformanceOption {
+        standard: 1024 * 2, // * ~64MB -> max at 128GB segment
```
seanses: Getting the fetch info has a high latency: 8s for 100 GB segment and ~40s for 500 GB segment. Setting the segment size to this high value means a long start delay. It's better to use smaller segment size and fetches more often to hide the latency.

cas_client/src/download_utils.rs
```diff
@@ -339,42 +363,67 @@ pub(crate) enum DownloadQueueItem<T> {
...
-            .map_err(CasClientError::from)
+    pub fn from_configurable_constants() -> Arc<Self> {
+        if *NUM_RANGE_IN_SEGMENT_BASE == 0 {
+            warn!(
+                "NUM_RANGE_IN_SEGMENT_BASE is set to 0, which means no segments will be downloaded.
+                   This is likely a misconfiguration. Please check your environment variables."
+            );
+        }
+        let max_num_segments = if *NUM_RANGE_IN_SEGMENT_MAX == 0 {
+            usize::MAX
```
seanses, line 385: HUGE start delay

cas_client/src/download_utils.rs
```diff
@@ -339,42 +363,67 @@ pub(crate) enum DownloadQueueItem<T> {
...
-    n_concurrent_download_task: Arc<Semaphore>,
+    max_segments: usize,
+    delta: usize,
 }
 
-impl DownloadScheduler {
-    pub fn new(n_concurrent_range_get: usize) -> Arc<Self> {
+impl DownloadSegmentLengthTuner {
+    pub fn new(n_range_in_segment_base: usize, max_segments: usize, delta: usize) -> Arc<Self> {
+        if *NUM_RANGE_IN_SEGMENT_BASE == 0 {
```
seanses: `NUM_RANGE_IN_SEGMENT_BASE` doesn't have an effect in this function, so no need to check or warn! here

cas_client/src/download_utils.rs
assafvayner: this value is not the initial value (that's the `BASE` config value). The initial is set to 16 (for standard, 256 for HP), not 2048.

This 2048 value is a maximum (previously uncapped) of the segment size to be used.

cas_client/src/download_utils.rs
assafvayner: Like in the other comment, this value is not the starting number for the segment size factor, it is a maximum. If `HF_XET_NUM_RANGE_IN_SEGMENT_MAX` were to be 0, and max_num_segments here => usize::MAX that would essentially mean that if you call tune_on a million times and keep growing the segment size you could reach an unreasonably large segment size.

Adding a max is more important given that I've added a non-1 delta by default to get the segment size increasing faster. This should swallow any delays in later segment reconstruction downloads if they are large since we will be downloading content at the same time.

Although, with this change using the defaults will enforce that you are never in a situation where you have too long of a time downloading the next segment reconstruction since it is at a max size.

cas_client/src/remote_client.rs
```diff
@@ -376,8 +387,10 @@ impl RemoteClient {
         let chunk_cache = self.chunk_cache.clone();
         let term_download_client = self.http_client.clone();
         let range_download_single_flight = self.range_download_single_flight.clone();
-        let download_scheduler = DownloadScheduler::new(*NUM_CONCURRENT_RANGE_GETS);
+        // min 1
```
seanses: what is this comment?

cas_client/src/remote_client.rs
```diff
@@ -102,6 +106,12 @@ impl RemoteClient {
             None
         };
         let range_download_single_flight = Arc::new(Group::new());
+        let num_range_gets = if *NUM_CONCURRENT_RANGE_GETS == 0 {
+            Semaphore::MAX_PERMITS // virtually no limit
+        } else {
+            NUM_CONCURRENT_RANGE_GETS.max(Semaphore::MAX_PERMITS)
```
seanses: ! `.min()`

assafvayner: good catch

cas_client/src/remote_client.rs
assafvayner: removed

cas_client/src/remote_client.rs
```diff
@@ -44,11 +44,14 @@ pub const CAS_ENDPOINT: &str = "http://localhost:8080";
 pub const PREFIX_DEFAULT: &str = "default";
 
 utils::configurable_constants! {
+// Env (HF_XET_NUM_CONCURRENT_RANGE_GETS) to set the number of concurrent range gets.
+// setting this value to 0 disables the limit, sets it to the max, this is not recommended as it may lead to errors
```
seanses, line 48: Now this is a global limit, we should adjust the value accordingly to match the previous speed. For example, previously in standard mode we have globally `MAX_CONCURRENT_DOWNLOADS = 8` * `NUM_CONCURRENT_RANGE_GETS = 16` = 128 concurrent range gets, now we only have 16.

assafvayner: updated to 128/512. I think we got above that in HP mode, but I would rather keep it conservative to not hit file limits. I think I'd be willing to set the default HP max at ~800

hoytak: 128 seems like a lot; at some point there isn't any more gain, right?

seanses (approved):
LGTM!

cas_client/src/remote_client.rs
seanses: Not sure. We can check updated benchmark results to see if we want to change this. I think there are two reasons that we don't want to set it too low:
1. Ajit observed that ranges are mostly short.
2. We don't gain more beyond some point, but likely no loss. We definitely lose if below that point. And we don't know what's that point

PR #362, opened 2025-05-30
Move batch file uploads to FileUploadSession.

hoytak:
This PR moves the batch file upload code to within the FileUploadSession.  This allows us to scan all of the file sizes and check for their existence on the start of an upload.  With this, we can set the total for the upload progress bar at the start of a session instead of setting it dynamically as new files are added.

### DIFF
--- a/data/src/data_client.rs
+++ b/data/src/data_client.rs
@@ -132,7 +132,16 @@ pub async fn upload_bytes_async(
     Ok(files)
 }
 
-#[instrument(skip_all, name = "data_client::upload_files", fields(session_id = tracing::field::Empty, num_files=file_paths.len()))]
+#[instrument(skip_all, name = "data_client::upload_files", 
+    fields(session_id = tracing::field::Empty,
+    num_files=file_paths.len(),
+    new_bytes = tracing::field::Empty,
+    deduped_bytes = tracing::field::Empty,
+    defrag_prevented_dedup_bytes = tracing::field::Empty,
+    new_chunks = tracing::field::Empty,
+    deduped_chunks = tracing::field::Empty,
+    defrag_prevented_dedup_chunks = tracing::field::Empty
+))]
 pub async fn upload_async(
     file_paths: Vec<String>,
     endpoint: Option<String>,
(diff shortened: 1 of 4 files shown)


data/src/file_upload_session.rs
```diff
@@ -176,6 +181,82 @@ impl FileUploadSession {
...
+            let span = info_span!("clean_file_task");
+
+            cleaning_tasks.push(tokio::spawn(
+                async move {
+                    // First, get a permit to process this file.
+                    let _processing_permit = ingestion_concurrancy_limiter.acquire().await?;
+
+                    // Enable tracing to record this file's ingestion speed.
+                    let span = Span::current();
+                    span.record("file.name", file_name.to_string());
```
jgodlew: Per the docs for [Span::record](https://docs.rs/tracing/latest/tracing/span/struct.Span.html#method.record):

> you cannot record a value for a field that was not specified when the span was created

So this won't work unless we've already specified `file.name` when the `info_span` is created on line 206. Since we already have the file_name and len, those are pretty easy to add to the `info_span` call.

For the dedup metrics, we will need to add those to the span creation as well, but setting the value to `tracing::field::Empty`:
```
let span = info_span!(
      "clean_file_task",
      "file.name" = file_name.to_string(),
      "file.len" = file_size,
      "file.new_bytes" = tracing::field::Empty,
      "file.deduped_bytes" = tracing::field::Empty,
      "file.deduped_chunks" = tracing::field::Empty,
      "file.new_chunks" = tracing::field::Empty,
      "file.defrag_prevented_dedup_bytes" = tracing::field::Empty,
      "file.defrag_prevented_dedup_chunks" = tracing::field::Empty,
 );
```

data/src/data_client.rs
```diff
@@ -145,31 +145,27 @@ pub async fn upload_async(
...
+    let metrics = upload_session.finalize().await?;
 
-    // TODO: Report on metrics
+    // Record dedup metrics.
+    span.record("new_bytes", metrics.new_bytes);
+    span.record("deduped_bytes ", metrics.deduped_bytes);
+    span.record("deduped_chunks", metrics.deduped_chunks);
+    span.record("new_chunks", metrics.new_chunks);
+    span.record("defrag_prevented_dedup_bytes", metrics.defrag_prevented_dedup_bytes);
+    span.record("defrag_prevented_dedup_chunks", metrics.defrag_prevented_dedup_chunks);
```
jgodlew: The same issue noted with the file-level span is here too. Since we didn't specify these in span creation (i.e. the `#[instrument(...)]` on the function signature), these `span.record`s are no-ops / won't add anything.

data/src/file_upload_session.rs
hoytak: Thanks!  Fixed.

data/src/data_client.rs
hoytak: Thanks!  Fixed.

jgodlew (approved):
LGTM! Thanks for adding the metrics, they'll be useful when debugging the repo migrators.

assafvayner (approved):

PR #363, opened 2025-05-30
chunk deserialization special error when chunk header is a xorb footer

assafvayner:
This PR includes a couple of components for an upcoming CAS server change to the accepted xorb payload format.

The validation code is moved to the server codebase, and it is possible to imagine with the removal of the cas_object from the client (all but the chunk format) that the CASObject definition may be moved to the server later.

There is a new error type introduced on chunk header deserialization to indicate the start of a V0/V1 footer rather than a generic failure parsing chunk headers. This allows us to distinguish in validation code whether this is an old xorb format that is still valid or the next bytes are bogus.

Lastly introduces standardized use of futures::io::AsyncRead traits rather than tokio traits. We need to be consistent across usages. There is a compatibility adaptor between the 2 AsyncRead traits, code using the utils crate will use the adaptors as needed.


cas_object/src/cas_chunk_format/deserialize_async.rs
assafvayner, line 1: In this file we were using tokio AsyncRead traits rather than futures traits

PR #365, opened 2025-06-02
Make shard session resume robust to multiple sessions.

hoytak:
Currently, the shard merging algorithm, which is part of the shard resume session, is not robust to shard files being deleted.  This can happen when multiple upload_files are being called, each from the same endpoint.  This PR makes this routine robust to these files disappearing.  

In addition, we add a configurable environment variable, SESSION_RESUME_ENABLE, that can be set to 0 / false to disable session resume behavior.  This is true by default.

### DIFF
--- a/mdb_shard/src/session_directory.rs
+++ b/mdb_shard/src/session_directory.rs
@@ -1,16 +1,16 @@
 use std::collections::HashSet;
-use std::io::{Cursor, Read};
+use std::io::Cursor;
 use std::mem::swap;
 use std::path::Path;
 use std::sync::Arc;
 
-use merklehash::MerkleHash;
 use tokio::task::JoinHandle;
-use tracing::debug;
+use tracing::{error, info};
 
 use crate::error::Result;
 use crate::set_operations::shard_set_union;
 use crate::shard_file_handle::{MDBShardFile, MDBShardFileVec};
+use crate::{MDBShardFileFooter, MDBShardFileHeader, MDBShardInfo};
 
 /// Merge a collection of shards, deleting the old ones.
 /// After calling this, the passed in shards may be invalid -- i.e. may refer to a shard that doesn't exist.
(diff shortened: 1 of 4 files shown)


mdb_shard/src/session_directory.rs
```diff
@@ -44,124 +56,108 @@ pub fn merge_shards(
...
-
-                if copy_preserved_source_shards {
-                    let copied_sfi = cur_sfi.copy_into_target_directory(&target_directory)?;
-                    finished_shards.push(copied_sfi);
-                    shards_to_remove.push(cur_sfi.clone());
-                } else {
-                    finished_shards.push(cur_sfi.clone());
-                }
+    let mut dest_shards = Vec::<Arc<MDBShardFile>>::with_capacity(shards.len());
+    let mut ingested_shards: Vec<Arc<MDBShardFile>> = Vec::with_capacity(shards.len());
```
assafvayner, line 76: style option, you can use `MDBShardFileVec::with_capacity` here, I don't really care which to use, as long as it's consistent for all `MDBShardFileVec`

don't change anything if you prefer the current version.

mdb_shard/src/shard_file_handle.rs
```diff
@@ -245,31 +264,71 @@ impl MDBShardFile {
...
+    pub fn read_into_buffer(&self, buffer: &mut Vec<u8>) -> Result<()> {
+        let mut call = || -> Result<()> {
+            let n = self.shard.num_bytes();
+            buffer.resize(n as usize, 0);
+
+            std::fs::File::open(&self.path)?.read_exact(buffer)?;
+            Ok(())
+        };
+
+        let res = call();
```
assafvayner: can you get rid of the `call` function? Unless there's a good reason for it?

mdb_shard/src/shard_file_handle.rs
```diff
@@ -245,31 +264,71 @@ impl MDBShardFile {
...
+                Err(e) => {
+                    if skip_on_error {
+                        info!("Error loading shard {file_name:?}; skipping.");
+                        return Ok(());
+                    } else {
+                        error!("Error reading shard {file_name:?}; skipping.");
+                        return Err(e);
+                    }
+                },
+            };
```
assafvayner: slight preference for a different syntax here:
```
let Ok(s) = s_res else {
  if skip_on_error {
    ...
    return Ok(());
  }
  // early returned, no else
  return Err(e)
}
```

mdb_shard/src/shard_file_handle.rs
hoytak: It just consolidates the error handling to one place -- both sources of errors are handled by the same code.  Seemed easier than copy-pasting the error logic.

mdb_shard/src/shard_file_handle.rs
hoytak: You don't have access to the error object in this method or I would do it that way!

mdb_shard/src/session_directory.rs
hoytak: I think I should get rid of the MDBShardFileVec type now that I have a struct holding the return values...

mdb_shard/src/shard_file_handle.rs
hoytak: But I guess I don't use the buffer.clear() functionality so I can drop it.

PR #371, opened 2025-06-04
Small optimizations for chunking / upload path

hoytak:
This PR converts the fundamental data type in the Chunk class from Arc<[u8]> to bytes::Bytes.  The latter functions like the former, but allows us to avoid a copy of the data on creation as conversion from a vector is trivial.  In addition, it allows creating slices by reference, so chunks can just refer to their original data without a copy.  This is now okay, as we convert the xorbs to their compressed form quickly so chunks do not hang around long, meaning the extra memory overhead of this is negligible.


data/src/file_cleaner.rs
```diff
@@ -27,7 +28,7 @@ pub struct SingleFileCleaner {
     // Common state
     session: Arc<FileUploadSession>,
 
-    // The chunker
+    // The chunker.  If None, then the chunking is being done on a background compute thread.
```
assafvayner: is it meant to be an option?

data/src/file_cleaner.rs
```diff
@@ -82,34 +83,48 @@ impl SingleFileCleaner {
...
         }
 
         Ok(())
     }
 
-    async fn add_data_impl(&mut self, data: &[u8]) -> Result<()> {
-        // Chunk the data.
-        let chunks: Arc<[Chunk]> = Arc::from(self.chunker.next_block(data, false));
+    #[instrument(skip_all, level="debug", name = "FileCleaner::add_data", fields(file_name=self.file_name.as_ref().map(|s|s.to_string()), len=data.len()))]
+    pub(crate) async fn add_data_detail(&mut self, data: Bytes) -> Result<()> {
```
assafvayner: I don't know if `_detail` is the best name, impl was fine imo. (naming is hard admittedly)

data/src/file_cleaner.rs
```diff
@@ -82,34 +83,48 @@ impl SingleFileCleaner {
...
+        // Put the chunking on a compute thread so it doesn't tie up the async schedulers
+        let chunk_data_jh = {
+            let mut chunker = std::mem::take(&mut self.chunker);
```
assafvayner, line 105: Does the chunker have a cheap default? A default value is placed instead of the taken value, so if we do this a lot, we should use some kind of option.

Another idea, not fully baked, but could the stored "Chunker" produce a sort of guard or some new memory cheap component that then we can next_block_bytes on? I'm not worried about safety or correctness, the code is safe/correct, so this is not important; it is fine as is

deduplication/src/chunk.rs
```diff
@@ -0,0 +1,24 @@
...
+use merklehash::{compute_data_hash, MerkleHash};
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Chunk {
+    pub hash: MerkleHash,
+    pub data: Bytes,
+}
+
+impl Chunk {
+    pub fn new(data: Bytes) -> Self {
```
assafvayner, line 11: whenever we call Chunk::new(), do we not already have the hash of the chunk? seems like extra work to recalculate if so

data/src/file_cleaner.rs
hoytak: Ah, from an earlier iteration.  Will swap out.

deduplication/src/chunk.rs
hoytak: We can still init through `Chunk { hash : ..., data : ... };` but currently we only use this path in the code.

data/src/file_cleaner.rs
hoytak: It is cheap; basically a block of zero bytes.  

I thought about putting this logic into the chunker, but currently it doesn't depend on tokio or have any async components, so I don't really think it's worth it to move it there...

deduplication/src/chunk.rs
assafvayner: I see, I am looking at where we do invoke Chunk::new(), the primary case I'm seeing is after finding the next chunk boundary, which returns a chunk size, but is the hash known by the next_boundary function/s? can it be returned too if so?

hoytak: No, the chunking hash is a different hash function.

PR #372, opened 2025-06-04
remove footer serialized from upload xorb payload on remote_client

assafvayner:
merging this PR requires updating CAS server first.

Updates the xorb serialized format to not include the xorb footer when specified.

This is not the cleanest solution necessarily, because the cas_object tests still require that we serialize the footer (those tests are useful for the cas server where xorbs will later actually contain the footer) but are not relevant for the remote_client upload path. Likewise the LocalClient relies on the xorb footer being serialized.

We may want to refactor this code to remove the footer from xet-core entirely. We would then have to re-implement LocalClient somehow (might be worth enough reason to keep the footer in xet-core).


jgodlew (approved):
LGTM! :shipit:

PR #378, opened 2025-06-10
Switch reqwest to rustls-tls from default; use hickory-dns for dns resolution.

hoytak:
This PR switches reqwest to use the native-tls package, which provides a more reliable abstraction over linux openssl implementation oddities.  In addition, it enables the trust-dns feature to embed a more robust dns resolution path, which hopefully will fix the dns name resolution errors such as 

Now, the cas_client package exposes three features.  The native-tls and native-tls-vendored are passed on to hf_xet for building wheels, while rustls-tls is the default. 

```
# rustls-ssl embeds all ssl stuff in a rust package.  This is the most portable option, but also may not respect local 
# network configurations.   Default.
rustls-tls = ["reqwest/rustls-tls"]

# Uses native tls in the request package; this uses the native-tls package to wrap openssl, which is a more robust and portable
# way of ensuring that tls just works. 
native-tls = ["reqwest/native-tls"]

# This uses the above, but statically compiles in openssl, which makes the result more portable at the expense of 
# library size. 
native-tls-vendored = ["reqwest/native-tls-vendored"]
```

To enable this, some dependencies were moved around so that the hf_xet package now depends directly on cas_client, which is the only package to depend on reqwest.


Cargo.toml
```diff
@@ -64,17 +64,13 @@ lz4_flex = "0.11"
...
 parking_lot = "0.12"
 pin-project = "1"
 prometheus = "0.14"
 rand = "0.9"
 rand_chacha = "0.9"
 rayon = "1.5"
 regex = "1"
-reqwest = { version = "0.12", features = ["json", "stream"] }
-reqwest-middleware = "0.4"
-reqwest-retry = "0.7"
```
jgodlew, line 77: I would prefer to keep the versions specified in the `workspace.dependencies` even though they're only being used in the cas_client since that will help us keep track of the dependency versions in the event that we want to add `reqwest*` elsewhere.

cas_client/Cargo.toml
```diff
@@ -29,16 +26,37 @@ futures = { workspace = true }
...
 tempfile = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true }
 tokio-retry = { workspace = true }
 tracing = { workspace = true }
 url = { workspace = true }
 
+reqwest-middleware = "0.4"
+reqwest-retry = "0.7"
+reqwest = { version = "0.12", features = ["json", "trust-dns", "stream"] }
```
jgodlew: `trust-dns` has been re-named to `hickory-dns`: https://docs.rs/reqwest/latest/reqwest/#optional-features, https://github.com/seanmonstar/reqwest/blob/master/Cargo.toml#L74.

jgodlew (approved):
Overall looks good. Just a few comments about the dependencies and features.

Cargo.toml
hoytak: I tried that, but unfortunately it's not compatible with using features; [features] cannot be specified in the workspace Cargo.toml so I needed to move it to here.  Previously, we did this in the hf_xet/Cargo.toml, but I wanted to move it here so that all dependencies on reqwest would go through the same specification.

PR #382, opened 2025-06-11
Shard interface updates

assafvayner:
Changes to be used potentially in a CAS server PR.

- consistent usage of futures::io::AsyncRead and import pattern
- add deserialize_async variants to cas info and file info used structs. The only difference is the use of async readers, but we still just read the whole struct worth (expect top level) of data and deserialize from slice.
- constants exports

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2052,7 +2052,7 @@ dependencies = [
  "async-trait",
  "blake3",
  "clap",
- "futures-io",
+ "futures",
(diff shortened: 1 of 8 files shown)


hoytak (approved):
LGTM! 

I think down the road we can have the streaming shard utilities build up a buffer using read_exact as we get new data, with the output being a minimal shard around a vec.  That way there would be fewer copies...  But this is good for now.

PR #384, opened 2025-06-12
Background loading for shards

hoytak:
Currently, loading all of the shards is done in a blocking manner, which means that a large number of shards causes the call to upload_files to take a long time to get started.  This PR optimizes this path by loading the lookup table sections of the shard directories in the background while the chunking and file reading can get started.

It also introduces a new utility class, RwTaskLock, which provides a RwLock-like interface around a value that can either be specified by the value itself or by a future that resolves to the value.  This makes it easy to background tasks when values like lookup tables are held behind an rwlock-like interface.  This utility is self-contained and unit tests are provided.

### DIFF
--- a/utils/src/rw_task_lock.rs
+++ b/utils/src/rw_task_lock.rs
@@ -27,7 +27,7 @@ pub struct RwTaskLockReadGuard<'a, T, E> {
     guard: RwLockReadGuard<'a, RwTaskLockState<T, E>>,
 }
 
-impl<'a, T, E> Deref for RwTaskLockReadGuard<'a, T, E> {
+impl<T, E> Deref for RwTaskLockReadGuard<'_, T, E> {
     type Target = T;
     fn deref(&self) -> &T {
         match &*self.guard {


utils/src/rw_task_lock.rs
```diff
@@ -0,0 +1,323 @@
...
+                    return Ok(RwTaskLockReadGuard { guard: state });
+                },
+                RwTaskLockState::Error => return Err(E::from(RwTaskLockError::CalledAfterError)),
+                RwTaskLockState::Pending(_) => {},
+            }
+        }
+        // Acquire write lock to initialize if necessary
+        let mut state = self.state.write().await;
+
+        match replace(&mut *state, RwTaskLockState::Error) {
```
assafvayner, line 112: just so I understand, this is `replace` takes out the value replacing it with error.
- this is fine imo, I would have also been ok with some kind of default/unset state, and use `take`
  - but then we would need to assert that anytime we don't have a write guard that state is invalid, and also there's then just more states to manage and that's not ideal.
- on line 125 we technically do not need to set the state to `Error` again, (I don't mind the tautology there, just asking to make sure I understand)

utils/src/rw_task_lock.rs
```diff
@@ -0,0 +1,323 @@
...
+            state: RwLock::new(RwTaskLockState::Pending(task)),
+        }
+    }
+
+    /// Awaitable read: yields a custom read guard or error.
+    pub async fn read(&self) -> Result<RwTaskLockReadGuard<'_, T, E>, E> {
+        // Fast path
+        {
+            let state = self.state.read().await;
+            match &*state {
```
assafvayner, line 101: is `&*` replaceable with some trait function i.e. AsRef or Deref? low-importance style nit, fine as is in the end

utils/src/rw_task_lock.rs
```diff
@@ -0,0 +1,323 @@
...
+    ///     assert_eq!(*lock.read().await?, 20);
+    ///     Ok(())
+    /// }
+    /// ```
+    pub async fn update<Fut, Updater>(&self, updater: Updater) -> Result<(), RwTaskLockError>
+    where
+        Updater: FnOnce(T) -> Fut + Send + 'static,
+        Fut: Future<Output = Result<T, E>> + Send + 'static,
+    {
+        use RwTaskLockState::*;
```
assafvayner, line 187: I see mixed use of RwTaskLockState::Error and standalone Pending, I don't care if to use this import or not (miniscule pref not to) but can you make it consistent i.e. all state variants use the enum prefix or all don't

utils/src/rw_task_lock.rs
```diff
@@ -0,0 +1,323 @@
...
+
+        let mut state_lg = self.state.write().await;
+
+        let state = replace(&mut *state_lg, RwTaskLockState::Error);
+
+        match state {
+            Pending(jh) => {
+                // Chain the old pending future, then the updater.
+                let new_task = tokio::spawn(async move {
```
assafvayner, line 196: I don't mind the spawn here, but for a more generalizable library we may want to avoid the spawn. Technically the update future will be polled when the `.read` function is called, but I think that from what I see in the usages of it, we want it running earlier than when the `.read` is called?

assafvayner (approved):
If you can try to answer my questions before merging that would be nice. I think I get the gist of it, only thing I'm not super sure of is if the update function needs to spawn and replace the pending future or if it can/should just poll/await the future to completion and end on a resolved state (does that make sense?)

utils/src/rw_task_lock.rs
hoytak: The importance of the spawn is that it causes the task to run in the background, which is the point here, whereas normally a future only begins execution on .await.  As for spawn, we're using it heavily elsewhere, so if we wanted to abstract that away we could do it through a wrapper function, but that's a different discussion than this.

utils/src/rw_task_lock.rs
hoytak: I could replace it with .deref().asref(), but this is exactly the same thing...

utils/src/rw_task_lock.rs
hoytak: I thought of that and went with the Error option, since it's local and it would be a programming error if we hit it due to this.   

re line 125 I like making it explicit, but you're right.

hoytak:
> I think I get the gist of it, only thing I'm not super sure of is if the update function needs to spawn and replace the pending future or if it can/should just poll/await the future to completion and end on a resolved state (does that make sense?)

The main idea here is that I want to enable an easy path to change the value in the background without the complexity and difficulty of getting it right on your own.   So here, the update function can chain updates in order to process a bunch in the background, and it's transparent to the caller.  This is the technique I rely on in the shard loading path.

hoytak:
Waiting to merge this until after we cut a release for the dns issues.

PR #392, opened 2025-06-18
streaming shard interface updates

assafvayner:
Changes the streaming_shard/MDBMinimalShard implementation with the following changes:

- minimal shard now holds vec's of MDBFileInfoView's and MDBCASInfoView's
- Each View type holds a bytes::Bytes object instead of `Arc<[u8]>` and an offset
- enables passing in a custom callback taking a reference when deserializing a MinimalShard from an `AsyncRead`
  - this will be used by CAS server.

Also removed deserialize_async functions defined in #382 .


mdb_shard/Cargo.toml
```diff
@@ -30,10 +31,7 @@ heapify = { workspace = true }
 uuid = { workspace = true, features = ["v4", "js"] }
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
 uuid = { workspace = true, features = ["v4"] }
-
-[target.'cfg(not(target_family = "wasm"))'.dev-dependencies]
-# for tokio::test
-tokio = { workspace = true}
+tokio = { workspace = true, features = ["rt-multi-thread"] }
```
assafvayner, line 34: had issues with the shard_benchmark.rs files, this fixed

PR #397, opened 2025-06-26
Generic retry wrapper to consolidate and streamline retry logic.

hoytak:
This PR consolidates the retry logic for the http connections around a single utility, RetryWrapper, and integrates it more cleanly with the logging and parsing logic.  The goal is to replace the retry middleware with this, which doesn't work with streaming connections. 

This PR introduces this and replaces the simpler connection paths in RemoteClient with this wrapper but leaves the previous logic intact.  The next step is to fully switch the remaining cases over to this wrapper to clean up the code.


cas_client/src/remote_client.rs
```diff
@@ -34,10 +34,10 @@ use utils::singleflight::Group;
 use crate::download_utils::*;
 use crate::error::{CasClientError, Result};
 use crate::http_client::{Api, ResponseErrorLogger, RetryConfig};
-use crate::interface::*;
+use crate::interface::{ShardDedupProber, *};
```
assafvayner: this is an odd import? is ShardDedupProber not part of "*"?

cas_client/src/retry_utils.rs
assafvayner, line 1: delete file?

cas_client/src/remote_client.rs
```diff
@@ -142,18 +142,19 @@ impl RemoteClient {
...
-            .await
-            .map_err(|e| CasClientError::Other(format!("request failed with error {e}")))?;
+        let client = self.authenticated_http_client_with_retry.clone();
+        let api_tag = "cas::query_dedup";
 
-        // Dedup shard not found, return empty result
-        if !response.status().is_success() {
+        let response = RetryWrapper::new(api_tag)
+            .with_429_no_retry()
+            .log_errors_as_info()
```
assafvayner, line 150: is this an override over the default log level? what is the default/is there a default?

cas_client/src/retry_wrapper.rs
```diff
@@ -0,0 +1,556 @@
...
+                } else {
+                    format!("{txt}: {api} api call failed: {err}")
+                }
+            };
+
+            if self.log_errors_as_info || log_as_info {
+                info!("{msg}");
+            } else {
+                error!("{msg}");
+            }
```
assafvayner, line 75: instead can we hold a tracing::Level, default it to tracing::Level::ERROR and emit `event!` here? if you want to override it by passing in log_as_info, then pass in an `Option<tracing::Level>`?

assafvayner: I would rather have formally more control, even if we use the same settings everywhere. Fine if you really prefer to not expose that option.

cas_client/src/retry_wrapper.rs
```diff
@@ -0,0 +1,556 @@
...
+    pub async fn run_and_process<T, ReqFut, ReqFn, ProcFut, ProcFn>(
+        self,
+        make_request: ReqFn,
+        process_fn: ProcFn,
+    ) -> Result<T, CasClientError>
+    where
+        ReqFn: Fn() -> ReqFut + Send + 'static,
+        ReqFut: std::future::Future<Output = Result<Response, reqwest_middleware::Error>> + 'static,
+        ProcFn: Fn(Response) -> ProcFut + Send + 'static,
+        ProcFut: std::future::Future<Output = Result<T, RetryableReqwestError>> + 'static,
```
assafvayner, line 176: Can you add comments on these types, they are quite long/complex-looking, they aren't too complex, but it's a minor chore to read. similar in run_and_extract_json

cas_client/src/remote_client.rs
hoytak: The default log level is to log fatal errors as errors, i.e. the connection couldn't be completed either due to non-transient errors or running out of retries.  The current logging -- prior to this PR -- is done with a log_error after the send, which isn't as consistent as it doesn't include a lot of the context, just the error.

cas_client/src/retry_wrapper.rs
hoytak: What's the need for that extra control here?  I'm wondering what scenarios would benefit from it.  Currently, each of the calls does a log_error call except the global dedup query, which swallows any errors; this reflects that behavior. 

Also, this logs retries as info, and the error is for fatal errors.

cas_client/src/retry_wrapper.rs
hoytak: Sure, happy to.

cas_client/src/retry_wrapper.rs
```diff
@@ -0,0 +1,587 @@
...
+
+            if self.log_errors_as_info || log_as_info {
+                info!("{msg}");
+            } else {
+                error!("{msg}");
+            }
+            CasClientError::ServerConnectionError(msg)
+        };
+
+        use RetryableReqwestError::{FatalError, RetryableError};
```
jgodlew: Should we move this up to the top of the file? I don't see much use having the import in what seems to be a common path (request is successful).

jgodlew (approved):
Overall, looks good!

cas_client/src/retry_wrapper.rs
hoytak: Done!

PR #403, opened 2025-07-02
xet session

assafvayner:
fix XET-528

adds XetSession export out of hf_xet.


rajatarya:
@assafvayner Is this ready for review?

My quick question: why separate into XetUploadSession and XetDownloadSession? As a user I'd like a `XetSession` - see the pseudocode from the Linear issue - so I can just have an active session and do uploads/downloads as I'd like.

rajatarya:
Also, shouldn't there be tests here?

assafvayner:
I tagged you to get a temp-check on this

- the separation between the download session and the upload session was an ask from @hoytak, I'm ambivalent on which interface to go with.

PR #405, opened 2025-07-02
Adding buffer to JWT token expiration

jgodlew:
Should help reduce the number of HTTP-401 users see by refreshing the tokens earlier than their expiry time.

### DIFF
--- a/utils/src/auth.rs
+++ b/utils/src/auth.rs
@@ -5,6 +5,9 @@ use std::time::{SystemTime, UNIX_EPOCH};
 
 use crate::errors::AuthError;
 
+/// Seconds before the token expires to refresh
+const REFRESH_BUFFER_SEC: u64 = 30;
+
 /// Helper type for information about an auth token.
 /// Namely, the token itself and expiration time
 pub type TokenInfo = (String, u64);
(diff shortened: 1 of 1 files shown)


utils/src/auth.rs
```diff
@@ -5,6 +5,9 @@ use std::time::{SystemTime, UNIX_EPOCH};
 
 use crate::errors::AuthError;
 
+/// Seconds before the token expires to refresh
+const REFRESH_BUFFER_SEC: u64 = 30;
```
assafvayner, line 9: note that the token validity time must be greater than this time, for now that's simple to enforce since tokens are valid for minutes, but this means our tokens cannot have a really short expiry < 30 secs.

assafvayner (approved):
:shipit:

utils/src/auth.rs
jgodlew: Yeah, considering that the token expiry is [15 min](https://github.com/huggingface-internal/moon-landing/blob/39575645c18aac9b575db352c92d8e8a55fc5592/server/lib/config.ts#L684), we should be fine with a 30s buffer. Even if the expiration becomes smaller (e.g. 15s), that just means that we will need unique tokens for each request to CAS, which adds latency to all calls, but doesn't cause errors.

PR #410, opened 2025-07-10
Automatic concurrency adjustment for parallel uploads and downloads 

hoytak:
This PR implements a robust controller for adaptively adjusting the concurrency of upload and download transfers.  

By default, the controller dynamically adjusts the concurrency so that 80% of the transfers are completed within 20 seconds; it increases the concurrency as long as this criteria is met.  When more than 20% of the transfers begin taking longer than that, concurrency is reduced.  Concurrency adjustments are throttled so that increasing the concurrency happens only every 500ms, and decreasing it happens at most once every 250ms. 

This PR implements this adaptive control on the upload path; the download changes will happen in a separate PR.


cas_client/src/constants.rs
```diff
@@ -11,4 +11,45 @@ utils::configurable_constants! {
...
+    /// Log the concurrency on this interval.
+    ref CONCURRENCY_CONTROL_LOGGING_INTERVAL_MS: u64 = 10 * 1000;
+
+    /// The maximum number of simultaneous xorb and/or shard upload streams permitted by
+    /// the adaptive concurrency control. Can be overwritten by environment variable "HF_XET_MAX_CONCURRENT_UPLOADS".
+    ref MAX_CONCURRENT_UPLOADS: usize = 100;
+
+    /// The minimum number of simultaneous xorb and/or shard upload streams that the
+    /// the adaptive concurrency control may reduce the concurrancy down to on slower connections.
+    ref MIN_CONCURRENT_UPLOADS: usize = 2;
```
seanses: It seems 1 is more reasonable for the worst case.

data/src/file_upload_session.rs
```diff
@@ -362,19 +341,17 @@ impl FileUploadSession {
...
+        let upload_permit = self.client.acquire_upload_permit().await?;
         let cas_prefix = session.config.data_config.prefix.clone();
         let completion_tracker = self.completion_tracker.clone();
 
         self.xorb_upload_tasks.lock().await.spawn(
             async move {
                 let n_bytes_transmitted = session
                     .client
-                    .upload_xorb(&cas_prefix, cas_object, Some(completion_tracker))
+                    .upload_xorb_with_permit(&cas_prefix, cas_object, Some(completion_tracker), upload_permit)
```
seanses: An `upload_xorb` function is exposed through the client interface that acquires the permit automatically. It's better to use this function instead of managing permits from outside. 

Is it possible to make the `upload_xorb_with_permit` function private?

data/src/shard_interface.rs
```diff
@@ -282,13 +281,10 @@ impl SessionShardInterface {
                     }
 
                     // Upload the shard.
-                    shard_client
-                        .upload_shard(&shard_prefix, &si.shard_hash, false, data, &salt)
+                    client
+                        .upload_shard_with_permit(&shard_prefix, &si.shard_hash, false, data, &salt, upload_permit)
```
seanses: Same as above, use the `upload_shard` version that acquires the permit automatically.

cas_client/src/constants.rs
```diff
@@ -11,4 +11,45 @@ utils::configurable_constants! {
...
+    /// The size of a large transfer.
+    ref CONCURRENCY_CONTROL_LARGE_TRANSFER_NUM_BYTES : u64 = 64_000_000;
+
+    /// The minimum time in milliseconds between adjustments when increasing the concurrency.
+    ref CONCURRENCY_CONTROL_MIN_INCREASE_WINDOW_MS : u64 = 500;
+
+    /// The minimum time in milliseconds between adjustments when decreasing the concurrency.
+    ref CONCURRENCY_CONTROL_MIN_DECREASE_WINDOW_MS : u64 = 250;
+
+    /// The maximum number of connection successes and failures to examine when adjusting the concurrancy.
```
seanses: This value and the below two `CONCURRENCY_CONTROL_TARGET_SUCCESS_RATIO_LOWER` and `CONCURRENCY_CONTROL_TARGET_SUCCESS_RATIO_UPPER` share the same comments but they mean different things.

utils/src/adjustable_semaphore.rs
```diff
@@ -0,0 +1,309 @@
...
+    pub fn decrement_total_permits(&self) -> bool {
+        // Make sure we can do the decrease.
+        if !attempt_decrement(&self.total_permits, self.min_permits) {
+            return false;
+        }
+
+        if let Ok(permit) = self.semaphore.clone().try_acquire_owned() {
+            permit.forget();
+        } else {
+            self.enqueued_permit_decreases.fetch_add(1, Ordering::Relaxed);
```
seanses, line 98: I don't think this effectively reduces the concurrency: when permits are acquired by tasks as soon as they become available, the attempts to decrement always get enqueued, and ignored (until there's a signal to increment). I think we need to check this queue in `acquire()` before the actual acquisition happens?

seanses (commented):
I understand the controller logic, but more experiments are needed before this becomes a convincing solution. Places I have doubt on:

- A success signal is defined as a 200 status code & transfer finishes in a linear interpolated hardcoded time limit. This yields different speed expectations for xorbs with different size, e.g. 6.9 Mbps for a 10 MB xorb and 25.6 Mbps for a 64 MB xorb. In practice they should have the same speed. 

- After 90% success signals within a observation window, controller attempts to increment concurrency by 1. It's unclear what's target for this controller to maximize. 25.6 Mbps bandwidth can be easily reached, this control logic means the concurrency will shoot to the max value(100). But the overall throughput with 100 concurrent flows is not guaranteed to be greater than, say 20 flows. And it's unknown that with 100 concurrent flows if a single flow speed will drop below 25.6 Mbps to trigger a decrease signal. This implies that users need to manually tune the speed expectation argument based on their network condition, and thus provides no better experience than tuning the concurrency directly.

data/src/file_upload_session.rs
hoytak: It's done this way (though more so true with upload_shard), as we first acquire a permit, then spawn the background task to upload the xorb/shard.  If we fold them both in here, we build up background tasks waiting on permit acquisition and run out of memory, as it's the permit acquisition that keeps the number of in-memory xorbs under control.  The non-permit versions are just here to make the testing code easy.

utils/src/adjustable_semaphore.rs
hoytak: The drop function in the permit checks this value before returning itself to the main semaphore; if it can decrease this value, it does so and then forgets itself, as this is the only way to decrease the number of permits in a semaphore.  

Checking it in acquire has two issues -- 1., it breaks fairness which we depend on for the uploads and downloads to run in order and not block other things, and 2. we need an active permit to actually decrease the concurrency. 

The tests below cover this pretty heavily, including a pretty robust stress test, so I'm quite confident in this design.

hoytak:
> I understand the controller logic, but experiments are needed before this becomes a convincing solution. Places I have doubt on:
> 
> * A success signal is defined as a 200 status code & transfer finishes in a linear interpolated hardcoded time limit. This yields different speed expectations for xorbs with different size, e.g. 6.9 Mbps for a 10 MB xorb and 25.6 Mbps for a 64 MB xorb. In practice they should have the same speed.

I definitely agree it should have more testing, and I'll try to set some things up to do that.  Note with the speed calculations, there is a constant time overhead due to server processing as well which I was trying to account for in those numbers.  Granted, though, they are empirically calculated.  

> * After 90% success signals within a observation window, controller attempts to increment concurrency by 1. It's unclear what's target for this controller to maximize. 25.6 Mbps bandwidth can be easily reached, this control logic means the concurrency will shoot to the max value(100). But the overall throughput with 100 concurrent flows is not guaranteed to be greater than, say 20 flows. And it's unknown that with 100 concurrent flows if a single flow speed will drop below 25.6 Mbps to trigger a decrease signal. This implies that users need to manually tune the speed expectation argument based on their network condition, and thus provides no better experience than tuning the concurrency directly.

It's not really accurate to think of this in terms of speed, but rather in terms of whether or not the bandwidth is saturated, and then to back off before it's overloaded.  There's a wide range of concurrency values where it remains saturated.  In this window, the time per transfer increases due to congestion, which is fine, until it becomes too congested and things start getting dropped or failing.  

The idea is to increase it to the point where we're confident it's saturated, but before problems arise.   The early detection of problems is indicated given by the latency for a connection increasing past a threshold.  That said, the value of this threshold is rather arbitrary and tuned for my connection, so I think I'm going to change that part to be a bit more automatic.

cas_client/src/constants.rs
```diff
@@ -11,4 +11,45 @@ utils::configurable_constants! {
...
+    ref CONCURRENCY_CONTROL_TRACKING_SIZE : usize = 20;
+
+    /// The maximum number of connection successes and failures to examine when adjusting the concurrancy.
```
assafvayner: small typo in a couple comments, `concurrancy` should be `concurrency`

data/src/shard_interface.rs
hoytak: See comment above.

cas_client/src/retry_wrapper.rs
```diff
@@ -204,10 +228,40 @@ impl RetryWrapper {
                         Ok(resp) => self_.process_ok_response(try_idx, resp),
                     };
 
-                    match checked_result {
-                        Ok(ok_response) => process_fn(ok_response).await,
-                        Err(e) => Err(e),
+                    let (reply_bytes, processing_result) = match checked_result {
+                        Ok(ok_response) => (ok_response.content_length().unwrap_or(0), process_fn(ok_response).await),
```
assafvayner, line 256: this size IIUC is the size of the body of the response, not the size of the data sent in the request.

If it is important here to know the size of the request payload, then we should enforce it in another way.

utils/src/adjustable_semaphore.rs
```diff
@@ -0,0 +1,309 @@
+use std::sync::atomic::Ordering::SeqCst;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Arc;
+
+use tokio::sync::{AcquireError, Semaphore};
+
+/// An adjustable semaphore in which the total number of permits can be adjusted at any time
+/// between a minimum and a maximum bound.  Adjustments do not affect any permits currently
+/// issued; if an adjustment cannot be permformed immediately, then it is resolved before any
+/// new permits are issued.
```
assafvayner: Can you add a comment as to why this is needed over the standard tokio::sync::Semaphore which also has the functionality to add and forget permits

cas_client/src/retry_wrapper.rs
hoytak: It's only used here for the download path; on upload this value is specified when the permit is given to this class.  So hopefully it accurately reflects the transfer size.  This distinction is made below, but I added a comment.

utils/src/adjustable_semaphore.rs
hoytak: The tokio Semaphore itself cannot forget permits; permits may be discarded and never returned, but there isn't any other way to decrease the number of permits.  This turned into quite the headache to manage in the controller class, so I created this utility class to implement and test the forgetting functionality reliably and cleanly.   This simplified the logic in the adaptive concurrency code a lot.

cas_client/src/remote_client.rs
```diff
@@ -733,6 +736,38 @@ impl Client for RemoteClient {
         Ok(Some(response.bytes().await?))
     }
 
+    async fn acquire_upload_permit(&self) -> Result<ConnectionPermit> {
+        self.upload_concurrency_controller.acquire_connection_permit().await
+    }
+
+    #[instrument(skip_all, name = "RemoteClient::upload_shard", fields(shard.len = shard_data.len()))]
+    async fn upload_shard_with_permit(&self, shard_data: Bytes, upload_permit: ConnectionPermit) -> Result<bool> {
```
seanses, line 744: Why don't we fold this function into the below `upload_shard` function?

cas_client/src/interface.rs
```diff
@@ -51,17 +52,40 @@ pub trait Client {
...
     ) -> Result<u64>;
 
+    /// Upload a new shard, acquiring the permit.
+    async fn upload_shard(&self, shard_data: bytes::Bytes) -> Result<bool> {
+        let permit = self.acquire_upload_permit().await?;
+        self.upload_shard_with_permit(shard_data, permit).await
+    }
+
+    /// Upload a new xorb, acquiring the permit.
+    async fn upload_xorb(
```
seanses: If acquiring a permit is a must then we can keep only one between `upload_xorb_with_permit` and `upload_xorb`, and between `upload_shard_with_permit` and `upload_shard`.

cas_client/src/remote_client.rs
hoytak: We have to separate the permit acquisition and upload step for the xorb upload as the blocking permit acquisition is how we control memory as well.

cas_client/src/interface.rs
hoytak: Happy to do that, but it meant changing a LOT of tests, so I felt like keeping these in here for that reason.  I could mark them as test only.

cas_client/src/remote_client.rs
seanses: Yeah I noticed the interface already implemented an automatic permit acquisition 
```
async fn upload_shard(&self, shard_data: bytes::Bytes) -> Result<bool> {
        let permit = self.acquire_upload_permit().await?;
        self.upload_shard_with_permit(shard_data, permit).await
    }
```
We can clean up the similar functions and repeated logic, maybe delete the `upload_shard` in this remote_client.rs

cas_client/src/remote_client.rs
```diff
@@ -225,6 +227,7 @@ impl RemoteClient {
             #[cfg(not(target_family = "wasm"))]
             range_download_single_flight: Arc::new(Group::new()),
             shard_cache_directory,
+            upload_concurrency_controller: AdaptiveConcurrencyController::new_upload("upload"),
```
seanses, line 230: Why do we differentiate `upload` and `download`? This `RemoteClient` is versatile and there's no way to tell whether it was constructed for upload or download or both. Is the underlying controller logic different between the two operations?

xet_config/src/groups/client.rs
```diff
@@ -94,4 +114,189 @@ crate::config_group!({
...
+    /// Use the environment variable `HF_XET_CLIENT_AC_LOGGING_INTERVAL_MS` to set this value.
+    ref ac_logging_interval_ms: u64 = 10 * 1000;
+
+    /// The maximum number of simultaneous xorb and/or shard upload streams permitted by
+    /// the adaptive concurrency control.
+    ///
+    /// The default value is 64.
+    ///
+    /// Use the environment variable `HF_XET_CLIENT_AC_MAX_UPLOAD_CONCURRENCY` to set this value.
+    ref ac_max_upload_concurrency: usize = 64;
```
seanses, line 260: I feel for this and the below ac values there's no need to differentiate `upload` and `download`, the algorithm will learn the pattern by itself

cas_client/src/remote_client.rs
hoytak: Currently, a lot of the reason is that we have specified different bounds on the minimum and maximum concurrencies, with the maximum parallel download much larger than the maximum parallel uploads.  But I'm quite happy to merge them -- I do agree it would simplify things.

cas_client/src/interface.rs
hoytak: Done.

Issue #400, opened 2025-06-30
Unable to download XET files from repositories


jsulz:
### Describe the bug

Opening this on behalf of @albertodepaola, notes below are from an internal conversation - feel free to add any additional information you have @albertodepaola!

Currently unable to download any Xet-backed files from repositories (example repo: https://huggingface.co/meta-llama/Meta-Llama-3-8B-Instruct). See error log below. 

Non-Xet files (e.g., any `json` file) is downloaded successfully. 

Logged in behind a corporate proxy; once `HF_HUB_DISABLE_XET` was turned on, download over HTTP was successful. 

### Reproduction

Using: 

`huggingface-cli download [repo-id]`

### Logs

```shell
{"timestamp":"2025-06-27T00:05:03.923029Z","level":"WARN","fields":{"message":"Reqwest(reqwest::Error { kind: Request, source: hyper_util::client::legacy::Error(Connect, ConnectFailed(ConnectError(\"dns error\", Custom { kind: Uncategorized, error: \"failed to lookup address information: Name or service not known\" }))) }). Retrying..."},"filename":"/home/runner/work/xet-core/xet-core/cas_client/src/http_client.rs","line_number":242}
```

### System info

```shell
- hf-xet: `1.1.5`
- No `HF_XET*` environment variables set
```

jsulz:
@albertodepaola - two follow-ups for you: 

1. Were there any more details you could provide about the proxy you were working behind?
2. Were you able to download successfully off the work network?

goodsonjr:
I have also experienced this issue where our systems do not have direct access to the internet or DNS. These require the use of an HTTP proxy (Squid) configured with the usual `http[s]_proxy` variables. I am used to Reqwest _just working_ with our setup, but this appears to be due to the inclusion of its `system-proxy` feature as a [default feature](https://github.com/seanmonstar/reqwest/blob/e6e2d8a25dfcdfae026946672d3aa015214a10c1/Cargo.toml#L30) allowing it to automatically pick up and utilize these proxies.

The xet-core crate brings in `reqwest` but specifies a list of features that doesn't include `system-proxy` and explicitly disables the default features: https://github.com/huggingface/xet-core/blob/8fbe9684fc13d6e576e72f384311c9bc392c5190/cas_client/Cargo.toml#L40-L43 I believe this is preventing the library from using the system proxy. I don't know whether this was intentional or not, perhaps something about the Xet transfers won't work through proxies.

hoytak:
[...]

PR #413, opened 2025-07-14
Enabling proxy support for reqwest

hoytak:
Adds the system-proxy feature for reqwest in order to handle proxies specified on the system.  This required a minor version upgrade to support.  Hopefully this is a fix for https://github.com/huggingface/xet-core/issues/400.

### DIFF
--- a/cas_client/Cargo.toml
+++ b/cas_client/Cargo.toml
@@ -37,9 +37,10 @@ serde_json = { workspace = true }
 
 reqwest-middleware = "0.4"
 reqwest-retry = "0.7"
-reqwest = { version = "0.12", features = [
+reqwest = { version = "0.12.22", features = [
     "json",
     "stream",
+    "system-proxy"
 ], default-features = false }
 hyper-util = "0.1.10"
 tower-service = "0.3.3"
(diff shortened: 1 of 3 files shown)


jgodlew (approved):
:shipit:

cas_client/Cargo.toml
```diff
@@ -37,9 +37,10 @@ serde_json = { workspace = true }
 
 reqwest-middleware = "0.4"
 reqwest-retry = "0.7"
-reqwest = { version = "0.12", features = [
+reqwest = { version = "0.12.22", features = [
     "json",
     "stream",
+    "system-proxy"
```
bpronan, line 43: FYI: this doesn't look necessary. It seems to be [enabled by default](https://docs.rs/reqwest/latest/reqwest/index.html#:~:text=system%2Dproxy%20(enabled%20by%20default)%3A%20Use%20Windows%20and%20macOS%20system%20proxy%20settings%20automatically.).

hoytak: We have the default-features set to off to trim the library size as we don't use http2 support.

PR #414, opened 2025-07-14
Move MDB v1 to reference test code; add standalone hash functions

hoytak:
This PR removes the old merkledb/ code, extracting the remaining functions still used for calculating the aggregate hash functions and moves them to merklehash.  This provides a significant gain in code clarity while also providing a speedup to hash computation as the entire merkle tree is not built to compute the hash.  Existing tests ensure correctness.


seanses (approved):
I think we can just drop the "merkledb_test_reference" crate, no code is referencing it and we can always find it in git history.

merklehash/src/aggregated_hashes.rs
coyotte508, line 1: Looks very nice!

Maybe a update/finish API would be great, if eg the file is 1TB and there are over 10M chunks, which is 32B+8B = 400MB in memory + overhead (esp if ported to WASM where there would be more overhead with JS storage etc)

And besides the memory footprint, to avoid blocking.

Not a priority for this PR, but to keep in mind for a follow up

hoytak: Yes, I have a plan to do that.  I agree it's quite a bit of overhead.

coyotte508:
the thin wasm goes from 137kB to <100kB with this!

PR #416, opened 2025-07-16
add verification hash and file hash functions

assafvayner:
Adding more functions to the thin wasm export, needed for forming shards in particular

CC @coyotte508

After the first draft, binary size is 97K for me

### DIFF
--- a/hf_xet_thin_wasm/src/lib.rs
+++ b/hf_xet_thin_wasm/src/lib.rs
@@ -1,4 +1,4 @@
-use merklehash::{MerkleHash, xorb_hash};
+use merklehash::{DataHashHexParseError, MerkleHash, xorb_hash};
 use serde::{Deserialize, Serialize};
 use wasm_bindgen::prelude::*;
 
@@ -65,6 +65,8 @@ fn serialize_result<T: Serialize>(result: &T) -> Result<JsValue, JsValue> {
     res
 }
 
+/// takes an Array of Objects of the form { "hash": string, "length": number }
+/// and returns a string of a hash
 #[wasm_bindgen]
 pub fn compute_xorb_hash(chunks_array: JsValue) -> Result<String, JsValue> {
     let js_chunks: Vec<JsChunk> =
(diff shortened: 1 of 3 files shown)


hf_xet_thin_wasm/src/lib.rs
```diff
@@ -65,20 +65,45 @@ fn serialize_result<T: Serialize>(result: &T) -> Result<JsValue, JsValue> {
     res
 }
 
+/// takes an Array of Objects of the form { "hash": string, "length": number }
```
seanses, line 68: Better to document the length type properly. I remember in JS int32 and int64 (BigInt) are different types. Not sure if `serde_wasm_bindgen::from_value` is going to silently cast the type or throw out errors if there's type mismatch.

hf_xet_thin_wasm/src/lib.rs
```diff
@@ -65,20 +65,45 @@ fn serialize_result<T: Serialize>(result: &T) -> Result<JsValue, JsValue> {
...
-        })
-        .collect::<Result<Vec<_>, _>>()?;
+        .map(|jsc| Ok((MerkleHash::from_hex(&jsc.hash)?, jsc.length as usize)))
+        .collect::<Result<_, DataHashHexParseError>>()
+        .map_err(|e| JsValue::from(e.to_string()))?;
 
     Ok(xorb_hash(&chunks).hex())
 }
+
+/// takes an Array of Objects of the form { "hash": string, "length": number }
```
seanses, line 84: Same as above.

hf_xet_thin_wasm/src/lib.rs
assafvayner: In this case `number` is correct (it's actually a f64 IIRC in JS) but gets converted inside this function to u32, this is chunk length so we will usually deal with it in u32 not u64.

I've added this comment to reflect the JS types, we will later perhaps want to revise/add docs to this package.

PR #418, opened 2025-07-18
add type hints to hf_xet

assafvayner:
This PR adds a typing file to hf_xet, my particular interest to have this is the lack/limitation of IDE support when working with hf_xet functions from huggingface_hub library in pycharm/vscode.

### DIFF
--- a/hf_xet/hf_xet.pyi
+++ b/hf_xet/hf_xet.pyi
@@ -1,0 +1,84 @@
+from typing import List, Optional, Tuple, Callable, Union
+
+class PyXetUploadInfo:
+    hash: str
+    file_size: int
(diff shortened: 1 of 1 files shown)


hoytak (approved):
I'm assuming you're testing this in the IDE and they work... 

Thanks for doing this!  LGTM.

assafvayner:
> I'm assuming you're testing this in the IDE and they work...

Yes, I tested with Pycharm, but I need to also probably verify this with VSCode. I'm holding off a bit on merging this to make sure that it doesn't interfere with huggingface_hub, I expect this will cause lints to fail on huggingface_hub.

PR #419, opened 2025-07-18
remove footer from upload shard payload

assafvayner:
Since the upload shard API hardening the server no longer uses the shard footer. This PR truncates the footer from the body of the upload shard api call when sending to CAS server.


data/src/shard_interface.rs
```diff
@@ -272,7 +272,16 @@ impl SessionShardInterface {
             shard_uploads.spawn(
                 async move {
                     debug!("Uploading shard {shard_prefix}/{:?} from staging area to CAS.", &si.shard_hash);
-                    let data: bytes::Bytes = std::fs::read(&si.path)?.into();
+                    let mut data: bytes::Bytes = std::fs::read(&si.path)?.into();
```
hoytak: You can just use `si.shard.metadata` -- the shard_info you load below is already present in `si.shard`, so you can just do 
```
let truncate_idx = si.shard.metadata.file_lookup_offset as usize;
```
and then read in just that many bytes from the file.

hoytak (approved):
LGTM

PR #420, opened 2025-07-22
set shard size limit as max, not target min

assafvayner:
Enforce that shards are cut at <= target_shard_max_size (64 MiB)

Previously we were enforcing that shards are cut after exceeding this limit. This enabled in theory shards <128MiB, all shards after this PR will be <= 64 MiB


hoytak (approved):
LGTM

PR #421, opened 2025-07-22
use v1 api paths

assafvayner:
Using API paths as specified in new API updates with `/v1` prefix and no shard hash.

https://docs.google.com/document/d/14CkiKiwX_y6m4oboh9rUrRCvVZtNqUTZbGzK8M5fh3o/edit?usp=sharing

### DIFF
--- a/cas_client/src/interface.rs
+++ b/cas_client/src/interface.rs
@@ -57,14 +57,7 @@ pub trait Client {
     ) -> Result<Option<Bytes>>;
 
     /// Upload a new shard.
-    async fn upload_shard(
-        &self,
-        prefix: &str,
-        hash: &MerkleHash,
-        force_sync: bool,
-        shard_data: bytes::Bytes,
-        salt: &[u8; 32],
-    ) -> Result<bool>;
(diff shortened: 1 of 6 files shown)


cas_client/src/interface.rs
```diff
@@ -57,14 +57,7 @@ pub trait Client {
...
     /// Upload a new shard.
-    async fn upload_shard(
-        &self,
-        prefix: &str,
-        hash: &MerkleHash,
-        force_sync: bool,
-        shard_data: bytes::Bytes,
-        salt: &[u8; 32],
-    ) -> Result<bool>;
+    async fn upload_shard(&self, shard_data: Bytes, force_sync: bool, salt: &[u8; 32]) -> Result<bool>;
```
jgodlew: Looking at the code, we might be able to remove the `force_sync` and `salt` params from this function as they aren't really used (I'm fine with removing salt as a separate PR since it might require a bit more refactoring for the local_client).

jgodlew (approved):
LGTM! Just a comment on some cleanup.

hoytak (approved):
LGTM

PR #422, opened 2025-07-22
spec draft

assafvayner:
fix XET-681

XET protocol specification initial draft

- documentation of core procedures required for file uploads and downloads
- format specifications for shards and xorbs


spec/chunking.md
```diff
@@ -0,0 +1,97 @@
...
+
+### For Deduplication
+
+- Identical content produces identical chunks regardless of file position
+- Small edits don't shift all subsequent chunk boundaries
+- Optimal granularity for detecting duplicate content blocks
+
+## Parallel Chunking (Advanced)
+
+[Proof for explanation of parallel chunking](deduplication/src/parallel%20chunking.pdf).
```
sirahd: [nit] the current link doesn't work 
```suggestion
[Proof for explanation of parallel chunking](../deduplication/src/parallel%20chunking.pdf).
```

spec/xorb_formation.md
```diff
@@ -0,0 +1,70 @@
+# Xorb Formation
+
+## Collecting Chunks
+
+Using the chunking algorithm a file is mapped to a series of chunks, once those chunks are found, they need to be collected into collections of chunks called each called a "Xorb" (Xet Orb, pronounced like "zorb").
```
sirahd:
```suggestion
Using the chunking algorithm a file is mapped to a series of chunks, once those chunks are found, they need to be collected into collections of chunks each called a "Xorb" (Xet Orb, pronounced like "zorb").
```

spec/xorb_formation.md
```diff
@@ -0,0 +1,70 @@
...
+#### Chunk Header Structure
+
+The chunk header is serialized as follows:
+
+- **Version** (1 byte): Protocol version, currently `0`
+- **Compressed Size** (3 bytes): Size of data after compression as a 3 byte little-endian unsigned integer.
+- **Compression Type** (1 byte): Algorithm used for compression (See mapping below)
+- **Uncompressed Size** (3 bytes): Size of raw chunk data (before compression) as a 3 byte little-endian unsigned integer.
+
+Both Compressed and Uncompressed Size can fit in a 3 byte integer, given that
```
sirahd: incomplete line?

spec/xorb_formation.md
```diff
@@ -0,0 +1,70 @@
...
+
+VERSION = 0
+buffer = bytes()
+
+for chunk in xorb.chunks:
+    uncompressed_length = len(chunk)
+    compressed, compression_scheme = pick_compression_scheme_and_compress(chunk)
+    header = Header(VERSION, len(compressed), compression_scheme, uncompressed_length)
+    buffer.write(header)
+    buffer.write(compressed)
```
sirahd: [nit] put this in a code block

spec/shard.md
```diff
@@ -0,0 +1,605 @@
...
+    })
+}
+```
+
+## Usage Patterns
+
+### Finding a File by Hash
+
+1. Truncate file hash to first 8 bytes
+2. Binary search in file lookup table
```
sirahd: If there are multiple shards stored in the system, how do I know which shard file to lookup? Our current implementation relies on DDB to provide this information, but without DDB do we need to iterate through all shard files?

spec/api.md
```diff
@@ -0,0 +1,129 @@
...
+- **Error Responses**:
+  - 400 bad request
+  - 404 File not found
+  - 416 Range Not Satisfiable: When requested byte range start exceeds the end of a file
+
+### 2. Batch Get Reconstruction
+
+**Description**: Retrieves reconstruction information for multiple files in a single request.
+
+- **Path**: `/v1/reconstructions`
```
bpronan: The convention for a RESTful API is to have both the singular and batch versions of the api using the same path. 

In other words, the above API should be `/v1/reconstructions/{id}` and this would be `/v1/reconstructions`.

spec/api.md
```diff
@@ -0,0 +1,129 @@
...
+- **Method**: `GET`
+- **Query Parameters**:
```
bpronan: What happens if you don't specify query parameters? Ex: `curl https://cas/v1/reconstructions`.

Does this return all of the reconstructions for the repo embedded in the JWT?

bpronan: If so, we would need pagination.

spec/api.md
```diff
@@ -0,0 +1,129 @@
...
+### 4. Upload XORB
+
+**Description**: Uploads a serialized CAS object (XORB) to the server with progress tracking. Review [how to compute xorb hash](spec/hashing.md#Xorb%20Hashes) to compute xorb hashes.
+
+- **Path**: `/v1/xorb/default/{hash}`
+- **Method**: `POST`
+- **Parameters**:
+  - `hash`: MerkleHash in hex format. Review [how to compute xorb hash](spec/hashing.md#Xorb%20Hashes) to compute xorb hashes.
+- **Headers**:
+  - `Content-Length`: Size of upload data
```
bpronan: What happens if this doesn't match with the content length of the Xorb? It should be mentioned in an `Error Responses` section.

spec/api.md
```diff
@@ -0,0 +1,129 @@
...
+  - 404 - Chunk not already tracked by global deduplication
+
+### 4. Upload XORB
+
+**Description**: Uploads a serialized CAS object (XORB) to the server with progress tracking. Review [how to compute xorb hash](spec/hashing.md#Xorb%20Hashes) to compute xorb hashes.
+
+- **Path**: `/v1/xorb/default/{hash}`
+- **Method**: `POST`
+- **Parameters**:
+  - `hash`: MerkleHash in hex format. Review [how to compute xorb hash](spec/hashing.md#Xorb%20Hashes) to compute xorb hashes.
```
bpronan: What can the user expect if this hash is incorrect? I would expect a 400 status code without knowing otherwise.

spec/api.md
```diff
@@ -0,0 +1,129 @@
...
+- **Body**: None
+- **Response**: Raw bytes in shard format (chunk data if exists)
+- **Error Responses**:
+  - 404 - Chunk not already tracked by global deduplication
+
+### 4. Upload XORB
+
+**Description**: Uploads a serialized CAS object (XORB) to the server with progress tracking. Review [how to compute xorb hash](spec/hashing.md#Xorb%20Hashes) to compute xorb hashes.
+
+- **Path**: `/v1/xorb/default/{hash}`
```
bpronan: What purpose does `default` serve here? If this were more RESTful, I would expect the path to be `/v1/xorbs/{hash}` where default is an optional parameter.

spec/api.md
```diff
@@ -0,0 +1,129 @@
...
+
+- **Error Responses**:
+  - 400 bad request
+  - 404 File not found if any file is not found
+
+### 3. Query Chunk Deduplication (Global Deduplication)
+
+**Description**: Checks if a chunk exists in the CAS for deduplication purposes.
+
+- **Path**: `/v1/chunk/default-merkledb/{hash}`
```
bpronan: What purpose is the default here? I would expect this to be `GET /v1/chunks/{hash}` unless `default-merkledb` is variable.

spec/api.md
```diff
@@ -0,0 +1,129 @@
...
+}
+```
+
+  was_inserted is false if the xorb already exists, this is not an error
+
+### 5. Check XORB Existence
+
+**Description**: Checks if an XORB exists in the Content Address Store (CAS). Use of this API is optional but can save the process of uploading all of the xorb data.
+
+- **Path**: `/v1/xorb/default/{hash}`
```
bpronan: Same question about `default` as above.

spec/api.md
```diff
@@ -0,0 +1,129 @@
...
+- **Body**: None
+- **Response**: Status code only
+  - `200 OK`: XORB exists
+  - `404 Not Found`: XORB does not exist
+
+### 6. Upload Shard
+
+**Description**: Uploads a shard to the CAS with optional forced synchronization.
+
+- **Path**: `/v1/shard`
```
bpronan: REST convention would use `POST /v1/shards`.

spec/api.md
```diff
@@ -0,0 +1,129 @@
...
+- **Method**: `POST`
+- **Headers**:
```
bpronan: Headers are annoying for some libraries. Could we also accept this as a query parameter? Just a thought...

spec/api.md
```diff
@@ -0,0 +1,129 @@
...
+    "fetch_info": {...}
+  }
+  ```
+
+- **Error Responses**:
+  - 400 bad request
+  - 404 File not found
+  - 416 Range Not Satisfiable: When requested byte range start exceeds the end of a file
+
+### 2. Batch Get Reconstruction
```
bpronan: Supporting batch get reconstruction will open up a bunch of feature expectations, so we want to be sure that this unlocks a specific use case before we support it. 

Most API users expect stuff like pagination and query parameter matching to be built-in on day 1 for a batch API.

spec/api.md
assafvayner: I can drop the mention for the header. Content-Length is a required header for all POST requests and is usually set implicitly by most libraries in any language (that I have used).

spec/api.md
assafvayner: is it worth to change/add that path now? (in cas server)

spec/api.md
assafvayner: I'm choosing to omit this API for now since it is not used.

today that API would return nothing if no files are specified by query parameters.

I agree that if we wanted to support "give me all files" then we would need pagination, but I am also not sure if we can filter based on a ref.

spec/api.md
assafvayner: dropped batch api for now

spec/auth.md
```diff
@@ -0,0 +1,122 @@
...
+   Use Bearer authentication with the `access_token` to authenticate with the Xet service at `endpoint` until `expiration` time.
+
+4. **Token refresh (when needed):**
+   Use the same API to generate a new token
+
+## Token Scope
+
+Xet tokens can have either a `read` or a `write` scope. The type of token issues is determined on the `token_type` URI path component when requesting the token from the HuggingFace Hub (see above).
+
+Revise API specification for what scope level is required to invoke each API (briefly, only `POST /shard` and `POST /xorb/*` API's require `write` scope).
```
bpronan: It would be helpful to mention that `write` scope allows for `read` on a resource.

spec/file_reconstruction.md
```diff
@@ -0,0 +1,95 @@
...
+      }
+    ]
+  }
+}
+```
+
+This example shows reconstruction of a file that requires:
+
+- Chunks `[0, 3)` from the first xorb (~264KB of unpacked data)
+- Chunks `[0, 1)` from the second xorb (~144KB of unpacked data)
```
jgodlew: Shouldn't these chunk indices be:
```suggestion
- Chunks `[0, 4)` from the first xorb (~264KB of unpacked data)
- Chunks `[0, 2)` from the second xorb (~144KB of unpacked data)
```

spec/download_protocol.md
```diff
@@ -0,0 +1,248 @@
...
+To download a file, first call the reconstruction API to get the metadata needed for reconstruction:
+
+**Endpoint**: `GET /v1/reconstruction/{file_id}`
+
+**Parameters**:
+
+- `file_id`: The file's MerkleHash in hex format (64-character hex string)
+- `Range` (header, optional): Byte range in format `bytes={start}-{end}` (inclusive end)
+
+**Authentication**: Bearer token required via `Authorization` header
```
jgodlew: nit: maybe link to `auth` doc?

spec/auth.md
```diff
@@ -0,0 +1,122 @@
...
+
+The scope of the xet tokens is limited to the repository and ref for which they were issued. To upload or download from different repositories or refs (different branches) clients will need to be issued different tokens.
+
+## Token Scope Relative to HuggingFace Hub Authentication Token
+
+When requesting a Xet token from the HuggingFace hub, you will only receive a xet token matching the requested parameters if you actually have access to them, based on the access afforded to your hub authentication token.
```
jsulz:
```suggestion
When requesting a Xet token from the HuggingFace Hub, you will only receive a xet token matching the requested parameters if you actually have access to them, based on the access afforded to your hub authentication token.
```

spec/shard.md
```diff
@@ -0,0 +1,605 @@
...
+├─────────────────────┤
+│        ...          │
+├─────────────────────┤
+│   Bookend Entry     │ ← All 0xFF hash + zeros
+└─────────────────────┘
+```
+
+#### File Info Bookend
+
+The end of the file infos section is marked by a bookend entry.
```
jsulz:
```suggestion
The end of the file info sections is marked by a bookend entry.
```
Since "section" is the head noun in "file info section" I believe the plural goes here.

spec/shard.md
```diff
@@ -0,0 +1,605 @@
...
+
+1. Seek to `footer.cas_info_offset`
+2. Read `CASChunkSequenceHeader`
+3. Check if `cas_hash` is all 0xFF (bookend marker) - if so, stop
+4. Read `num_entries` × `CASChunkSequenceEntry` structures
+5. Repeat from step 2 until bookend found
+
+#### CAS Info Bookend
+
+The end of the cas infos section is marked by a bookend entry.
```
jsulz:
```suggestion
The end of the cas info sections is marked by a bookend entry.
```
Same comment as file info above.

spec/shard.md
```diff
@@ -0,0 +1,605 @@
...
+
+The bookend entry is 48 bytes long where the first 32 bytes are all 0xFF, followed by 16 bytes of all 0x00.
+
+Suppose you were attempting to deserialize a CASChunkSequenceHeader and it's hash was all 1 bits then this entry is a bookend entry and the next bytes start the next section.
+
+Since the file info section immediately follows the file info section, a client does not need to deserialize the footer to know where it starts deserialize this section, it begins right after the file info section bookend and ends when the cas info bookend is reached.
```
jsulz:
```suggestion
Since the cas info section immediately follows the file info section, a client does not need to deserialize the footer to know where it starts deserializing this section, it begins right after the file info section bookend and ends when the cas info bookend is reached.
```

Same comment as above about breaking up the sentence as well.

spec/shard.md
```diff
@@ -0,0 +1,605 @@
...
+
+#### File Info Bookend
+
+The end of the file infos section is marked by a bookend entry.
+
+The bookend entry is 48 bytes long where the first 32 bytes are all 0xFF, followed by 16 bytes of all 0x00.
+
+Suppose you were attempting to deserialize a FileDataSequenceHeader and it's file hash was all 1 bits then this entry is a bookend entry and the next bytes start the next section.
+
+Since the file info section immediately follows the header, a client does not need to deserialize the footer to know where it starts deserialize this section, it begins right after the header and ends when the bookend is reached.
```
jsulz:
```suggestion
Since the file info section immediately follows the header, a client does not need to deserialize the footer to know where it starts deserializing this section, it begins right after the header and ends when the bookend is reached.
```
Verb agreement. Could also be "begin deserialization" 🤷 

Could also break this sentence up a bit. 

```suggestion
Since the file info section immediately follows the header, a client does not need to deserialize the footer to know where it starts deserializing this section. It begins right after the header and ends when the bookend is reached.
```

spec/chunking.md
```diff
@@ -0,0 +1,97 @@
...
+
+### The Boundary Condition
+
+Define MASK as the 64 bit number 0xffff000000000000 (bitwise 16 1's followed by 48 0's; 18446462598732840960)
+
+The byte position when the rolling hash `& MASK == 0` (bitwise AND is equal to 0) is a chunk boundary.
+
+### Boundary Probability
+
+The mask is designed so that approximately 1 in every `target_chunk_size` positions will trigger a boundary, making the average chunk size equal to the target.
```
seanses:
```suggestion
The mask is designed so that approximately 1 in every `target_chunk_size` bytes will trigger a boundary, making the average chunk size equal to the target.
```

seanses: Use "bytes" instead of "positions" to be precise

spec/chunking.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Algorithm Flow
+
+### 1. Initialize
+
+- Set up gearhash with 64-byte window
```
seanses: Is the target of this spec to make sure 3rd party clients are compatible with our server and our dedup info? If so we should also fix the [hash table](https://github.com/srijs/rust-gearhash/blob/adad44e7141cfd29d898cf6e0858f50b995db286/src/table.rs#L5) so any client produce the same hash for the same content. Essentially, their hasher implementation should be logically identical to https://github.com/srijs/rust-gearhash

spec/chunking.md
```diff
@@ -0,0 +1,97 @@
...
+
+- Linear time complexity: O(n) for data size n
+- Constant space: only needs 64-byte window and counters
+- Fast rolling hash updates
+
+## Algorithm Benefits
+
+### For Deduplication
+
+- Identical content produces identical chunks regardless of file position
```
seanses: Re "regardless of file position": I can't say this is always true. A hash windows of 64 bytes already implies a rolling hash is affected by the order of bytes.

spec/chunking.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Determinism
+
+- Same content always produces same chunk boundaries
+- No randomization or external dependencies
+- Boundaries depend only on the 64-byte content window
+
+### Content Sensitivity
+
+- Small changes affect only nearby boundaries
```
seanses:
```suggestion
- Small changes within a `target_chunk_size` window affect only a few next boundaries 
```

spec/file_reconstruction.md
```diff
@@ -0,0 +1,95 @@
+# File Reconstruction
+
+After a file is mapped into a series of chunks, and those chunks are tracked by xorbs, we can define the file reconstruction or the recipe to re-materialize the file.
+
+For every chunk of the file we need to know which xorb this chunk is in and its index (starting at 0) in that xorb and list out all the chunks in this way to re-create the file.
+
+Now since xorbs contain multiple chunks, we can condense the listing of each range of chunks that are contiguous in the same xorb together and this map the listing of chunks that form the file to a series of sub-ranges of chunks within xorbs.
```
jgodlew: We might want to indicate that this contiguous range of chunks within a xorb is referred to as a **term**

spec/chunking.md
```diff
@@ -0,0 +1,97 @@
...
+- Most chunk boundaries remain stable when content changes
```
seanses:
```suggestion
- Most chunk boundaries remain stable when content changes within a `target_chunk_size` window
```

spec/chunking.md
```diff
@@ -0,0 +1,97 @@
...
+- Output current chunk (from last boundary to current position)
+- Reset for next chunk (set rolling hash to 0)
+- Continue processing
+
+## Size Constraints
+
+### Minimum Chunk Size
+
+- Prevents tiny chunks that hurt compression efficiency
+- `target_size / 8` (8KiB)
```
seanses: Let's unify the notation, choose either `target_chunk_size` or `target_size` and use it consistently.

spec/deduplication.md
```diff
@@ -0,0 +1,183 @@
...
+- Multiple versions of datasets with incremental changes
+- Model checkpoints that share common layers or parameters
+- Documentation and configuration files with similar content
+- Large files where only portions have changed between versions
+
+## Core Concepts
+
+### Chunks
+
+A **chunk** is a variable-sized content block derived from files using Content-Defined Chunking (CDC) with a rolling hash function. Chunks are the fundamental unit of deduplication in Xet.
```
jgodlew, line 20: Add a link to the `chunking` doc?

spec/chunking.md
```diff
@@ -0,0 +1,97 @@
...
+- `target_size × 2` (128KiB)
+- Forces boundary creation regardless of hash value
+
+## Key Properties
+
+### Determinism
+
+- Same content always produces same chunk boundaries
+- No randomization or external dependencies
+- Boundaries depend only on the 64-byte content window
```
seanses: Also depends on min and max chunk size

spec/chunking.md
seanses: Actually, this reads just like a paraphrase of the above item

spec/deduplication.md
```diff
@@ -0,0 +1,183 @@
...
+Not all chunks are eligible for global deduplication queries to manage system load:
+
+1. **First chunk**: The first chunk of every file is always eligible.
+2. **Hash pattern matching**: Chunks are eligible if: the last 8 bytes of the hash interpreted as a little-endian 64 bit integer % 1024 == 0.
+3. **Spacing constraints**: There must be a minimum of 4MB between global deduplication queries within a file.
+
+#### Query Process
+
+1. **Background Query**: Global deduplication queries run asynchronously to avoid blocking upload
+2. **HMAC Protection**: Chunk hashes are protected using repository-specific HMAC keys
```
jgodlew: Q: Is there a particular reason why these keys are "repository-specific"?

spec/deduplication.md
```diff
@@ -0,0 +1,183 @@
...
+3. **Shard Response**: When a match is found, the API returns a shard containing:
+   - **CAS Info Section**: Contains metadata about many xorbs that store chunks
+   - **HMAC Key**: Included in the shard metadata header for decryption purposes
```
jgodlew: Is this supposed to be for *encryption* purposes? Step 5 indicates that this key is used for encrypting the local chunk hashes to compare.

spec/chunking.md
seanses: So we should mention here that `target_chunk_size` is 64 KiB, and that because the probability for all 16 bits to be 0 is 1/2^16 given a [hash table](https://github.com/srijs/rust-gearhash/blob/adad44e7141cfd29d898cf6e0858f50b995db286/src/table.rs#L5) that produces uniform hashes.

spec/api.md
```diff
@@ -0,0 +1,125 @@
...
+## Endpoints
+
+### 1. Get File Reconstruction
+
+**Description**: Retrieves reconstruction information for a specific file, optionally with byte range support.
+
+- **Path**: `/v1/reconstructions/{file_id}`
+- **Method**: `GET`
+- **Parameters**:
+  - `file_id`: MerkleHash in hex format (64 hexadecimal character string)
```
rajatarya: Add example here, is this character string case sensitive? Is BADBEEF ok or is it badbeef?

spec/api.md
```diff
@@ -0,0 +1,125 @@
...
+  ```json
+  {
+    "offset_into_first_range": 0,
+    "terms": [...],
+    "fetch_info": {...}
+  }
+  ```
+
+- **Error Responses**:
+  - 400 bad request (such as the file_id not being )
```
rajatarya: need to finish thought: "such as the file_id not being ...)

spec/api.md
```diff
@@ -0,0 +1,125 @@
...
+
+  ```json
+  {
+    "offset_into_first_range": 0,
+    "terms": [...],
+    "fetch_info": {...}
+  }
+  ```
+
+- **Error Responses**:
```
rajatarya: Error responses should indicate retry-ability. What should the caller do when these are received?

spec/api.md
```diff
@@ -0,0 +1,125 @@
...
+**Description**: Checks if a chunk exists in the CAS for deduplication purposes.
+
+- **Path**: `/v1/chunks/{prefix}/{hash}`
+- **Method**: `GET`
+- **Parameters**:
+  - `prefix`: Currently only acceptable prefix for the Global Deduplication API is `default-merkledb`
+  - `hash`: Chunk hash in hex format (64 hexadecimal character string). Review [how to compute chunk hash](../hashing.md#Chunk%20Hashes) to compute chunk hashes
+- **Minimum Token Scope**: `read`
+- **Body**: None
+- **Response**: Raw bytes in shard format (chunk data if exists)
```
rajatarya: As a caller, what should I do given this information? How should I interpret this response?

spec/api.md
```diff
@@ -0,0 +1,125 @@
...
+
+```json
+{
+  "was_inserted": true
+}
+```
+
+  was_inserted is false if the xorb already exists, this is not an error
+
+- **Error Responses**:
```
rajatarya: I think there should be a central spec area on retries and error codes, and each of these API Error Responses section should reference it.

spec/api.md
```diff
@@ -0,0 +1,125 @@
...
+- **Body**: None
+- **Response**: Status code only
+  - `200 OK`: Xorb exists
+  - `404 Not Found`: Xorb does not exist
+- **Error Responses**:
+  - `400 Bad Request`: Xorb hash path component is incorrectly formatted.
+
+### 5. Upload Shard
+
+**Description**: Uploads a shard to the CAS with optional forced synchronization.
```
rajatarya: I guess this should also link off to something explaining WHY Xorbs and Shards exist, what they represent, etc.

spec/api.md
```diff
@@ -0,0 +1,125 @@
...
+  ```json
+  {
+    "result": 0 | 1
+  }
+  ```
+
+  Where 0 indicates the shard already exists and 1 indicates "SyncPerformed" meaning that the shard was registered (UploadShardResponseType).
+
+- **Error Responses**:
+  - `400 Bad Request`: shard is incorrectly serialized, shard contents failed verification
```
rajatarya: What about 416 Size Too Large? We should have an error code for a shard upload that is too big.

spec/api.md
```diff
@@ -0,0 +1,125 @@
...
+  - `403 Forbidden`: authentication token missing `write` scope.
+
+## Common Error Cases/Codes
```
rajatarya: Add 500 to this list.

Also mention no other HTTP codes are returned.

spec/auth.md
```diff
@@ -0,0 +1,122 @@
+# Authentication and Authorization
+
+To invoke any API's mentioned in this specification a client must first acquire a token (and the url) to authenticate against the server which serves these API's.
+
+The Xet protocol server uses bearer authentication via a token generated by the huggingface hub (<https://huggingface.co>).
```
rajatarya: hugging face hub -> Hugging Face Hub

spec/auth.md
```diff
@@ -0,0 +1,122 @@
...
+```typescript
+{
+   "accessToken": string,
+   "exp": number,
+   "casUrl": string,
+}
+```
+
+- accessToken is the token to be used when invoking API's on the Xet CAS service (any Xet API denoted in this specification)
+- exp is the unix timestamp of when this token expires
```
rajatarya, line 55: Is this seconds since epoch?

spec/auth.md
```diff
@@ -0,0 +1,122 @@
...
+
+```typescript
+{
+   "accessToken": string,
+   "exp": number,
+   "casUrl": string,
+}
+```
+
+- accessToken is the token to be used when invoking API's on the Xet CAS service (any Xet API denoted in this specification)
```
rajatarya, line 54: expected size/length of this string? max/min length?

spec/auth.md
```diff
@@ -0,0 +1,122 @@
...
+- exp is the unix timestamp of when this token expires
+- casUrl is the API service endpoint URL
```
rajatarya, line 56: Any boundary conditions for this value?

spec/auth.md
```diff
@@ -0,0 +1,122 @@
...
+   Host: huggingface.co
+   Authorization: Bearer xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+   ```
+
+2. **Parse the response**
+
+   ```python
+   endpoint = response_json["casUrl"]
+   access_token = response_json["accessToken"]
+   expiration = response_json["exp"]
```
rajatarya, line 95: I wonder if the response should also include the current time (as reported by the server) so the client can understand how many seconds are remaining in the token.

spec/auth.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Token Scope
+
+Xet tokens can have either a `read` or a `write` scope. The type of token issues is determined on the `token_type` URI path component when requesting the token from the HuggingFace Hub (see above).
+
+Revise API specification for what scope level is required to invoke each API (briefly, only `POST /shard` and `POST /xorb/*` API's require `write` scope). Note that all `read` scope API's can be invoked when using a `write` scope token.
+
+The scope of the xet tokens is limited to the repository and ref for which they were issued. To upload or download from different repositories or refs (different branches) clients will need to be issued different tokens.
+
+## Token Scope Relative to HuggingFace Hub Authentication Token
```
rajatarya: HuggingFace -> Hugging Face

spec/auth.md
rajatarya: also, HuggingFace -> Hugging Face

spec/auth.md
```diff
@@ -0,0 +1,122 @@
...
+
+When requesting a Xet token from the HuggingFace hub, you will only receive a xet token matching the requested parameters if you actually have access to them, based on the access afforded to your hub authentication token.
+
+If you require a `write` scope xet token, then you must request it using a HuggingFace Hub token that has write access to the particular repository and ref that you want to access.
```
rajatarya:
```
s/HuggingFace/Hugging Face/g
```

spec/chunking.md
```diff
@@ -0,0 +1,97 @@
+# Content-Defined Chunking Algorithm
+
+## Overview
+
+The first step in uploading a file is to convert the file into a series of chunks. In order to "chunk" a file we use a rolling hash using the gearhash algorithm over the contents of the file and then set chunk boundaries through a few conditions.
```
rajatarya: Link to gearhash algorithm.

spec/chunking.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Boundary Detection
+
+### The Boundary Condition
+
+Define MASK as the 64 bit number 0xffff000000000000 (bitwise 16 1's followed by 48 0's; 18446462598732840960)
```
rajatarya: Endian-ness needs to be mentioned here, right?

spec/chunking.md
rajatarya: We should link to gearhash implementations and document how to test/verify your implementation (use rust-gearhash repo with these steps to verify...)

spec/chunking.md
rajatarya: Maybe the phrasing that is more accurate is something like: "Identical content (in identical file positions) produces identical chunks."

spec/deduplication.md
```diff
@@ -0,0 +1,183 @@
...
+- **Identification**: Each chunk is uniquely identified by its cryptographic hash (MerkleHash)
+- **Content addressing**: Chunks are stored and retrieved using their hash as the key
+
+### Xorbs (Extended Object Blocks)
+
+**Xorbs** are containers that aggregate multiple chunks for efficient storage and transfer:
+
+- **Maximum size**: 64MB
+- **Maximum chunks**: 8,192 chunks per xorb
+- **Purpose**: Batch multiple chunks together to reduce metadata overhead and improve compression efficiency
```
rajatarya: Might want to mention reducing network overhead as part of purpose.

spec/deduplication.md
```diff
@@ -0,0 +1,183 @@
...
+
+#### Level 3: Global Deduplication API
+
+**Scope**: Entire Xet ecosystem
+**Mechanism**: Global deduplication service with HMAC protection
+**Purpose**: Discover deduplication opportunities across all users and repositories
+
+### 3. Global Deduplication Process
+
+The global deduplication system provides deduplication capabilities across all data that is managed by the xet system:
```
rajatarya: xet -> Xet

spec/deduplication.md
rajatarya: I believe this is to ensure that querying doesn't leak information you shouldn't have access to.

spec/deduplication.md
rajatarya: I believe the idea is that you encrypt your hashes with this key, and then compare it against the encrypted hashes you received in the response - to see if any chunk deduplicates.

spec/deduplication.md
```diff
@@ -0,0 +1,183 @@
...
+
+Global deduplication uses HMAC (Hash-based Message Authentication Code) to protect chunk hashes while enabling deduplication.
+
+**Security Properties**:
+
+Raw chunk hashes are never transmitted; a client has to encrypt their raw chunk hash and find a match to know a raw chunk hash exists in the system. They may know this chunk hash because they own this data, the match has made them privy to know which xorb has this chunk hash and the position in the xorb, but has not revealed any other raw chunk hashes in that xorb or other xorbs.
+
+## Deduplication Effectiveness and Metrics
+
+### Tracking Metrics
```
rajatarya: Not sure this section is needed for client creation. Might want to consider pushing to appendix or supplemental section instead.

spec/deduplication.md
```diff
@@ -0,0 +1,183 @@
...
+
+The system tracks comprehensive deduplication statistics including:
+
+- Total bytes and chunks processed
+- Bytes and chunks saved through deduplication
+- New bytes and chunks that couldn't be deduplicated
+- Savings achieved specifically through global deduplication
+- Instances where deduplication was skipped to prevent fragmentation
+
+### Fragmentation Prevention
```
rajatarya: This is a good paragraph - but isn't actionable. I think the next paragraph here should explain how to think about Fragmentation in building a client. Suggest an example of the skipping practice, or the once every 4MB dedupe query.

spec/deduplication.md
```diff
@@ -0,0 +1,183 @@
...
+When new chunks need to be stored, they are aggregated into xorbs based on size and count limits. If adding a new chunk would exceed the maximum xorb size or chunk count, the current xorb is finalized and uploaded. [See section about xorb formation](../xorb_formation.md)
+
+### File Reconstruction Information
+
+When chunks are deduplicated, the system creates file reconstruction information that includes:
+
+- Hash of the xorb containing the chunks
+- Flags for the CAS block
+- Total bytes in the segment
+- Start and end indices within the xorb
```
rajatarya: inclusive/exclusive indices?

spec/download_protocol.md
```diff
@@ -0,0 +1,242 @@
...
+To download a file, first call the reconstruction API to get the metadata needed for reconstruction:
+
+**Endpoint**: `GET /v1/reconstructions/{file_id}`
+
+**Parameters**:
+
+- `file_id`: The file's MerkleHash in hex format (64-character hex string)
+- `Range` (header, optional): Byte range in format `bytes={start}-{end}` (inclusive end)
+
+**Authentication**: Bearer token required via `Authorization` header
```
rajatarya: Authorization Bearer token requires READ token

spec/download_protocol.md
```diff
@@ -0,0 +1,242 @@
...
+- Type: `Map<Xorb Hash (64 character hex string), Array<CASReconstructionFetchInfo>>`
+- Maps xorb hashes to download information
+- Each fetch info entry contains:
+  - `url`: HTTP URL for downloading the xorb data, presigned url containing authorization information
+  - `url_range`: Byte range `{ start: number, end: number }` for the Range header; end-inclusive `[start, end]`
+    - The range header must be set as `Range: bytes=<start>-<end>` when downloading this chunk range
+  - `range`: Chunk range `{ start: number, end: number }` that this URL provides; end-exclusive `[start, end)`
+    - This range indicates which range of chunk indices within this xorb that this fetch info term is describing
+
+## Stage 3: Downloading and Reconstructing the File
```
rajatarya, line 94: This section could really benefit from an example diagram with chunks defined, byte ranges, and the example ReconstructionInfo JSON.

spec/download_protocol.md
```diff
@@ -0,0 +1,242 @@
...
+
+1. **Preserve order**: Process terms in the exact order they appear in the `terms` array
+2. **Handle range offsets**: For the first term, skip `offset_into_first_range` bytes
+3. **Concatenate**: Append each term's data to build the final file content
+4. **Validate**: Ensure the final file size matches expectations
+
+### Error Handling
+
+- **403 Forbidden**: Presigned URLs may expire; re-fetch reconstruction info and retry
+- **416 Range Not Satisfiable**: Requested range exceeds file size
```
rajatarya: What should the client do if they get this error?

rajatarya (commented):
This is a great start to the protocol. I did a solid pass through the first 10 files or so, the last 9 I went through quickly. Probably need to spend more time with them on another pass.

I think diagrams and examples would really help with the protocol. Sequence diagram of the download process, upload process.

I also think an explanation of why shards, what they encompass, and how they are used would be beneficial.

spec/deduplication.md
```diff
@@ -0,0 +1,183 @@
...
+## Core Concepts
+
+### Chunks
+
+A **chunk** is a variable-sized content block derived from files using Content-Defined Chunking (CDC) with a rolling hash function. Chunks are the fundamental unit of deduplication in Xet.
+
+- **Target size**: 64KB (configurable)
+- **Size range**: 8KB to 128KB (minimum and maximum constraints)
+- **Identification**: Each chunk is uniquely identified by its cryptographic hash (MerkleHash)
+- **Content addressing**: Chunks are stored and retrieved using their hash as the key
```
seanses: Actually, we don't have an API to retrieve chunks by hash

spec/deduplication.md
```diff
@@ -0,0 +1,183 @@
...
+
+```mermaid
+graph TD
+    A[File Input] --> B[Content-Defined Chunking]
+    B --> C[Hash Computation]
+    C --> D[Chunk Creation]
+    D --> E[Deduplication Query]
+```
+
+1. **File Reading**: Files are read in blocks (default 8MB) to manage memory usage
```
seanses: Probably no need to define how to read a file in a spec

spec/deduplication.md
```diff
@@ -0,0 +1,183 @@
...
+
+The global deduplication system provides deduplication capabilities across all data that is managed by the xet system:
+
+#### Eligibility Criteria
+
+Not all chunks are eligible for global deduplication queries to manage system load:
+
+1. **First chunk**: The first chunk of every file is always eligible.
+2. **Hash pattern matching**: Chunks are eligible if: the last 8 bytes of the hash interpreted as a little-endian 64 bit integer % 1024 == 0.
+3. **Spacing constraints**: There must be a minimum of 4MB between global deduplication queries within a file.
```
seanses: This is not a constraint. The server has no way to enforce this. It's just a client-side performance optimization. So I'd not put this in "eligibility criteria" but rather under a section like "recommendations for implementation"

spec/deduplication.md
seanses: No they are not repo-specific. It's generated from system time.

spec/deduplication.md
```diff
@@ -0,0 +1,183 @@
...
+   - Track the original (non-encrypted) chunk hash while noting which xorb contains that chunk
+6. **Metadata Caching**: Client downloads and caches shard metadata for future deduplication
+
+#### HMAC Security Mechanism
+
+Global deduplication uses HMAC (Hash-based Message Authentication Code) to protect chunk hashes while enabling deduplication.
+
+**Security Properties**:
+
+Raw chunk hashes are never transmitted; a client has to encrypt their raw chunk hash and find a match to know a raw chunk hash exists in the system. They may know this chunk hash because they own this data, the match has made them privy to know which xorb has this chunk hash and the position in the xorb, but has not revealed any other raw chunk hashes in that xorb or other xorbs.
```
seanses:
```suggestion
Raw chunk hashes are never transmitted from servers to clients; a client has to encrypt their raw chunk hash and find a match to know a raw chunk hash exists in the system. They may know this chunk hash because they own this data, the match has made them privy to know which xorb has this chunk hash and the position in the xorb, but has not revealed any other raw chunk hashes in that xorb or other xorbs.
```

spec/shard.md
```diff
@@ -0,0 +1,605 @@
...
+**Location**: `footer.cas_info_offset` to `footer.file_lookup_offset`
+
+This section contains CAS (Content Addressable Storage) block information:
+
+#### CASChunkSequenceHeader
+
+```rust
+struct CASChunkSequenceHeader {
+    cas_hash: Hash,           // 32-byte CAS block hash
+    cas_flags: u32,           // CAS flags
```
jgodlew: We should probably list out what the actual flags are and where they're located in this bitmap.

spec/shard.md
```diff
@@ -0,0 +1,605 @@
...
+- `MDB_FILE_FLAG_WITH_METADATA_EXT` (0x40000000 or 1 << 30): Has metadata extension
+
+Given the `file_data_sequence_header.file_flags & MASK` (bitwise AND) operations, if the result != 0 then the effect is true.
+
+#### FileDataSequenceEntry
+
+```rust
+struct FileDataSequenceEntry {
+    cas_hash: [u64; 4],           // 32-byte CAS block hash
+    cas_flags: u32,               // CAS flags
```
jgodlew: Again, we should probably list out what valid flags are and where they're located in this bitmap (like in the `FileDataSequenceHeader` above.

spec/deduplication.md
jgodlew: yeah, just pointing out that the line says `decryption` instead of `encryption`.

spec/spec.md
```diff
@@ -0,0 +1 @@
+# Xet Protocol Specification
```
jgodlew, line 1: Is this page expected to be the top-level entry point into the docs? We might want to flesh this out with links to the various sections.

spec/shard.md
```diff
@@ -0,0 +1,498 @@
...
+
+```txt
+┌────────────────────────────────────────────────────────────────┬────────────────────────────────┐
+│                      sha256 (32 bytes)                         │       _unused (16 bytes)       │
+│                     SHA256 Hash                                │         Reserved Space         │
+└────────────────────────────────────────────────────────────────┴────────────────────────────────┘
+0                                                               32                               48
+```
+
+### File Info Section Layout
```
jgodlew: nit: I would prefer we put the layout at the start of the `2. File Info Section` as a way to get an overview of the section layout before diving into each of the specific types.

spec/shard.md
```diff
@@ -0,0 +1,498 @@
...
+│                     chunk_hash (32 bytes)                      │chunk_   │unpacked │    _unused      │
+│                        Chunk Hash                              │byte_    │segment_ │   (8 bytes)     │
+│                                                                │range_   │bytes    │                 │
+│                                                                │start    │(4 bytes)│                 │
+│                                                                │(4 bytes)│         │                 │
+└────────────────────────────────────────────────────────────────┴─────────┴─────────┴─────────────────┘
+0                                                               32        36        40               48
+```
+
+### CAS Info Section Layout
```
jgodlew: nit: I would prefer we put the layout at the start of the `3. CAS Info Section` as a way to get an overview of the section layout before diving into each of the specific types.

spec/shard.md
assafvayner: yes but also no.

yes -> the client could attempt to look for a file in its shard cache.
no -> when the client wants to download a file reconstruction it has to go through the reconstruction api. There's no plans for the client to have the reconstruction cached except for the latest steps in design for git-xet.

This is my understanding, but for more clarification pinging @seanses to add more if needed.

spec/chunking.md
assafvayner: > , and that because the probability for all 16 bits to be 0 is 1/2^16 given a [hash table](https://github.com/srijs/rust-gearhash/blob/adad44e7141cfd29d898cf6e0858f50b995db286/src/table.rs#L5) that produces uniform hashes.

I'm not 100% sure I understand why that fraction causes the 64 KiB target, can you elaborate?

spec/shard.md
assafvayner: 0 is the only valid value for now

spec/auth.md
assafvayner: that information is not usually included in specifications for tokens (that I have seen), the reason being that the token provider maintains the freedom to change this at any time.

spec/auth.md
assafvayner: yes this implied by "unix timestamp"

spec/auth.md
assafvayner: IMO, besides it being a URL, no. That enforces a very large length constraint and doesn't specify if the endpoint has some path prefix (in our case it does not, but the user should not make an assumption that it doesn't).

spec/auth.md
assafvayner: IMO no, the transmission time/clock skew makes all of this not really matter.
- client will need to refresh certainly after receiving a 401 error code
- the duration of token validity is ~1hr so that's plenty of time to forget issues of clock skew and build in a factor on the order of minutes, I will list that we use 30s (that's what we use for hf_xet) as the buffer

spec/auth.md
assafvayner: I'm adding a statement on upper limit, but setting it to 64000 (firefox max). I don't think this helps users too much but it's explicit.

spec/auth.md
assafvayner: Should/can we specify a minimum validity time i.e. the server promises to return a token that is at valid for at least X minutes. I don't want to lock us to then have to provide tokens for that duration.

@bpronan + @jgodlew do you think there is some reasonable minimum that we can "publish"

spec/chunking.md
assafvayner: I want to also get an opinion from @seanses on that.

I didn't know so I looked into how the gearhash is done, IIUC, the hash at any time is a u64, the mask is also a u64, both numbers will have the same "value" in a big/little endian system, so their bitwise `(a & b) == 0` operation will have the same result no matter with either endianness always.

i.e. `1 & 3 == 0 -> false` is the same with either endianness, regardless of if the 1 bits are in the most or least significant bytes.

spec/chunking.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Pseudocode
+
+```text
+Inputs:
+  data: byte array
+  MIN_CHUNK_SIZE = 8 * 1024
+  MAX_CHUNK_SIZE = 128 * 1024
+  MASK = 0xffff000000000000
+  TABLE[256]: Array of 256 pseudo-random 64-bit constants
```
rajatarya: Don't we have to share these in the protocol document in order for everyone to have the same numbers in the table?

spec/chunking.md
```diff
@@ -0,0 +1,135 @@
...
+
+  if size < MIN_CHUNK_SIZE:
+    continue
+
+  if size >= MAX_CHUNK_SIZE or (h & MASK) == 0:
+    emit chunk [start_offset, i + 1)
+    start_offset = i + 1
+    h = 0
+
+if start_offset < len(data):
```
rajatarya, line 78: Doesn't this need to be in the for loop?

spec/chunking.md
```diff
@@ -0,0 +1,135 @@
...
+  MIN_CHUNK_SIZE = 8 * 1024
+  MAX_CHUNK_SIZE = 128 * 1024
+  MASK = 0xffff000000000000
+  TABLE[256]: Array of 256 pseudo-random 64-bit constants
+
+State:
+  h = 0
+  start_offset = 0 // start of the "current chunk"
+
+for i in 0..len(data):
```
rajatarya: Do you want to add the check that len(data) < MIN_CHUNK_SIZE?

spec/api.md
```diff
@@ -0,0 +1,119 @@
...
+## Authentication
+
+To authenticate, authorize, and obtain the API base URL, follow the instructions in [Authentication](./auth.md).
+
+## Endpoints
+
+### 1. Get File Reconstruction
+
+- **Description**: Retrieves reconstruction information for a specific file, optionally with byte range support.
+- **Path**: `/v1/reconstructions/{file_id}`
```
severo, line 37: I guess it's


```suggestion
- **Path**: `/v1/reconstruction/{file_id}`
```

assafvayner: We're planning on changing the paths to match the spec, server&clients haven't got updated yet

spec/chunking.md
assafvayner: this condition check cuts the last chunk if there's anymore data left. should not be in loop

spec/chunking.md
assafvayner: yes, I linked it above on line 14. I can't add links inside ``` sections. Instead I'll remove the constants and reference the sections above.

spec/chunking.md
```diff
@@ -0,0 +1,135 @@
...
+
+```text
+h = (h << 1) + TABLE[b]
+```
+
+### Boundary test and size constraints
+
+At each position after updating `h`, let `size = current_offset - start_offset + 1`.
+
+- If `size < MIN_CHUNK_SIZE`: do not test `MASK`; continue
```
seanses, line 33: Actually we don't compute the hash / update `h` when `size < MIN_CHUNK_SIZE`. This is called cut-point skipping and helps us get higher speed.

spec/chunking.md
```diff
@@ -0,0 +1,135 @@
...
+    start_offset = i + 1
+    h = 0
+
+if start_offset < len(data):
+  emit chunk [start_offset, len(data))
+```
+
+### Boundary probability and mask selection
+
+- With 16 one-bits in `MASK`, the probability a random 64-bit `h` matches is `1 / 2^16`, yielding an average of 64 KiB between matches.
```
seanses: `h` matches what? Better say "Given that `MASK` has 16 one-bits, for a random 64-bit hash `h`, the chance that all those 16 bits are zero is `1 / 2^16`. On average, that means you’ll see a match about once every 64 KiB."

spec/chunking.md
```diff
@@ -0,0 +1,135 @@
...
+- Placing the one-bits in MSBs or LSBs (or random) yields the same probability; we use MSB placement (`0xffff000000000000`).
```
seanses: This is not an arbitrary choice, but due to how the rolling hash is computed. At each step (per byte), `h` is left shifted by 1 and added with another value, so the right most bit only carries information from the last byte, and the second last bit only from the last two bytes, and so on. We don't want to check for boundary criterion based on information from the last few bytes but rather from the entire window, so the mask is put at the left most bits.

assafvayner: oh good point I didn't think of that. Thanks for pointing this out.

mdb_shard/src/cas_structs.rs
```diff
@@ -87,8 +87,8 @@ impl CASChunkSequenceHeader {
 #[derive(Clone, Debug, Default, PartialEq)]
 pub struct CASChunkSequenceEntry {
     pub chunk_hash: MerkleHash,
-    pub unpacked_segment_bytes: u32,
     pub chunk_byte_range_start: u32,
+    pub unpacked_segment_bytes: u32,
```
rajatarya, line 91: Do we have to be careful about reordering these? Does this change the physical layout on the disk?

spec/auth.md
```diff
@@ -0,0 +1,127 @@
...
+
+Users may assume the "accessToken" and "casUrl" fields lengths have an upper limit of 64000 characters.
+
+#### Example Response Object
+
+```json
+{
+   "accessToken": "xet_xxxxxxxxxxx",
+   "exp": 1848535668,
+   "casUrl": "https://cas-server.xethub.hf.co"
```
rajatarya, line 66: Is this the actual URL? I thought it was https://cas-server.transfer.hf.co now?

spec/auth.md
```diff
@@ -0,0 +1,127 @@
...
+   Use Bearer authentication with the value for the `accessToken` key to authenticate with the Xet service at `endpoint` until `expiration` time.
+
+4. **Token refresh (when needed):**
+   Use the same API to generate a new token.
+  
+   > In `xet-core` we use and recommend to add in 30 seconds of buffer time before the provided `expiration` time to refresh the token.
+
+## Token Scope
+
+Xet tokens can have either a `read` or a `write` scope. The type of token issues is determined on the `token_type` URI path component when requesting the token from the Hugging Face Hub (see above).
```
rajatarya: I would add a line in this paragraph about how write tokens can perform all the actions read tokens can perform, along with write operations. I think that point can be easily skipped as it is written two paragraphs below.

spec/chunking.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Boundary test and size constraints
+
+At each position after updating `h`, let `size = current_offset - start_offset + 1`.
+
+- If `size < MIN_CHUNK_SIZE`: do not test `MASK`; continue
+- Else if `size >= MAX_CHUNK_SIZE`: force a boundary
+- Else if `(h & MASK) == 0`: boundary at this position
+
+When a boundary is taken:
```
rajatarya: Maybe rephrase: 'When a boundary is found or taken:' instead?

spec/deduplication.md
```diff
@@ -0,0 +1,182 @@
+# (TODO iterate on this, very drafty) Xet Chunk-Level Deduplication Specification
```
rajatarya: Ready to drop the TODO part of this line?

spec/deduplication.md
```diff
@@ -0,0 +1,182 @@
...
+- Multiple versions of datasets with incremental changes
+- Model checkpoints that share common layers or parameters
+- Documentation and configuration files with similar content
+- Large files where only portions have changed between versions
+
+## Core Concepts
+
+### Chunks
+
+[chunking.md](../spec/chunking.md)
```
rajatarya: I'd move this after the paragraph of what a chunk is, and maybe:

```
[Detailed chunking description](../spec/chunking.md)
```

spec/deduplication.md
```diff
@@ -0,0 +1,182 @@
...
+
+[See section about file reconstruction](../file_reconstruction.md).
+
+## Fragmentation Prevention
+
+While deduplication is valuable for saving space, doing it too aggressively can cause file fragmentation—meaning a file’s chunks end up scattered across many different xorbs. This can make reading files slower and less efficient.
+To avoid this, in xet-core we aim (and encourage implementors) to keep long, continuous runs of chunks together in the same xorb whenever possible.
+Instead of always deduplicating every possible chunk, the system sometimes chooses to reference a straight run of chunks in a single xorb, even if it means skipping deduplication for a few chunks.
+This approach balances the benefits of deduplication with the need to keep files easy and fast to read.
+Consider for example referencing a deduplicated chunks in a minimum run of chunks (e.g. at least 8 chunks) or targeting an average term length >= 1MB.
```
rajatarya: Has term been defined before? A term is a contiguous set of chunks, rights?

spec/download_protocol.md
```diff
@@ -0,0 +1,329 @@
...
+  - `unpacked_length`: Expected length after decompression (for validation)
+
+#### fetch_info
+
+- Type: `Map<Xorb Hash (64 character lowercase hex string), Array<CASReconstructionFetchInfo>>`
+- Maps xorb hashes to required information to download some of their chunks.
+  - The mapping is to an array of 1 or more `CASReconstructionFetchInfo`
+- Each `CASReconstructionFetchInfo` contains:
+  - `url`: HTTP URL for downloading the xorb data, presigned url containing authorization information
+  - `url_range`: Byte range `{ start: number, end: number }` for the Range header; end-inclusive `[start, end]`
```
rajatarya: I wonder if we want to rename to differentiate from the range parameters. Right now both are start/end: number, but one is byte range and one is chunk index.

Maybe `url_range(bytes_start, bytes_end)`

and then `range(index_start, index_end)`

This, I think is clearer. And it avoids simple issues where the types are the same.

spec/download_protocol.md
```diff
@@ -0,0 +1,329 @@
...
+    - This range indicates which range of chunk indices within this xorb that this fetch info term is describing
+
+## Stage 3: Downloading and Reconstructing the File
+
+### Process Overview
+
+1. Process each `CASReconstructionTerm` in order from the `terms` array
+2. For each `CASReconstructionTerm`, find matching fetch info using the term's hash
+  i. get the list of fetch_info items under the xorb hash from the `CASReconstructionTerm`. The xorb hash is guaranteed to exist as a key in the fetch_info map.
+  ii. linearly iterate through the list of `CASReconstructionFetchInfo` and find one which refers to a chunk range that is equal or encompassing the term's chunk range.
```
rajatarya: This doesn't render correctly in markdown - I see it on the same line as i.

spec/download_protocol.md
```diff
@@ -0,0 +1,329 @@
...
+offset_into_first_range = reconstruction["offset_into_first_range"]
+```
+
+#### Match Terms to Fetch Info
+
+For each `CASReconstructionTerm` in the `terms` array:
+
+- Look up the term's `hash` in the `fetch_info` map to get a list of `CASReconstructionFetchInfo`
+- Find a `CASReconstructionFetchInfo` entry where the fetch info's `range` contains the term's `range`
+  - linearly search through the array of CASReconstructionFetchInfo and find the element where the range block ({ "start": number, "end": number }) of the `CASReconstructionFetchInfo` has start <= term's range start AND end >= term's range end.
```
rajatarya: Is it guaranteed that there will always exist a range block that meets the criteria described here?

If not the client should fail the download, right?

spec/download_protocol.md
```diff
@@ -0,0 +1,329 @@
...
+For partial file downloads, the reconstruction API supports range queries:
+
+- Include `Range: bytes=start-end` header in reconstruction request
+- The `offset_into_first_range` field indicates where your range starts within the first term
+- The end of the content will need to be truncated to fit the requested range.
+  - Except if the requested range exceeds the total file length, then the returned content will be shorter and no truncation is necessary.
+
+When downloading individual term data:
+
+A client must include the range header formed with the values from the url_range field to specify the exact range of data of a xorb that they are accessing. Not specifying this header will cause result in an authorization failure.
```
rajatarya: Might want to include language here of the form: "Xet global deduplication requires that access to xorbs is only granted to authorized ranges. Not specifying this header will result in an authorization failure."

spec/spec.md
rajatarya: I agree. Might want to structure with subsections like: "Building a client library for xet storage", "Overall Xet architecture", "Deduplication explained".

Each of these headings can just present the same links in different order with slightly different context. Just creating a TOC that allows a reader to jump in based on their intent/use case.

spec/shard.md
rajatarya, line 1: This file is a little stiff (seems AI generated). I think a humanizing pass on this file would make it easier. On first read I wasn't sure why this file/format was necessary for a client builder.

rajatarya (commented):
Huge progress on this set of documents, take a bow @assafvayner!

I think you are _really_ close. I have a really hard time reviewing like this - I think it is time to get these .md files rendered properly on an HTML site (HF Space, or whatever) so they can be reviewed with the eyes of a user first coming to them.

I think the content is all there and written clearly - I believe the remaining pieces are minor context and organization to simplify the consumption/understanding for the audience.

spec/auth.md
assafvayner: `https://cas-server.xethub.hf.co` is the real url

spec/auth.md
assafvayner: I'm not sure I understand, do I add/move it? it is stated in the next paragraph as you mentioned.

spec/download_protocol.md
assafvayner: I've got it separated by a space, these are the keys into the json object for the fetch info struct.

But specified this

mdb_shard/src/cas_structs.rs
assafvayner: no, the serialization/deserialization is explicit in the code, in fact this re-order is to make it consistent with the ordering in the code.

spec/auth.md
rajatarya: Yes the suggestion is to move it into the same paragraph so it isn't skipped.

You could rephrase to: 'Xet tokesn can have either a `read` or a `write` scope. `write` scope supercedes `read` and can do all `read` operations along with `write` operations.'

PR #423, opened 2025-07-24
add whether chunk should be checked against global dedup

coyotte508:
https://huggingface.slack.com/archives/C07KCK52LBY/p1753369740405649

It's not trivial in JS - need to either create a dataview on the Uint8Arr (we have string) or make little endian conversion manually on the specific two bytes involved from the  hash

At the moment it's simpler to output this information from the WASM


hf_xet_thin_wasm/src/lib.rs
```diff
@@ -16,20 +16,26 @@ extern "C" {
 pub struct JsChunk {
     pub hash: String,
     pub length: u32,
+    pub dedup: bool,
```
assafvayner, line 25: This is good for sending the chunk back to JS, but is it always guaranteed that JS will also include it when sending a chunk object to WASM (for xorb/file hash calculation). If yes, then this is good as is. If not then with this current version deserialization will fail if that field is missing but it's simple to make it resilient:

- convert the type of the field from `bool` to `Option<bool>`
- add this line above that field `#[serde(skip_serializing_if = "Option::is_none")]`
- later when you set the `dedup` field, wrap it in `Some(<value>)`

Then both: {hash: "<real hash>", length: 0, dedup: true/false} and {hash: "<real hash>", length: 0} deserialize correctly.

But if you prefer to just enforce that the `dedup` field is included when sent from JS, then that is also viable IMO.

hf_xet_thin_wasm/src/lib.rs
```diff
@@ -40,19 +46,29 @@ impl JsChunker {
...
         JsChunker {
             inner: deduplication::Chunker::new(target_chunk_size),
+            first_chunk_outputted: false,
         }
     }
 
     pub fn add_data(&mut self, data: Vec<u8>) -> Result<JsValue, JsValue> {
         let result = self.inner.next_block(&data, false);
-        let serializable_result: Vec<JsChunk> = result.into_iter().map(JsChunk::from).collect();
+        let mut serializable_result: Vec<JsChunk> = vec![];
```
assafvayner: very minor optimization

```suggestion
        let mut serializable_result: Vec<JsChunk> = Vec::with_capacity(result.len());
```

assafvayner (approved):
LGTM, few minor suggestions

hf_xet_thin_wasm/src/lib.rs
coyotte508: Good catch, I added `    #[serde(skip_deserializing)]` on the `dedup` field, should I still make the field optional?

coyotte508: Ended up making two different types in https://github.com/huggingface/xet-core/pull/423/commits/0e53c8777a77b7c7928a6c2fddefc556f744ce60

assafvayner: > Good catch, I added     #[serde(skip_deserializing)] on the dedup field, should I still make the field optional?

We would have needed it to be optional (or define a default; but then it become impossible to differentiate a false from a null/unset) in case we get passed in a JsChunk with no dedup field. I think the current version is fine.

coyotte508: ok I don't have write perms and this repo so merge when you think it's ok!

PR #424, opened 2025-07-24
Errors on shard reading are now logged and ignored.

hoytak:
Reading the lookup hashes from a shard could cause an error to be propagated when it should simply be ignored; This PR logs the error but ignores it.

### DIFF
--- a/mdb_shard/src/shard_file_manager.rs
+++ b/mdb_shard/src/shard_file_manager.rs
@@ -6,7 +6,7 @@ use std::sync::Arc;
 
 use merklehash::{HMACKey, MerkleHash};
 use tokio::sync::RwLock;
-use tracing::{debug, info, instrument, trace};
+use tracing::{debug, info, instrument, trace, warn};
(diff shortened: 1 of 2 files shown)


mdb_shard/src/shard_file_manager.rs
```diff
@@ -276,7 +276,19 @@ impl ShardFileManager {
...
+                            warn!(
+                                "Join error on task reading hashes from {:?}; skipping dedup lookup insert.",
+                                &s.path
+                            );
+                            return Ok(sbkp_lg);
+                        };
+
+                        let Ok(insert_hashes) = insert_hashes_task_result.map_err(|e| {
+                            warn!("Error reading hashes from {:?}: {e:?}. Skipping dedup lookup insert.", &s.path);
+                        }) else {
```
assafvayner, line 289: this looks like a good use of `ErrorPrinter::warn_error`

assafvayner (approved):

seanses (approved):

Issue #417, opened 2025-07-17
Cache directory env vars ignored, causing failures when ~/.cache isn't usable


half-duplex:
### Describe the bug

xet-core does not obey [`$CACHE_DIRECTORY`](https://www.freedesktop.org/software/systemd/man/257/systemd.exec.html#RuntimeDirectory=) or [`$XDG_CACHE_HOME`](https://specifications.freedesktop.org/basedir-spec/latest/), instead [defaulting](https://github.com/huggingface/xet-core/blob/5afd1eeef1b91d59d735a7d5d2c7712c1ad2ab46/data/src/data_client.rs#L36-L47) to `$HOME/.cache/` . This goes very poorly when it (via hf_xet, at least) is used by services running under users with a non-writable home directory, like immich on NixOS: https://github.com/NixOS/nixpkgs/issues/418799 . It also likely affects services using systemd.exec hardening tools like `DynamicUser=true`.

Beyond not working, the resulting error, `Operation not permitted (os error 1)`, gives no information about the problem. I wasted quite a bit of time assuming the download destination wasn't accessible before finding the actual problem in strace.

When used from a normal user account, checking `$XDG_CACHE_HOME` before using a default ensures you respect the user's preferences. When used from a systemd service, ~`$XDG_CACHE_HOME`~ `$CACHE_DIRECTORY` is set based on [`CacheDirectory`](https://www.freedesktop.org/software/systemd/man/257/systemd.exec.html#RuntimeDirectory=), so you'll much more reliably have a writable directory that's appropriate for cache instead of a home directory that may be nonexistent, read-only, or otherwise an inappropriate place for cache.

Combined with the existing logic, I think the ideal would be:
- `$HF_XET_CACHE`
- else `$HF_HOME/xet`
- else `$CACHE_DIRECTORY/huggingface/xet` (new)
- else `$XDG_CACHE_HOME/huggingface/xet` (new)
- else `$HOME/.cache/huggingface/xet`
- else `(PWD)/.cache/huggingface/xet`

### Reproduction

1. Choose an application that uses xet-core, e.g. [immich](https://immich.app/) (with hf_xet)
2. Create a system user with a home directory that isn't writeable or doesn't exist, e.g. `/var/empty`
3. Create a service for the application, with `CacheDirectory=immich` and `User=immich`
4. Start the service
5. See that downloading models fails with an error like below

### Logs

<details>
<summary>Service log</summary>

```
Traceback (most recent call last):
[...]

PR #426, opened 2025-07-25
Respect XDG_CACHE_HOME and ~/ when setting cache directory.

hoytak:
Currently, we default the cache directory to `home_dir()/.cache`, but as pointed out in https://github.com/huggingface/xet-core/issues/417, this is the incorrect behavior.  This PR switches this behavior to use the [cache_dir()](https://docs.rs/dirs/latest/dirs/fn.cache_dir.html) function to properly determine this.    

The side effect of this, however, is that on OSX the cache dir will change to the more standard `$HOME/Library/Caches/huggingface/xet`, which means nothing in `~/.cache/huggingface/xet` will be valid anymore. 

Fixes https://github.com/huggingface/xet-core/issues/417.

### DIFF
--- a/data/src/data_client.rs
+++ b/data/src/data_client.rs
@@ -9,13 +9,14 @@ use cas_client::remote_client::PREFIX_DEFAULT;
 use cas_client::{CacheConfig, FileProvider, OutputProvider, CHUNK_CACHE_SIZE_BYTES};
 use cas_object::CompressionScheme;
 use deduplication::DeduplicationMetrics;
-use dirs::cache_dir;
+use dirs::home_dir;
 use parutils::{tokio_par_for_each, ParallelError};
 use progress_tracking::item_tracking::ItemProgressUpdater;
 use progress_tracking::TrackingProgressUpdater;
-use tracing::{info_span, instrument, Instrument, Span};
+use tracing::{info, info_span, instrument, Instrument, Span};
 use ulid::Ulid;
 use utils::auth::{AuthConfig, TokenRefresher};
+use utils::normalized_path_from_user_string;
 
 use crate::configurations::*;
 use crate::constants::{INGESTION_BLOCK_SIZE, MAX_CONCURRENT_DOWNLOADS, MAX_CONCURRENT_FILE_INGESTION};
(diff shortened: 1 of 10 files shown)


jsulz:
Thanks for the PR explanation, @hoytak, I missed some nuance when we were chatting in Slack 😅 

I think our goal should be to follow the `huggingface_hub` logic so the `xet` cache is always nested inside that directory. This is a better developer experience, even if it's not exactly following OS conventions to a tee. 

That would still address #417 (just followed up in that issue to confirm) since the construction of the cache directory in `huggingface_hub` respects the `XDG_CACHE_HOME` environment variable: https://github.com/huggingface/huggingface_hub/blob/9d258315ce580c74366cee5ae492f22111ce3cdd/src/huggingface_hub/constants.py#L129-L138

Does that make sense?

data/src/data_client.rs
```diff
@@ -35,16 +36,31 @@ pub fn default_config(
...
     // HF_XET_CACHE takes precedence over HF_HOME
-    let cache_root_path = if env::var("HF_XET_CACHE").is_ok() {
-        PathBuf::from(env::var("HF_XET_CACHE").unwrap())
-    } else if env::var("HF_HOME").is_ok() {
-        let home = env::var("HF_HOME").unwrap();
-        PathBuf::from(home).join("xet")
-    } else {
-        let home = home_dir().unwrap_or(current_dir()?);
-        home.join(".cache").join("huggingface").join("xet")
+    let cache_root_path = {
```
jsulz, line 39: I can really only speak to this section (and so I will 😄 ) - this looks good to me and appears to match https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/constants.py#L129-L138

jsulz (approved):
The directory logic looks good to me!

data/src/data_client.rs
```diff
@@ -35,16 +36,31 @@ pub fn default_config(
...
+        if let Ok(cache) = env::var("HF_XET_CACHE") {
+            normalized_path_from_user_string(cache)
+
+        // If HF_HOME is set, use the $HF_HOME/xet
+        } else if let Ok(hf_home) = env::var("HF_HOME") {
+            normalized_path_from_user_string(hf_home).join("xet")
+        } else {
+            // If XDG_CACHE_HOME is set, use the $XDG_CACHE_HOME/huggingface/xet, otherwise
+            // use $HOME/.cache/huggingface/xet
+            let cache_home = match env::var("XDG_CACHE_HOME") {
```
assafvayner: this makes more sense to be the next bit of the `if let Ok()..` chain rather than a match. Then the last default is the else case

utils/src/file_paths.rs
```diff
@@ -0,0 +1,190 @@
...
+///
+/// // Assuming current directory is "/work"
+/// let p = normalized_path_from_user_string("data/file.txt");
+/// assert!(p.is_absolute());
+/// // Typically, p == PathBuf::from("/work/data/file.txt") on Unix-like systems.
+/// ```
+pub fn normalized_path_from_user_string(path: impl AsRef<str>) -> PathBuf {
+    let path = path.as_ref();
+
+    #[cfg(target_family = "unix")]
```
assafvayner: I think this can be replaced with `#[cfg(unix)]` and below `#[cfg(not(unix))]`

utils/Cargo.toml
```diff
@@ -21,12 +21,19 @@ thiserror = { workspace = true }
 tokio = { workspace = true, features = ["time", "rt", "macros"] }
 tracing = { workspace = true }
 
+# Only import this on non-wasm unix systems
+[target.'cfg(target_family = "unix")'.dependencies]
```
assafvayner: similar to below, there is a macro for `unix` rather than having to check target_family

assafvayner (approved):

seanses (approved):

utils/src/file_paths.rs
```diff
@@ -0,0 +1,190 @@
...
+    {
+        use expanduser::expanduser;
+        let user_expanded_path = expanduser(path).unwrap_or_else(|_| PathBuf::from(path));
```
seanses: From the test failure on Windows looks like this crate doesn't have a good support for Windows. Try [shellexpand](https://docs.rs/crate/shellexpand/latest)?

PR #428, opened 2025-07-29
Logging improvements

hoytak:
Added logging improvements to hf_xet: 
- When HF_XET_LOG_FILE is set to a valid file path, then events are logged to that file in a non-blocking manner.  
- When HF_XET_LOG_FORMAT is set to either "json" or "text", it overrides the log format default.  By default, when done to a file, logging is done in json, and to console it is done using pretty printing.  
- Added in debug log statements in the upload and download paths to help trace activity.

### DIFF
--- a/hf_xet/src/lib.rs
+++ b/hf_xet/src/lib.rs
@@ -292,8 +292,6 @@ pub fn hf_xet(py: Python<'_>, m: &Bound<'_, PyModule>) -> PyResult<()> {
     // huggingface_hub.
     m.add_class::<PyPointerFile>()?;
 
-    eprintln!("Initializing hf_xet, PID = {}", std::process::id());
-
     // Init the threadpool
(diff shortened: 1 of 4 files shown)


hf_xet/src/lib.rs
```diff
@@ -259,6 +292,8 @@ pub fn hf_xet(py: Python<'_>, m: &Bound<'_, PyModule>) -> PyResult<()> {
     // huggingface_hub.
     m.add_class::<PyPointerFile>()?;
 
+    eprintln!("Initializing hf_xet, PID = {}", std::process::id());
```
assafvayner: does this always print? do we want this to always print?

assafvayner (approved):

hf_xet/src/lib.rs
hoytak: Ah, should change this to debug.

Issue #415, opened 2025-07-15
Error downloading from Xet repository in Zero GPU Space


jsulz:
> I encountered a similar situation in Hugging Face's Gradio Zero GPU space, so I am reporting it just in case. I [tried to reproduce the issue with the minimum configuration, but I was unable to reproduce it with just `hf_hub_download`](https://huggingface.co/spaces/John6666/yolotest). There must be some prerequisites, but I have not been able to identify them. The details of the situation are unclear.
> I am attaching the log with the `HF_DEBUG` environment variable enabled.
> 
> ### Actual repo (dedicated for Zero GPU)
> https://huggingface.co/spaces/John6666/DiffuseCraftDetailfixTest (Only when Detailfix is enabled)
> 
> ### Related code
> https://github.com/R3gm/stablepy/blob/main/stablepy/diffusers_vanilla/adetailer.py#L45
> 
> ### Libraries
> ```
> hf_transfer               0.1.9
> hf-xet                    1.1.5
> httpcore                  1.0.9
> httpx                     0.28.1
> huggingface-hub           0.33.2
> ```
> 
> ### When HF_HUB_DISABLE_XET == "0"
> ```
> [DEBUG] >> detailfix inpaint only
> Send: curl -X HEAD -H 'Accept: */*' -H 'Accept-Encoding: identity' -H 'Connection: keep-alive' -H 'user-agent: unknown/None; hf_hub/0.33.2; python/3.10.13; torch/2.2.0' https://huggingface.co/Bingsu/adetailer/resolve/main/face_yolov8n.pt
> Request 70440b92-03c0-42f3-896f-7696aaeacb1f: HEAD https://huggingface.co/Bingsu/adetailer/resolve/main/face_yolov8n.pt (authenticated: False)
> Downloading 'face_yolov8n.pt' to '/home/user/cache/huggingface/hub/models--Bingsu--adetailer/blobs/70b640f8f60b1cf0dcc72f30caf3da9495eb2fb6509da48c53374ad6806e6a9c.incomplete'
> Xet Storage is enabled for this repo. Downloading file from Xet Storage..
[...]

PR #429, opened 2025-07-30
Make hf_xet fork-exec safe

hoytak:
Currently, some environments (e.g. vllm) use spawn or fork-exec to create a child process.  However, this is known to cause issues within the tokio runtime and lead to hangs, as only the calling thread survives across spawns but the runtime assumes these threads exist and are accessible.  This PR detects when a fork-exec has occurred and silently discards the old runtime, creating a new one and reinstalling the signal handlers and restarting the background threads. 

Possible fix for https://github.com/huggingface/xet-core/issues/415; also issue in https://github.com/vllm-project/vllm/pull/21539.

### DIFF
--- a/hf_xet/src/runtime.rs
+++ b/hf_xet/src/runtime.rs
@@ -13,7 +13,7 @@ use crate::log;
 
 lazy_static! {
     static ref SIGINT_DETECTED: Arc<AtomicBool> = Arc::new(AtomicBool::new(false));
-    static ref SIGINT_HANDLER_INSTALLED: (AtomicU32, Mutex<()>) = (AtomicU32::new(0), Mutex::new(()));
+    static ref SIGINT_HANDLER_INSTALL_PID: (AtomicU32, Mutex<()>) = (AtomicU32::new(0), Mutex::new(()));
     static ref MULTITHREADED_RUNTIME: RwLock<Option<(u32, Arc<ThreadPool>)>> = RwLock::new(None);
 }
 
@@ -49,7 +49,7 @@ fn check_sigint_handler() -> PyResult<()> {
     // will have to press it again.
     SIGINT_DETECTED.store(false, Ordering::SeqCst);
 
-    let stored_pid = SIGINT_HANDLER_INSTALLED.0.load(Ordering::SeqCst);
+    let stored_pid = SIGINT_HANDLER_INSTALL_PID.0.load(Ordering::SeqCst);
     let pid = std::process::id();
 
     if stored_pid == pid {
@@ -57,10 +57,10 @@ fn check_sigint_handler() -> PyResult<()> {
     }
 
     // Need to install it; acquire a lock to do so.
-    let _install_lg = SIGINT_HANDLER_INSTALLED.1.lock().unwrap();
+    let _install_lg = SIGINT_HANDLER_INSTALL_PID.1.lock().unwrap();
 
     // If another thread beat us to it while we're waiting for the lock.
-    let stored_pid = SIGINT_HANDLER_INSTALLED.0.load(Ordering::SeqCst);
+    let stored_pid = SIGINT_HANDLER_INSTALL_PID.0.load(Ordering::SeqCst);
     if stored_pid == pid {
         return Ok(());
     }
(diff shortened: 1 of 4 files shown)


hoytak:
Fix available in an RC as  `pip install hf_xet==1.1.6rc5`

hf_xet/src/runtime.rs
```diff
@@ -12,8 +12,8 @@ use crate::log;
 
 lazy_static! {
     static ref SIGINT_DETECTED: Arc<AtomicBool> = Arc::new(AtomicBool::new(false));
-    static ref SIGINT_HANDLER_INSTALLED: (AtomicBool, Mutex<()>) = (AtomicBool::new(false), Mutex::new(()));
-    static ref MULTITHREADED_RUNTIME: RwLock<Option<Arc<ThreadPool>>> = RwLock::new(None);
+    static ref SIGINT_HANDLER_INSTALLED: (AtomicU32, Mutex<()>) = (AtomicU32::new(0), Mutex::new(()));
```
rajatarya: If AtomicU32 is a counter, would it make more sense to have this key be the PID of the process?

(Instead of having a counter can we use the process PID)?

hf_xet/src/runtime.rs
```diff
@@ -84,9 +87,11 @@ fn signal_check_background_loop() {
             // meaining that all the tasks have completed or been cancelled.
```
rajatarya: typo: meaining -> meaning

hf_xet/src/runtime.rs
```diff
@@ -106,18 +111,33 @@ pub fn init_threadpool(py: Python) -> PyResult<Arc<ThreadPool>> {
...
+
+    let mut swapping_after_fork = false;
+
+    if let Some((runtime_pid, existing)) = guard.take() {
+        if runtime_pid == pid {
+            // We're OK, so reset it here.
+            *guard = Some((pid, existing.clone()));
+            return Ok(existing);
+        } else {
+            // Ok, discard the previous runtime, as it's effectively poisoned by the
```
rajatarya, line 130: Should we log this - seems like something worth logging (maybe at INFO level)?

rajatarya (approved):
I really like these changes, suggesting some logging so easier to know when this codepath is executing.

xet_threadpool/src/threadpool.rs
```diff
@@ -204,6 +204,28 @@ impl ThreadPool {
...
+
+        // When a task is shut down, it will stop running at whichever .await it has yielded at.  All local
+        // variables are destroyed by running their destructor.
+        let Ok(mut rt_lock) = self.runtime.write() else {
+            return;
+        };
+
+        let maybe_runtime = rt_lock.take();
+
+        let Some(runtime) = maybe_runtime else {
```
assafvayner: I think you can inline the maybe_runtime variable

xet_threadpool/src/threadpool.rs
```diff
@@ -204,6 +204,28 @@ impl ThreadPool {
         drop(runtime);
     }
 
+    /// Discards the runtime without shutdown; to be used after fork-exec or spawn.
+    pub fn discard_runtime(&self) {
+        // This function makes a best effort attempt to clean everything up.
+
+        // When a task is shut down, it will stop running at whichever .await it has yielded at.  All local
+        // variables are destroyed by running their destructor.
+        let Ok(mut rt_lock) = self.runtime.write() else {
```
assafvayner, line 216: what is the meaning of an error here, would this be an error we need to report up, or ignoring is correct. If ignore is right, then add it to the comment right above.

hf_xet/src/runtime.rs
assafvayner: It is the PID of the process, however might be worth a rename of the this variable.

jsulz (approved):
Confirmed this resolves #415 by testing in duplicated spaces.

hf_xet/src/runtime.rs
hoytak: This does use the process ID.  The zero is because that is never used as a process ID, so it's effectively the null value.

hf_xet/src/runtime.rs
hoytak: We can't actually do that, as the logging requires the runtime and the runtime isn't valid at this point...  But maybe later?

hf_xet/src/runtime.rs
```diff
@@ -146,10 +168,15 @@ fn get_threadpool(py: Python) -> PyResult<Arc<ThreadPool>> {
     // First try a read lock to see if it's already initialized.
     {
         let guard = MULTITHREADED_RUNTIME.read().unwrap();
-        if let Some(ref existing) = *guard {
-            return Ok(existing.clone());
+        if let Some((runtime_pid, ref existing)) = *guard {
```
seanses, line 179: Nit: we can wrap multiple copies of this pid check logic into the MULTITHREADED_RUNTIME such that a `get()` returns None if there's a pid mismatch.

seanses (approved):
LGTM!

hf_xet/src/runtime.rs
assafvayner: IIUC this is that check; getting the runtime we check the pid matches what's stored. right?

assafvayner (approved):

hf_xet/src/runtime.rs
hoytak: I tried it that way first, but I need the new pid anyway in one place, and I need to know if the swap happened in one place, so I decided to just make it explicit.  Less code that way.

PR #431, opened 2025-07-30
Limit number of async worker threads on large CPUs

hoytak:
Currently, tokio spins up async worker threads equal to the number of cores, which can be quite large on huge machines, e.g. 128.  This isn't needed to keep everything running; we already offload much of the compute to blocking threads and so here the number is a significant overkill, especially if hf_xet is used for downloading only a few files.

This PR limits the number of async worker threads to 32 unless TOKIO_WORKER_THREADS is set, in which case that value is used.  It also removes the cap on the number of blocking threads tokio can spin up as needed as there is no real reason to not use tokio's default value there.


xet_threadpool/src/threadpool.rs
```diff
@@ -6,13 +6,54 @@ use std::sync::{Arc, OnceLock};
...
+/// 2) Otherwise, use `min(available_parallelism, THREADPOOL_MAX_ASYNC_THREADS)`, with a floor of 2.
+#[cfg(not(target_family = "wasm"))]
+fn get_num_tokio_worker_threads() -> usize {
+    use std::num::NonZeroUsize;
+
+    // Allow TOKIO_WORKER_THREADS to override this value.
+    if let Ok(val) = std::env::var("TOKIO_WORKER_THREADS") {
+        match val.parse::<usize>() {
+            Ok(n) if n > 0 => {
+                use tracing::info;
```
assafvayner: here and below any good reason to not just import at the top of the file?

xet_threadpool/src/threadpool.rs
```diff
@@ -6,13 +6,54 @@ use std::sync::{Arc, OnceLock};
...
+/// Note that the compute intensive parts of the code get offloaded to blocking threads, which don't count against this
+/// limit.
+const THREADPOOL_MAX_ASYNC_THREADS: usize = 32;
+
+/// Returns the number of Tokio worker threads to use:
+/// 1) If `TOKIO_WORKER_THREADS` is set to a positive integer, use that.
+/// 2) Otherwise, use `min(available_parallelism, THREADPOOL_MAX_ASYNC_THREADS)`, with a floor of 2.
+#[cfg(not(target_family = "wasm"))]
+fn get_num_tokio_worker_threads() -> usize {
+    use std::num::NonZeroUsize;
```
assafvayner, line 28: again, any reason to not import at the top? (is it for wasm? for linting?)

assafvayner (approved):

xet_threadpool/src/threadpool.rs
hoytak: It's just how my vscode does it automatically -- happy to switch.

xet_threadpool/src/threadpool.rs
```diff
@@ -6,13 +6,54 @@ use std::sync::{Arc, OnceLock};
...
-use tracing::debug;
+use tracing::{debug, info};
 
 use crate::errors::MultithreadedRuntimeError;
 
 const THREADPOOL_THREAD_ID_PREFIX: &str = "hf-xet"; // thread names will be hf-xet-0, hf-xet-1, etc.
 const THREADPOOL_STACK_SIZE: usize = 8_000_000; // 8MB stack size
-const THREADPOOL_MAX_BLOCKING_THREADS: usize = 100; // max 100 threads can block IO
+
+/// Cap the number of tokio threads to 16 to avoid massive expansion on huge CPUs; can be overridden with
```
seanses:
```suggestion
/// Cap the number of tokio threads to 32 to avoid massive expansion on huge CPUs; can be overridden with
```

PR #433, opened 2025-07-30
Associate static semaphores with runtime

hoytak:
If a parent process spawns a child process while permits from a static semaphore are issued, the number of permits available to the child process will be reduced for the entirety of the child's lifetime, even when the parent process permits are returned.  This could potentially cause a deadlock or painful slowdown on upload or download.   This PR moves all our static semaphores to ones associated with the runtime, so after a spawn they are reset.


seanses (approved):
LGTM! Only that it seems there's no need for the global semaphore lookup to be a hash map, a Vec seems enough (and faster).

PR #435, opened 2025-08-01
[do not merge] add shard deserialization to wasm

coyotte508:
Just to compare with local deserialization and debug


coyotte508:
for info this PR makes the wasm grow from 100kB to 115kB

PR #437, opened 2025-08-04
Make hf_xet safe(ish) across python os.fork()

hoytak:
This PR ensures that none of the tokio thread state exists through a call to python's os.fork() as used in the multiprocessing library.  For an explanation of the issue, see https://github.com/vllm-project/vllm/blob/main/docs/design/multiprocessing.md#tradeoffs. 

It does this by offloading all the async calls to a separate and transient OS thread, which would not exist after the spawn process.  Thus any possible restart of the tokio runtime due to a spawn would occur in a clean environment and without thread-local storage causing issues. 

To accomplish this, this PR refactors the hf_xet logging layer to separate it out from the python runtime, as the python runtime is not Send/Sync.  This also simplifies this layer somewhat and isolates the telemetry reporting logic so that only the background sending thread of the telemetry logic is restarted after a spawn. 

In addition, this PR removes the use of parking_lot, both in singleflight.rs and as part of tokio.  The library is not safe across fork(); in particular, note https://github.com/Amanieu/parking_lot/blob/9c810e4a118fd592d528c8659ac599ba2049c824/core/src/parking_lot.rs#L51.


hf_xet/src/logging.rs
```diff
@@ -83,11 +85,41 @@ fn init_logging_to_file(path: &Path) -> Result<(), std::io::Error> {
...
+            {
+                version_info.push_str(&format!("{package_name}/{version}; "));
+            }
+        }
+    }
+    version_info
+}
+
+fn init_global_logging(py: Python) {
+    let version_info = get_version_info_string(py);
```
jgodlew, line 116: nit: can we push this down to only be done if we have telemetry enabled (i.e. move to line 140 inside the HF_HUB_ENABLE_TELEMETRY block)?

hf_xet/src/runtime.rs
```diff
@@ -161,9 +169,7 @@ pub fn init_threadpool(py: Python) -> PyResult<Arc<ThreadPool>> {
     drop(guard);
 
     // Initialize the logging.
```
jgodlew: nit: we can remove the comment since we aren't initializing logging anymore.

hf_xet/src/lib.rs
```diff
@@ -293,7 +293,12 @@ impl From<PyXetDownloadInfo> for (XetFileInfo, DestinationPath) {
     }
 }
 
+lazy_static::lazy_static! {
+    static ref current_process : Mutex<u32> = Mutex::new(0);
+}
+
```
sirahd: I don't see this being used anywhere. Was it added for debugging?

hf_xet/src/logging.rs
```diff
@@ -103,45 +135,50 @@ fn init_global_logging(py: Python) -> Option<TelemetryTaskInfo> {
...
 
-    // Spawn the telemetry logging.
-    if let Some(ref tti) = telemetry_task_info {
-        let telemetry_task = get_telemetry_task(tti.clone());
-        let _telemetry_task = runtime.spawn(telemetry_task);
+    let pid = std::process::id();
+
+    if logger_pid == 0 {
+        init_global_logging(py);
+        INITIALIZED_LOGGING_ID.store(pid, Ordering::SeqCst);
```
jgodlew: `init_global_logging` needs to only be called once or else it will panic. As it is currently, we might have simultaneous threads call `check_logging_state` in parallel, which leads to a race-condition here (both threads detect that logger_pid == 0 before either one can initialize logging).

Instead of the AtomicU32, we probably want the ID to be a OnceCell to guarantee this is only run once?

jgodlew: Not sure how often this race will occur though, but if something changes with pyo3 (especially with some non-GIL support), we might become susceptible to the race.

jgodlew (approved):
Overall looks good. Just a few comments / nits.

hf_xet/src/logging.rs
```diff
@@ -103,45 +135,50 @@ fn init_global_logging(py: Python) -> Option<TelemetryTaskInfo> {
...
+    } else if logger_pid != pid {
+        if let Err(e) = restart_telemetry_task_after_spawn() {
+            error!("Error restarting telemetry task in subprocess; telemtry may not work: {e:?}");
```
sirahd:
```suggestion
            error!("Error restarting telemetry task in subprocess; telemetry may not work: {e:?}");
```

hf_xet/src/logging.rs
hoytak: Ah!  I realized in other places we never record this information, which would be useful for debugging, so I was going to add an info! statement to the other paths to log this.  Put those in.

hf_xet/src/logging.rs
hoytak: Good thought.  Changed it to a lock.

hf_xet/src/lib.rs
hoytak: good call.  Artifact from a previous implementation.

hf_xet/src/runtime.rs
```diff
@@ -132,12 +138,12 @@ pub fn init_threadpool(py: Python) -> PyResult<Arc<ThreadPool>> {
             // resources will be freed up when the child exits.
             existing.discard_runtime();
 
-            swapping_after_fork = true;
+            info!("Runtime restarted due to detected process ID change, likely due to fork-exec call.");
```
seanses: "likely due to a fork call" (exec not needed to change pid)

hf_xet/src/runtime.rs
```diff
@@ -197,22 +197,30 @@ pub fn async_run<Out, F>(py: Python, execution_call: F) -> PyResult<Out>
...
-    let runtime_internal = runtime.clone();
-    let result: PyResult<Out> = py
-        .allow_threads(move || runtime_internal.external_run_async_task(execution_call))
+    // Make sure the logger is set up.
+    check_logging_state(py);
+
+    let result: PyResult<Out> = py.allow_threads(move || {
+        // Now, without the GIL, spawn the task on a new OS thread.  This avoids having tokio cache stuff in
+        // thread-local storage that is invalidated after a fork-exec.
+        spawn_os_thread(move || {
```
seanses, line 208: If I understand correctly this thread is only where the tokio scheduler runs, right, i.e. tokio runtime itself spawns other threads?

utils/src/singleflight.rs
```diff
@@ -130,7 +129,7 @@ where
     /// and notifying all waiters that there is a value.
     fn complete(&self, res: SingleflightResult<T, E>) {
         // write-lock
-        let mut val = self.res.write();
+        let mut val = self.res.write().unwrap();
```
seanses, line 132: I think we can change the return value of `complete()` to a `Result<_>` and propagate the error? This is called in `fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<T, SingleflightError<E>>>` which we can return the error

seanses: If we can do this maybe also do for the below "unwrap()"s

hf_xet/src/runtime.rs
hoytak: Yes, correct.

utils/src/singleflight.rs
hoytak: I agree, but trying to not make the changes too invasive.  This is just to get rid of the parking_lot rwlock.

PR #445, opened 2025-08-06
Fix DataHash hex string serde to little endian

seanses:
Different programming languages and platforms may have different byte order layout of a u64 in memory, this is not an issue when we typecast a [u8;32] to [u64;4] (which is essentially a pointer typecasting and thus doesn't reorder bytes at all), until these bytes are actually used as u64s. For such cases we make it explicit to use the little-endian order.


hoytak (approved):
LGTM.

PR #447, opened 2025-08-09
fix: singleflight owner task not removing Call from Group if dropped

jgodlew:
When a singleflight `Group` is called by many tasks for a particular key, one of these tasks is chosen as the `owner` to actually perform the work. The other tasks are considered `waiters` and will wait until the `owner` completes the work.

In the normal case, the `owner` runs the work, takes the result and provides it to any `waiter` tasks. It is also the responsibility of the `owner` to clean up the state it added to the singleflight `Group`, namely, the `Call` record in the `Group`'s `CallMap`. 

However, if the `owner` is `drop`'ed  while waiting for the work to complete (e.g. by its task being canceled), then the work will still finish and any waiter tasks will be notified (as those actions are part of a separately spawned task), but the state the owner added to the `Group` is not cleaned up (i.e. there is still a lingering mapping of key -> `Call`).

What this means is that if the `owner` is dropped before it can remove the mapping, then the results of the work are permanently "cached" in the singleflight `Group` and any subsequent tasks looking for the key will always get back those results until the `Group` is dropped (usually on process restart).

Since much of the work we use singleflight for is downloading immutable blobs of data, "caching" the `Call` results doesn't sound that terrible (we already cache some stuff outside of singleflight). However, this is caching the `Result`, not just the `Ok(…)` variant, so if the work returned an `Error`, then that becomes what is permanently cached, which, for intermittent networking issues can cause problems. 

This PR fixes the issue by adding a RAII guard struct when the `Call` is added to the `CallMap` so that the `work()` function will remove the `Call` when the owner exits the function (either successfully, due to error/panic, or if it is dropped).

See the `test_dropped_owner` test for an example of how this situation can occur.

### DIFF
--- a/utils/src/singleflight.rs
+++ b/utils/src/singleflight.rs
@@ -303,7 +303,7 @@ where
     Waiter,
 }
 
-impl<'a, T, E> Drop for CreateGuard<'a, T, E>
+impl<T, E> Drop for CreateGuard<'_, T, E>
 where
     T: ResultType + 'static,
     E: ResultError + 'static,
--- a/utils/src/singleflight.rs
+++ b/utils/src/singleflight.rs
@@ -42,30 +42,31 @@ use std::future::Future;
 use std::marker::PhantomData;
 use std::pin::Pin;
 use std::sync::atomic::{AtomicBool, AtomicU16, Ordering};
-use std::sync::{Arc, RwLock};
+use std::sync::{Arc, Mutex, RwLock};
 use std::task::{ready, Context, Poll};
 
 use futures::future::Either;
 use pin_project::{pin_project, pinned_drop};
 use tokio::runtime::Handle;
-use tokio::sync::{Mutex, Notify};
-use tracing::debug;
+use tokio::sync::Notify;
+use tracing::{debug, error};
 
 pub use crate::errors::SingleflightError;
 
 type SingleflightResult<T, E> = Result<T, SingleflightError<E>>;
 type CallMap<T, E> = HashMap<String, Arc<Call<T, E>>>;
+type CallCreate<'a, T, E> = (Arc<Call<T, E>>, CreateGuard<'a, T, E>);
 
 // Marker Traits to help make the code a bit cleaner.
 
-/// ResultType indicates the success type of a singleflight [Group].
+/// ResultType indicates the success type for a singleflight [Group].
 /// Since the actual processing might occur on a separate thread,
 /// we need to type to be [Send] + [Sync]. It also needs to be [Clone]
 /// so that we can clone the response across many tasks
 pub trait ResultType: Send + Clone + Sync + Debug {}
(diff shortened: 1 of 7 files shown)


utils/src/singleflight.rs
```diff
@@ -42,30 +42,31 @@ use std::future::Future;
...
 use std::pin::Pin;
 use std::sync::atomic::{AtomicBool, AtomicU16, Ordering};
-use std::sync::{Arc, RwLock};
+use std::sync::{Arc, Mutex, RwLock};
 use std::task::{ready, Context, Poll};
 
 use futures::future::Either;
 use pin_project::{pin_project, pinned_drop};
 use tokio::runtime::Handle;
-use tokio::sync::{Mutex, Notify};
```
jgodlew, line 51: Note: this PR switches from a Tokio Mutex to a std Mutex since we can't call async code in `drop()` (and we need to lock the `CallMap` to remove the Call on `CreateGuard::drop`). 

This is safe to do since the locking sections are only over data (i.e. the `CallMap`) and not across `await` boundaries or I/O.

utils/src/singleflight.rs
```diff
@@ -258,31 +259,69 @@ where
...
-    async fn get_call_or_create(&self, key: &str) -> (Arc<Call<T, E>>, bool) {
-        let mut m = self.call_map.lock().await;
+    /// not.
+    ///
+    /// Returns an error if the underlying `call_map` Mutex is poisoned.
+    fn get_call_or_create<'a>(&'a self, key: &'a str) -> Result<CallCreate<'a, T, E>, SingleflightError<E>> {
+        let mut m = self
+            .call_map
+            .lock()
+            .inspect_err(|err| error!(?err, "Failed to lock call map"))
```
assafvayner: use log_error() from ErrorPrinter if available

utils/src/singleflight.rs
```diff
@@ -258,31 +259,69 @@ where
...
+            .map_err(|_| SingleflightError::GroupLockPoisoned)?;
```
assafvayner, line 268: We can add an impl From<LockPoisonError> for SingleFlightError via thiserror or explicitly to make this simpler

jgodlew: Unfortunately, I can't add a From implementation because the generic error variant (`InternalError(E)`) will conflict with any From implementation that I try to add since `PoisonError<T>` could implement the `SingleflightError`'s `E` type in future versions (i.e. if MutexGuard gets `Send`).

jgodlew:
> Is there any possible issue with switching the mutex type?

I don't think there is any functional issue here (e.g. correctness or deadlocks) as we don't have an `await` while the CallMap is locked.

However, we are making the process of locking blocking, which means that tokio won't context-switch tasks if there is contention. I don't expect this to cause much slowdown as the critical sections are fairly short (cost of a HashMap get + insert or remove).

seanses (approved):
LGTM!

PR #449, opened 2025-08-12
Simple singleflight implementation.

hoytak:
This PR proposes a simple singleflight implementation that can be associated with the runtime and used globally to disambiguate client calls using singleflight. 

The idea is to associate this with the runtime so that multiple clients can use a common instance.

### DIFF
--- a/xet_threadpool/src/simple_singleflight.rs
+++ b/xet_threadpool/src/simple_singleflight.rs
@@ -250,22 +250,4 @@ mod singleflight_tests {
         }
         assert_eq!(CALLS.load(Ordering::SeqCst), 1);
     }
-
-    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
-    async fn type_mismatch_requires_clear_then_reuse_works() {
-        let mgr = SingleFlightManager::new();
-        let key = "same-key";
-
-        // First use key with u32
-        let v1 = mgr.run_single_flight(key, || async { 7u32 }).await.unwrap();
-        assert_eq!(v1, 7);
-
-        // Reuse same key with a *different* T (String) -> expect error
-        match mgr.run_single_flight::<_, _, String>(key, || async { "ok".to_string() }).await {
-            Err(MultithreadedRuntimeError::Other(msg)) => {
-                assert!(msg.contains("already registered with a different result type"), "unexpected error: {msg}");
-            },
-            other => panic!("expected type-mismatch error, got {other:?}"),
-        }
-    }
 }
--- a/xet_threadpool/src/simple_singleflight.rs
+++ b/xet_threadpool/src/simple_singleflight.rs
@@ -1,0 +1,286 @@
+use std::any::Any;
+use std::collections::HashMap;
+use std::future::Future;
+use std::sync::Arc;
+
+use tokio::sync::{oneshot, Mutex};
+use tracing::info;
+
+use crate::errors::MultithreadedRuntimeError;
+
+/// A simple single-flight group: concurrent calls to `run` will coalesce so that
+/// only one invocation of the retrieval function executes; all other callers
+/// await the same result.
+///
+/// - The first caller becomes the "leader", executes the retrieval, and then broadcasts the value to all "followers".
+/// - Followers wait on a oneshot to receive a clone of the produced value.
+///
+/// Notes:
+/// - If the retrieval future panics, followers will receive a canceled oneshot and this function will panic in
+///   followers with a clear message.
+/// - If you want error handling instead of panics, change `T` to `Result<U, E>` and broadcast the result.
+pub struct SingleFlightGroup<T: Clone + Send + Sync + 'static> {
+    state: Arc<Mutex<Option<Vec<oneshot::Sender<T>>>>>,
+}
+
+impl<T: Clone + Send + Sync + 'static> SingleFlightGroup<T> {
+    pub fn new() -> Self {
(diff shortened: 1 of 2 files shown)


xet_threadpool/src/simple_singleflight.rs
```diff
@@ -0,0 +1,253 @@
...
+/// only one invocation of the retrieval function executes; all other callers
+/// await the same result.
+///
+/// - The first caller becomes the "leader", executes the retrieval, and then broadcasts the value to all "followers".
+/// - Followers wait on a oneshot to receive a clone of the produced value.
+///
+/// Notes:
+/// - If the retrieval future panics, followers will receive a canceled oneshot and this function will panic in
+///   followers with a clear message.
+/// - If you want error handling instead of panics, change `T` to `Result<U, E>` and broadcast the result.
```
jgodlew, line 21: This requires that the error type `E` implement `Clone`, which is not the case for many of the places we use singleflight.

xet_threadpool/src/simple_singleflight.rs
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    /// Coalesces concurrent calls by `key`. A key must always be used with the same `T`
+    /// unless you `clear(key)` first. If a different `T` is used for an existing key,
+    /// this returns an error prompting a clear.
+    pub async fn run_single_flight<F, Fut, T>(
+        &self,
+        key: &str,
+        retrieval_function: F,
+    ) -> Result<T, MultithreadedRuntimeError>
```
jgodlew, line 108: Callers may (and in our case do) want to know if they were the leader/owner or follower/waiter (e.g. for monitoring purposes), so we need to return that back.

xet_threadpool/src/simple_singleflight.rs
```diff
@@ -0,0 +1,253 @@
...
+                    let msg = format!(
+                        "Singleflight leader failed before broadcasing result, possible shutdown in progress. {e:?}"
+                    );
+                    info!("{msg}");
+                    Err(MultithreadedRuntimeError::Other(msg))
+                },
+            }
+        } else {
+            // We're the leader: get and run the retrieval *without* holding the lock.
+            let value = retrieval_function().await;
```
jgodlew, line 74: If the leader panics or is cancelled here, then the followers will never be informed and will thus wait forever.

xet_threadpool/src/simple_singleflight.rs
```diff
@@ -0,0 +1,253 @@
...
+                }
+            } else {
+                let g: Arc<SingleFlightGroup<T>> = Arc::new(SingleFlightGroup::new());
+                map.insert(key.to_owned(), g.clone() as Arc<dyn Any + Send + Sync>);
+                g
+            }
+        };
+
+        // lock released; delegate
+        let res = group.run(retrieval_function).await;
```
jgodlew, line 135: If the retrieval_function is run as a part of the leader task, then we are susceptible to deadlock situations where a chain of dependencies occurs. This is a situation we ran into in the past, leading to us needing to spawn the leader as a new tokio task.

xet_threadpool/src/simple_singleflight.rs
```diff
@@ -0,0 +1,253 @@
...
+
+        // Now, clear out our key here; no need to keep a completed group around.
+        let _ = self.groups.lock().await.remove(key);
```
jgodlew, line 138: This is going to be done by every task (leader + followers) that runs. Ideally, the leader should be the one to remove this to minimize the contention.

xet_threadpool/src/simple_singleflight.rs
jgodlew: See singleflight's [test_deadlock](https://github.com/huggingface/xet-core/blob/main/utils/src/singleflight.rs#L662) for an example of how such a situation can arise.

xet_threadpool/src/simple_singleflight.rs
```diff
@@ -0,0 +1,253 @@
...
+///
+/// Notes:
+/// - If the retrieval future panics, followers will receive a canceled oneshot and this function will panic in
+///   followers with a clear message.
+/// - If you want error handling instead of panics, change `T` to `Result<U, E>` and broadcast the result.
+pub struct SingleFlightGroup<T: Clone + Send + Sync + 'static> {
+    state: Arc<Mutex<Option<Vec<oneshot::Sender<T>>>>>,
+}
+
+impl<T: Clone + Send + Sync + 'static> SingleFlightGroup<T> {
```
assafvayner, line 26: enforcing the `Clone` binding is iffy, I would like to do it, but it forces us to make all Result/Error types clone-able, which is not always available cleanly.

any aws/http errors tend to contain a streaming or at least non-clonable body, so we have to Arc wrap a bunch of errors possibly.

That being said, I'm on board with this

assafvayner (commented):
There's possibly a race condition that causes an issue (maybe/right?):

There's a leader (L) running, any number of waiters on the same key, and in the meanwhile a new call (N) to a SingleFlightManager.run_single_flight finds a group for the same key.

both L and N need to acquire the group lock, L when work is finished, N in entering SingleFlightGroup.run, L finishes the work right when N has SingleFlightGroup.run called and manages to acquire the lock first, it then sends all the results to the waiters and releases the lock. Now N adds itself to the waiters and never receives a finish signal, rather is should get an error when the oneshot sender is dropped.

PR #451, opened 2025-08-13
WIP more logging for debugging

assafvayner:
to get all the logs need to set these to 1

`HF_XET_LOG_PROGRESS_UPDATES` logs GIL status on progress updates
`HF_XET_LOG_TOKEN_REFRESH` logs GIL status on token refresh
`LOG_DURATION` logs duration of semaphore acquisition for download tasks, also logs if the task takes over 5 seconds
`LOG_QUEUE_STATUS` logs the size of download task queue and number of running download LOG_QUEUE_STATUS for every received task item after processing some all ready download

Not sure we want to merge this in the end


utils/src/auth.rs
```diff
@@ -89,7 +97,11 @@ impl TokenProvider {
     pub fn new(cfg: &AuthConfig) -> Self {
         Self {
             token: cfg.token.clone(),
-            expiration: cfg.token_expiration,
+            // expiration: cfg.token_expiration,
+            expiration: (SystemTime::now() + Duration::from_secs(REFRESH_BUFFER_SEC + 10))
```
assafvayner: hack, forcing token refresh often

PR #452, opened 2025-08-13
Cache clients with threadpool for recycling between calls to download_files

hoytak:
This PR allows the different clients to be cached between calls to download_files, instead of each one creating a new client on the fly.  Because each client keeps connections open after use for the keep alive time, this can cause an explosion in the number of open file handles even though most of the connections aren't being used due to global constraints on the number of parallel downloads.  

To enable this, it also adds in a caching layer to the xet runtime that allows arbitrary values to be cached with the runtime; sorta a OnceLock type of mechanism but not using statics.

### DIFF
--- a/cas_client/src/http_client.rs
+++ b/cas_client/src/http_client.rs
@@ -1,6 +1,7 @@
 #[cfg(not(target_family = "wasm"))]
 mod dns_utils;
 
+use std::fmt::Debug;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -27,6 +28,7 @@ pub(crate) const BASE_RETRY_DELAY_MS: u64 = 3000; // 3s
 pub(crate) const BASE_RETRY_MAX_DURATION_MS: u64 = 6 * 60 * 1000; // 6m
 
 /// A strategy that doesn't retry on 429, and defaults to `DefaultRetryableStrategy` otherwise.
+#[derive(Debug)]
 pub struct No429RetryStrategy;
 
 impl RetryableStrategy for No429RetryStrategy {
@@ -42,6 +44,7 @@ impl RetryableStrategy for No429RetryStrategy {
     }
 }
 
+#[derive(Debug)]
 pub struct RetryConfig<R: RetryableStrategy> {
     /// Number of retries for transient errors.
     pub num_retries: u32,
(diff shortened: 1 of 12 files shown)


seanses (changes requested):
`session_id` is unique for each call to download_files so nothing is actually cached.

jgodlew (commented):
~We may have an issue with caching the CAS clients due to the auth likely being different across calls to hf_xet upload/download. For example, when downloading, we get read-access tokens, but if someone (in the same python program) uploads after downloading, they need write-access tokens.

Along a similar (but much less important) vein, this will group all requests across hf_xet calls within the same `session_id`. IIRC, we currently treat the session_id as per-hf_xet call.~

Nevermind, I didn't see that the key name incorporates the auth config and session_id. Also, I didn't see Di's comments, but agree with them.

hoytak:
> How about we cache only the reqwest Client? I think that caches and shares the connection pool

Great idea, I like that.  Changed.

hoytak:
Good catches, should be fixed now.

assafvayner:
I think this caching system is a bit overkill for this use case. So a reqwest::Client is just cheaply clonable wrapper over a hyper legacy client (also cheaply clonable) which maintains a connection pool.

I would prefer if we just keep 1 single reqwest::Client and then we can clone it freely and wrap middlewares as needed, but I don't think we need a more complex dynamic cache.

something as dumb as a static OnceLock in the reqwest_client function or in the threadpool if we want to destroy it on the threadpool recreation.

That all being said, I recall their being an issue with connection recycling down to the hyper_util legacy client which is what reqwest is using. Will need to dig more into this.

cas_client/src/http_client.rs
```diff
@@ -79,30 +79,43 @@ impl RetryConfig<No429RetryStrategy> {
...
 /// Builds authenticated HTTP Client to talk to CAS.
 /// Includes retry middleware with exponential backoff.
-pub fn build_auth_http_client<R: RetryableStrategy + Send + Sync + 'static>(
+pub async fn build_auth_http_client<R: RetryableStrategy + Send + Sync + 'static>(
     auth_config: &Option<AuthConfig>,
     retry_config: RetryConfig<R>,
     session_id: &str,
+    cache_tag: Option<&str>,
 ) -> Result<ClientWithMiddleware, CasClientError> {
+    let cache_tag = cache_tag.map(|tag| format!("{tag}::{auth_config:?}"));
```
seanses, line 112: I don't think we need to use different reqwest client for different tag & auth_config? The `auth_config` is eventually incorporated into a middleware layer added on top of the reqwest client. It seems to me that using a single reqwest client for all wrapped clients is enough.

seanses:
Replaced by #457

PR #453, opened 2025-08-14
Raise soft file handle limits to hard limits on OSX. 

hoytak:
On OSX, raise the soft file handle limits to the hard limits (which cannot be changed in the process).

### DIFF
--- a/xet_threadpool/Cargo.toml
+++ b/xet_threadpool/Cargo.toml
@@ -15,5 +15,5 @@ oneshot = { workspace = true }
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
 tokio = { workspace = true, features = ["rt-multi-thread"] }
 
-[target.'cfg(any(target_os = "macos", target_os = "linux"))'.dependencies]
-libc = "0.2"
+[target.'cfg(target_os = "macos")'.dependencies]
+libc = { workspace = true}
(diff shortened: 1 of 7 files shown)


xet_threadpool/src/file_handle_limits.rs
```diff
@@ -0,0 +1,39 @@
+use tracing::info;
+
+#[cfg(any(target_os = "macos", target_os = "linux"))]
+pub fn raise_nofile_soft_to_hard() {
```
rajatarya, line 2: Do we want to wrap this in some error handling? How does Rust handle errors in unsafe blocks?

assafvayner: This all looks fine in terms of error handling since we shouldn't fail if this function fails. Any time there is an error it is logged and we return silently back to the caller.

Rust doesn't do anything special in unsafe blocks in regards to error handling.

xet_threadpool/Cargo.toml
```diff
@@ -14,3 +14,6 @@ oneshot = { workspace = true }
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
 tokio = { workspace = true, features = ["rt-multi-thread"] }
+
+[target.'cfg(any(target_os = "macos", target_os = "linux"))'.dependencies]
```
seanses:
```suggestion
[target.'cfg(unix)'.dependencies]
```

xet_threadpool/Cargo.toml
```diff
@@ -14,3 +14,6 @@ oneshot = { workspace = true }
...
+libc = "0.2"
```
seanses:
```suggestion
libc = { workspace = true }
```

seanses: And add this to the workspace Cargo.toml

seanses (approved):

PR #454, opened 2025-08-14
parutils makeover remove async_scoped

assafvayner:
removing async_scoped dep and creating new parallel util to replace tokio_par_for_each that relied on async_scoped.

Any usage of tokio_par_for_each which is the only fn used out of parutils has been replaced with the new `tokio_run_max_concurrency_fold_result_with_semaphore`

TODO:
- [ ] add more tests
- [ ] use semaphore acquired from the global semaphore provider where/if relevant.


data/src/file_upload_session.rs
```diff
@@ -36,6 +37,11 @@ use crate::remote_client_interface::create_remote_client;
 use crate::shard_interface::SessionShardInterface;
 use crate::{prometheus_metrics, XetFileInfo};
 
+lazy_static! {
+    pub static ref CONCURRENT_FILE_INGESTION_LIMITER: GlobalSemaphoreHandle =
```
hoytak, line 41: So static attributes are the thing that's bad for fork safety, but the global_semaphore_handle! is cheap to call on construction and returns the exact same one each time.

data/src/data_client.rs
```diff
@@ -10,23 +10,30 @@ use cas_client::{CacheConfig, FileProvider, OutputProvider, CHUNK_CACHE_SIZE_BYT
...
 
 utils::configurable_constants! {
     ref DEFAULT_CAS_ENDPOINT: String = "http://localhost:8080".to_string();
 }
 
+lazy_static! {
+    static ref DOWNLOAD_FILE_CONCURRENCY_LIMITER: GlobalSemaphoreHandle =
+        global_semaphore_handle!(*MAX_CONCURRENT_DOWNLOADS);
+}
+
```
hoytak: Static semaphores are what's bad for fork safety, so the global_semaphore_handle! is a cheap call that returns the same runtime-global semaphore when called on construction.

parutils/src/lib.rs
```diff
@@ -1,4 +1,185 @@
...
+use std::sync::Arc;
 
-mod parallel_utils;
-pub use parallel_utils::*;
+use thiserror::Error;
+use tokio::sync::{AcquireError, Semaphore};
+use tokio::task::{JoinError, JoinSet};
+
+#[derive(Debug, Error)]
+pub enum ParutilsError<E: Send + Sync + 'static> {
```
hoytak: Feel free to move this to a util in xet_threadpool and then repurpose the MultithreadedRuntimeError class

parutils/src/lib.rs
```diff
@@ -1,4 +1,185 @@
...
+/// Like tokio_run_max_concurrency_fold_result_with_semaphore but does not require
+/// that the futures in the iterator parameter produce a Result. The futures may produce any type
+/// that will be returned.
+///
+/// tokio_run_max_concurrency_fold_result_with_semaphore returns an error as soon as any task returns
+/// an error.
+///
+/// This function can be used to ensure that all tasks are finished and returns Vec<Result<>>
+/// if the futures passed in still return results.
+pub async fn tokio_run_max_concurrency_with_semaphore<Fut, T>(
```
jgodlew: nit: I am not huge on having these variants of the function (i.e. Fut doesn't return a Result) since they are unused. I would rather add them when needed instead.

parutils/src/lib.rs
```diff
@@ -1,4 +1,185 @@
...
+///     let semaphore = Arc::new(Semaphore::new(2)); // Limit concurrency to 2 tasks.
+///     let futures = (1..=3).map(|n| async move { Ok::<_, ()>(n) });
+///
+///     let results =
+///         tokio_run_max_concurrency_fold_result_with_semaphore(futures.into_iter(), semaphore)
+///             .await;
+///     assert_eq!(results.unwrap(), vec![1, 2, 3]);
+/// }
+/// ```
+pub async fn tokio_run_max_concurrency_fold_result_with_semaphore<Fut, T, E>(
```
jgodlew: nit: IMO, the `tokio_run_max_concurrency*` name is a bit lengthy. We could probably rename to just `run_constrained_*` instead?

parutils/src/lib.rs
```diff
@@ -1,4 +1,185 @@
...
+    Join(#[from] JoinError),
+    #[error(transparent)]
+    Acquire(#[from] AcquireError),
+    #[error(transparent)]
+    Task(E),
+}
+
+/// Runs all the futures in the provided iterator with a maximum concurrency limit.
+///
+/// This function requires that a permit is acquired from the provided semaphore before any work
```
jgodlew:
```suggestion
/// This function ensures that a permit is acquired from the provided semaphore before any work
```

jgodlew (commented):
Overall, looks good. We might want to consider removing the parutils crate and moving these helper functions to the xet_threadpool crate.

jgodlew (approved):

PR #455, opened 2025-08-14
Add back retry for connection setup and sending request

seanses:
This PR fixes the regression that part of the retry logic for downloading was accidentally removed. The added back retry logic complements the retry for deserializing the data stream of responses.
Fix XET-696

### DIFF
--- a/cas_client/src/http_client.rs
+++ b/cas_client/src/http_client.rs
@@ -98,7 +98,7 @@ pub fn build_auth_http_client<R: RetryableStrategy + Send + Sync + 'static>(
     retry_config: RetryConfig<R>,
     session_id: &str,
 ) -> Result<ClientWithMiddleware, CasClientError> {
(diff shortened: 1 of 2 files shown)


cas_client/src/http_client.rs
```diff
@@ -98,7 +98,7 @@ pub fn build_auth_http_client<R: RetryableStrategy + Send + Sync + 'static>(
     retry_config: RetryConfig<R>,
     session_id: &str,
 ) -> Result<ClientWithMiddleware, CasClientError> {
-    let auth_middleware = auth_config.as_ref().map(AuthMiddleware::from).info_none("CAS auth disabled");
+    let auth_middleware = auth_config.as_ref().map(AuthMiddleware::from);
```
seanses, line 101: This emits error messages which shouldn't be there at all, because apparently `build_http_client` calls this function which auth isn't supposed to be enabled.

assafvayner (approved):

PR #457, opened 2025-08-15
Cache and reuse reqwest Client

seanses:
This PR caches the reqwest::Client in a runtime if it exists and re-uses it in all wrapped clients in the `RemoteClient` object. This effectively shares the connection pool and thus reduces opening sockets.

Fix XET-704

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4393,6 +4393,7 @@ version = "0.1.0"
 dependencies = [
  "error_printer",
  "oneshot",
+ "reqwest",
  "thiserror 2.0.12",
  "tokio",
  "tracing",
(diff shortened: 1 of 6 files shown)


seanses:
This seems to effectively reduces the number of opened sockets and unblocks Github CI running on `macos-latest` that downloads files using hf-xet: https://github.com/huggingface/xet-scenario-tests/actions/runs/16999971729

seanses:
This PR aims to replace https://github.com/huggingface/xet-core/pull/452

assafvayner (approved):

jgodlew (approved):

PR #458, opened 2025-08-15
tokio console setup

assafvayner:


hf_xet/README.md
```diff
@@ -2,3 +2,28 @@
...
+```
+
+Then while hf_xet is running (via a `hf` cli command or `huggingface_hub` python code), `tokio-console` will be able to connect.
+
+### Ex.
+
+```bash
+# In one terminal:
+pip install huggingface_hub
+RUSTFLAGS="--cfg tokio_unstable" maturin develop -r --features tokio-console
```
jgodlew, line 23: Is it fine to just have this flag in the .cargo config? AFAICT, it doesn't really impact anything, but it makes this env unnecessary.

assafvayner: 😬 I am not sure like it to be on by default, if not for anything functional, just because of optics.

I tried cajoling it to have it such that if this feature is enabled then set this RUSTFLAGS value. This allegedly works with a plain `cargo build` but not with `maturin` based on how `maturin` builds the RUSTFLAGS variable. May need to open an issue against pyo3/maturin

PR #460, opened 2025-08-18
run_and_extract_custom: remove use of explicit tokio_retry without utility

assafvayner:
we had 1 case of using "raw" tokio_retry rather than the retry utility. This was due to using a special custom parsing logic for chunks, rather than built in json functionality. This PR adds a run_and_extract custom that let's a user specify the function to parse the response body.

### DIFF
--- a/cas_client/src/remote_client.rs
+++ b/cas_client/src/remote_client.rs
@@ -249,10 +249,7 @@ impl RemoteClient {
             .run(move || client.get(url.clone()).with_extension(Api(api_tag)).send())
             .await;
 
-        if result
-            .as_ref()
-            .is_err_and(|e| e.status().is_some_and(|status| status == StatusCode::NOT_FOUND))
-        {
+        if result.as_ref().is_err_and(|e| e.status().is_some()) {
             return Ok(None);
         }
         Ok(Some(result?))
--- a/cas_client/src/remote_client.rs
+++ b/cas_client/src/remote_client.rs
@@ -243,16 +243,19 @@ impl RemoteClient {
         let client = self.authenticated_http_client.clone();
         let api_tag = "cas::query_dedup";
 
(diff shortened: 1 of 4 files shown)


cas_client/src/retry_wrapper.rs
```diff
@@ -107,22 +106,20 @@ impl RetryWrapper {
...
-            CasClientError::ServerConnectionError(msg)
+            CasClientError::from(err)
         };
 
         let retriability = default_on_request_success(&resp);
 
         match (resp.error_for_status(), retriability) {
             (Err(e), Some(Retryable::Fatal)) => {
-                let cas_err = process_error("Fatal Server Error", e, false);
+                let cas_err = process_error("Fatal Error", e, false);
```
seanses, line 122: We received a valid response from the server, so it should be ok to say it's a "Fatal Server Error" right?

cas_client/src/remote_client.rs
```diff
@@ -243,16 +243,19 @@ impl RemoteClient {
...
 
-        let response = RetryWrapper::new(api_tag)
+        let result = RetryWrapper::new(api_tag)
             .with_429_no_retry()
             .log_errors_as_info()
             .run(move || client.get(url.clone()).with_extension(Api(api_tag)).send())
             .await;
 
-        if matches!(response, Err(CasClientError::ServerConnectionError(_))) {
+        if result
```
seanses: We don't want to propagate the error, especially due to 429, since this is a best effort query.

cas_client/src/retry_wrapper.rs
assafvayner: I think Server Error here implies a 5xx error code, which is not correct here, this is just an error from the response. The actual error message will be included where "Fatal Error" string is included.

seanses (approved):
LGTM!

PR #466, opened 2025-08-21
Add ppc64le target

mgiessing:
This PR adds `ppc64le` CI build

- Extend matrix with ppc64le (manylinux_2_28).

- Install binutils-powerpc64le-linux-gnu and alias objcopy accordingly.

- Keep existing aarch64 handling; create hf_xet/dbg across targets.

- Enables publishing manylinux wheels for ppc64le without impacting other targets.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,6 +33,10 @@ jobs:
             target: aarch64
             python-target: arm64
             manylinux: manylinux_2_28
+          - runner: ubuntu-22.04
+            target: ppc64le
+            python-target: ppc64le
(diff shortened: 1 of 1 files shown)


rajatarya:
Hi @mgiessing : I do not believe GitHub offers ppc64le runners - so even if we add this build variant to the Release Workflow, it won't be able to run anywhere.

Does GitHub offer ppc64le runners or do they have to be self-hosted? We don't have any self-hosted runners right now for `hf-xet` builds.

rajatarya:
Closing because no response, re-open if GitHub supports running on this machine architecture.

PR #468, opened 2025-08-22
Adding retry for unhandled io errors when sending requests

jgodlew:
Allows us to retry errors when we receive I/O errors when sending requests. For example, on macOS, when we see:
```
No buffer space available (os error 55)
```
when downloading from S3, we should wait and retry once the system has more network resources.

### DIFF
--- a/cas_client/src/retry_wrapper.rs
+++ b/cas_client/src/retry_wrapper.rs
@@ -323,7 +323,7 @@ impl RetryWrapper {
 ///
 /// Unfortunately, those errors don't translate to a defined [std::io::ErrorKind], and are thus
 /// not able to be effectively filtered.
-fn on_request_failure(error: &reqwest_middleware::Error) -> Option<Retryable> {
+pub fn on_request_failure(error: &reqwest_middleware::Error) -> Option<Retryable> {
     let reqwest_middleware::Error::Reqwest(error) = error else {
         // If something fails in the middleware we're screwed.
         return Some(Retryable::Fatal);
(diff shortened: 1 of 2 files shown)


seanses:
I just ran a test on this PR, it seems that the "os error 55" is not caught or retried: https://github.com/huggingface/xet-scenario-tests/actions/runs/17216048201/job/48839818576#step:13:196

cas_client/src/retry_wrapper.rs
```diff
@@ -317,6 +317,75 @@ impl RetryWrapper {
...
+/// Like [request_middleware::default_on_request_failure], but retries all IOErrors instead of a
+/// subset. There are a few errors that can occur on certain systems that we will want to retry
+/// (e.g. `No buffer space available: (os error 55)`).
+///
+/// Unfortunately, those errors don't translate to a defined [std::io::ErrorKind], and are thus
+/// not able to be effectively filtered.
+fn on_request_failure(error: &reqwest_middleware::Error) -> Option<Retryable> {
+    match error {
+        // If something fails in the middleware we're screwed.
+        reqwest_middleware::Error::Middleware(_) => Some(Retryable::Fatal),
```
assafvayner: might as well de-indent by doing:
```
if let reqwest_middleware::Error::Reqwest(error) =  error else {
    return Some(Retryable::Fatal);
};

the rest
```

assafvayner (approved):

jgodlew:
@seanses the errors are now being caught and retried properly: [ci-run](https://github.com/huggingface/xet-scenario-tests/actions/runs/17622222674/job/50070359884#step:13:209). Downloading the logs from that run shows that there are `Retry attempt #...` logs being printed and no `ERROR`s. 

However, in testing, I did notice that [sometimes](https://github.com/huggingface/xet-scenario-tests/actions/runs/17597085880/job/49991683004#step:13:202), retrying 5 times was insufficient. Lowering the number of concurrent range gets (`export HF_XET_NUM_CONCURRENT_RANGE_GETS=32`) should help reduce the number of connections established, reducing the likelihood of running out of available mbufs on macOS (likely source of the os error 55 / ENOBUFS).

PR #471, opened 2025-08-25
Add lz4 compression/decompression to thin wasm

coyotte508:
Investigating an issue where I think the `lz4` package used for node is the culprit

thin wasm goes from 100kB => 130kB


coyotte508:
nvm the issue was I set the compression flag wrongly

PR #473, opened 2025-08-25
CI test on macos

seanses:
We test on Ubuntu and Windows, so it seems reasonable to test on macOS too. This also gets CI prepared for git-xet tests.


assafvayner (commented):
optionally remove the lint step on macos, I don't think we need lint on more than 1 system

assafvayner (approved):

rajatarya (approved):

PR #474, opened 2025-08-26
Enable socks5 proxy support

SuperKenVery:


rajatarya:
Hi @SuperKenVery : Can you describe the use case for adding socks5 proxy support? Can you add some tests for this use case in this PR as well?

seanses:
@rajatarya AFAIK some VPN services use SOCKS5 tunnel.

assafvayner:
So I see enabling this feature adds proxy support, but we will need to add some docs on how to use proxy by environment variable. (we don't need to setup our own special way to handle proxies then, at least not yet)

https://github.com/seanmonstar/reqwest/blob/v0.12.23/src/lib.rs#L137-L158

SuperKenVery:
I should definitely add some tests to this, as well as adding some docs.

The reason for socks5 support is that, my school server doesn't have an internet access. And socks5 is more convenient than http(s) proxy in that I could use `ssh -R 1080 host` to setup a socks proxy server.

hoytak:
@rajatarya @assafvayner:  Looking at this PR, this seems like a low risk change and we should just get it in.  Note that it is simply enabling a feature flag in the reqwests crate. 
 
 To address the concerns: 
- Proxy support is enabled through standard environment variables that are well documented and independent of any specific program, so we don't need to document them ourselves.  See the docs on HTTP_PROXY or HTTPS_PROXY.  
- Tests would already be present in the reqwest layer, so I don't think we need to add specific tests for this.

hoytak:
@SuperKenVery -- just verifying here.  When you enabled that feature flag in reqwest, were you able to get the connection to work in your environment?

SuperKenVery:
Yes, it's able to use socks5 proxy in my environment.

Given that you've address the concerns, I'm converting this from draft to pr...

hoytak:
> Yes, it's able to use socks5 proxy in my environment.
> 
> Given that you've address the concerns, I'm converting this from draft to pr...

Thank you.  Before we merge it, could you edit the description of the PR to include this information and how you tested it.  Also, I think we need to check in the Cargo.lock that have changed as well.  If you could do the following and then check in the changed lock files, that would be great: 

- Run cargo build in the main directory.
- In hf_xet/, run `maturin develop`.
- In hf_xet_wasm, run `./build_wasm.sh`

SuperKenVery:
Okay I'll run those commands and edit PR description. I previously tested by doing an editable install in pip.

---
EDIT: I checked `Cargo.lock`, it seems it's indeed unchanged. This is further verified by this: https://asciinema.org/a/2QPqTXaBvliqsKHAVkns64nPD

---
EDIT2: I ran the commands you mentioned, and they ran without problems. 🥂

PR #481, opened 2025-09-02
Thin wasm: do not automatically set is_dedup to true for first chunk

coyotte508:
Related to https://github.com/huggingface/huggingface.js/pull/1718

We'll want to edit parts of file while loading old data's dedup info

In those case we don't always want to load dedup info for the first chunk (since it may not be at the beginning of the file)

### DIFF
--- a/hf_xet_thin_wasm/src/lib.rs
+++ b/hf_xet_thin_wasm/src/lib.rs
@@ -28,12 +28,12 @@ pub struct JsChunkOut {
 
 
 impl JsChunkOut {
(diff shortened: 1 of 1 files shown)


assafvayner:
> So the is_dedup = true for first chunk is handled client side

Best to ensure this is still done, the best chance to dedupe large parts of files is on first chunk.

coyotte508:
it is: 

https://github.com/huggingface/huggingface.js/blob/36d9d94ed867ba11d387f5608edf12992a61046c/packages/hub/src/utils/createXorbs.ts#L187-L192

PR #482, opened 2025-09-05
update api paths to use plural nouns

assafvayner:
Updates paths used by the clients to use latest CAS paths as defined in the spec.

All paths now use plural nouns and shard upload no longer uses the hash, removes the prefix and hash from the client trait upload_shard function.

### DIFF
--- a/hf_xet_wasm/Cargo.toml
+++ b/hf_xet_wasm/Cargo.toml
@@ -39,7 +39,7 @@ tokio_with_wasm = { version = "0.8.2", features = ["rt"] }
 tokio-stream = "0.1.17"
 uuid = { version = "1", features = ["v4", "js"] }
 wasm_thread = "0.3"
-wasm-bindgen = "0.2.100"
+wasm-bindgen = "=0.2.100"
 wasm-bindgen-futures = "0.4.50"
 web-sys = { version = "0.3.72", features = [
     "File",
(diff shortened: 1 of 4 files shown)


cas_client/src/download_utils.rs
```diff
@@ -613,7 +613,7 @@ mod tests {
         let server = MockServer::start();
         server.mock(|when, then| {
             when.method(GET)
-                .path(format!("/reconstruction/{}", MerkleHash::default()))
+                .path(format!("/reconstructions/{}", MerkleHash::default()))
```
rajatarya, line 616: Don't you have to keep both for backcompat?

assafvayner: On the server we have both and test both, on the client I'm only testing the path the client actually uses. This is mocking the CAS server such that the path the client uses is served.

hf_xet_wasm/Cargo.toml
```diff
@@ -39,7 +39,7 @@ tokio_with_wasm = { version = "0.8.2", features = ["rt"] }
 tokio-stream = "0.1.17"
 uuid = { version = "1", features = ["v4", "js"] }
 wasm_thread = "0.3"
-wasm-bindgen = "0.2.100"
+wasm-bindgen = "=0.2.100"
```
seanses, line 42: Why pin the version? What change breaks the build?

assafvayner: there was a new version of wasm-bindgen and wasm-bindgen-cli released. We need to keep them matched, in CI and locally.

I chose to pin at the version we've been using rather than upgrading everywhere.

We need `=` pin since otherwise cargo treats it as `^` and tries to bump.

seanses: I see.SG

PR #485, opened 2025-09-08
use u64 rather than usize in file hashing paths

assafvayner:
Using the file hashing components in WASM found a bug that using 32 but usize causes errors when hashing the file.

This PR enforces the use of u64 everywhere along that path (and also pins the wasm-bindgen version)

### DIFF
--- a/cas_client/src/download_utils.rs
+++ b/cas_client/src/download_utils.rs
@@ -613,7 +613,7 @@ mod tests {
         let server = MockServer::start();
         server.mock(|when, then| {
             when.method(GET)
-                .path(format!("/reconstruction/{}", MerkleHash::default()))
+                .path(format!("/reconstructions/{}", MerkleHash::default()))
                 .header(RANGE.as_str(), HttpRange::from(file_range).range_header());
             let response = QueryReconstructionResponse {
                 offset_into_first_range: 0,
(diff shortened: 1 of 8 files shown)


seanses:
Our chunk size can never exceed u32::MAX because there's a upper limit at 128 KiB, why are these causing problems?

assafvayner:
> Our chunk size can never exceed u32::MAX because there's a upper limit at 128 KiB, why are these causing problems?

The issue isn't with the chunk size, it's with the sizes of "segment" cuts made in the aggregated node hash computation. I went the simple route and just changed everything to u64 rather than only the particular piece. Is this a non-negotiable? we can keep u32 around with some extra work to convert when it matters.

hf_xet_thin_wasm/src/lib.rs
```diff
@@ -2,15 +2,16 @@ use merklehash::{DataHashHexParseError, MerkleHash, xorb_hash};
...
-#[wasm_bindgen]
-extern "C" {
-    #[wasm_bindgen(js_namespace = console)]
-    fn log(s: &str);
-}
+// #[wasm_bindgen]
+// extern "C" {
+//     #[wasm_bindgen(js_namespace = console)]
+//     fn log(s: &str);
+// }
```
hoytak, line 14: Should this be commented?

hoytak:
LGTM

hf_xet_thin_wasm/src/lib.rs
assafvayner: yes, we do not log anything in the thin wasm right now.

It is extern "C" but really this is importing JS functionality

PR #487, opened 2025-09-09
Updated chunker to eliminate spurious boundary triggering.

hoytak:
We must enforce that the next boundary is actually past the minimum chunk size.   In rare cases, a boundary could be triggered before the minimum chunk size has passed, and this triggering would be based not on the content of the file but on the previous state of the rolling hash function.  This PR fixes this case.

### DIFF
--- a/deduplication/src/chunking.rs
+++ b/deduplication/src/chunking.rs
@@ -138,6 +138,14 @@ impl Chunker {
         }
     }
 
+    fn reset_state(&mut self) {
+        // Strictly speaking, this is unneccesary, as we should always hash 64 bytes out making the previous state
+        // of the hasher irrelevant.  However, this explicitly declares we're resetting things to the
+        // initial state.
+        self.hash.set_hash(0);
+        debug_assert!(self.chunkbuf.is_empty());
+    }
+
     /// Process more data; this is a continuation of any data from before when calls were
     ///
     /// Returns the next chunk, if available, and the amount of data that was digested.
@@ -145,7 +153,7 @@ impl Chunker {
     /// If is_final is true, then it is assumed that no more data after this block will come,
     /// and any data currently present and at the end will be put into a final chunk.
     pub fn next(&mut self, data: &[u8], is_final: bool) -> (Option<Chunk>, usize) {
-        let (data, consume): (Bytes, usize) = {
+        let (chunk_data, consume): (Bytes, usize) = {
             if let Some(next_boundary) = self.next_boundary(data) {
                 if self.chunkbuf.is_empty() {
                     (Bytes::copy_from_slice(&data[..next_boundary]), next_boundary)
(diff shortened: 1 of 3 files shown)


deduplication/src/chunking.rs
```diff
@@ -98,12 +98,26 @@ impl Chunker {
...
 
         let mut bytes_to_next_boundary;
-        if let Some(boundary) = self.hash.next_match(&data[consume_len..read_end], self.mask) {
-            // If we trigger a boundary before the end, create a chunk.
-            bytes_to_next_boundary = boundary;
-            create_chunk = true;
-        } else {
-            bytes_to_next_boundary = read_end - consume_len;
+        loop {
+            if let Some(boundary) = self.hash.next_match(&data[consume_len..read_end], self.mask) {
```
seanses: This looks buggy: the `consume_len` is never updated in the loop. So if `next_match` finds a boundary but `boundary + cur_chunk_len < self.minimum_chunk`, the hasher is run over the same content again

seanses: So this can run indefinitely

seanses (changes requested):
Bug

deduplication/src/chunking.rs
hoytak: Ah, yes, good catch.

assafvayner (approved):
LGTM

deduplication/src/chunking.rs
```diff
@@ -80,43 +81,58 @@ impl Chunker {
...
-            cur_chunk_len += max_advance;
+        if previous_len + HASH_WINDOW_SIZE < self.minimum_chunk {
+            let skip = min(self.minimum_chunk - previous_len - HASH_WINDOW_SIZE - 1, n_bytes);
+            cur_index += skip;
         }
 
         // If we have a lot of data, don't read all the way to the end when we'll stop reading
         // at the maximum chunk boundary.
-        let read_end = n_bytes.min(consume_len + self.maximum_chunk - cur_chunk_len);
+        let read_end = n_bytes.min(cur_index + self.maximum_chunk - previous_len);
```
seanses, line 98: The number `cur_index + self.maximum_chunk - previous_len` doesn't seem to be identical to the previous value `consume_len + self.maximum_chunk - cur_chunk_len`. This result seems to pass the max chunk size boundary.

seanses: But I guess this doesn't impact the below logic to cut at the max chunk size

seanses (commented):
Optionally, also explicitly reset the hash when is_final is true and document for the finish() function that the chunker can be reused.

seanses (approved):
This leads to dedup loss against existing Xet data. But as we expect more data in the future and to conform to the OSS spec, this is fine.

Experiments against existing models (obtained using xtool dedup command):

dedup loss

deduplication/src/chunking.rs
hoytak: It should be the same -- those variables were mainly renamed.

seanses: See the above if statement, previously we have both `consume_len += max_advance` and `cur_chunk_len += max_advance;` but now only `cur_index += skip;`

hoytak: Ah, true...  I think that was actually in error too.  Which would cause even more spurious results, as when there is data in the buffer we start chunking earlier, allowing even smaller chunks.

PR #492, opened 2025-09-15
Adding README to few crates for documentation

assafvayner:
Added README to a few crates so that we can link to the crate directory to link to individual crates and have something rendered for reference implementation when linking to a specific file doesn't make sense.


assafvayner:
~~re: Cargo.lock, I rebuilt to check all Cargo.lock's are updated after merging main in too. The wasm package lock file must have lagged behind some change.~~

nevermind it was fixed, I was behind.

PR #494, opened 2025-09-16
upgrade rust edition to 2024; upgrade rustc to 1.89

seanses:
- Upgrade Rust edition and rustc version to bring in some nice features, e.g. let chains instead of nested if block.
- Fix clippy and format due to the upgrade.
- Fix a bug identified by the new rustc: https://github.com/huggingface/xet-core/blob/6cb0a7fb4e4ca01daebddc0058241590c33ebbc8/xet_runtime/src/runtime.rs#L195
```
{
    // A new multithreaded runtime with a capped number of threads
    TokioRuntimeBuilder::new_multi_thread().worker_threads(get_num_tokio_worker_threads())
}
```
here the end curly bracket drops the temporary builder while a `&mut Self` to the dropped value is returned. (this may be due to a difference between compilers regarding how they treat the scope of "{...}" of `#[cfg(...))] {...}`?)

### DIFF
--- a/hf_xet/src/logging.rs
+++ b/hf_xet/src/logging.rs
@@ -86,12 +86,11 @@ fn get_version_info_string(py: Python<'_>) -> String {
     let mut version_info = String::new();
 
     // Get Python version
-    if let Ok(sys) = py.import("sys") {
-        if let Ok(version) = sys.getattr("version").and_then(|v| v.extract::<String>()) {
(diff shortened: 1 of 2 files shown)


assafvayner (approved):

sirahd (approved):
🚀

hoytak (approved):
LGTM.

PR #495, opened 2025-09-16
Allow Duration and byte sizes in constants for easier use.

hoytak:
- Duration: Currently, we use a lot of _MS and _SEC suffixes in constants to denote duration.  This PR allows std::time::Duration to be used directly, with values such as "10sec" or "100ms" or "1d" translated directly into std::time::Duration. 

- ByteSize: It also introduces a new utility type, ByteSize, that simply wraps a u64 but allows the user to specify "1mb" or "45gb" as the value when setting constant values.  The suffixes mb, mib, kb, kib, gb, gib, b, etc. are all supported, with the default being the raw value.

### DIFF
--- a/hf_xet/Cargo.lock
+++ b/hf_xet/Cargo.lock
@@ -878,6 +878,20 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7454e41ff9012c00d53cf7f475c5e3afa3b91b7c90568495495e8d9bf47a1055"
 
 [[package]]
+name = "duration-str"
+version = "0.17.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9add086174f60bcbcfde7175e71dcfd99da24dfd12f611d0faf74f4f26e15a06"
+dependencies = [
+ "chrono",
+ "rust_decimal",
+ "serde",
+ "thiserror 2.0.15",
+ "time",
+ "winnow",
+]
+
+[[package]]
 name = "either"
 version = "1.15.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


mdb_shard/src/constants.rs
```diff
@@ -17,11 +21,11 @@ utils::configurable_constants! {
     ///
     /// Note the cache is pruned to below this value at the beginning of a session,
     /// but during a single session new shards may be added such that this limit is exceeded.
-    ref SHARD_CACHE_SIZE_LIMIT : u64 = 16_000_000_000;
+    ref SHARD_CACHE_SIZE_LIMIT : ByteSize = ByteSize::from("10gb");
 
     /// The amount of time a shard should be expired by before it's deleted, in seconds.
     /// By default set to 7 days.
-    ref MDB_SHARD_EXPIRATION_BUFFER_SECS: u64 = 7 * 24 * 3600;
+    ref MDB_SHARD_EXPIRATION_BUFFER: Duration = Duration::from_secs(7 * 24 * 3600);
```
seanses, line 28: `Duration::from_days(7)`

mdb_shard/src/constants.rs
```diff
@@ -17,11 +21,11 @@ utils::configurable_constants! {
     ///
     /// Note the cache is pruned to below this value at the beginning of a session,
     /// but during a single session new shards may be added such that this limit is exceeded.
-    ref SHARD_CACHE_SIZE_LIMIT : u64 = 16_000_000_000;
+    ref SHARD_CACHE_SIZE_LIMIT : ByteSize = ByteSize::from("10gb");
```
seanses: Is this a value in any of our public doc? If so better not change (or update the doc)

Cargo.toml
```diff
@@ -48,6 +48,7 @@ csv = "1"
 ctor = "0.4"
 derivative = "2.2"
 dirs = "5.0"
+duration-str = "0.17.0"
```
seanses: It seems the convention is to ignore the patch number, so `duration-str = "0.17"`

mdb_shard/src/constants.rs
hoytak: Unfortunately, that's not in stable rust yet.

mdb_shard/src/constants.rs
hoytak: Ooops, I meant to keep the same value; it should still be 16gb.

mdb_shard/src/constants.rs
seanses: Oh my.. Didn't realize that they haven't stabilized this useful constructor.

PR #497, opened 2025-09-17
MacOS diag scripts

rajatarya:
Adds corresponding Diagnostics script for MacOS.

Lightly tested with hf-xet 1.1.10 and MacOS 15.6.1 - correctly takes stack traces on interval and writes out to diagnostics folder as expected.


jsulz (approved):
Nice! Works for me as well on MacOS 14.5. Did not test on latest MacOS - need to find a good time to walk away from my computer for an hour 😅

rajatarya:
> Nice! Works for me as well on MacOS 14.5. Did not test on latest MacOS - need to find a good time to walk away from my computer for an hour 😅

I tested on MacOS Sequoia 15.6.1 (24G90) so that is a newer one. My computer is now asking me to update to latest MacOS (renumbered, now 26.0 - sigh.)

PR #498, opened 2025-09-17
use refresh route sent by server if exists

seanses:


hub_client/src/types.rs
```diff
@@ -9,6 +14,67 @@ pub struct CasJWTInfo {
...
+impl Display for HFRepoType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(self.as_str())
+    }
+}
+
+#[derive(Debug, PartialEq)]
+pub struct RepoInfo {
+    // The type of a repo, one of "model | dataset | space"
+    pub repo_type: HFRepoType,
```
assafvayner, line 58: Do we care about knowing this, RepoType, can we just have a String instead, they are short strings, cheap clones + we can be resilient to the case of new repo types in the futures.

Maybe add a warn! log if the string is not in ["model", "dataset", "space"]

I ask this in reference to the possibility that new ones could be added [ctx](https://huggingface.slack.com/archives/C023JAKTR2P/p1758555207478369). (I ack that adding a new repo type will be very laborious in pretty much any other HF package, I suggest we slightly alleviate here since we don't hard rely on this info)

hub_client/src/client.rs
```diff
@@ -63,12 +64,29 @@ impl HubClient {
...
+        // The reference may contain "/" but the "xet-[]-token" API only parses "rev" from a single component,
+        // thus we encode the reference. It defaults to "main" if not specified by caller because the
+        // API route expects a "rev" component.
+        let rev = encode(self.reference.as_deref().unwrap_or("main"));
+
+        // Clients can get a xet write token, if
+        // - the "rev" is a regular branch, with a HF write token;
+        // - the "rev" is a pr branch, with a HF write or read token;
+        // - it intends to create a pr and repo is enabled for discussion, with a HF write or read token.
+        let query = if let Operation::Upload = operation
```
assafvayner: personal style pref (and I think commonly used in xet-core/CAS) use `matches!()` rather than `if let` when not capturing an inner data value

assafvayner: also personal style pref here, make `url` `mut` above this condition, then if the condition is true, [push_str](https://doc.rust-lang.org/std/string/struct.String.html#method.push_str) onto `url`, it's not that expensive

git_xet/src/app/xet_agent.rs
```diff
@@ -70,13 +71,25 @@ impl TransferAgent for XetAgent {
...
-        )?;
+
+        let token_refresher: Arc<dyn TokenRefresher> =
+            if let Some(refresh_route) = req.action.header.get(XET_REFRESH_ROUTE) {
+                Arc::new(DirectRefreshRouteTokenRefresher::new(
+                    repo,
+                    self.remote_url.clone(),
+                    refresh_route,
+                    Operation::Upload,
+                    "",
```
assafvayner: should we create a session id for these? if it's not something we can get from git, we can at least have it be unique for each run of the custom transfer agent

git_xet/src/app/xet_agent.rs
```diff
@@ -7,14 +7,15 @@ use data::FileUploadSession;
 use data::data_client::{clean_file, default_config};
 use hub_client::Operation;
 use progress_tracking::{ProgressUpdate, TrackingProgressUpdater};
+use utils::auth::TokenRefresher;
 
-use crate::constants::{HF_ENDPOINT_ENV, XET_ACCESS_TOKEN_HEADER, XET_TOKEN_EXPIRATION_HEADER};
+use crate::constants::{HF_ENDPOINT_ENV, XET_ACCESS_TOKEN_HEADER, XET_REFRESH_ROUTE, XET_TOKEN_EXPIRATION_HEADER};
 use crate::errors::{Result, config_error, internal, not_supported};
```
assafvayner: these free floating functions un-namespaced are a bit odd (to me), can we namespace them on the error type and import the error type?

git_xet/src/app/xet_agent.rs
```diff
@@ -7,14 +7,15 @@ use data::FileUploadSession;
...
 use crate::git_repo::GitRepo;
 use crate::git_url::{GitUrl, Scheme};
-use crate::hub_client_token_refresher::HubClientTokenRefresher;
 use crate::lfs_agent_protocol::errors::bad_syntax;
```
assafvayner: same as above, would be nice to know which error type is related to which function that creates an error

git_xet/src/app/xet_agent.rs
```diff
@@ -95,10 +108,21 @@ impl TransferAgent for XetAgent {
...
-        let token = req.action.header[XET_ACCESS_TOKEN_HEADER].clone();
-        let token_expiry: u64 = req.action.header[XET_TOKEN_EXPIRATION_HEADER].parse().map_err(internal)?;
-
-        let config = default_config(cas_url, None, Some((token, token_expiry)), Some(Arc::new(token_refresher)))?
+        let token = req
+            .action
+            .header
+            .get(XET_ACCESS_TOKEN_HEADER)
+            .ok_or_else(|| internal("Hugging Face Hub didn't provide a CAS access token"))?
+            .clone();
```
assafvayner, line 113: why this clone? we can own this String

hub_client/src/types.rs
seanses: This struct exists to make translating information between different APIs a bit easier. For example, 
- "model", "dataset" and "space" between single and plural forms,
-  empty repo type on some APIs default to "model" while repo type is mandatory on others.

Declaring it as an enum can benefit from Rust compiler checking that all "match" blocks should be updated to compile.

hub_client/src/client.rs
seanses: Updated to use matches!

git_xet/src/app/xet_agent.rs
seanses: I have a plan to get it from the moon-landing LFS Batch API, this way all spawned git-xet processes can (and they should) have the same session id. Will make the change to moon-landing first.

git_xet/src/app/xet_agent.rs
seanses: Updated to access them by `errors::*`

git_xet/src/app/xet_agent.rs
seanses: Updated to access it by `protocol_errors::bad_syntax`

hub_client/src/types.rs
seanses: I'll add logging on a higher level in another PR when the logging to file utility is ready. Right now git-lfs captures the stdout and stderr of git-xet so logging has no effect.

git_xet/src/app/xet_agent.rs
seanses: Because req is pass by reference so we can't move, so there's no difference between `clone()` and `.to_owned()`.

git_xet/src/app/xet_agent.rs
assafvayner: can you leave a TODO here if you want to merge this first?

git_xet/src/app/xet_agent.rs
assafvayner: my misread, thought this was a String

git_xet/src/app/xet_agent.rs
assafvayner: this is better, but preferably this is namespace on the error type itself, like we do mostly everywhere else e.g.:

https://github.com/huggingface/xet-core/blob/main/cas_client/src/error.rs#L81-L95
https://github.com/huggingface/xet-core/blob/main/utils/src/errors.rs#L51-L55
https://github.com/huggingface/xet-core/blob/main/mdb_shard/src/error.rs#L63-L71
https://github.com/huggingface/xet-core/blob/main/chunk_cache/src/error.rs#L32-L40

and in CAS [xetcas](https://github.com/huggingface-internal/xetcas/blob/main/cas_server/src/error.rs#L60-L79)

git_xet/src/app/xet_agent.rs
seanses: Added TODO to update token refresher and default_config to use the session id from action header.

seanses: Implemented in the following PR https://github.com/huggingface/xet-core/pull/506

hub_client/src/types.rs
```diff
@@ -9,6 +14,67 @@ pub struct CasJWTInfo {
...
+    // The full name of a repo, formatted as "owner/name"
+    pub full_name: String,
+}
+
+impl RepoInfo {
+    pub fn from(repo_type: &str, repo_id: &str) -> Result<Self> {
```
sirahd: [nit] maybe rename it to `try_form` since it could return an error?
```suggestion
    pub fn try_from(repo_type: &str, repo_id: &str) -> Result<Self> {
```

git_xet/src/app/xet_agent.rs
seanses: Moved them into enum impls. The change affects many files so stacking the PR at https://github.com/huggingface/xet-core/pull/507

hub_client/src/types.rs
seanses: updated

PR #499, opened 2025-09-17
Build and release git-xet

seanses:
The defines the workflow to build git-xet on Linux (amd64 & arm64), macOS (amd64 & arm64) and Windows (amd64). For the macOS build the compiled binary is signed and notarized in place.

### DIFF
--- a/.github/workflows/git-xet-release.yml
+++ b/.github/workflows/git-xet-release.yml
@@ -108,4 +108,4 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          gh release create ${{ github.event.inputs.tag}} git-xet* --generate-notes --prerelease --target ${{github.sha}}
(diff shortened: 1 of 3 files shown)


assafvayner (approved):

assafvayner:
do we need to some signing for the windows binary too?

seanses:
@assafvayner I think so but I haven't figured it out, will do in another PR.

.github/workflows/git-xet-release.yml
```diff
@@ -0,0 +1,113 @@
...
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Semantic version for release (tag will shard the same name)"
+        required: true
+        default: "v0.1.0"
+  pull_request:
```
seanses: Remove this trigger `pull_request` before merge

PR #502, opened 2025-09-22
Logging to directory + log file management; default to log directory for hf_xet 

hoytak:
This PR switches the default logging to log events to a file in '~/.cache/huggingface/xet/logs' (or 'xet/logs' under the specified cache directory if not `~/.cache/huggingface/`).  

In this directory, log files older than 2 weeks are cleaned up on process start, and if the total size of files in the directory is larger than 1gb, then log files are deleted by age to get the directory size under 1gb.  Log files are named with a timestamp and PID; by default, logs newer than 1 day or logs with an active associated PID are never deleted.   All of these are user configurable constants.


xet_runtime/src/log_file_management.rs
assafvayner, line 1: delete file?

xet_runtime/src/lib.rs
assafvayner, line 1: why does this need to be in xet_runtime? Can we spin this into a 3rd crate?

I 👍 this to be default behavior for hf_xet users, 👎 for all users of xet_runtime, specifically the repo-scanner use of xet-core (needs to be updated but here: https://github.com/huggingface-internal/repository-scanner/blob/main/apps/Cargo.toml#L52). I don't think this functionality needs to be imported in that use case, so shouldn't live in this crate.

xet_runtime/src/logging.rs
```diff
@@ -0,0 +1,488 @@
...
+use tracing_subscriber::layer::SubscriberExt;
+use tracing_subscriber::util::SubscriberInitExt;
+use tracing_subscriber::{EnvFilter, Layer};
+use utils::{ByteSize, normalized_path_from_user_string};
+
+use crate::xet_cache_root;
+
+utils::configurable_constants! {
+
+    /// The log destination.  By default, logs to the logs/ subdirectory in the huggingface cache directory.
```
assafvayner: specify in the huggingface xet cache rather than just huggingface cache directory, the former implies the hf cache directory root

xet_runtime/src/logging.rs
```diff
@@ -0,0 +1,488 @@
...
+        let filtered_fmt_layer = fmt_layer_base.pretty().with_filter(fmt_filter);
+        registry.with(filtered_fmt_layer).init();
+    }
+}
+
+fn init_logging_to_file(path: &Path, use_json: bool) -> Result<(), std::io::Error> {
+    // Set up logging to a file.
+    use std::ffi::OsStr;
+
+    use tracing_appender::{non_blocking, rolling};
```
assafvayner: should these be imported at top of file?

xet_runtime/src/logging.rs
```diff
@@ -0,0 +1,488 @@
...
+        if prefix != *LOG_PREFIX {
+            debug!("ignoring log file {:?} with differing prefix {prefix} (!={})", path, *LOG_PREFIX);
+            continue;
+        }
+
+        let meta = match entry.metadata() {
+            Ok(m) => m,
+            Err(e) => {
+                warn!("metadata failed for {:?}: {}", path, e);
+
```
assafvayner: rm line?

xet_runtime/src/logging.rs
```diff
@@ -0,0 +1,488 @@
...
+    let mut n_log_files = 0usize;
+
+    for entry in std::fs::read_dir(log_dir)? {
+        let entry = match entry {
+            Ok(e) => e,
+            Err(e) => {
+                warn!("read_dir entry error: {}", e);
+                continue;
+            },
+        };
```
assafvayner: personal style nit, when you want to log an Err(e) case to use the ErrorPrinter trait with `warn_error`

so changes to:
```
let Ok(entry) = entry.warn_error("read_dir entry error") else {
  continue;
}
```

xet_runtime/src/logging.rs
```diff
@@ -0,0 +1,488 @@
...
+    }
+}
+
+/// The main entry point to set up logging.  Should only be called once.
+pub fn init_logging(cfg: LoggingConfig) {
+    let maybe_log_file: Option<PathBuf> = {
+        match &cfg.logging_mode {
+            LoggingMode::Directory(log_dir) => {
+                if cfg.enable_log_dir_cleanup && log_dir.exists() && log_dir.is_dir() {
+                    run_log_directory_cleanup_background(log_dir);
```
assafvayner: this function spawns a thread that will log things, is it a problem if it starts logging before the collector is set up?

should we instead first call `log_file_in_dir` and if it succeeds then we can spawn the cleanup thread.

seanses:
>From a troubleshooter's perspective, the value is almost always in the `DEBUG` log statements. Asking a user to rerun a download/upload with the log level set to something new can be painful depending on the user. Any way we can get these by default?

Agree with @jsulz that default to `DEBUG` when logging to file

xet_runtime/src/logging.rs
```diff
@@ -0,0 +1,488 @@
...
+    pub fn new(version: String) -> LoggingConfig {
+        // Choose the logging mode.
+        let logging_mode = {
+            if let Some(log_dest) = &*LOG_DEST {
+                if log_dest.is_empty() {
+                    LoggingMode::Console
+                } else {
+                    let path = normalized_path_from_user_string(log_dest);
+
+                    if log_dest.ends_with('/') || log_dest.ends_with('\\') || (path.exists() && path.is_dir()) {
```
seanses:
```suggestion
                    if log_dest.ends_with('/') || (cfg!(windows) && log_dest.ends_with('\\')) || (path.exists() && path.is_dir()) {
```

xet_logging/src/logging.rs
```diff
@@ -0,0 +1,488 @@
...
+        } else {
+            true
+        }
+    });
+
+    // 2) Size trimming: if above the limit, delete oldest eligible (unprotected) first.
+    let mut n_pruned = 0;
+    if total_bytes - deleted_bytes > size_limit {
+        // Sort by oldest first.
+        candidates.sort_by_key(|lf| lf.age);
```
seanses: This is sort by youngest first?

xet_logging/src/logging.rs
```diff
@@ -0,0 +1,488 @@
...
+    let mut deleted_bytes: u64 = 0;
+    candidates.retain(|lf| {
+        if lf.age > max_retention {
+            let path = &lf.path;
+            match std::fs::remove_file(path) {
+                Ok(_) => {
+                    deleted_bytes += lf.size;
+                    debug!("Log Directory Cleanup: Removed old log file {path:?})");
+                },
+                Err(e) => {
```
seanses, line 296: What if multiple processes run concurrently? If the file is already gone (ErrorKind::NotFound), consider it already deleted (`deleted_bytes += lf.size;`)

xet_runtime/src/logging.rs
```diff
@@ -0,0 +1,488 @@
...
+                break;
+            }
+
+            match std::fs::remove_file(&lf.path) {
+                Ok(()) => {
+                    deleted_bytes += lf.size;
+                    n_pruned += 1;
+                    debug!("Log Directory cleanup: Pruned log file {:?}.", lf.path);
+                },
+                Err(e) => info!("Log Directory Cleanup: Error removing size-pruned log file {:?}: {}", lf.path, e),
```
seanses: Same as above

xet_logging/src/logging.rs
```diff
@@ -0,0 +1,488 @@
...
+struct CandidateLogFile {
+    path: PathBuf,
+    size: u64,
+    age: Duration,
+}
+
+fn run_log_directory_cleanup_background(log_dir: &Path) {
+    // Spawn run_log_directory_cleanup as background thread, logging any errors as a warn!
+    let log_dir = log_dir.to_path_buf();
+    std::thread::spawn(move || {
```
seanses, line 192: In the case of git-xet running as a transfer agent of git-lfs, git-lfs will spawn [multiple](https://github.com/huggingface/xet-core/blob/15942e295ee2b45bd96029bfe53debc131719437/git_xet/src/lfs_agent_protocol/protocol_spec.rs#L46) git-xet processes at the same time. If we can elect a leader to run the cleanup that's better, e.g. all processes with the same name as myself and with the smallest pid gets to clean.

seanses, line 192: What happens if the calling thread finishes before the spawned thread ends and the process termination shuts down the spawned thread? It seems to me that it's Ok here, as the spawned thread doesn't reference global memory data and either writes temporary data to disk. Just want to throw out the question in case I miss anything.

xet_runtime/src/lib.rs
hoytak: I'm happy to move this into another crate and make it more explicit.  Note currently that a user has to explicitly call the init_logging function with a LoggingConfig instance to set it up; nothing is set up without that explicit call and config instance, so there isn't really default behavior.

I'm a little confused how this relates to the repo scanner, as it looks like that sets up its own logging and doesn't import the xet_runtime part of xet-core directly (it's there through the data crate, but only used parts get compiled in).

hoytak: Updated the description to clarify this.

hoytak:
Replying to your thoughts @jsulz :

 > 1. A `1GB` max size on this directory is quite large. Outside of the size it takes up on disk it means there will be many log files for users to dig through when forwarding reports. Any reason for it to be this large?

I suspect that for most people, the two week maximum age will be what actually causes the files to be deleted, so most directories will not be anywhere close to that large.  However, I'm happy to decrease this. 

> 2. From a troubleshooter's perspective, the value is almost always in the `DEBUG` log statements. Asking a user to rerun a download/upload with the log level set to something new can be painful depending on the user. Any way we can get these by default?

The problem here is that debug logs a LOT of things that aren't relevant for many things, and it's still a bit expensive to write to a file.  Enabling debug logging will noticeably slow down things like dedup lookup, cache querying, etc. quite significantly.  We could mitigate this by going through and eliminating the debug statements in lower loops, or changing some to be enabled only for debug builds, but I'd rather go through the current debug statements that are part of the upload or download paths and change them to info statements. 

> 3. Having this on by default is good for us (and for the user, IMO), but some people may want to turn these off for yet-to-be-known reasons. Might be good to include that as a config option.

It is all controlled by config variables.  `HF_XET_LOG_DEST=""` changes it to log to console, `HF_XET_LOG_DEST=<file name>` logs it to a file, and `HF_XET_LOG_DEST=<dir name>` logs it to a directory without cleanup work. 

> 
> I can take on documenting this and updating our issue intake to include mention of this as well once it's released. Great stuff @hoytak

xet_logging/src/logging.rs
hoytak: That should be fine, and it's intentionally safe like that.  Cleaning up the logs should be best-effort.

hoytak: I think it's fine to ensure the cleanup process is safe for multiple processes calling it; reading a directory listing shouldn't be expensive.  I suspect negotiating between different processes would be a much more difficult endeavor, especially on systems with multiple users or where different processes could have different log directories.  Let's just make this really quick and robust.  

It does mean we should probably limit the number of log files as well as the total size in order to keep the directory listing manageable...

jsulz:
Thanks @hoytak!

> I suspect that for most people, the two week maximum age will be what actually causes the files to be deleted

🤦 if only I had read your description or the code more thoroughly. Agreed! Happy to strike my comment on that front. 

> but I'd rather go through the current debug statements that are part of the upload or download paths and change them to info statements.

Sounds good to me and you make fair points about your concerns here. I think we should ensure the logs print information that helps us adequately debug a customer's issue. I'm not sure the current INFO logs are sufficient on that front. 

> It is all controlled by config variables. HF_XET_LOG_DEST="" changes it to log to console, HF_XET_LOG_DEST=<file name> logs it to a file, and HF_XET_LOG_DEST=<dir name> logs it to a directory without cleanup work.

Apologies, I might be misunderstanding your comment/the PR. I don't see a way to turn off writing to a log file. Is there a way to do that? If not, it might be good to consider it.

xet_logging/src/logging.rs
seanses: There's no need for negotiation, only need to list the running processes. But I think I can actually handle this on a higher level for the special case of git-xet.

hoytak:
> > It is all controlled by config variables. HF_XET_LOG_DEST="" changes it to log to console, HF_XET_LOG_DEST= logs it to a file, and HF_XET_LOG_DEST= logs it to a directory without cleanup work.
> 
> Apologies, I might be misunderstanding your comment/the PR. I don't see a way to turn off writing to a log file. Is there a way to do that? If not, it might be good to consider it.

When HF_XET_LOG_DEST="" is set to an empty string, it logs straight to the console, i.e. not a file. Happy to consider something clearer if you think this is an issue. 

Also, RUST_LOG=off turns off logging completely.

xet_logging/src/constants.rs
```diff
@@ -0,0 +1,43 @@
+use std::time::Duration;
+
+use utils::ByteSize;
+
+utils::configurable_constants! {
```
assafvayner, line 5: Is there a way at all to disable logging altogether? I think we should have that option but have logging enabled by default.

This setting should also skip the cleanup and creating/verifying the log directory.

xet_logging/src/config.rs
```diff
@@ -0,0 +1,90 @@
...
+                        || (cfg!(windows) && log_dest.ends_with('\\'))
+                        || (path.exists() && path.is_dir())
+                    {
+                        LoggingMode::Directory(path)
+                    } else {
+                        LoggingMode::File(path)
+                    }
+                }
+            } else {
+                LoggingMode::Directory(log_directory.as_ref().to_path_buf())
```
assafvayner, line 68: not sure where to comment, I noticed the log directory is not separated by the CAS environment. Is there a way to do that? To have a different logs directory per environment. If it's not really possible (since we don't know the endpoint at the start, then don't worry about it, it doesn't really matter a lot, especially since we have less endpoints now.

xet_logging/src/constants.rs
hoytak: Yes, already built in -- `RUST_LOGS=off`  (see above).

hoytak: I could detect this and skip the log directory stuff completely in this case :-).

xet_logging/src/config.rs
hoytak: We could, but I'm not sure what the advantage of this would be above having the endpoint logged at the beginning of a file?  What's the use case?

assafvayner: If I have on the same day talked to 2 different cas endpoints, then the logs for the interactions with those 2 endpoints are separated so I can easily tell which logs came from which interactions

seanses: I think CAS endpoint is not important because that's not explicit to users, rather, they care more about the hub endpoint? Users in us and eu access the same hf.co but get different CAS server names, and if we deploy more CAS servers in us-west in the future users in us also get different CAS server names. Placing the logs under different folders and asking them to find the correct folder can be a bit confusing.

seanses (approved):
LGTM!

PR #503, opened 2025-09-22
git-xet install script

seanses:


git_xet/install.sh
```diff
@@ -0,0 +1,105 @@
...
+chmod +x "${BINARY_NAME}"
+
+# Move the binary to the install directory
+echo "Installing binary to ${INSTALL_DIR}..."
+# Use sudo if the user doesn't have write permissions to the directory
+if [ -w "${INSTALL_DIR}" ]; then
+    mv "${BINARY_NAME}" "${INSTALL_DIR}/" || handle_error "Failed to move binary."
+else
+    echo "This script requires sudo permissions to install to ${INSTALL_DIR}."
+    sudo mv "${BINARY_NAME}" "${INSTALL_DIR}/" || handle_error "Failed to move binary with sudo."
```
rajatarya, line 100: I guess we need to test this on Windows - sudo won't work unless running git-bash or under WSL, right?

rajatarya (commented):
Probably should add a TODO / OS check to say this script doesn't work for Windows today.

git_xet/install.sh
seanses: This isn't for Windows (CMD or PowerShell). I'm not sure how to install on Windows yet.

git_xet/install.sh
```diff
@@ -0,0 +1,105 @@
...
+            arm64)
+                DOWNLOAD_URL="${URL_MACOS_ARM64}"
+                ;;
+        esac
+        ;;
+esac
+
+# Check if a download URL was found
+if [ -z "${DOWNLOAD_URL}" ]; then
+    handle_error "Unsupported OS/Architecture combination: ${OS}/${ARCH}"
```
seanses, line 60: > Probably should add a TODO / OS check to say this script doesn't work for Windows today.

@rajatarya  On Windows, (if they run in git-bash which is based on MINGW), `$ uname -s` outputs `MINGW64_NT-10.0-26100` and `$uname -m` outputs `x86_64`. I think this should stop and warn the user. If they run in WSL, then a linux binary should work (I need to try it).

git_xet/install.sh
rajatarya: Sounds good, let's resolve this comment for now.

rajatarya (approved):

PR #506, opened 2025-09-23
Update in accordance with LFS Batch Api change

seanses:
LFS Batch Api response is updated in https://github.com/huggingface-internal/moon-landing/pull/15121 that
- makes token refresh route mandatory (in field `action.href`) and puts CAS URL in the headers map;
- sends a session id in the headers map;

This PR builds on top of https://github.com/huggingface/xet-core/pull/498 and updates how client uses the information according to the above changes.

### DIFF
--- a/git_xet/src/lfs_agent_protocol/protocol_spec.rs
+++ b/git_xet/src/lfs_agent_protocol/protocol_spec.rs
@@ -124,7 +124,7 @@ impl LFSProtocolRequestEvent {
                 }
 
                 if req.action.href.is_empty() {
-                    return Err(bad_argument("empty action.href in server response"));
+                    return Err(GitLFSProtocolError::bad_argument("empty action.href in server response"));
                 }
 
                 Ok(())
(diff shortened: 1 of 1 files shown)


git_xet/src/app/xet_agent.rs
```diff
@@ -73,27 +75,14 @@ impl TransferAgent for XetAgent {
...
-                    // TODO: use session id from the action header
-                    "",
-                )?)
-            };
-
+        let session_id = req.action.header.get(XET_SESSION_ID).map(|s| s.as_str()).unwrap_or_default();
+        let token_refresher: Arc<dyn TokenRefresher> = Arc::new(DirectRefreshRouteTokenRefresher::new(
+            repo,
+            self.remote_url.clone(),
+            &req.action.href,
```
sirahd, line 82: Do we need to validate that `&req.action.href` is not empty in the caller?

seanses: Updated to check in the protocol parser.

PR #509, opened 2025-09-24
Convert status code to error for get_cas_jwt

seanses:
Let the error status code be caught early instead of later during json deserialization.


sirahd (approved):
🚀

PR #510, opened 2025-09-24
Added lazy evaluation functionality to error printer.

hoytak:
Currently, the full message given to a function in error_printer must be always created. This causes extra work when there are no errors if the message should contain additional data. 

This PR introduces `_fn` versions of the existing functions that call a given function on demand to obtain the message.  Thus `.warn_error_fn(|| format!("Error processing {context}"))` would only allocate and create the error string when an error needs to be logged.


assafvayner:
FYI @jgodlew too

seanses (approved):

jgodlew (approved):
LGTM!

PR #513, opened 2025-09-25
Cache rust build in actions

seanses:
In response to [A Joint Statement on Sustainable Stewardship](https://openssf.org/blog/2025/09/23/open-infrastructure-is-not-free-a-joint-statement-on-sustainable-stewardship/) and [Rust Foundation Signs Joint Statement on Open Source Infrastructure Stewardship](https://rustfoundation.org/media/rust-foundation-signs-joint-statement-on-open-source-infrastructure-stewardship/), implements caching of dependency and build artifact, and reduces some CI runtime. Cache entry keys are formed by `os_type`-`arch_type`-`hash of Cargo.lock`, cache configuration adapts from https://docs.github.com/en/actions/tutorials/build-and-test-code/rust#caching-dependencies.

### DIFF
--- a/.github/actions/cache-rust-build/action.yml
+++ b/.github/actions/cache-rust-build/action.yml
@@ -1,0 +1,16 @@
+name: 'Cache Rust Build'
+description: 'Cache Rust dependency and build artifact'
+runs:
+  using: "composite"
+  steps:
+    - name: Cache
+      uses: actions/cache@v4
+      with:
+        path: |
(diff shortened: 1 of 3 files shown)


jgodlew:
> LGTM should add to CAS as well, it has only 1 target directory so simpler

CAS already has this caching being done (in `.github/actions/common-rust/action.yaml`).

sirahd (approved):

PR #517, opened 2025-09-30
rm all docs files

assafvayner:
Removes docs in xet-core in favor of docs in hub-docs for the xet protocol: https://github.com/huggingface/hub-docs/pull/1963


jsulz (approved):
🫡

PR #518, opened 2025-10-01
openapi spec and Makefile for it

assafvayner:
fix XET-741

Creates an openapi specification for all CAS API's following the first version of the protocol specification.

Makefile to generate different language clients for CAS APIs.

### DIFF
--- a/openapi/.gitignore
+++ b/openapi/.gitignore
@@ -1,0 +1,2 @@
+openapitools.json
+generated/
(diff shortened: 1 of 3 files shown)


rajatarya (approved):
This looks fine. My main concern is if we aren't using these ourselves then they might regress over time without us knowing. I'd like a scenario test using this definition to generate a client and then verify it works against prod - that is something we could run nightly to catch any API regressions.

rajatarya:
I think we should add an issue to build an end-to-end test that uses this to generate a CAS client and verify it is working as expected - and have that issue reference this PR.

assafvayner:
https://linear.app/xet/issue/XET-764/create-end-to-end-tests-for-cas-client-openapi-specification

PR #520, opened 2025-10-02
Fix python 314 compat

Qubitium:
Fix python 3.14 build compat

1. `pyo3` depend updated to `0.26`: this is required or else it can't be compiled for python 3.14
2. version update to `1.1.11-dev0`
```
# removed
Fix compat with Rustc 1.86.0
3. change rust conditions that was throwing `unstable` `errors` in `rustc 1.86.0 (05f9846f8 2025-03-31)` (fairly new version, not the latest) _
```
@hoytak  @seanses  @assafvayner

### DIFF
--- a/hf_xet/src/progress_update.rs
+++ b/hf_xet/src/progress_update.rs
@@ -151,7 +151,7 @@ impl WrappedProgressUpdaterImpl {
         // increment.
         //
         // Run on compute thread that doesn't block async workers
-        Python::with_gil(|py| {
+        Python::attach(|py| {
             let func = py_func.bind(py);
 
             // Test if it's enabled first; if None is passed in, then this is disabled.
@@ -223,7 +223,7 @@ impl WrappedProgressUpdaterImpl {
     async fn register_updates_impl(self: Arc<Self>, updates: ProgressUpdate) -> PyResult<()> {
         // Run on compute thread that doesn't block async workers
         tokio::task::spawn_blocking(move || {
-            Python::with_gil(|py| {
+            Python::attach(|py| {
                 let f = self.py_func.bind(py);
 
                 if self.update_with_detailed_progress {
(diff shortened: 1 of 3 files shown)


Qubitium:
> Please upgrade your Rust toolchain to 1.89.0+. We really like the [let chains](https://doc.rust-lang.org/edition-guide/rust-2024/let-chains.html) feature.

Done.

hf_xet/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "hf_xet"
-version = "1.1.10"
+version = "1.1.11-dev0"
```
assafvayner: Please revert this to the original 1.1.10, we bump this up as part of the release process, presumably this change was used for testing.

assafvayner (changes requested):

hf_xet/Cargo.toml
Qubitium: @assafvayner  Will do but `main` branch should have a different version than `release` or else you can't tell if a user is using a self compiled main/dev commit vs release.

Qubitium: Done. reverted the version in `cargo.toml` and `cargo.lock` file.

seanses:
Thanks! Can you also fix the lint errors introduced by the upgrade (see the below failed workflow)?

Qubitium:
@seanses  Now using latest pyo3 non-deprecated semantics:

py::allow_threads -> py::detach
py::with_gil -> py::attach

https://github.com/huggingface/xet-core/pull/520/commits/7984038a4d104917802f9002bd8abceb14264c6d

Please re-trigger workflow.

rajatarya:
This is awesome, thanks @Qubitium for the contribution!

PR #522, opened 2025-10-03
fix: explicitly specify main branch for hub client in migration utility

sirahd:
The current hub client does not pass revision into the argument, which causes the moon-landing call to append `create_pr=1` query param to the token API and returns 403 error.


jsulz (approved):

seanses:
Root cause: repos disable Discussion and thus `create_pr=1` query param fails [pullRequestGuard](https://github.com/huggingface-internal/moon-landing/blob/9cee7bae3141ca2ff5900d12196b927fc42041c9/server/app/Middlewares.ts#L1349).

PR #528, opened 2025-10-14
separating output providers, sequential output providers

assafvayner:
This PR does a refactor of how we pass in the catch all "OutputProvider" to the download mechanism.

It separates the download system to supporting "Sequential" and "Seeking" operations:

- Seeking e.g. opening the file multiple times and seeking to location -- this is the standard writing mechanism hf_xet uses today.
- Sequential e.g. opening a file once and writing data in order -- this is to be used in a set of upcoming PR's/features to use the parallel-download/sequential-write mechanism to support writing to Stdout and to a channel buffer in memory.

To support an in memory channel with backpressure the Channel{Writer, Stream, Reader} are introduced (re-introduced?) in utils. This particularly could be useful in the mount functionality.

### DIFF
--- a/.github/workflows/pre-release-testing.yml
+++ b/.github/workflows/pre-release-testing.yml
@@ -45,7 +45,6 @@ jobs:
           cd ${{ matrix.target-repo }}
           git config user.name "Hugging Face Bot (Xet RC Testing)"
           git config user.email "bot+xet@huggingface.co"
-
       - name: Wait for prerelease to be out on PyPI
         run: |
           VERSION=${{ steps.get-version.outputs.VERSION }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,17 +28,19 @@ jobs:
         platform:
           - runner: ubuntu-22.04
             target: x86_64
-            python-target: x86_64
             manylinux: auto
           - runner: ubuntu-22.04
             target: aarch64
-            python-target: arm64
             manylinux: manylinux_2_28
+        python-version:
+          - 3.14
+          - 3.13t
+          - 3.14t
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: 3.x
+          python-version: ${{ matrix.python-version }}
           cache: 'pip'
       - name: Install wheel tooling
         run: |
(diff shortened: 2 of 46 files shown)


seanses (commented):
Declaring enum values and then implicitly casting between variants in lower level functions makes it difficult to reason about the code. Also it's not very friendly to throw error in runtime when these casts aren't logically correct while the types can be checked in compile time.

My suggestions:
1. Drop the OutputProvider enum, and instead make the functions take explicit sequential writer or parallel writer: e.g. `get_file_with_sequential_writer(SequentialOutputProvider)` and `get_file_with_parallel_writer(SeekingOutputProvider)`. Then in FileDownloader, calls `get_file_with_sequential_writer` when `RECONSTRUCT_WRITE_SEQUENTIALLY` is true.

2. Drop the SequentialOutputProvider enum, instead make it an alias of `Box<dyn AsyncWrite + Send + Unpin>` and change the lower level function to use `writer.write_all(...).await`.

data/src/file_downloader.rs
```diff
@@ -37,19 +38,55 @@ impl FileDownloader {
...
-        output: &OutputProvider,
+        output: SeekingOutputProvider,
         range: Option<FileRange>,
         progress_updater: Option<Arc<ItemProgressUpdater>>,
     ) -> Result<u64> {
         let file_progress_tracker = progress_updater.map(|p| ItemProgressUpdater::item_tracker(&p, file_name, None));
 
         // Currently, this works by always directly querying the remote server.
-        let n_bytes = self.client.get_file(file_id, range, output, file_progress_tracker).await?;
+        let n_bytes = if *RECONSTRUCT_WRITE_SEQUENTIALLY {
```
seanses: Actually, let's move this if-statement to one level higher in data_client.rs so we directly call the correct function `smudge_file_from_hash_sequential` instead of doing the `let sequential = output.try_into()?;` below.

hoytak:
Looking at this now, and the main question I have is whether this entire path can be simplified.  I'm honestly not clear what the conceptual distinction is between all the different types involved (multiple types of channels, output provider, etc.).

For the sake of discussion, I just want to posit a simplification.  I'm open to try to understand why it wouldn't work if it wouldn't: 

- Drop OutputProvider and SeekingOutputProvider completely.
- use `impl AsyncWrite` and `impl AsyncWrite + AsyncSeek + Clone` in their place (for the sequential and parallel downloaders).
- Move all the implementation details (e.g. output to a file or output to a stream) to isolated types that higher level calling routines can use directly.

PR #530, opened 2025-10-20
Fix breaking build changes due to git safety checks.

hoytak:
Fix build break due to git safety checks. 

The recent logging PR added the automatic logging of the git repository commit hash that was used to build the wheel.  However, this requires querying for this hash at build time, which requires running git.  It turns out that in the CI, the user checking out the repository and the user building the wheel are different, which makes git upset.  This PR adds code to tell git everything will be ok.

### DIFF
--- a/.github/actions/git-disable-checks/action.yml
+++ b/.github/actions/git-disable-checks/action.yml
@@ -1,8 +1,0 @@
-name: 'Disable git user checks'
-description: 'Allow git user and repository owner to differ'
-runs:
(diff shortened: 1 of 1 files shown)


seanses (commented):
Turns out it's the user in the docker container is different from the owner of the repo. So the fix is to run `git config --global --add safe.directory "*"` in the docker container. See successful build: https://github.com/huggingface/xet-core/actions/runs/18661456190

PR #532, opened 2025-10-23
Shard utilities needed for GC pass and server-side xorb rewriting.

hoytak:
This PR adds a utility that rewrites a shard to include only the relevant xorb information, dropping unreferenced file information.  

In addition, to preserve the global dedup tracking information associated with the files, this PR also adds a backwards-compatible flag to the chunk metadata that marks a specific chunk as global dedup eligible.  This allows the global dedup information to be tracked independently of the file metadata.

### DIFF
--- a/mdb_shard/src/cas_structs.rs
+++ b/mdb_shard/src/cas_structs.rs
@@ -278,4 +278,21 @@ impl MDBCASInfoView {
         writer.write_all(&self.data[..n_bytes])?;
         Ok(n_bytes)
     }
+
+    #[inline]
+    pub fn serialize_with_chunk_rewrite<W: Write>(
+        &self,
+        writer: &mut W,
+        chunk_rewrite_fn: impl Fn(usize, CASChunkSequenceEntry) -> CASChunkSequenceEntry,
+    ) -> std::io::Result<usize> {
+        let mut n_out_bytes = 0;
+        n_out_bytes += self.header.serialize(writer)?;
+
+        for idx in 0..self.num_entries() {
+            let rewritten_chunk = chunk_rewrite_fn(idx, self.chunk(idx));
+            n_out_bytes += rewritten_chunk.serialize(writer)?;
+        }
+
+        Ok(n_out_bytes)
+    }
 }
--- a/mdb_shard/src/streaming_shard.rs
+++ b/mdb_shard/src/streaming_shard.rs
@@ -350,7 +350,7 @@ impl MDBMinimalShard {
         // a lookup of which chunks occur at the start of a file.  These are the ones for which we set the
         // global dedup eligibility flag.
         //
-        // In addition, we propegate the global dedup eligibility flag if it is already present.
+        // In addition, we propagate the global dedup eligibility flag if it is already present.
         //
         let file_start_chunks = self.file_start_entries();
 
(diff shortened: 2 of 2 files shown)


mdb_shard/src/cas_structs.rs
```diff
@@ -106,13 +111,31 @@ impl CASChunkSequenceEntry {
...
-            #[cfg(test)]
-            _unused: 216944691646848u64,
-            #[cfg(not(test))]
+            flags: 0,
             _unused: 0,
         }
     }
 
+    /// Mark this chunk as a candidate for population in the global dedup table.
+    pub fn with_global_dedup_flag(self, is_global_dedup_chunk: bool) -> Self {
```
assafvayner, line 120: The idea for this is that we would turn on this flag if this chunk is the first chunk of a file?

mdb_shard/src/streaming_shard.rs
```diff
@@ -307,22 +346,60 @@ impl MDBMinimalShard {
...
+                        chunk = chunk.with_global_dedup_flag(true);
+                        next_idx += 1;
+                        next_c_idx = *gde_indices.get(next_idx).unwrap_or(&usize::MAX);
+                    }
+
+                    bytes += chunk.serialize(writer)?;
+                }
+            } else {
+                bytes += cas_info.serialize(writer)?;
+            }
```
assafvayner, line 391: Can make cas_info.serialize() take an Option<Vec<usize>> that then adds the metadata when serializing the chunk information. Rather than have this logic in this function.

mdb_shard/src/cas_structs.rs
hoytak: Exactly.  With this PR, it's done server side and allows us to persist this information across separating out the files and xorb definitions.

mdb_shard/src/streaming_shard.rs
```diff
@@ -307,22 +346,60 @@ impl MDBMinimalShard {
         let mut stored_bytes = 0;
         let mut materialized_bytes = 0;
 
+        // When adding in the global dedup flags based on the files present in the shard, we first need to get
+        // a lookup of which chunks occur at the start of a file.  These are the ones for which we set the
+        // global dedup eligibility flag.
+        //
+        // In addition, we propegate the global dedup eligibility flag if it is already present.
```
seanses: typo "propegate"

mdb_shard/src/streaming_shard.rs
```diff
@@ -350,19 +427,86 @@ impl MDBMinimalShard {
 
         Ok(bytes)
     }
+
+    /// Serialize out a xorb without any of the file information and a subset of xorb data that is given
```
seanses:
```suggestion
    /// Serialize out a shard without any of the file information and a subset of xorb data that is given
```

mdb_shard/src/streaming_shard.rs
```diff
@@ -350,19 +427,86 @@ impl MDBMinimalShard {
...
+    /// by the xorb_filter_fn.  Global deduplication chunk information is preserved.
+    pub fn serialize_xorb_subset_only<W: Write>(
+        &self,
+        writer: &mut W,
+        xorb_filter_fn: impl Fn(&MDBCASInfoView) -> bool,
+    ) -> Result<usize> {
+        self.serialize_impl(writer, false, false, xorb_filter_fn)
+    }
+
+    /// Serialize out the given xorb, sanitizing and updating the global dedup chunk flags and optionally
```
seanses:
```suggestion
    /// Serialize out the given shard, sanitizing and updating the global dedup chunk flags and optionally
```

mdb_shard/src/streaming_shard.rs
```diff
@@ -350,19 +427,86 @@ impl MDBMinimalShard {
 
         Ok(bytes)
     }
+
+    /// Serialize out a xorb without any of the file information and a subset of xorb data that is given
+    /// by the xorb_filter_fn.  Global deduplication chunk information is preserved.
+    pub fn serialize_xorb_subset_only<W: Write>(
+        &self,
+        writer: &mut W,
+        xorb_filter_fn: impl Fn(&MDBCASInfoView) -> bool,
```
seanses, line 425: What's the purpose of `xorb_filter_fn`? Why don't we want to keep all xorb info of an uploaded shard?

hoytak: For GC, we want to only keep xorbs that are currently referenced by an active file, so this allows us to filter the unused xorb references out to queue for later deletion.

seanses: I see

mdb_shard/src/streaming_shard.rs
hoytak: Updated the interface to something similar to your idea.

PR #537, opened 2025-10-24
Improved logging for cas_client crate

hoytak:
This PR adds improved logging for the cas_client crate in light of the changes to log file capturing, upgrading a number of debug! calls to info! and adding begin and completion logging for all the remote operations.  There should be no functionality changes.

### DIFF
--- a/cas_client/src/download_utils.rs
+++ b/cas_client/src/download_utils.rs
@@ -399,7 +399,7 @@ impl DownloadSegmentLengthTuner {
         let mut num_range_in_segment = self.n_range_in_segment.lock()?;
         debug_assert!(*num_range_in_segment <= self.max_segments);
 
-        info!("Download metrics: retries_on_403={}, duration={:?}", metrics.n_retries_on_403, metrics.duration);
+        info!(retried_on_403=metrics.n_retries_on_403, duration=?metrics.duration, "Download metrics");
 
         if metrics.n_retries_on_403 > 0 {
             if *num_range_in_segment > 1 {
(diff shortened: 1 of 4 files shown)


cas_client/src/download_utils.rs
```diff
@@ -449,7 +452,10 @@ pub(crate) async fn get_one_fetch_term_data(
             hash,
         };
         if let Ok(Some(cached)) = cache.get(&key, &fetch_term.range).await.log_error("cache error") {
+            info!("Cache hit for hash {} range {:?}", hash, fetch_term.range);
```
rajatarya: This might be noisy. Cache hit & cache miss info! statements might be too verbose.

cas_client/src/download_utils.rs
```diff
@@ -474,6 +480,8 @@ pub(crate) async fn get_one_fetch_term_data(
             .await
         {
             info!("Writing to local cache failed, continuing. Error: {}", e);
+        } else {
+            info!("Cache write successful for hash {} range {:?}", hash, fetch_term.range);
```
rajatarya: This might be too verbose as well.

cas_client/src/http_client.rs
```diff
@@ -222,7 +227,7 @@ impl Middleware for LoggingMiddleware {
                 // to check if we are retrying or not.
                 let status_code = res.status().as_u16();
                 let request_id = request_id_from_response(res);
-                debug!(request_id, status_code, "Received CAS response");
+                info!(request_id, status_code, "Received CAS response");
```
rajatarya, line 231: _I'm really torn on this one - it does seem too verbose - but super useful when debugging_

rajatarya (approved):
I'm slightly concerned about the logs being too verbose from logging each cache hit/miss and each CAS response, along with each reconstruction request start/end.

But, I think the value of having these for debugging/triaging issues is greater than the size of the logs (and the logs get cleaned anyway) so I think we should merge this!

PR #542, opened 2025-10-27
Updated diagnostics scripts to collect logs

rajatarya:
- also updated README
- added analysis script to load latest dump collected


jsulz (approved):
Looks good! Ran this on Mac and Windows following the instructions without issue.

PR #546, opened 2025-10-30
Adds User-Agent header in cas_client requests to CAS

rajatarya:
Adds User-Agent when making requests to CAS.

* sets to (project) / (version)
* version is picked from Cargo.toml
* project is hf-xet crates, git-xet crates, (also hard-coded xtool)

The reason for this change is to add better observability on the server - so we can segment reqeusts by client and understand client versions in the wild.

@assafvayner : did these changes bork WASM work?
@seanses : did these changes bork git-xet?

Context: https://linear.app/xet/issue/XET-384/send-agent-name-with-version-to-cas-server
Fixes: XET-384

### DIFF
--- a/git_xet/src/token_refresher.rs
+++ b/git_xet/src/token_refresher.rs
@@ -3,13 +3,11 @@ use std::sync::Arc;
 use async_trait::async_trait;
 use cas_client::{Api, RetryConfig, build_http_client};
 use hub_client::{CasJWTInfo, CredentialHelper, Operation};
-use reqwest::header;
 use reqwest_middleware::ClientWithMiddleware;
 use utils::auth::{TokenInfo, TokenRefresher};
 use utils::errors::AuthError;
 
 use crate::auth::get_credential;
-use crate::constants::GIT_LFS_CUSTOM_TRANSFER_AGENT_PROGRAM;
 use crate::errors::Result;
 use crate::git_repo::GitRepo;
 use crate::git_url::GitUrl;
@@ -47,10 +45,7 @@ impl DirectRefreshRouteTokenRefresher {
 #[async_trait]
 impl TokenRefresher for DirectRefreshRouteTokenRefresher {
     async fn refresh(&self) -> std::result::Result<TokenInfo, AuthError> {
-        let req = self
-            .client
-            .get(&self.refresh_route)
-            .with_extension(Api("xet-token"));
+        let req = self.client.get(&self.refresh_route).with_extension(Api("xet-token"));
         let req = self
             .cred_helper
             .fill_credential(req)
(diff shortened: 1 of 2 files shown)


cas_client/src/http_client.rs
```diff
@@ -324,6 +331,22 @@ impl Middleware for SessionMiddleware {
...
+#[cfg_attr(not(target_family = "wasm"), async_trait::async_trait)]
+#[cfg_attr(target_family = "wasm", async_trait::async_trait(?Send))]
+impl Middleware for UserAgentMiddleware {
+    async fn handle(
+        &self,
+        mut req: Request,
+        extensions: &mut Extensions,
+        next: Next<'_>,
+    ) -> reqwest_middleware::Result<Response> {
+        req.headers_mut().insert(USER_AGENT, HeaderValue::from_str(&self.0).unwrap());
```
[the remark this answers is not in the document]
assafvayner: can you `?` instead of unwrap here?
you could also store a `HeaderValue` inside the Middleware struct rather than a string if you want

assafvayner:
instead of a middleware would you mind just setting this in the client when we build a client: https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.user_agent

git_xet/src/app/xet_agent.rs
```diff
@@ -12,6 +12,8 @@ use utils::auth::TokenRefresher;
 use crate::constants::{
     HF_ENDPOINT_ENV, XET_ACCESS_TOKEN_HEADER, XET_CAS_URL, XET_SESSION_ID, XET_TOKEN_EXPIRATION_HEADER,
 };
+
+const USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
```
assafvayner, line 16: can you add something like this for xtool too?, pretty much we should always attach a version, we should be bumping that version (of the data crate) when we make significant changes.

rajatarya: Sounds good, done!

rajatarya:
@seanses Are you good with me merging this?

git_xet/src/token_refresher.rs
seanses: We can drop this line as it's already set for the entire client

hub_client/src/client.rs
seanses: We can drop this line as it's already set for the entire client

git_xet/src/token_refresher.rs
rajatarya: Great catch, removing!

hub_client/src/client.rs
rajatarya: Great, catch, removing!

PR #551, opened 2025-11-03
refactor to rewrite shard header with footer_len set to 0

assafvayner:
related to https://github.com/huggingface/xet-core/issues/547

When xet-core uploads the shard the header field where the footer length is specified is set to 200 where it should be 0 according to the specification.

Note: this value is ignored by the server today, but ideally we would set this right since it can be useful to know if there is a footer on the shard when reading the shard as a whole in a non-streaming fashion.

### DIFF
--- a/data/src/shard_interface.rs
+++ b/data/src/shard_interface.rs
@@ -349,6 +349,7 @@ mod tests {
 
     use super::*;
 
+    #[test]
     fn test_read_shard_to_bytes_remove_footer() -> Result<()> {
         let tmp_dir = TempDir::with_prefix("test_read_shard_to_bytes_remove_footer")?;
         let tmp_dir_path = tmp_dir.path();


data/src/shard_interface.rs
```diff
@@ -328,3 +322,23 @@ impl SessionShardInterface {
...
+    // Read only the portion of the shard file up to the file_lookup_offset,
+    // which excludes the footer and lookup sections.
+    let mut buf = {
+        let mut file = File::open(&si.path)?;
+        let mut buf = vec![0u8; split_off_index];
+        file.read_exact(&mut buf)?;
+        buf
+    };
+    {
+        let mut cursor = Cursor::new(&mut buf);
```
seanses: `si.shard` already holds a deserialized shard header at field `si.shard.header` so no need to deser again.

seanses: Now we can remove this cursor and let `header` directly serialize to `&mut buf` because `Write` is implemented for `&mut [u8]` already.

data/src/shard_interface.rs
```diff
@@ -328,3 +322,17 @@ impl SessionShardInterface {
...
+    let split_off_index = si.shard.metadata.file_lookup_offset as usize;
+    // Read only the portion of the shard file up to the file_lookup_offset,
+    // which excludes the footer and lookup sections.
+    let mut file = File::open(&si.path)?;
+    let mut buf = vec![0u8; split_off_index];
+    file.read_exact(&mut buf)?;
+    // re-write the header to set footer_size to 0.
+    let mut header = si.shard.header.clone();
+    header.footer_size = 0;
+    header.serialize(&mut (&mut buf[..]))?;
```
hoytak: Maybe use `buf[..sizeof::<MDBShardHeader>()]` to make it explicit that's the only part of the buffer you're rewriting.

hoytak (approved):
LGTM, minor comment.  

Maybe add a test?

assafvayner:
P.S. Also tested with this version that uploads work and CAS does accept this

data/src/shard_interface.rs
```diff
@@ -328,3 +322,47 @@ impl SessionShardInterface {
...
+    Ok(Bytes::from(buf))
+}
+
+#[cfg(test)]
+mod tests {
+    use std::io::Cursor;
+
+    use super::*;
+
+    fn test_read_shard_to_bytes_remove_footer() -> Result<()> {
```
seanses, line 353: Missing the `#[test]` attribute

seanses (approved):
LGTM!

PR #553, opened 2025-11-03
Better support of authentication through SSH

seanses:
To review this PR, first review https://github.com/huggingface/xet-core/pull/565 and then https://github.com/huggingface/xet-core/pull/566.

This PR finally enables `git-xet` on Windows authenticating to remote Git server using SSH URL. This is a crucial part as access tokens to the CAS server expire every 900 s and `git-xet` needs to re-authenticate with the Git server by itself during push/pull (whereas the first authentication is handled by `git-lfs`).

This PR builds on top of https://github.com/huggingface/xet-core/pull/565 and uses the same SSH connect utility to authenticate over SSH repo remote URL on both *nix and Windows OS.

Resolves XET-731


git_xet/src/auth/ssh.rs
```diff
@@ -32,39 +34,40 @@ struct GitLFSAuthenticateResponse {
...
+            user_and_host: self.remote_url.user_and_host()?,
+            port: self.remote_url.port(),
+            arg_list: vec![
+                "git-lfs-authenticate".into(),
+                self.remote_url.full_repo_path(),
+                self.operation.as_str().into(),
+            ],
+        };
+
+        // Access to "ssh" and "sh" is provided by the `git` -> `git-lfs` -> `git-xet`, see
```
seanses:
```suggestion
        // Access to "ssh" and "sh" is provided by the `git` -> `git-lfs` -> `git-xet` call chain, see
```

PR #556, opened 2025-11-05
add zstd compression support

azr:


hoytak:
Can resurrect if planned; would be too big of a breaking change with old clients to do without a game plan.

PR #559, opened 2025-11-06
Disable disk cache in hf_xet by default through cargo system. 

hoytak:
This PR disables the disk cache by default in hf_xet using cargo features instead of in-code logic.

Reverts https://github.com/huggingface/xet-core/pull/535

### DIFF
--- a/chunk_cache/Cargo.toml
+++ b/chunk_cache/Cargo.toml
@@ -32,3 +32,4 @@ required-features = ["analysis"]
 
 [features]
 analysis = ["dep:clap"]
+no-default-cache = []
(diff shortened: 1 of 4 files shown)


rajatarya:
Oh, now I get this - love it! I think the only thing left is to back out the `Some(0)` parts of the previous commit. Do you want to handle that or do you want me to add to this PR/do a separate PR?

hoytak:
> Love this!
> 
> I think the other part is to back out the `Some(0)` changes from my previous PR, and then do some quick sanity testing to verify things remain working as expected.

This PR should fully revert that PR.  Did I miss something?  I don't see any Some(0) leftover here.

rajatarya:
> > Love this!
> > I think the other part is to back out the `Some(0)` changes from my previous PR, and then do some quick sanity testing to verify things remain working as expected.
> 
> This PR should fully revert that PR. Did I miss something? I don't see any Some(0) leftover here.

My mistake, I clicked on the [commit](https://github.com/huggingface/xet-core/pull/559/commits/02343e73789d179ef3ee9e07b67773347c32045a) in the Slack message and only reviewed those changes - didn't see the full PR right away. Sorry about that.

PR #560, opened 2025-11-07
deduplicate downloads for sequential output mode

assafvayner:
This PR ensures that we only download each fetch info term once no matter how many separate ranges it fulfills in sequential output mode. The data is interned in memory until the last usage but is dropped after the last reference to it is dropped.

CC @co42 I confirmed this reduces the 10GB downloaded for the 6GB file (on my instance this wasn't faster but I was writing to EBS so that was likely my bottleneck).

### DIFF
--- a/cas_client/src/download_utils.rs
+++ b/cas_client/src/download_utils.rs
@@ -172,7 +172,6 @@ pub(crate) struct FetchTermDownloadInner {
     pub range_download_single_flight: RangeDownloadSingleFlight,
 }
 
-
 #[derive(Debug)]
 pub(crate) struct FetchTermDownload {
     fetch_term_download: FetchTermDownloadInner,
@@ -252,8 +251,8 @@ impl<T: Clone> Clone for TermDownloadResult<T> {
     fn clone(&self) -> Self {
         Self {
             payload: self.payload.clone(),
-            duration: self.duration.clone(),
-            n_retries_on_403: self.n_retries_on_403.clone(),
+            duration: self.duration,
+            n_retries_on_403: self.n_retries_on_403,
         }
     }
 }
@@ -798,7 +797,6 @@ mod tests {
             term: terms[0].clone(),
             skip_bytes: offset_info_first_range,
             take: file_range.length(),
-
         };
 
         let handle = tokio::spawn(async move { download_task.run().await });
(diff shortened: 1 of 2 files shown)


co42:
Thanks @assafvayner !

seanses (commented):
The imminent problem I see is that:
1. The memory usage has the same upper bound as https://github.com/huggingface/xet-core/pull/558, the only improvement is that it drops cached values earlier. Imagine a case where the first term and the last term in a segment share the same coalesced range, and the second term and the second-to-last term share the same coalesced range, and so on, then the entire segment is kept in memory.
2. If the persistent chunk cache is enabled, this still cache the coalesced range in memory, which is unnecessary.
3. It makes sense to keep all the coalesced ranges in memory for the "High performance" mode, but this is definitely not appropriate for regular hf-xet/git-xet users.

I don't like either https://github.com/huggingface/xet-core/pull/558 or this. I think we should have a better solution.

cas_client/src/download_utils.rs
```diff
@@ -172,10 +172,29 @@ pub(crate) struct FetchTermDownload {
     pub range_download_single_flight: RangeDownloadSingleFlight,
 }
 
+#[derive(Debug)]
+pub(crate) struct FetchTermDownload {
```
hoytak, line 176: Is there a reason to have two structs here?  I'm not sure what nesting these gives us...

assafvayner: I realize now maybe not, if I rename things, the parallel download path doesn't need to use the once_cell at all so we could reuse what was originally called "FetchTermDownload" and then the once_cell version only in the sequential path.

Or I can fold everything to 1 struct with the once_cell, I think that's what I'll do.

assafvayner: I retract that. the current version is that `OnceCell` is only used inside `FetchTermDownload` which contains a `FetchTermDownloadInner` and there is a separate struct type used in the parallel case with no `OnceCell`'s called `FetchTermDownloadOnceAndWriteEverywhereUsed` that contains an inner `FetchTermDownloadInner`. I could get rid of the inner struct by having `FetchTermDownloadOnceAndWriteEverywhereUsed` hold a `FetchTermDownload` with a `OnceCell`.

PR #564, opened 2025-11-15
Rework of the constant configuration system. 

hoytak:
This PR refactors the current configuration system built on `configurable_constants!{}` to instead use an explicit `XetConfig` struct defined in a new `xet_config `package and instantiated as part of the xet runtime.   All of the config values that were previously configurable by environment variables have now been moved into this structure. 

Each of the modules having config values here has a separate group that is loaded as part of the config.

On runtime start, environment variables are still checked in the same way as before, so this PR should largely be backwards compatible.

The previous `configurable_constants!{}` macro is now `test_configurable_constants!{}` and is meant for testing constants that can only be changed in debug mode for specific tests.  This allows us to drop the `release_fixed` decorator.  

In addition, high performance mode is handled by an explicit function on XetConfig, with_high_performance_mode(), that sets the defaults of some of the constants to the high performance values.  This is called if high performance mode is enabled.

### DIFF
--- a/data/src/shard_interface.rs
+++ b/data/src/shard_interface.rs
@@ -12,7 +12,7 @@ use mdb_shard::cas_structs::MDBCASInfo;
 use mdb_shard::file_structs::{FileDataSequenceEntry, MDBFileInfo};
 use mdb_shard::session_directory::{ShardMergeResult, consolidate_shards_in_directory, merge_shards_background};
 use mdb_shard::shard_in_memory::MDBInMemoryShard;
-use mdb_shard::{MDBShardFile, MDBShardFileHeader, MDB_SHARD_LOCAL_CACHE_EXPIRATION, ShardFileManager};
+use mdb_shard::{MDB_SHARD_LOCAL_CACHE_EXPIRATION, MDBShardFile, MDBShardFileHeader, ShardFileManager};
 use merklehash::MerkleHash;
 use tempfile::TempDir;
 use tokio::sync::Mutex;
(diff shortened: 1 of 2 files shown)


xet_config/src/macros.rs
```diff
@@ -0,0 +1,86 @@
...
+            /// The group name is derived from the module path. For example, in module `xet_config::groups::data`,
+            /// the env var for TEST_INT would be HF_XET_DATA_TEST_INT.
+            pub fn apply_env_overrides(&mut self) {
+                // Get the module name from the module path
+                // This works by getting the last segment of the module path
+                let module_path = module_path!();
+                let group_name = module_path
+                    .split("::")
+                    .last()
+                    .unwrap_or("unknown");
```
sirahd: Can `group_name` ever be `unknown` (e.g. if the macro is defined outside of any module scope)? If so I think it would be more intuitive to ignore the module name in the `env_var_name` if it is unspecified instead of using `unknown` prefix, thoughts?

xet_logging/tests/integration_tests.rs
assafvayner, line 1: why did these tests need to be async? (maybe needs to revert something?)

hoytak: The config variables are now stored with the runtime, which can be instantiated on the fly from within an existing tokio environment -- that's how we make the async tests work currently -- but xet_config() can't be called in a non-asynchronous environment...   I could change that logic but it was easier to make these async.

xet_config/src/macros.rs
hoytak: The only time it can be unknown is in the bin source file of a program, so no, it can't be unknown here.  I can change this to reflect that.

hoytak:
> Does this change the env var names we already exported? Are old env var names still functional after this change? I think it's better that we don't change or keep compatible to the already published names.

Most of the env var names should be the same.  I'll go through and verify that this is the case.

seanses (approved):
LGTM!

PR #565, opened 2025-11-17
Add ssh connect utility according to git standard

seanses:
This is the first PR to review for https://github.com/huggingface/xet-core/pull/553.

This implements an utility to help set up SSH connection according to Git standards.

1. Env vars `$GIT_SSH_COMMAND`,  `$GIT_SSH` and git config entry `core.sshCommand` define
which ssh executable to use for an SSH connection. `$GIT_SSH_COMMAND` takes precedence over `core.sshCommand` and both are interpreted by the shell (e.g. `GIT_SSH_COMMAND = "ssh -i ~/.ssh/key"`), which allows additional arguments to be included. They both takes precedence over `$GIT_SSH`, which on the other hand must be just the path to a program (which can be a wrapper shell script, if additional arguments are needed). When none of these is given, the default ssh program to use is `ssh`.

2. Env var `$GIT_SSH_VARIANT` takes precedence over git config entry `ssh.variant` and they both define whether
`$GIT_SSH`/`$GIT_SSH_COMMAND`/`core.sshCommand` refer to OpenSSH, plink/putty or tortoiseplink, or instruct git to automatically detect the ssh program type. Valid values are "ssh" (to use OpenSSH options), "plink", "putty",
"tortoiseplink", "simple" (no options except the host and remote command). The default auto-detection can be
explicitly requested using the value "auto". Any other value is treated as "ssh".

This implementation follows the git standard and how the same functionality is handled in
git-lfs (https://github.com/git-lfs/git-lfs/blob/071e19e8ea03b1e40b181706909fb8c18d928e29/ssh/ssh.go#L41).

### DIFF
--- a/git_xet/src/utils/ssh_connect.rs
+++ b/git_xet/src/utils/ssh_connect.rs
@@ -1,0 +1,523 @@
+#![allow(unused)]
+// A utility to help establish SSH connection to a remote Git server.
+
+use crate::errors::{GitXetError, Result};
+use crate::git_repo::GitRepo;
+
+// The type of SSH program to use for SSH connections, valid values are given
+// at https://git-scm.com/docs/git-config#Documentation/git-config.txt-sshvariant.
+enum Variant {
+    Auto,
+    Ssh,
+    Simple,
+    Putty,
+    Tortoise,
+}
+
+impl From<&str> for Variant {
+    fn from(value: &str) -> Variant {
+        match value.to_ascii_lowercase().as_str() {
+            "" | "auto" => Variant::Auto,
+            "simple" => Variant::Simple,
+            "putty" | "plink" => Variant::Putty,
+            "tortoiseplink" => Variant::Tortoise,
+            _ => Variant::Ssh,
+        }
+    }
+}
+
+pub struct SSHMetadata {
+    pub user_and_host: String,
+    pub port: Option<u16>,
+    pub arg_list: Vec<String>,
+}
+
+const DEFAULT_SSH_CMD: &str = "ssh";
+const DEFAULT_GIT_BASH: &str = "sh";
+
+// Return the executable name to execute an SSH command on this machine and the base args.
+// Format the args as needed if the command needs to run in a shell.
+//
+// Git allows using a list of environment variables and git configs to control how to establish
+// an SSH connection to the remote server.
+//
+// 1. Env vars $GIT_SSH_COMMAND and $GIT_SSH and git config entry "core.sshCommand" define
+// which ssh executable to use for SSH connection. $GIT_SSH_COMMAND takes precedence over "core.sshCommand"
+// and both are interpreted by the shell, which allows additional arguments to be included. They both takes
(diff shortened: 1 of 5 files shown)


git_xet/src/utils/ssh_connect.rs
assafvayner, line 1: Are we using this just for windows git-xet or are we planning on using this for macos/linux too?

git_xet/src/utils/ssh_connect.rs
```diff
@@ -0,0 +1,523 @@
...
+    }
+}
+
+// Find out which type of SSH program is used, this allows constructing the call args accordingly.
+// See https://git-scm.com/docs/git-config#Documentation/git-config.txt-sshvariant for details.
+fn get_ssh_variant(repo_config: &git2::Config, sshexe: &str) -> Variant {
+    let variant_str = std::env::var("GIT_SSH_VARIANT")
+        .or_else(|_| repo_config.get_string("ssh.variant"))
+        .unwrap_or_default();
+    let variant = Variant::from(variant_str.as_str());
```
assafvayner, line 147: if the variant_str is the default "" then the Variant chosen will be auto. Is this the correct behavior? As opposed to SSH?

git_xet/src/utils/ssh_connect.rs
```diff
@@ -0,0 +1,523 @@
...
+}
+
+impl From<&str> for Variant {
+    fn from(value: &str) -> Variant {
+        match value.to_ascii_lowercase().as_str() {
+            "" | "auto" => Variant::Auto,
+            "simple" => Variant::Simple,
+            "putty" | "plink" => Variant::Putty,
+            "tortoiseplink" => Variant::Tortoise,
+            _ => Variant::Ssh,
```
assafvayner, line 24: nitpick but I think this should be a `TryFrom<>` with `"ssh" mapping to Ssh variant. and then set `Ssh` variant as default and use `unwrap_or_default()` later.

I find is slightly confusing that the "" and _ => cases lead to different outcomes. maybe needs a comment to why "" is auto

git_xet/src/utils/ssh_connect.rs
seanses: For all platforms. This is the Git standard and should be followed everywhere. The previous implementation that uses the openssh crate was not following the spec.

git_xet/src/utils/ssh_connect.rs
seanses: Yes this is what the spec says: https://git-scm.com/docs/git-config#Documentation/git-config.txt-sshvariant:~:text=The%20default%20auto%2Ddetection%20can%20be%20explicitly%20requested%20using%20the%20value%20auto.%20Any%20other%20value%20is%20treated%20as%20ssh.

git_xet/src/utils/ssh_connect.rs
seanses: So I thought about this and feel it's more clear to group all cases in one place. Otherwise people need to check the TryFrom impl, which defines an error type that will never be used because there's always a default to use, and check the Default impl value -- the logic will be scattered over places.

PR #566, opened 2025-11-17
Integration test for ssh access on Windows

seanses:
This is the second PR to review for https://github.com/huggingface/xet-core/pull/553.

This PR builds on top of https://github.com/huggingface/xet-core/pull/565 and builds an integration test to test access to "ssh" and "sh" on Windows through the "git" (-> "git-lfs") -> "git-xet" call chain.

Out of all the ssh variants, access to programs like "plink", "putty", "tortoiseplink" or "simple" should be given by the env var `$GIT_SSH_COMMAND` or `$GIT_SSH`, or by git config entry `core.sshCommand`. Direct access to the mostly used utility "ssh" and in-direct access to "ssh" via "sh -c" on Windows is provided by the "git" (-> "git-lfs") -> "git-xet" call chain, see git_xet/tests/test_ssh.rs for details.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -127,7 +127,7 @@ winapi = { version = "0.3", features = [
 criterion = { version = "0.5", features = ["html_reports"] }
 httpmock = "0.7"
 rand_core = "0.6"
-russh = "0.54.6"
+russh = "0.54"
 serial_test = "3"
 tempdir = "0.3"
 tracing-test = { version = "0.2", features = ["no-env-filter"] }


git_xet/src/app.rs
```diff
@@ -38,6 +38,11 @@ Remove "lfs.concurrenttransfers" from the global Git config."#)]
     /// Run this program as a LFS custom transfer agent. This is not meant
     /// to be used directly by users, but instead to be invoked by git-lfs.
     Transfer,
+
+    /// Run any arguments passed in as a command. This is a feature only for
+    /// integration tests.
+    #[cfg(feature = "git-xet-for-integration-test")]
+    Runany(RunanyArg),
```
assafvayner: should this be RunAny?

seanses: Renamed

PR #568, opened 2025-11-18
add v1 to api paths

assafvayner:
Not sure how this got reverted or if it wasn't added during the spec writing.

All API paths used by xet-core clients should use the latest API path documented. This includes adding the version to the path.


rajatarya:
The server side for these have already been deployed, correct?

PR #570, opened 2025-11-19
No re-computation of sha256 if provided

seanses:
This PR let File cleaner skip re-computing sha256 if provided for files -- sha256 computation isn't blazing fast, and there's no need to re-compute. The sha256 is not verified anywhere, only serving as a foreign key in the Xet file table to the file id in moon-landing lfs-files table.

- Updates git-xet and migration utility to pass in the sha256 which is already available. 
- Will update huggingface_hub for the `upload_large_folder` case later.
- Related PR in repo-scanner: https://github.com/huggingface-internal/repository-scanner/pull/368 (update the commit id when this merges).

Fix XET-200

### DIFF
--- a/data/src/sha256.rs
+++ b/data/src/sha256.rs
@@ -27,7 +27,7 @@ impl ShaGenerator {
     }
 }
 
-/// Helper struct to generate a sha256 hash as a MerkleHash.
+/// Helper struct to generate a sha256 hash.
 #[derive(Debug, Default)]
 pub struct Sha256Generator {
     hasher: Option<JoinHandle<Result<sha2Sha256, JoinError>>>,


data/src/file_upload_session.rs
```diff
@@ -189,10 +190,14 @@ impl FileUploadSession {
         }))
     }
 
-    pub async fn upload_files(self: &Arc<Self>, files: &[impl AsRef<Path>]) -> Result<Vec<XetFileInfo>> {
+    pub async fn upload_files(
+        self: &Arc<Self>,
+        files: &[impl AsRef<Path>],
+        sha256s: impl IntoIterator<Item = Option<MerkleHash>> + Send,
```
sirahd: would it be more ergonomic if we modify the signature to accept a single list of `(file_path, Option<sha256>)` as input instead of two parameters, one for list of file paths and the other one for sha256? With a single input param we don't need to check the list size matches, and don't need to `zip` them back together.

data/src/sha256.rs
```diff
@@ -2,17 +2,34 @@ use merklehash::MerkleHash;
 use sha2::{Digest, Sha256};
 use tokio::task::{JoinError, JoinHandle};
 
-/// Helper struct to generate a sha256 hash as a MerkleHash.
-#[derive(Debug)]
-pub struct ShaGenerator {
-    hasher: Option<JoinHandle<Result<Sha256, JoinError>>>,
+pub enum ShaGen {
```
hoytak: I don't think these names make it clear what's going on here -- I'd rather have ShaGenerator for the outward facing one and ShaGeneratorImpl for the inner one.  

Or better yet, could we just have the SingleFileCleaner hold an Option<ShaGenerator>?

data/src/data_client.rs
```diff
@@ -168,7 +171,21 @@ pub async fn upload_async(
...
+    // Parse sha256 hex string and ignore invalid ones, or if no sha256 is provided,
+    // create an iterator of infinite number of "None"s.
+    let sha256s: Box<dyn Iterator<Item = Option<MerkleHash>> + Send> = match &sha256s {
+        Some(v) => {
+            if v.len() != file_paths.len() {
+                return Err(DataProcessingError::ParameterError(
+                    "mistached length of the file list and the sha256 list".into(),
+                ));
+            }
+            Box::new(v.iter().map(|s| MerkleHash::from_hex(s).ok()))
```
hoytak: I'm not a huge fan of storing the Sha256 hash in a MerkleHash, as I think it's a bit confusing to lay out.  I'd rather have a type alias from DataHash (as MerkleHash already is) that makes it clear what we're storing.

assafvayner: That works IMO, I do wonder if we can skip the 4 u64's for this with a new alias over [u8; 32], if we ever display it, it should print using the standard hex output in python or other environments.

Edit: we do need it to convert to hash correctly when added to a shard (correctly meaning as a merklehash would)

seanses: yeah I don't like it either

hoytak: It should be displayed through the fmt specifiers, which do print it that way...

seanses: Updated all places that meant to store a sha256 to use an alias type `Sha256 = DataHash`.

data/src/sha256.rs
seanses: Updated the naming. 

I think the other suggestion would make the logic a bit weird. The SingleFileCleaner should hold either a provided sha256 value or a sha256 generator, but if we make it hold an Option of sha256 value and an Option of sha256 generator, the logic reads as they can both be optional (and we need to write code to take care of this case).

data/src/file_upload_session.rs
seanses: Updated the signature of this function to take an iterator of a tuple, but kept the higher level callers still taking separated list of file_path and sha256. Some external callers (especially huggingface_hub upload_folder) don't compute sha256 before xet upload, asking them to always provide a tuple seems inconvenient.

PR #587, opened 2025-12-03
SIGINT handler registration controller (supports deregister)

### LINKED ISSUE #585: hf-xet SIGINT handler remains registered after completing operation (download/upload)
### Describe the bug

Context is here: https://github.com/huggingface/huggingface_hub/issues/3429

### Reproduction

Repro steps are here: https://github.com/huggingface/huggingface_hub/issues/3429

### Logs

```shell

```

### System info

```shell
System info is here: https://github.com/huggingface/huggingface_hub/issues/3429

Reproduces with huggingface_hub 1.1.6 and hf-xet 1.2.0 on Windows 11 amd64
```

rajatarya:
- On Unix restores the previous SIGINT handler, on Windows deregisters handler
- attempts to call winapi directly as unsafe block to allow deregistering ctrl+c handler

Fixes #585 

Coded with assistance from AI (Cursor/Composer).

### DIFF
--- a/hf_xet/Cargo.lock
+++ b/hf_xet/Cargo.lock
@@ -1411,7 +1411,6 @@ dependencies = [
  "serde",
  "serde_json",
  "signal-hook",
- "signal-hook-registry",
  "tracing",
  "utils",
  "winapi",


rajatarya:
I've tested this with a local hf-xet build following the GH issue repro steps - and now CTRL+C correctly stops Uvicorn process on Windows. Will test on MacOS and then clean up the code before marking PR ready for review.

rajatarya:
OK tested on MacOS and see the expected CTRL+C behavior, when running Univorn after download is complete, hitting CTRL+C successfully exits the Python process.

Also verified that while downloading model file pressing CTRL+C results in download cancelation.

hf_xet/src/runtime.rs
```diff
@@ -17,30 +17,85 @@ lazy_static! {
...
+fn install_sigint_handler() -> Result<signal_hook_registry::SigId, MultithreadedRuntimeError> {
     use signal_hook::consts::SIGINT;
-    use signal_hook::flag;
 
     // Register the SIGINT handler to set our atomic flag.
-    // Using `signal_hook::flag::register` allows us to set the atomic flag when SIGINT is received.
-    flag::register(SIGINT, SIGINT_DETECTED.clone()).map_err(|e| {
+    // Using signal_hook_registry directly to get a SigId that we can unregister later.
+    let flag = SIGINT_DETECTED.clone();
+    let signal_id = unsafe {
```
rajatarya: @hoytak I'm concerned about this `unsafe` block vs the older code. But having the SigId is what allows us to deregister this handler when done with operations.

hf_xet/src/runtime.rs
```diff
@@ -17,30 +17,85 @@ lazy_static! {
...
-    })?;
+    use winapi::um::consoleapi::SetConsoleCtrlHandler;
+    use winapi::um::wincon::CTRL_C_EVENT;
+
+    // Install our handler using Windows API directly (instead of ctrlc)
+    // Our handler checks if operations are active:
+    // - If active: handles Ctrl+C and returns TRUE (stops propagation)
+    // - If not active: returns FALSE (allows Python's handler or default to handle it)
+    // This way we don't need to save/restore Python's handler - we just delegate to it
+    unsafe {
```
rajatarya, line 57: I think this `unsafe` is required because directly calling into native Windows API here.

hoytak (changes requested):
So I don't think registering and deregistering the signal handle is the correct approach for two reasons:
1. There is a fundamental race condition here in that external_executor_count can be incremented at any point.  As written, it can increment to one while we're deregistering the signal handler, which would leave it in a bad state.  So it's problematic to do this. 
2. Registering a signal handler in the way we do it doesn't replace the previous ones, it just chains them.  So the others get called.  However, this is more subtle and here we find the root of the windows issues people discovered: 
- On unix, all the installed signal handlers are called when they are registered using the current method.  Since our signal handler does nothing but set an atomic, there's no reason to deregister it; the python or other signal handlers will get called immediately after that.  That's one of the reasons it is so simple and all the logic is elsewhere. 
- On windows, other signal handlers are also chained, but it's more subtle here as signal handlers are called in reverse order of installation until one returns true, which is the of the problem that we're trying to fix.  

Digging into this a bit more, the ctrlc package signal handler returns true on windows, which blocks the other handers from being called.  Thus the actual here fix is to change the windows handler to just set the SIGINT_DETECTED flag and always return false, allowing other handers to do the rest of the work there.  We shouldn't have to change the 
unix code at all.

hf_xet/src/runtime.rs
```diff
@@ -30,17 +30,57 @@ fn install_sigint_handler() -> Result<(), MultithreadedRuntimeError> {
...
+                runtime_pid == std::process::id() && runtime.external_executor_count() > 0
+            } else {
+                false
+            }
+        };
+
+        if has_active_ops {
+            // We have active operations, handle it ourselves
+            SIGINT_DETECTED.store(true, Ordering::SeqCst);
+            winapi::shared::minwindef::TRUE
```
hoytak: This is the key thing here.  We always want to return FALSE.  Returning true blocks the python signal handler from running as well.

rajatarya: This makes sense, updated in latest revision.

rajatarya:
Removed `signal-hook-registry` crate dependency, will merge once all the tests pass.

PR #589, opened 2025-12-05
Added delayed option to RWTaskLock utility

hoytak:
The RWTaskLock utility allows easy handling of a value produced by a background task.  This PR adds in more control over when the task is executed, in particular the ability to delay the execution of the task to when the value is actually needed or to when run_delayed() is called.


assafvayner:
I would slightly prefer to include this change with whatever context you plan to use the delayed variant. Would be ok to merge early but would I'd like to know why you want to add this.

hoytak:
This could be useful but was not needed in the end; closing for now.

PR #591, opened 2025-12-10
Move get_file_term_data and get_reconstruction into Client interface. 

hoytak:
This function adds get_file_term_data and get_reconstruction to the Client interface, including adding those functions to the LocalClient.  It also rearranges some things in the RemoteClient struct to call the methods from there.  This is the first step in a larger effort to separate out the file writer methods from the RemoteClient class in order to allow for more thorough testing and simulation using the LocalClient backend.

### DIFF
--- a/cas_client/src/download_utils.rs
+++ b/cas_client/src/download_utils.rs
@@ -574,7 +574,6 @@ mod tests {
 
     use super::*;
     use crate::remote_client::RemoteClient;
-    use crate::{RetryConfig, build_http_client};
 
     #[tokio::test]
     async fn test_fetch_info_query_and_find() -> Result<()> {


hoytak:
I tried to make the changes to the actual file download paths minimal in this PR, which means passing around an Arc<dyn Client> trait in a lot of cases.  The next PR will restructure a lot of this so that it makes more sense.  

The main things in this one that I want to get eyes on is making sure the behavior of the LocalClient implementations of get_file_term_data and get_reconstruction match how the CAS server functions so that the test coverage of the download code can be adequate.

cas_client/src/interface.rs
```diff
@@ -20,16 +25,32 @@ use crate::{SeekingOutputProvider, SequentialOutput};
...
 #[cfg_attr(target_family = "wasm", async_trait::async_trait(?Send))]
-pub trait Client {
-    /// Get an entire file by file hash with an optional bytes range.
-    ///
-    /// The http_client passed in is a non-authenticated client. This is used to directly communicate
-    /// with the backing store (S3) to retrieve xorbs.
-    ///
-    /// Content is written in-order to the provided SequentialOutput
+pub trait Client: Send + Sync {
+    async fn get_file_reconstruction_info(
```
seanses, line 29: It seems this function is never used, might as well remove it.

cas_client/src/download_utils.rs
seanses: Now that the dispatch `client` is passed in as a function parameter to `Self::run()`, we should drop this field.

cas_client/src/local_client.rs
```diff
@@ -35,19 +45,23 @@ pub struct LocalClient {
...
         let tmp_dir = TempDir::new().unwrap();
         let path = tmp_dir.path().to_owned();
-        let mut s = Self::new(path)?;
+        let s = Self::new_internal(path, Some(tmp_dir))?;
+        Ok(Arc::new(s))
+    }
 
-        s.tmp_dir = Some(tmp_dir);
-        Ok(s)
+    pub fn new(path: impl AsRef<Path>) -> Result<Arc<Self>> {
```
seanses, line 72: Fill in some comments to explain the difference between `LocalClient::temporary()` and `LocalClient::new()`

cas_client/src/local_client.rs
```diff
@@ -233,22 +262,343 @@ impl LocalClient {
...
+                chunk_range: (segment.chunk_index_start, segment.chunk_index_end),
+                byte_range: (byte_start, byte_end),
+            };
+
+            fetch_info_map.entry(segment.cas_hash).or_default().push(fetch_info_intemediate);
+
+            cumulative_bytes += segment.unpacked_segment_bytes as u64;
+            s_idx += 1;
+        }
+
```
seanses, line 519: If `bytes_range` is not None, the CAS prod code further trims within the first term `terms[0]` and the last term `terms.last()` and removing chunks that don't overlap with `bytes_range`. It's up to you how precisely you want this to match the real behavior.

cas_client/src/local_client.rs
```diff
@@ -401,25 +716,129 @@ impl Client for LocalClient {
...
-        true
-    }
+    /// Insert a random file into the local CAS.
+    ///
+    /// This function is used to test the local CAS client.
+    ///
+    /// It generates a random file with a given number of chunks and chunk size.
+    /// It then creates all the xorbs and shard definitions needed, returning
+    /// the file data and the file hash.
+    async fn insert_random_file(
```
seanses: Maybe this function should be outside of LocalClient and calls the pub APIs to upload a random file.

cas_client/src/interface.rs
hoytak: Fair, let's do that in a separate PR as it's not related to this rearrangement, but I agree.  Though perhaps it would be useful for the reconstruction stuff... ?

cas_client/src/local_client.rs
hoytak: Hmm, not sure why that changed in the PR...  Will see.  But new takes a path and temporary() constructs it in a temporary directory.   Will document.

cas_client/src/local_client.rs
hoytak: Ah, good catch.  Yes, I want it to mimic it exactly.

cas_client/src/local_client.rs
hoytak: Ah, good thought.

hoytak: Done; now put as a trait -- ClientTestingUtils

cas_client/src/local_client.rs
```diff
@@ -408,22 +391,330 @@ impl Client for LocalClient {
     fn use_shard_footer(&self) -> bool {
         true
     }
+
+    #[cfg(not(target_family = "wasm"))]
```
assafvayner: I think the local_client module is entirely not available in wasm, so you shouldn't need this macro inside this file

cas_client/src/local_client.rs
```diff
@@ -408,22 +391,330 @@ impl Client for LocalClient {
...
+                // Go through and merge adjascent or overlapping ranges,
+                // then form the full CASReconstructionFetchInfo structs.
+
+                let mut new_fi = fi.clone();
+
+                while idx + 1 < fi_vec.len() {
+                    let next_fi = &fi_vec[idx + 1];
+                    if next_fi.chunk_range.0 <= fi.chunk_range.1 {
+                        new_fi.chunk_range.1 = next_fi.chunk_range.1;
+                        new_fi.byte_range.1 = next_fi.byte_range.1;
```
assafvayner: can these `.0`/`.1` be named fields; if these are `CASReconstructionFetchInfo.{chunk_range/bytes_range}` they should already be, so I'm confused why there's 0, 1

seanses: It's from a struct defined within this function:
```
struct FetchInfoIntemediate {
    xorb_hash: MerkleHash,
    chunk_range: (u32, u32),
    byte_range: (u32, u32),
}
```

But yeah I agree with @assafvayner it's better to replace it with
```
struct FetchInfoIntemediate {
    xorb_hash: MerkleHash,
    chunk_range: ChunkRange,
    byte_range: HttpRange,
}
```

Also note that the conversion from
```
let (byte_start, byte_end) =
                xorb_footer.get_byte_offset(segment.chunk_index_start, segment.chunk_index_end)?;
```
to a HttpRange is an "end exclusive" range to an "end inclusive" range. To do this cleanly you can:

```
HttpRange::from(FileRange::new(byte_start as u64, byte_end as u64))
```
.

Or even better, to prevent future mistakes, let's change the return type of `get_byte_offset` from `(u32, u32)` to `FileRange`?

hoytak: Sounds good; made these changes except for: 
- The get_byte_offset function deals with u32s, so I'd rather not change the return type now.  
- For the intermediate struct, I'd rather just go with FileRange as that's what I'm dealing with, then convert to HttpRange when putting into a CASReconstructionInfo object.

cas_client/src/local_client.rs
hoytak: Resolved.

cas_client/src/local_client.rs
```diff
@@ -407,22 +389,365 @@ impl Client for LocalClient {
...
+                s_idx += 1;
+            }
+        }
+
+        // Now, prepare the response by iterating over the segments and
+        // adding the terms and fetch info to the response.
+        let mut terms = Vec::new();
+
+        #[derive(Clone)]
+        struct FetchInfoIntemediate {
```
assafvayner: typo `Intermediate`

cas_client/src/local_client.rs
```diff
@@ -407,22 +389,365 @@ impl Client for LocalClient {
...
+            while idx < fi_vec.len() {
+                // Go through and merge adjascent or overlapping ranges,
+                // then form the full CASReconstructionFetchInfo structs.
+                let mut new_fi = fi_vec[idx].clone();
+
+                while idx + 1 < fi_vec.len() {
+                    let next_fi = &fi_vec[idx + 1];
+                    if next_fi.chunk_range.start <= new_fi.chunk_range.end {
+                        new_fi.chunk_range.end = next_fi.chunk_range.end;
+                        new_fi.byte_range.start = next_fi.byte_range.end;
```
assafvayner: So if the fi_vec list is sorted on chunk_range.start, it's guaranteed that the new_fi.{chunk/byte}_range.start is at the right minimum value for the condensed range. With each other fetch info term you're coalescing, the end needs to be re-set not the start (you are setting the chunk_range.end, but you're setting byte_range.start which I think is meant to be end.

Also, it is possible if I have chunk ranges (0, 3) and (0, 2) that I start with the first and then I end up setting the range to (0, 2) you should set the end to the max of the 2 .end fields.

cas_client/src/local_client.rs
```diff
@@ -407,22 +389,365 @@ impl Client for LocalClient {
...
+                        break;
+                    }
+                }
+
+                merged.push(CASReconstructionFetchInfo {
+                    range: new_fi.chunk_range,
+                    url: format!(
+                        "{:?}:{}:{}:{}",
+                        file_path, new_fi.byte_range.start, new_fi.byte_range.end, timestamp_ms
+                    ),
```
assafvayner: minor nit, would like this format!() and the code that parses this url to be their own independent functions, can be `#[inline]`

seanses (approved):

cas_client/src/local_client.rs
hoytak: Great catch.

cas_client/src/local_client.rs
hoytak: Done.

PR #598, opened 2025-12-14
Add Unix domain socket support for CAS HTTP connections

jedisct1:
Support `HF_XET_UNIX_SOCKET=/path/to/socket` to route traffic through a Unix socket instead of TCP.

This is useful when running a XET client in a sandbox that doesn't have direct access to the network.

### DIFF
--- a/cas_client/src/http_client.rs
+++ b/cas_client/src/http_client.rs
@@ -13,34 +13,90 @@ use utils::auth::{AuthConfig, TokenProvider};
 
 use crate::{CasClientError, error};
 
+/// Returns the Unix socket path if configured via `HF_XET_CLIENT_UNIX_SOCKET_PATH` env var.
+#[cfg(not(target_family = "wasm"))]
+pub fn get_unix_socket_path() -> Option<String> {
+    use xet_runtime::xet_config;
+    xet_config().client.unix_socket_path.clone()
+}
+
+/// Middleware that rewrites https:// URLs to http:// when using Unix socket.
+/// This allows the proxy to parse plain HTTP and upgrade to HTTPS when forwarding.
+#[cfg(unix)]
+struct HttpsToHttpMiddleware;
+
+#[cfg(unix)]
+#[async_trait::async_trait]
+impl Middleware for HttpsToHttpMiddleware {
+    async fn handle(
+        &self,
+        mut req: Request,
+        extensions: &mut Extensions,
+        next: Next<'_>,
+    ) -> Result<Response, reqwest_middleware::Error> {
+        let url = req.url_mut();
+        if url.scheme() == "https" {
+            let original_scheme = url.scheme().to_string();
+            url.set_scheme("http").ok();
+            info!(original_scheme, new_scheme = "http", url = %url, "Rewriting URL scheme for Unix socket proxy");
+        }
+        next.run(req, extensions).await
+    }
+}
+
 #[cfg(not(target_family = "wasm"))]
 fn reqwest_client(user_agent: &str) -> Result<reqwest::Client, CasClientError> {
     use xet_runtime::{XetRuntime, xet_config};
 
-    let client = XetRuntime::get_or_create_reqwest_client(|| {
-        let mut builder = reqwest::Client::builder()
-            .pool_idle_timeout(xet_config().client.idle_connection_timeout)
-            .pool_max_idle_per_host(xet_config().client.max_idle_connections)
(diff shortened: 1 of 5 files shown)


hoytak:
Hello @jedisct1 -- thank you for this contribution.  We'll look into this and get back to you.

cas_client/src/unix_client.rs
```diff
@@ -0,0 +1,323 @@
...
+            client: self.client.clone(),
+            socket_path: self.socket_path.clone(),
+        }
+    }
+}
+
+/// Returns the Unix socket path if configured via environment variable or config.
+pub fn get_unix_socket_path() -> Option<String> {
+    // Check environment variable first (takes precedence)
+    if let Ok(path) = std::env::var("HF_XET_UNIX_SOCKET")
```
seanses: The xet_config crate actually handles parsing the `unix_socket_path` from the `HF_XET_UNIX_SOCKET_PATH` env var.

cas_client/src/http_client.rs
```diff
@@ -19,6 +19,10 @@ use utils::auth::{AuthConfig, TokenProvider};
 use xet_runtime::xet_config;
 
 use crate::retry_wrapper::on_request_failure;
+#[cfg(unix)]
+pub use crate::unix_client::{
```
seanses: How is this client actually plugged into the actual file upload/download functions?

jedisct1:
Hi @seanses, and thanks for your guidance and feedback!

Parsing is now handled by the `xet_config` crate as you suggested, and the integrations should be complete. I also added tests for this feature.

seanses:
Thank you for adding unit tests and integration! Can you also share instructions how to test this end to end?

jedisct1:
Hi @seanses ,

Here's a simple way to test this end-to-end.

Save this simple Python proxy:

```python
#!/usr/bin/env python3
import asyncio, os, httpx

SOCKET = "/tmp/xet-proxy.sock"

async def handle(r, w):
    line = await r.readline()
    method, url, _ = line.decode().split()
    hdrs, clen = {}, 0
    while (h := await r.readline()) not in (b'\r\n', b''):
        k, v = h.decode().split(':', 1)
        hdrs[k.strip()] = v.strip()
        if k.lower().strip() == 'content-length': clen = int(v)
    body = await r.read(clen) if clen else None
    host = hdrs.get('Host', '') or hdrs.get('host', '')
    if url.startswith('http://'):
        url = 'https://' + url[7:]
    elif url.startswith('/'):
        url = f"https://{host}{url}"
    print(f"-> {method} {url[:80]}...")
    async with httpx.AsyncClient(timeout=120, follow_redirects=True) as c:
        resp = await c.request(method, url, headers={k:v for k,v in hdrs.items() if k.lower() not in ('host','connection','content-length')}, content=body)
    print(f"<- {resp.status_code} ({len(resp.content)} bytes)")
    w.write(f"HTTP/1.1 {resp.status_code} OK\r\nContent-Length: {len(resp.content)}\r\n\r\n".encode() + resp.content)
    await w.drain(); w.close()

async def main():
    os.path.exists(SOCKET) and os.unlink(SOCKET)
    srv = await asyncio.start_unix_server(handle, path=SOCKET)
    print(f"Proxy listening on {SOCKET}")
    async with srv: await srv.serve_forever()

asyncio.run(main())
```

Start the proxy:

```sh
uv run --with httpx unix_proxy.py
```

Export `HF_XET_CLIENT_UNIX_SOCKET_PATH`:

```sh
export HF_XET_CLIENT_UNIX_SOCKET_PATH=/tmp/xet-proxy.sock
```

Then download a model, for example with:

```python
from huggingface_hub import hf_hub_download

path = hf_hub_download(
    repo_id='Qwen/Qwen2.5-0.5B',
    filename='model.safetensors',
    force_download=True
)
print(f'Downloaded: {path}')
```

The proxy terminal will show all requests.

hoytak:
@jedisct1 : With a bunch of recent tooling added to simulate connections, it made it doable to directly incorporate tests for full unix socket support.   I took this PR and added a commit on top of it that updates some of the API to make it easier to test, then incorporated a full test suite that uses a proxy around a local test server.  Could you check out https://github.com/huggingface/xet-core/pull/618 and make sure that incorporates the functionality you need?  

Also, I made sure to preserve your commit history so you'll get credit on the PR.

jedisct1:
Perfect! This is very cool, thank you!

PR #602, opened 2025-12-22
Mock CAS server built on LocalClient for testing and simulation.

hoytak:
This PR adds a fully functional CAS server built around a LocalClient instance.  This allows full testing of the RemoteClient interface without hitting the actual CAS backend.

For testing, it can either be run as a standalone executable, or it can be started using a LocalTestServer instance that exposes both a RemoteClient interface as client, or direct access to the state through a stored LocalClient instance. 

Numerous tests are added to also cover existing functionality as well as the new server functioning.

(Also, it exposed that when using a lot of tests with wiremock or this server, the testing would often hit a "Too many open files" error; this was fixed by consolidating these tests to reduce the number of separate testing servers running at once.

### DIFF
--- a/cas_client/Cargo.toml
+++ b/cas_client/Cargo.toml
@@ -69,7 +69,7 @@ native-tls-vendored = ["reqwest/native-tls-vendored"]
 
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
-axum = "0.8"
+axum = { workspace = true }
 futures-util = { workspace = true }
 heed = { workspace = true }
 hyper = { workspace = true }
--- a/cas_client/src/local_server/server.rs
+++ b/cas_client/src/local_server/server.rs
@@ -112,13 +112,12 @@ impl LocalServer {
 
     /// Builds the Axum router with all CAS API routes.
     ///
-    /// Routes are organized in two groups:
-    /// 1. `/v1/` prefixed routes - The canonical API paths
-    /// 2. Root-level routes - Legacy aliases for compatibility
+    /// Routes follow the pattern used by RemoteClient:
+    /// - `/v1/` prefixed routes for chunks, xorbs, reconstructions, and files
+    /// - Root-level `/reconstructions` for batch queries and `/shards` for uploads
     fn create_router(&self) -> Router {
         Router::new()
             .route("/health", get(handlers::health_check))
-            // Canonical v1 API routes
             .nest(
                 "/v1",
                 Router::new()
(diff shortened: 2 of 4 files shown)


cas_client/Cargo.toml
```diff
@@ -68,8 +69,11 @@ native-tls-vendored = ["reqwest/native-tls-vendored"]
 
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
+axum = "0.8"
```
seanses: Let's move these deps (axum and tower-http) to the workspace root.

cas_client/src/local_server/server.rs
```diff
@@ -0,0 +1,646 @@
...
+            .route("/reconstructions/{file_id}", get(handlers::get_reconstruction))
+            .route("/chunks/{prefix}/{hash}", get(handlers::get_dedup_info_by_chunk))
+            .route("/chunk/{prefix}/{hash}", get(handlers::get_dedup_info_by_chunk))
+            .route("/xorbs/{prefix}/{hash}", head(handlers::head_xorb).post(handlers::post_xorb))
+            .route("/xorb/{prefix}/{hash}", head(handlers::head_xorb).post(handlers::post_xorb))
+            .route("/shards", post(handlers::post_shard))
+            .route("/shard", post(handlers::post_shard))
+            .route("/shard/{prefix}/{hash}", post(handlers::post_shard))
+            .route("/files/{file_id}", head(handlers::head_file))
+            .route("/get_xorb/{prefix}/{hash}/", get(handlers::get_file_term_data))
```
seanses: I don't think we should expose this route (and it's not used anywhere). AI probably misread this from some remote client unit tests.

cas_client/src/local_server/server.rs
```diff
@@ -0,0 +1,646 @@
...
+            .nest(
+                "/v1",
+                Router::new()
+                    .route("/reconstructions", get(handlers::batch_get_reconstruction))
+                    .route("/reconstructions/{file_id}", get(handlers::get_reconstruction))
+                    .route("/chunks/{prefix}/{hash}", get(handlers::get_dedup_info_by_chunk))
+                    .route("/xorbs/{prefix}/{hash}", head(handlers::head_xorb).post(handlers::post_xorb))
+                    .route("/files/{file_id}", head(handlers::head_file)),
+            )
+            // Legacy route aliases (for compatibility with older clients)
```
seanses: Probably just remove all legacy routes -- this is for unit/integration tests and thus there's no old client.

PR #610, opened 2026-01-16
Remove need for xorb and shard footers in testing code.

hoytak:
Currently, the remote server rebuilds the footers for xorbs and shards from streaming versions of the uploaded data, but the testing code doesn't actually follow the same pattern.   This updates the local testing code to do the same, unifying the API.

### DIFF
--- a/cas_client/src/simulation/local_client.rs
+++ b/cas_client/src/simulation/local_client.rs
@@ -477,32 +477,16 @@ impl Client for LocalClient {
         // Rebuild a full in-memory shard with proper lookup tables
         let mut in_memory_shard = MDBInMemoryShard::default();
 
-        // Add file info by deserializing from the view's raw bytes
+        // Add file info from the views
         for i in 0..minimal_shard.num_files() {
             let file_view = minimal_shard.file(i).unwrap();
-            let mut cursor = Cursor::new(file_view.bytes());
-            if let Some(file_info) = mdb_shard::file_structs::MDBFileInfo::deserialize(&mut cursor)? {
-                in_memory_shard.add_file_reconstruction_info(file_info)?;
-            }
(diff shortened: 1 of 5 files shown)


cas_client/src/simulation/local_client.rs
```diff
@@ -465,13 +469,51 @@ impl Client for LocalClient {
...
+                    cas_view.cas_hash(),
+                    chunks.len() as u32,
+                    total_bytes,
+                ),
+                chunks,
+            };
+            in_memory_shard.add_cas_block(cas_info)?;
+        }
+
+        // Write the rebuilt shard to disk (creates proper lookup tables)
```
seanses, line 492: I think this meant to simulate the CAS server behavior -- the CAS server doesn't rebuild the lookup tables because they are useless. The `global_dedup_table` should function as the DDB chunk->shard table.

cas_client/src/simulation/memory_client.rs
```diff
@@ -537,26 +538,57 @@ impl Client for MemoryClient {
...
+        let minimal_shard = MDBMinimalShard::from_reader(&mut reader, true, true)?;
 
         // Merge file info into our shard
         {
             let mut shard_lg = self.shard.write().await;
-            reader.set_position(0);
-            for file_info in shard_info.read_all_file_info_sections(&mut reader)? {
+
+            // Convert file views to MDBFileInfo and add to shard
+            for i in 0..minimal_shard.num_files() {
```
seanses, line 550: I noticed this shares a lot of similar code with the above `LocalClient::upload_shard` to convert `MDBFileInfoView` to `MDBFileInfo` and `MDBCASInfoView` to `MDBCASInfo`. Let's unify them to helpers functions in `MDBFileInfoView` and `MDBCASInfoView`.

PR #611, opened 2026-01-21
Hash table with pass-through hasher for MerkleHashes

hoytak:
Currently, the rust HashMap uses a randomized hasher for input, which prevents hash collision attacks.  However, in our code, we don't need that protection in the client, and a MerkleHash is already a cryptographic hash.   This PR adds a MerkleHashMap type that just passes the hash through to the HashMap, providing a substantial speedup:

```
=================================================================
PERFORMANCE SUMMARY (times in ms, lower is better)
=================================================================
Test                                  HashMap         PassThrough
-----------------------------------------------------------------
--- 100K ---
  Insert                                  2.1                 0.7
  Lookup                                  2.1                 1.3
  Insert+Lookup                           4.4                 1.6
  Serialize                               1.6                 0.9
  Deserialize                             4.3                 1.2


--- 10M ---
  Insert                                433.2               204.1
  Lookup                                615.3               255.5
  Insert+Lookup                         951.6               460.4
  Serialize                             117.2                93.4
  Deserialize                           599.5                89.3

=================================================================
```

It also replaces HashMap<MerkleHash, ...> everywhere in the code to provide an across-the-board improvement.

### DIFF
--- a/utils/src/data_structures/passthrough_hasher.rs
+++ b/utils/src/data_structures/passthrough_hasher.rs
@@ -5,7 +5,7 @@ use merklehash::MerkleHash;
 use serde::{Deserialize, Serialize};
 
 /// Trait for types that can efficiently provide a u64 hash value.
-/// Types implementing this trait are optimized for use with DHPassthroughHasher,
+/// Types implementing this trait are optimized for use with `U64DirectHasher`,
 /// which avoids extra computation by directly using the u64 value as the hash.
 pub trait U64HashExtractable {
     fn u64_hash_value(&self) -> u64;
--- a/utils/src/data_structures/passthrough_hashmap.rs
+++ b/utils/src/data_structures/passthrough_hashmap.rs
@@ -8,7 +8,7 @@ use super::passthrough_hasher::{U64DirectHasher, U64HashExtractable};
 
 /// A HashMap wrapper optimized for keys that implement `U64HashExtractable`.
 ///
-/// This structure uses `DHPassthroughHasher` which avoids extra hash computation
+/// This structure uses `U64DirectHasher` which avoids extra hash computation
 /// by directly using the u64 value extracted from the key. This is particularly
 /// efficient for cryptographic hashes (like `MerkleHash`) where the first 8 bytes
 /// already provide excellent distribution.


utils/src/data_structures/passthrough_hasher.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    fn write(&mut self, bytes: &[u8]) {
+        debug_assert_eq!(bytes.len(), 32);
+
+        unsafe {
+            let dest = &mut self.state as *mut u64 as *mut u8;
+            bytes.as_ptr().copy_to_nonoverlapping(dest, 8);
+        }
+    }
```
assafvayner, line 64: If we can assert it's a merklehash and that the full merklehash is passed in, then this is fine.

I wonder about a couple of cases or mis-use or maybe out-of-our-control issues

- if write is called with a slice less than 8 bytes, what should happen?
- if write is called multiple times what should happen? Do we only want to save the first result?

utils/src/data_structures/passthrough_hashmap.rs
```diff
@@ -0,0 +1,484 @@
...
+impl<Key, Value, const N: usize> From<[(Key, Value); N]> for PassThroughHashMap<Key, Value>
+where
+    Key: U64HashExtractable + Hash + Eq,
+{
+    fn from(arr: [(Key, Value); N]) -> Self {
+        arr.into_iter().collect()
+    }
+}
+
+impl<Key, Value> Serialize for PassThroughHashMap<Key, Value>
```
assafvayner, line 175: should these not just serialize (and other operations) for the inner value, without needing to loop explicitly?

utils/src/data_structures/passthrough_hasher.rs
hoytak: This here implements the traits methods needed by the Hasher used in the HashMap, but this here is just implemented for the MerkleHash and the truncated hash.  I don't know if there's a way to enforce them being used only in those cases, but the idea is that the types implementing things explicitly for MerkleHash and TruncatedMerkleHash are all that get easily exported.

utils/src/data_structures/passthrough_hashmap.rs
hoytak: I tried that, but it turns out to be really slow -- it seems the inner HashMap implementation doesn't preallocate anything, so this method turned out to be 2-3x faster.

utils/src/data_structures/passthrough_hasher.rs
```diff
@@ -0,0 +1,143 @@
+use std::hash::{BuildHasher, Hasher};
+use std::marker::PhantomData;
+
+use merklehash::MerkleHash;
+use serde::{Deserialize, Serialize};
+
+/// Trait for types that can efficiently provide a u64 hash value.
+/// Types implementing this trait are optimized for use with DHPassthroughHasher,
```
seanses: I see "DHPassthroughHasher" is mentioned a couple of times here and in "passthrough_hashmap.rs". What is it?

utils/src/data_structures/passthrough_hashmap.rs
```diff
@@ -0,0 +1,484 @@
...
+/// - `Key`: The key type (must implement `U64HashExtractable + Hash + Eq`)
+/// - `Value`: The value type stored in the map
+pub struct PassThroughHashMap<Key, Value>
+where
+    Key: U64HashExtractable + Hash + Eq,
+{
+    inner: HashMap<Key, Value, U64DirectHasher<Key>>,
+}
+
+impl<Key, Value> PassThroughHashMap<Key, Value>
```
seanses, line 47: Curious if this can be implemented by just replacing the RandomSate of HashMap so we don't have to redefine and mimic all the functions of HashMap?

```
type PassThroughHashMap<Key, Value> = HashMap<Key, Value, U64DirectHasher<Key>>;
```

utils/src/data_structures/passthrough_hasher.rs
hoytak: A missed rename -- it should say U64DirectHasher.  I'll change that.

utils/src/data_structures/passthrough_hasher.rs
assafvayner: worth asking is it guaranteed that when a MerkleHash is passed through that it's passed in as the full 32 byte slice, and never called multiple times?

reading through the standard library Hash impl for slices, which should be what gets used. I'm not sure that it's guaranteed to only be called once

utils/src/data_structures/passthrough_hashmap.rs
hoytak: Yes, actually, just using a type alias is doable.  However, I can't implement traits for a type alias, and the error messages if you tried to use it with a Key type not having a matching implementation of the U64DirectHasher trait would be weird.

However, the method currently uses Deref and DerefMut traits to mimic the HashMap, so all the methods automatically pass through to the HashMap without reimplementing them, so it's not actually that much code.

The other downside to the type alias is that the default serialization and deserialization routines for the HashMap are actually really, really slow as you can see in the benchmarks, and this implementations speed that up significantly.

PR #612, opened 2026-01-22
Parallel HashMap/HashSet for large-scale MerkleHash keyed collections.

hoytak:
This PR adds a base data structure, ParallelHashMap, that uses a two-level hash table, where a fixed-size first level maps to a Mutex around a regular hash table.  This structure works efficiently in multithreaded code for large HashSet and HashMap implementations, as the first level can be adjusted to be large enough to make lock contention minimal.  Versions for MerkleHash with a pass through hasher are also provided, making this efficient as a building block for large scale MerkleHash keyed operations, like those needed for the planned cache improvements and garbage collection.  

In addition, it also provides efficient serialization and deserialization methods that allow it to be serialized to an archive directory, using multiple threads to save and load the contents.  Set operations can thus be implemented efficiently against a saved archive, which allows for reduced memory use.

A benchmark program shows improvement for this data structure over the base Mutex<HashMap> and Mutex<MerkleHashMap> in a parallel environment: 

For basic insert and lookup operations: 

```
=====================================================================================
PARALLEL 4 THREADS PERFORMANCE (times in ms, lower is better)
=====================================================================================
Test                           Mutex<HashMap>  Mutex<PassThrough>            Parallel
-------------------------------------------------------------------------------------
--- 100K ---
  Insert                                 14.3                 6.5                 5.6
  Lookup                                  8.6                 4.4                 1.9
  Insert+Lookup                          20.0                11.1                 6.5

--- 10M ---
  Insert                               2410.7              1589.5               653.7
  Lookup                               2133.0              1446.4               463.3
  Insert+Lookup                        4367.0              2988.3              1093.0

=====================================================================================

=====================================================================================
PARALLEL 8 THREADS PERFORMANCE (times in ms, lower is better)
=====================================================================================
Test                           Mutex<HashMap>  Mutex<PassThrough>            Parallel
-------------------------------------------------------------------------------------
--- 100K ---
  Insert                                 16.6                10.4                 7.7
  Lookup                                  6.1                 4.6                 2.2
  Insert+Lookup                          20.3                18.4                 9.5

--- 10M ---
  Insert                               1973.8              1314.2               785.0
  Lookup                               1645.6              1148.1               356.8
  Insert+Lookup                        3650.8              2442.2              1129.2

=====================================================================================
``` 

And for serialization (larger improvements possible with more entries): 

```
=====================================================================================
Test                                  HashMap         PassThrough            Parallel
-------------------------------------------------------------------------------------
--- 100K ---
  Serialize                               1.3                 0.9                 2.9
  Deserialize                             3.3                 1.0                 2.8

--- 10M ---
  Serialize                             109.0                88.7                53.5
  Deserialize                           756.3                81.0                50.7

=====================================================================================
```

### DIFF
--- a/utils/src/data_structures/parallel_hash_map_impl.rs
+++ b/utils/src/data_structures/parallel_hash_map_impl.rs
@@ -426,77 +426,72 @@ pub(super) mod archive {
                 let entry_count = location.entry_count;
                 let num_disk_submaps = metadata.num_submaps;
 
-                join_set.spawn(async move {
-                    tokio::task::spawn_blocking(move || {
-                        let file = File::open(&path)?;
-                        let mut reader = BufReader::with_capacity(READ_BUFFER_SIZE, file);
+                join_set.spawn_blocking(move || {
+                    let file = File::open(&path)?;
+                    let mut reader = BufReader::with_capacity(READ_BUFFER_SIZE, file);
 
-                        // Seek to the submap's byte offset
-                        reader.seek(io::SeekFrom::Start(byte_offset))?;
+                    // Seek to the submap's byte offset
+                    reader.seek(io::SeekFrom::Start(byte_offset))?;
 
-                        // Read submap index and count
-                        let disk_submap_idx: u64 =
-                            bincode::deserialize_from(&mut reader).map_err(invalid_data_error)?;
-                        let count: u64 = bincode::deserialize_from(&mut reader).map_err(invalid_data_error)?;
+                    // Read submap index and count
+                    let disk_submap_idx: u64 = bincode::deserialize_from(&mut reader).map_err(invalid_data_error)?;
+                    let count: u64 = bincode::deserialize_from(&mut reader).map_err(invalid_data_error)?;
 
-                        // Sanity check
-                        if count as usize != entry_count {
-                            return Err(invalid_data_error(format!(
-                                "Entry count mismatch: metadata says {}, file says {}",
-                                entry_count, count
-                            )));
+                    // Sanity check
+                    if count as usize != entry_count {
+                        return Err(invalid_data_error(format!(
+                            "Entry count mismatch: metadata says {}, file says {}",
+                            entry_count, count
+                        )));
+                    }
+
+                    let total_inserted = if num_disk_submaps == FIRST_LEVEL_SIZE {
+                        // Direct load: all entries go to the same submap, acquire lock once
+                        let target_idx = disk_submap_idx as usize;
+                        if target_idx >= FIRST_LEVEL_SIZE {
+                            return Err(invalid_data_error(format!("Invalid submap index: {target_idx}")));
                         }
 
-                        let total_inserted = if num_disk_submaps == FIRST_LEVEL_SIZE {
-                            // Direct load: all entries go to the same submap, acquire lock once
-                            let target_idx = disk_submap_idx as usize;
-                            if target_idx >= FIRST_LEVEL_SIZE {
-                                return Err(invalid_data_error(format!("Invalid submap index: {target_idx}")));
+                        let mut guard = inner.submaps[target_idx].lock().unwrap();
+                        let mut inserted = 0usize;
+
+                        for _ in 0..count {
+                            let k: Key = bincode::deserialize_from(&mut reader).map_err(invalid_data_error)?;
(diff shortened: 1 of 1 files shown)


utils/src/data_structures/parallel_hash_map_impl.rs
```diff
@@ -0,0 +1,527 @@
...
+
+            let mut join_set: JoinSet<io::Result<Vec<SubmapLocation>>> = JoinSet::new();
+
+            // Spawn parallel tasks, one for each data file
+            for file_idx in 0..num_files {
+                let inner = self.clone();
+                let dir = dir.clone();
+
+                join_set.spawn(async move {
+                    tokio::task::spawn_blocking(move || {
```
assafvayner: Should this just be join_set.spawn_blocking?

utils/src/data_structures/parallel_hash_map_impl.rs
```diff
@@ -0,0 +1,527 @@
...
+
+    fn hasher_path(dir: &Path) -> PathBuf {
+        dir.join(HASHER_FILENAME)
+    }
+
+    fn data_file_path(dir: &Path, file_index: usize) -> PathBuf {
+        dir.join(format!("data_{file_index}.bin"))
+    }
+
+    fn invalid_data_error(msg: impl Into<String>) -> io::Error {
```
assafvayner: May want to use `impl ToString`

utils/src/data_structures/parallel_hash_map_impl.rs
```diff
@@ -0,0 +1,527 @@
...
+        io::Error::new(ErrorKind::InvalidData, msg.into())
+    }
+
+    /// Reads hasher from archive directory.
+    fn read_hasher<S>(dir: &Path) -> io::Result<S>
+    where
+        S: for<'de> Deserialize<'de>,
+    {
+        let hasher_bytes = fs::read(hasher_path(dir))?;
+        let hasher: S = bincode::deserialize(&hasher_bytes).map_err(|e| invalid_data_error(e.to_string()))?;
```
assafvayner: In all of these closures: `|e| invalid_data_error(e.to_string())` since `e` as an error type implements `ToString/Into<String>` you should be able to use the `invalid_data_error` function as the whole function passed into `map_err`, no need for those closures.

utils/src/data_structures/parallel_hash_map_impl.rs
hoytak: Ah, yes!

utils/src/data_structures/parallel_hash_map_impl.rs
hoytak: Good insight.  Done.

utils/src/data_structures/parallel_hash_map_impl.rs
```diff
@@ -0,0 +1,510 @@
...
+                }
+
+                let permit = semaphore.clone().acquire_owned().await.unwrap();
+                let inner = self.clone();
+                let path = data_file_path(dir, location.file_index);
+                let byte_offset = location.byte_offset;
+                let entry_count = location.entry_count;
+                let num_disk_submaps = metadata.num_submaps;
+
+                join_set.spawn(async move {
```
seanses: Here

utils/src/data_structures/parallel_hash_map_impl.rs
seanses: There's a similar problem below

utils/Cargo.toml
```diff
@@ -22,36 +22,40 @@ derivative = { workspace = true }
...
 pin-project = { workspace = true }
 serde = { workspace = true }
 shellexpand = { workspace = true }
 thiserror = { workspace = true }
+tracing = { workspace = true }
+
+[target.'cfg(not(target_family = "wasm"))'.dependencies]
+bincode = { workspace = true }
+rand = { workspace = true }
+tempfile = { workspace = true }
```
seanses, line 35: Maybe put this under dev dependencies

hoytak: It's needed for the benchmark program, so not really in testing.  But I will do another PR here as we discussed to add simulation and benchmarking profiles to the compilation so it's not compiled all the time.

hoytak:
Decided to hold off on this until we get caching work done for this.

PR #614, opened 2026-01-23
Add hash_files() function to hf_xet Python package

rajatarya:
## Summary
Adds a `hash_files(file_paths: List[str])` function to the hf_xet Python package that computes xet hashes for files without uploading them. This enables fast, local-only file hashing without requiring authentication or server connection.

## Key Features
- **No authentication** or server connection required
- **Pure local computation** - no deduplication queries or network I/O
- **Results in same order** as input file paths
- **API consistency** - returns `PyXetUploadInfo` like `upload_files`

## Implementation
- Added `hash_single_file()` in data/src/data_client.rs for single file hashing
- Added `hash_files_async()` for parallel processing of multiple files
- Added Python binding `hash_files()` in hf_xet/src/lib.rs
- Reuses existing `Chunker` and `file_hash` infrastructure
- Uses `CONCURRENT_FILE_INGESTION_LIMITER` for controlled concurrency

## Testing
- ✅ 5 unit tests added (empty files, small files, determinism, errors, async)
- ✅ All tests passing
- ✅ Manual Python testing confirms correctness

## Usage Example
```python
import hf_xet

# Compute hashes without uploading
file_paths = ["/path/to/file1.txt", "/path/to/file2.txt"]
results = hf_xet.hash_files(file_paths)

for path, info in zip(file_paths, results):
    print(f"File: {path}, Hash: {info.hash}, Size: {info.file_size}")
```

### DIFF
--- a/data/src/data_client.rs
+++ b/data/src/data_client.rs
@@ -303,7 +303,7 @@ pub async fn clean_file(
 /// - Verify that downloaded files are correctly reassembled
 /// - Check if a file needs to be uploaded (by comparing hashes)
 /// - Generate cache keys for local file operations
-async fn hash_single_file(filename: String, buffer_size: usize) -> errors::Result<XetFileInfo> {
+fn hash_single_file(filename: String, buffer_size: usize) -> errors::Result<XetFileInfo> {
     let mut reader = File::open(&filename)?;
     let filesize = reader.metadata()?.len();
 
@@ -312,6 +312,10 @@ async fn hash_single_file(filename: String, buffer_size: usize) -> errors::Resul
     let mut chunk_hashes: Vec<(MerkleHash, u64)> = Vec::new();
 
     loop {
+        if XetRuntime::current().in_sigint_shutdown() {
+            return Err(std::io::Error::new(std::io::ErrorKind::Interrupted, "Operation interrupted by SIGINT").into());
+        }
+
         let bytes_read = reader.read(&mut buffer)?;
         if bytes_read == 0 {
             break;
@@ -364,9 +368,14 @@ pub async fn hash_files_async(file_paths: Vec<String>) -> errors::Result<Vec<Xet
     let semaphore = XetRuntime::current().global_semaphore(*CONCURRENT_FILE_INGESTION_LIMITER);
     let buffer_size = *xet_config().data.ingestion_block_size as usize;
 
-    let hash_futures = file_paths
-        .into_iter()
-        .map(|file_path| hash_single_file(file_path, buffer_size).instrument(info_span!("hash_file")));
+    let hash_futures = file_paths.into_iter().map(|file_path| {
+        async move {
+            tokio::task::spawn_blocking(move || hash_single_file(file_path, buffer_size))
+                .await
+                .map_err(|e| std::io::Error::other(e.to_string()))?
+        }
+        .instrument(info_span!("hash_file"))
+    });
 
     let files = run_constrained_with_semaphore(hash_futures, semaphore).await?;
 
@@ -487,7 +496,7 @@ mod tests {
         std::fs::write(&file_path, b"").unwrap();
 
         let buffer_size = 8 * 1024 * 1024; // 8MB
-        let result = hash_single_file(file_path.to_str().unwrap().to_string(), buffer_size).await;
+        let result = hash_single_file(file_path.to_str().unwrap().to_string(), buffer_size);
         assert!(result.is_ok());
 
         let file_info = result.unwrap();
@@ -503,7 +512,7 @@ mod tests {
         std::fs::write(&file_path, content).unwrap();
 
         let buffer_size = 8 * 1024 * 1024; // 8MB
-        let result = hash_single_file(file_path.to_str().unwrap().to_string(), buffer_size).await;
+        let result = hash_single_file(file_path.to_str().unwrap().to_string(), buffer_size);
         assert!(result.is_ok());
 
         let file_info = result.unwrap();
@@ -525,12 +534,12 @@ mod tests {
         let file_path_str = file_path.to_str().unwrap().to_string();
 
         // Hash with 8MB buffer size
-        let result1 = hash_single_file(file_path_str.clone(), 8 * 1024 * 1024).await;
+        let result1 = hash_single_file(file_path_str.clone(), 8 * 1024 * 1024);
         assert!(result1.is_ok());
         let file_info1 = result1.unwrap();
 
         // Hash with 4MB buffer size
-        let result2 = hash_single_file(file_path_str, 4 * 1024 * 1024).await;
+        let result2 = hash_single_file(file_path_str, 4 * 1024 * 1024);
         assert!(result2.is_ok());
         let file_info2 = result2.unwrap();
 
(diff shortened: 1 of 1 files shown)


data/src/data_client.rs
```diff
@@ -280,6 +282,94 @@ pub async fn clean_file(
...
+/// # Performance
+/// - Uses `CONCURRENT_FILE_INGESTION_LIMITER` to control parallelism
+/// - No authentication or server connection required
+/// - Pure local computation
+#[instrument(skip_all, name = "data_client::hash_files", fields(num_files=file_paths.len()))]
+pub async fn hash_files_async(file_paths: Vec<String>) -> errors::Result<Vec<XetFileInfo>> {
+    let semaphore = XetRuntime::current().global_semaphore(*CONCURRENT_FILE_INGESTION_LIMITER);
+
+    let hash_futures = file_paths.into_iter().map(|file_path| {
+        async move { tokio::task::spawn_blocking(move || hash_single_file(&file_path)).await? }
```
seanses: No need to spawn_blocking inside the future

data/src/data_client.rs
```diff
@@ -280,6 +282,94 @@ pub async fn clean_file(
...
+    let mut chunk_hashes: Vec<(MerkleHash, u64)> = Vec::new();
+
+    loop {
+        let bytes_read = reader.read(&mut buffer)?;
+        if bytes_read == 0 {
+            break;
+        }
+
+        let data = Bytes::copy_from_slice(&buffer[0..bytes_read]);
+        let is_final = bytes_read < buffer_size;
```
seanses: Bug: if a file size is multiple of buffer_size, this condition is never met, so there's a risk that the final chunk is never produced (if the hash of last 64 bytes of the file does't meet the cut condition). 

So in the loop always call `chunker.next_block_bytes(&data, false)`, and after the loop call `chunker.finish()` to get an `Option<Chunk>`.

data/src/data_client.rs
```diff
@@ -280,6 +282,96 @@ pub async fn clean_file(
...
+/// - Pure local computation
+#[instrument(skip_all, name = "data_client::hash_files", fields(num_files=file_paths.len()))]
+pub async fn hash_files_async(file_paths: Vec<String>) -> errors::Result<Vec<XetFileInfo>> {
+    let semaphore = XetRuntime::current().global_semaphore(*CONCURRENT_FILE_INGESTION_LIMITER);
+
+    let hash_futures = file_paths
+        .into_iter()
+        .map(|file_path| hash_single_file(file_path).instrument(info_span!("hash_file")));
+
+    let files = run_constrained_with_semaphore(hash_futures, semaphore).await?;
```
rajatarya, line 380: Because these are running async and the file sizes can be different, will the collection of results be in order?

The goal is this should have the same behavior as `upload_files()` - so if that returns the hash/file_size in the same order than this should as well.

rajatarya: Had Claude explain it to me - using `run_constrained_with_semaphore()` ensures ordering of results.

seanses: Yes, `run_constrained_with_semaphore` preserves the order of the input

data/src/data_client.rs
```diff
@@ -386,4 +478,113 @@ mod tests {
...
+        std::fs::write(&file_path, &content).unwrap();
+
+        let result = hash_single_file(file_path.to_str().unwrap().to_string()).await;
+        assert!(result.is_ok());
+
+        let file_info = result.unwrap();
+        assert_eq!(file_info.file_size(), file_size as u64);
+        assert!(!file_info.hash().is_empty());
+
+        // Verify the hash is deterministic - this would fail if chunker.finish()
```
seanses: I don't think this verifies what we intend to verify because `result` and `result2` are produced using the same config so there'll be no difference in the actual asm code execution.

I'd recommend to change the read buffer size (`xet_config().data.ingestion_block_size`) and test if this produces the same hash as the above. And maybe do the same thing for the `test_hash_determinism` test above.

rajatarya: Great feedback / catch! Updated in the latest commit!

data/src/data_client.rs
```diff
@@ -280,6 +282,96 @@ pub async fn clean_file(
...
+/// * `XetFileInfo` containing the hex-encoded hash and file size
+///
+/// # Errors
+/// * `IoError` if the file cannot be opened or read
+///
+/// # Use Cases
+/// - Verify that downloaded files are correctly reassembled
+/// - Check if a file needs to be uploaded (by comparing hashes)
+/// - Generate cache keys for local file operations
+async fn hash_single_file(filename: String) -> errors::Result<XetFileInfo> {
```
seanses: nit: This is a private function, so we can actually change the API to pass `buffer_size` in the parameter list, this would make it easier to write the tests below.

rajatarya: Sounds good.

rajatarya:
@Wauplin Does this interface match your thinking? Specifically the python `hash_files()` function in `hf_xet/src/lib.rs`?

rajatarya:
@seanses : Fanstastic feedback on this PR. Really appreciate you guiding me on how to improve this part of the codebase. My first time touching anything related to the chunker so really helped me. And, thanks for patient with me getting Claude to write this.

data/src/data_client.rs
```diff
@@ -280,6 +282,97 @@ pub async fn clean_file(
...
+///
+/// # Errors
+/// * `IoError` if the file cannot be opened or read
+///
+/// # Use Cases
+/// - Verify that downloaded files are correctly reassembled
+/// - Check if a file needs to be uploaded (by comparing hashes)
+/// - Generate cache keys for local file operations
+async fn hash_single_file(filename: String, buffer_size: usize) -> errors::Result<XetFileInfo> {
+    let mut reader = File::open(&filename)?;
```
XciD, line 307: should we use `tokio::fs::File` here ? maybe intentional ?

Wauplin:
> @Wauplin Does this interface match your thinking? Specifically the python `hash_files()` function in `hf_xet/src/lib.rs`?

Yes, that's perfect! Thanks for adding this :)

data/src/data_client.rs
hoytak: tokio::fs::File just wraps every operation in a spawn_blocking call around the std::fs version; it's not more efficient in general, so our standard is to not use it and manually control the spawn_blocking parts.  That said, I don't think this function needs to be async.

data/src/data_client.rs
```diff
@@ -280,6 +282,97 @@ pub async fn clean_file(
...
+/// * `XetFileInfo` containing the hex-encoded hash and file size
+///
+/// # Errors
+/// * `IoError` if the file cannot be opened or read
+///
+/// # Use Cases
+/// - Verify that downloaded files are correctly reassembled
+/// - Check if a file needs to be uploaded (by comparing hashes)
+/// - Generate cache keys for local file operations
+async fn hash_single_file(filename: String, buffer_size: usize) -> errors::Result<XetFileInfo> {
```
hoytak: So this function will read the entire file without a .await call, which means that a CTRL-C interrupt won't kick in until after a file is finished hashing.  I think we should change this function to non-async, wrap the call to it below in spawn_blocking so we don't tie up the async runtime with blocking threads, and then inside the loop below we check xet_runtime().in_sigint_shutdown() and break if so (fine to return an error).

data/src/data_client.rs
XciD: oh nice to know. thanks

data/src/data_client.rs
rajatarya: Oooh, good point, let me incorporate and update!

PR #617, opened 2026-01-28
Feature to monitor client process system usage

seanses:
Introduces a client benchmark utility to track system resource usage (CPU, memory, disk I/O, and network I/O) of a process, so we don't need to write scripts to capture usage stats according to different OS standards. This becomes extremely helpful when I benchmark on Python notebook instances, e.g. Google Colab, where system monitor is not easily accessible or when running a separate monitor script is not easy.

# Usage #
Users can enable monitoring by setting `HF_XET_SYSTEM_MONITOR_ENABLED` to true, set usage sample interval using `HF_XET_SYSTEM_MONITOR_SAMPLE_INTERVAL`, this outputs metrics to the tracing stream at `INFO` level by default. In addition, these metrics can be redirected to a separate file by setting sample log path using `HF_XET_SYSTEM_MONITOR_LOG_PATH`. Example:

Screenshot 2026-02-10 at 6 21 51 PM

# Output #
The stats are output in JSON format, which can be queried using tools like `jq`, e.g.
1. Trace of peak memory usage: `jq '.memory.peak_used_bytes' [HF_XET_SYSTEM_MONITOR_LOG_PATH]`
2. Trace of disk write speed: `jq '.disk.average_write_speed' [HF_XET_SYSTEM_MONITOR_LOG_PATH]`
3. Trace of network receive speed: `jq '.network.average_rx_speed' [HF_XET_SYSTEM_MONITOR_LOG_PATH]`


xet_runtime/src/runtime.rs
```diff
@@ -353,15 +372,32 @@ impl XetRuntime {
         F: Future + Send + 'static,
         F::Output: Send + Sync,
     {
-        self.external_executor_count.fetch_add(1, Ordering::SeqCst);
+        let _prev_count = self.external_executor_count.fetch_add(1, Ordering::SeqCst);
```
hoytak: I actually think, given our caches and other things that persist between external calls, that it would actually be best to tie this to the runtime itself instead of whether there is an external count.  The other thing is that technically the start and stop below can be out of order if one thread starts executing between lines 391 and 395 below.   Since the overhead of this sampling -- e.g. once per second or so -- is relatively low, I think it makes more sense to have it start when the runtime starts and drop when that drops.

xet_config/src/groups/system_monitor.rs
```diff
@@ -0,0 +1,20 @@
...
+
+    /// The path to write the system monitor output to.
+    /// If not set, the output will be written to tracing log at "Error" level.
+    /// If the pattern "#PID#" is present in the given value, it will be replaced
+    /// by the id of the current process.
+    ///
+    /// The default value is None.
+    ///
+    /// Use the environment variable `HF_XET_SYSTEM_MONITOR_OUTPUT_PATH` to set this value.
+    ref output_path: Option<String> = None;
```
hoytak: Let's use `{pid}` instead of `#PID#` for replacement -- I think that's more standard and also conventional with rust.  Also, let's also search for "{timestamp}" and insert the ISO timestamp there.

In fact, let's put the replacement logic in utils/src/file_paths.rs so this behavior is standardized.

utils/src/system_monitor.rs
```diff
@@ -0,0 +1,651 @@
...
+            .monitor_loop
+            .lock()
+            .map_err(|e| SystemMonitorError::Internal(e.to_string()))?;
+        self.stop.store(false, Ordering::Relaxed);
+
+        let sample_interval = self.sample_interval;
+        let output_path = self.output_path.clone();
+        let stop_clone = self.stop.clone();
+
+        *inner_runner = Some(std::thread::spawn(move || {
```
hoytak, line 471: Let's have one call to sample before the thread starts so that errors like a bad output_path show up immediately in the caller.

utils/src/system_monitor.rs
```diff
@@ -0,0 +1,651 @@
...
+                std::thread::sleep(sample_interval);
+                sampler.sample(&output_path)?;
+            }
+            Ok(())
+        }));
+
+        Ok(())
+    }
+
+    /// Stops the monitoring thread.
```
hoytak, line 511: Let's also add this to a Drop trait -- or just have the drop trait store true in stop flag, even if it doesn't

xet_config/src/groups/system_monitor.rs
```diff
@@ -0,0 +1,20 @@
+use std::time::Duration;
+
+crate::config_group!({
```
hoytak, line 5: Instead of a system profile, let's actually have a binary config option, enable, that enables sampling this.  I could see it useful to log this to debug client issues.

Thought there: 
- If it's not enabled, do nothing. 
- If it is enabled, and the path is specified in output_path below, write the json to that specific file.  
- If it is enabled, but no path is specified, dump it to the log using info!().

xet_config/src/groups/system_monitor.rs
hoytak: Second thought -- let's use log_path instead of output_path to make it clear it's logging the behavior.

seanses: Made the changes in a separate PR: https://github.com/huggingface/xet-core/pull/643

xet_runtime/src/runtime.rs
seanses: Updated

utils/src/system_monitor.rs
seanses: Updated

utils/src/system_monitor.rs
seanses: Done

xet_config/src/groups/system_monitor.rs
seanses: Updated

xet_config/src/groups/system_monitor.rs
seanses: Done

xet_runtime/src/runtime.rs
```diff
@@ -138,6 +140,10 @@ pub struct XetRuntime {
 
     // Primary configuration struct
     config: Arc<XetConfig>,
+
+    //  System monitor instance if enabled, monitor starts on initiation
+    #[cfg(not(target_family = "wasm"))]
+    _system_monitor: Option<SystemMonitor>,
```
hoytak: Is the _ prefix needed?  Seems it's actually used.

seanses: Oops, you're right. Previously not used because I forgot to stop the monitor in `perform_sigint_shutdown()`, and I forgot to change it.

PR #618, opened 2026-01-28
Unix Socket Proxy support.

hoytak:
This PR updates https://github.com/huggingface/xet-core/pull/598, which adds Unix domain socket support for the RemoteClient interface.  This version adds extensive testing using the LocalTestServer interface and creating a local proxy for it.

Use the environment variable HF_XET_CLIENT_UNIX_SOCKET_PATH=/path/to/socket to route all traffic through a Unix socket.  This is useful when running a XET client in a sandbox that doesn't have direct access to the network or for using tools like noxious_client to simulate bad network conditions.

### DIFF
--- a/cas_client/src/http_client.rs
+++ b/cas_client/src/http_client.rs
@@ -14,13 +14,6 @@ use xet_runtime::{XetRuntime, xet_config};
 
 use crate::{CasClientError, error};
 
-/// Returns the Unix socket path if configured via `HF_XET_CLIENT_UNIX_SOCKET_PATH` env var.
-#[cfg(not(target_family = "wasm"))]
-pub fn get_unix_socket_path() -> Option<String> {
-    use xet_runtime::xet_config;
-    xet_config().client.unix_socket_path.clone()
-}
-
 /// Middleware that rewrites https:// URLs to http:// when using Unix socket.
 /// This allows the proxy to parse plain HTTP and upgrade to HTTPS when forwarding.
 #[cfg(unix)]


hub_client/src/auth/interface.rs
```diff
@@ -5,6 +5,12 @@ use reqwest_middleware::RequestBuilder;
 #[async_trait]
 pub trait CredentialHelper: Send + Sync {
     async fn fill_credential(&self, req: RequestBuilder) -> Result<RequestBuilder>;
+
```
seanses: Let's make the change minimal  -- I don't think this added function is used anywhere?

hoytak: Weird, not sure how these snuck in here...  AI is weird sometimes.   But eliminated.

cas_client/src/http_client.rs
```diff
@@ -10,45 +10,99 @@ use reqwest_middleware::{ClientBuilder, ClientWithMiddleware, Middleware, Next};
...
+use xet_runtime::{XetRuntime, xet_config};
 
 use crate::{CasClientError, error};
 
+/// Returns the Unix socket path if configured via `HF_XET_CLIENT_UNIX_SOCKET_PATH` env var.
 #[cfg(not(target_family = "wasm"))]
-fn reqwest_client(user_agent: &str) -> Result<reqwest::Client, CasClientError> {
-    use xet_runtime::{XetRuntime, xet_config};
+pub fn get_unix_socket_path() -> Option<String> {
+    use xet_runtime::xet_config;
```
seanses: Probably no need to re-use, there is one at the top?

cas_client/src/http_client.rs
```diff
@@ -10,45 +10,99 @@ use reqwest_middleware::{ClientBuilder, ClientWithMiddleware, Middleware, Next};
...
+        if url.scheme() == "https" {
+            let original_scheme = url.scheme().to_string();
+            url.set_scheme("http").ok();
+            info!(original_scheme, new_scheme = "http", url = %url, "Rewriting URL scheme for Unix socket proxy");
+        }
+        next.run(req, extensions).await
+    }
+}
+
+#[allow(unused_variables)]
```
seanses, line 41: rm `#[allow(unused_variables)]`?

cas_client/src/http_client.rs
```diff
@@ -57,24 +111,38 @@ fn reqwest_client(user_agent: &str) -> Result<reqwest::Client, CasClientError> {
 }
 
 /// Builds authenticated HTTP Client to talk to CAS.
+#[allow(unused_mut)]
```
seanses, line 107: rm `#[allow(unused_mut)]`?

cas_client/src/http_client.rs
hoytak: Parts are not enabled on windows so the argument is ignored; this was easiest.

cas_client/src/http_client.rs
hoytak: Same reason -- variable doesn't need to be mutable on windows, just unix.

PR #619, opened 2026-01-28
Simulation tools for benchmarking xet-core

hoytak:
This PR adds in tooling for better simulation of local processes.  It's currently a work in progress, but the goal is to create tooling to easily simulate, test, and benchmark the client.  The goal is to ensure the upload and download paths work reliably in all environments.  

This PR: 
- Adds functionality to the LocalTestServer to simulate delays and server congestion.
- Reworks the API to allow for more flexibility in test scenarios.
- Adds dummy upload paths to allow simulations of raw data transfers.


hoytak:
Superseded by https://github.com/huggingface/xet-core/pull/648

Issue #604, opened 2025-12-30
Race condition in FileProvider causes corrupted blob content during parallel downloads


volodymyrjf:
## Description

This is a follow-up investigation to https://github.com/huggingface/huggingface_hub/issues/3643, which reported blob corruption during downloads. This issue identifies the root cause within xet-core.

## Summary

Files downloaded via XET storage have correct sizes but incorrect SHA-256 checksums due to a race condition in cas_client/src/output_provider.rs. The corruption is caused by parallel writers racing on the same output file.

## Race conditions occur because:
1. Multiple file descriptors to the same file race on metadata updates (file extension)
2. Seek and write are not atomic - another thread can interfere between them
3. OS-level buffering with multiple FDs can cause write ordering issues

### Proof-of-Concept Fix

To confirm the root cause, we implemented a mutex-based single shared file handle:

```
  struct FileProviderInner {
      filename: PathBuf,
      file: Mutex<Option<File>>,
  }

  struct SynchronizedFileWriter {
      inner: Arc<FileProviderInner>,
      position: u64,
  }

  impl Write for SynchronizedFileWriter {
      fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
          let mut file_guard = self.inner.file.lock()?;
          let file = file_guard.as_mut().ok_or_else(||
              std::io::Error::other("File not initialized"))?;

          file.seek(SeekFrom::Start(self.position))?;
          let written = file.write(buf)?;
          self.position += written as u64;

          Ok(written)
      }
  }
```

### Test Results
```
  | Version         | Iterations | Failed Iterations | File Corruption Rate | Avg Time |
  |-----------------|------------|-------------------|----------------------|----------|
  | Original hf_xet | 10         | 50% (5/10)        | ~9.1% (10/110)       | 46s      |
  | PoC fix (mutex) | 100        | 0% (0/100)        | 0% (0/1100)          | ~54s     |
```

The PoC fix eliminates corruption completely, confirming the root cause is in FileProvider.

Note: Our mutex-based solution is a proof-of-concept to demonstrate the issue, not a production-ready fix.

### Reproduction

download script
```
  from huggingface_hub import snapshot_download

  snapshot_download(
      repo_id="Qwen/Qwen-Image-Layered",
      repo_type="model",
      max_workers=24
  )
```

Running this repeatedly shows ~50% of iterations have at least one corrupted file (~9% file corruption rate).

[...]

PR #641, opened 2026-02-06
Fix writev exceeding limits

seanses:
On macOS and Linux, `writev(int fildes, const struct iovec *iov, int iovcnt)` may return EINVAL if
- the sum of the iov_len values in the iov array overflows a 32-bit integer (macOS) or an ssize_t value (Linux);
- iovcnt is less than or equal to 0, or greater than UIO_MAXIOV (POSIX standard IOV_MAX, value 1024); and specially on Linux, the glibc wrapper functions do some extra work if they detect that the underlying kernel system call failed because this limit was exceeded. The wrapper function would allocate a temporary buffer large enough for all of the items specified by iov, copies data from iov to this buffer, and passes the buffer in a call to write().

To avoid these potential syscall failures or performance degradation, we put a limit on the total number of bytes and number of slices to call `writev()`. Also adding unit tests for these two limits.

Fix https://github.com/huggingface/xet-core/issues/604, fix XET-813. Special thanks to @volodymyrjf for the investigation.

### DIFF
--- a/file_reconstruction/src/data_writer/sequential_writer.rs
+++ b/file_reconstruction/src/data_writer/sequential_writer.rs
@@ -20,10 +20,9 @@ use crate::{ErrorState, FileReconstructionError, Result};
 //  system call failed because this limit was exceeded. The wrapper function would allocate a temporary buffer large
 //  enough for all of the items specified by iov, copies data from iov to this buffer, and passes the buffer in a
 //  call to write().
-// To avoid these potential syscall failures or performance degradation, we put a limit on the sum of the iov_len values
-// in the iov array and on iovcnt.
-const WRITEV_MAX_BYTES: usize = 1024 * 1024 * 1024; // 1 GB
-const WRITEV_MAX_SLICE: usize = 128;
+// To avoid these potential syscall failures or performance degradation, we limit iovcnt to 24. Given our max Xorb size
+// 64 MiB, this effectively limits total number of bytes in iov to 64 MiB * 24 = 1.5 GiB.
+const WRITEV_MAX_SLICE: usize = 24;
 
 /// Items sent to the background writer thread.
 enum QueueItem {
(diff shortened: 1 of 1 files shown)


hoytak:
So, I'm looking into this, and it turns out that with our buffer sizes -- 64KB typical minimum, 64MB common and maximal -- calls to `writev` likely have diminishing returns past more than a handful of buffers.   So, given that our buffers have a maximum size, I'm inclined to just put in a hard limit of 24 vectored writes as a constant, which would be still testable and not require any config settings or arguments passed around.  I feel like it's overkill to make this a config setting since it's not really something that should have an impact on the user experience.

seanses:
> So, I'm looking into this, and it turns out that with our buffer sizes -- 64KB typical minimum, 64MB common and maximal -- calls to `writev` likely have diminishing returns past more than a handful of buffers. So, given that our buffers have a maximum size, I'm inclined to just put in a hard limit of 24 vectored writes as a constant, which would be still testable and not require any config settings or arguments passed around. I feel like it's overkill to make this a config setting since it's not really something that should have an impact on the user experience.

Ah it actually meant to just use a constant setting like below
```
pub fn new_vectorized<W: Write + Send + 'static>(writer: W) -> Self {
        // with default limits for Linux and macOS
        Self::new_internal(writer, true, WRITEV_MAX_BYTES, WRITEV_MAX_SLICE)
    }
```
, then configurable constructor is only there for writing tests:
```
#[cfg(test)]
    fn new_vectorized_with_config<W: Write + Send + 'static>(
        writer: W,
        writev_max_bytes: usize,
        writev_max_slice: usize,
    ) -> Self {
        Self::new_internal(writer, true, writev_max_bytes, writev_max_slice)
    }
```
I'm happy to reduce `WRITEV_MAX_SLICE` from 128 to 24.

PR #642, opened 2026-02-09
feat(data): introduce streaming client

kszucs:
Several query engines use the datfusion/rust ecosystem eventually depending on Apache OpenDAL. We also use opendal in the dataset viewer. 

In order to support seamless hf:// experience throughout the ecosystem I added the missing features to opendal's huggingface backend, one particular is Xet downlod/upload support hence this PR. 

Opendal has a specific API requirements for streaming upload and download so I created a `XetClient` and corresponding `XetWriter` and `XetReader` to stream downloads/uploads using the existing xet machinery and utilities. 

This changeset is actually used by https://github.com/apache/opendal/pull/7185


rajatarya:
Hey @kszucs : @seanses is working on a new session-based interface that will simplify making a Rust xet crate right now. We would want that interface to be used for adding streaming support. Can you work with him to align the efforts?

data/src/xet_file.rs
```diff
@@ -20,7 +24,26 @@ impl XetFileInfo {
...
+    }
+
+    /// Creates a new `XetFileInfo` instance with SHA256.
+    ///
+    /// # Arguments
+    ///
+    /// * `hash` - The Xet hash of the file. This is a Merkle hash string.
+    /// * `file_size` - The size of the file.
+    /// * `sha256` - The SHA256 hash of the file.
+    pub fn with_sha256(hash: String, file_size: u64, sha256: String) -> Self {
```
kszucs, line 41: In order to commit to the hub I need the calculated sha256 hash.

data/src/file_upload_session.rs
```diff
@@ -260,13 +260,14 @@ impl FileUploadSession {
     pub async fn start_clean(
         self: &Arc<Self>,
         file_name: Option<Arc<str>>,
-        size: u64,
+        size: Option<u64>,
```
kszucs, line 259: When writing using opendal the size is not known ahead of time, so in order to avoid collecting the data I had to implement this workaround. Ideally we could turn of progress tracking but seems like its API is more coupled with the general upload process.

hoytak: Turning it off is done by simply using the NoOp version, so it's possible to do.  We should definitely take that situation into account.

kszucs: Great, thanks for the hint! Trying it out.

kszucs: Took a closer look. While I can turn off the actual reporting by passing noop, `CompletionTracker` is still running and verifying (in dev mode) the upload so I ended up making total_bytes optional essentially signaling that total_bytes is not known before uploading.

hoytak: Right. I had to handle the known/unknown case in the download progress tracking in a parallel PR.  There it explicitly tries to update the total as more data is streamed if it's not known. The issue with making it optional is that reporting functions elsewhere use the ratio heavily, which would be problematic.  Let me put up a PR quick to add this same feature to the upload tracking.

hoytak: Put a PR up for this at https://github.com/huggingface/xet-core/pull/651.  This should satisfy your use case and satisfy the needed invariants for the reporting UX in both cases.

kszucs:
In the meantime I tree-shaked it into a more easily publishable single crate at https://github.com/kszucs/subxet using https://github.com/kszucs/cargo-subset.

rajatarya:
Circling back to this - @kszucs is this still relevant? I believe with hf-xet 1.5 @seanses we added streaming support using the XetSession classes.

kszucs:
No, not relevant anymore just forgot to close it.

PR #643, opened 2026-02-09
TemplatedPathBuf utility

seanses:
As requested in https://github.com/huggingface/xet-core/pull/617#discussion_r2765636811, this implements a utility for configuring path-like parameters. 

This folds inside the existing function `fn normalized_path_from_user_string` that expands `~` to home directory and converts to absolute paths, and evaluates a path template by substituting **case-insensitive** placeholders with corresponding values:
- `{pid}` for process ID,
- `{timestamp}` for ISO 8601 local timestamp with offset

For example,
```
let template = TemplatedPathBuf::new("~/logs/app_{PID}_{TIMESTAMP}.txt");
let path = template.as_path();
/// Returns an absolute path like "/home/user/logs/app_12345_2024-01-15T10-30-45-0500.txt"
```
or to be used directly in config groups:
```
crate::config_group!({
    ref log_path: Option<TemplatedPathBuf> = None;
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,15 +3,6 @@
 version = 4
 
 [[package]]
-name = "addr2line"
-version = "0.24.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
-dependencies = [
- "gimli",
-]
-
-[[package]]
 name = "adler2"
 version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -33,7 +24,7 @@ version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
- "cfg-if 1.0.0",
+ "cfg-if 1.0.4",
  "cipher",
  "cpufeatures",
 ]
@@ -166,15 +157,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
-name = "ascii-canvas"
-version = "3.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8824ecca2e851cec16968d54a01dd372ef8f95b244fb84b84e70128be347c3c6"
-dependencies = [
- "term",
-]
-
-[[package]]
 name = "assert-json-diff"
 version = "2.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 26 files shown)


utils/src/file_paths.rs
```diff
@@ -1,47 +1,188 @@
...
+        self.eval_impl(&substitutes)
+    }
+
+    /// Note: Templates can come from environment variables or other non-UTF-8 sources.
+    /// We work at the byte level using `into_encoded_bytes()` to preserve all path data
+    /// across all platforms (Unix raw bytes, Windows WTF-8), only interpreting ASCII
+    /// patterns like {PID} and {TIMESTAMP} which are guaranteed to be ASCII.
+    fn eval_impl(&self, substitutes: &[Box<dyn Substitute>]) -> PathBuf {
+        // Get platform-specific encoded bytes (Unix: raw bytes, Windows WTF-8)
+        let path_bytes = self.0.as_os_str().as_encoded_bytes();
```
hoytak: Does this work on windows?  There I believe the characters in an OsStr are UTF-16, which means that each character is actually 2 bytes...

seanses: Yes. According to [wtf-8.codeberg.page](https://wtf-8.codeberg.page/#wtf-8:~:text=On%20Windows%20(which%20uses%20potentially%20ill%2Dformed%20UTF%2D16%20in%20its%20APIs)%2C%20the%20Rust%20standard%20library%20uses%20WTF%2D8%20internally%20for%20OS%20strings)
> On Windows (which uses potentially ill-formed UTF-16 in its APIs), the Rust standard library [uses WTF-8](https://github.com/rust-lang/rfcs/blob/master/text/0517-io-os-reform.md#string-handling) internally for OS strings

WTF-8, like UTF-8, is ASCII transparent so scanning for ASCII characters is safe. Also see one example in the official doc:
https://doc.rust-lang.org/std/ffi/struct.OsString.html#example

Added more tests for non-ASCII paths.

hoytak: Ah, nice.

utils/src/file_paths.rs
```diff
@@ -1,47 +1,188 @@
...
+    ///
+    /// ```
+    /// use utils::TemplatedPathBuf;
+    ///
+    /// let template = TemplatedPathBuf::new("~/logs/app_{PID}_{TIMESTAMP}.txt");
+    /// let path = template.evaluated();
+    /// // Returns an absolute path like "/home/user/logs/app_12345_2024-01-15T10-30-45-0500.txt"
+    /// // (timestamp in local timezone with offset appended)
+    /// ```
+    pub fn evaluated(&self) -> PathBuf {
```
hoytak: I think we should do all the conversion on construction -- we want the timestamp to be constant across uses of the instance.

utils/src/file_paths.rs
```diff
@@ -1,47 +1,188 @@
...
+        let substitutes: Vec<Box<dyn Substitute>> = vec![
+            Box::new(SubstitutePid::default()),
```
hoytak: Seems a little heavy to have new structs defined for each of this -- could we just use

```
substitutes : [(&str, fn() -> String); 2] = [ ("pid", || {std::process::id().to_string() }), ("timestamp" : || {...})]; 
```

utils/src/file_paths.rs
seanses: Updated, now a evaluated copy is stored in the instance and returned by `as_path(&self)`.
Also introduced a `re_evaluate()` function in case re-evaluation is intended, e.g. after a fork.

utils/src/file_paths.rs
seanses: Updated to use `type Substitute = (&'static str, Box<dyn Fn() -> String>);`. `Box<dyn >` is needed because each closure is different type.

PR #644, opened 2026-02-09
Upgrade package dependencies

hoytak:
This PR updates all the package dependencies that would not cause significant API breakages to the current version.  The package versions in hf_xet_wasm and hf_xet are also updated to match the versions in the base package.   There should be no functional change.

### DIFF
--- a/cas_client/Cargo.toml
+++ b/cas_client/Cargo.toml
@@ -48,21 +48,19 @@ strict = []
 default = ["rustls-tls"]
 elevated_information_level = []
 
-# Three options for compliation here.
-rustls-tls = [
-    "reqwest/rustls",
-    "reqwest/rustls-native-certs",
-    "reqwest/webpki-roots",
-]
+# rustls-tls uses the rustls package with rustls-platform-verifier, which verifies TLS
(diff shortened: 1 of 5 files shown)


cas_client/Cargo.toml
```diff
@@ -50,9 +50,9 @@ elevated_information_level = []
 
 # Three options for compliation here.
 rustls-tls = [
-    "reqwest/rustls-tls",
-    "reqwest/rustls-tls-webpki-roots",
-    "reqwest/rustls-tls-native-roots",
+    "reqwest/rustls",
+    "reqwest/rustls-native-certs",
```
seanses: It seems `reqwest-0.13.2` has dropped features `rustls-native-certs` and `webpki-roots` and adopted a more robust dep `rustls_platform_verifier` and is enabled by default. Can we switch to this version?

hoytak: So we can upgrade to `reqwest-0.13.1` and still have the rustls-platform-verifier feature, but 0.13.2 actually requires a wasm-bindgen upgrade, which @assafvayner mentioned requires riskier upstream changes so would be better to do separately.   So I'll do that.

PR #645, opened 2026-02-10
Progress reporting for downloads.

hoytak:
This PR adds detailed progress reporting to the download path. 
- Transfer progress is reported as soon as the download streams start; actual bytes written are reported as the reconstructed file is written out. 
- Currently, each call to download_file creates a separate progress tracker, but this sets up for download groups with grouped download progress tracking. 
 
To support this, the UploadProgressStream was split into three classes; a common StreamProgressReporter and download and upload specific versions.  This also allows us to simplify the API to RetryWrapper.

More tracking was added to the file reconstruction paths to properly report progress.

### DIFF
--- a/cas_client/src/remote_client.rs
+++ b/cas_client/src/remote_client.rs
@@ -369,7 +369,9 @@ impl Client for RemoteClient {
                                         expected,
                                         buffer.len()
                                     );
-                                    warn!("get_file_term_data: expected {} bytes, got {}", expected, buffer.len());
+                                    if expected != buffer.len() {
+                                        warn!("get_file_term_data: expected {} bytes, got {}", expected, buffer.len());
+                                    }
                                 }
                                 Ok((Bytes::from(buffer), chunk_byte_indices))
                             },
--- a/cas_client/src/progress_tracked_streams.rs
+++ b/cas_client/src/progress_tracked_streams.rs
@@ -5,9 +5,9 @@ use std::task::{Context, Poll};
 
 use bytes::Bytes;
 use futures::Stream;
-use more_asserts::*;
+use more_asserts::debug_assert_le;
 
-// The progress callback function type which reports delta (change since previous), portion completed, total.
+/// Progress callback receiving (delta, completed, total) in bytes.
 pub type ProgressCallback = Arc<dyn Fn(u64, u64, u64) + Send + Sync + 'static>;
 
 /// Helper for progress reporting with reset/resume semantics. When a stream is
--- a/hf_xet_thin_wasm/Cargo.lock
+++ b/hf_xet_thin_wasm/Cargo.lock
@@ -12,6 +12,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "android_system_properties"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "anstream"
 version = "0.6.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -140,6 +149,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "bstr"
+version = "1.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "63044e1ae8e69f3b5a92c736ca6269b8d12fa7efe39bf34ddb06d102cf0e2cab"
+dependencies = [
+ "memchr",
+ "regex-automata",
+ "serde",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -190,7 +210,11 @@ version = "0.4.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "145052bdd345b87320e369255277e3fb5152762ad123a901ef5c262dd38fe8d2"
 dependencies = [
+ "iana-time-zone",
+ "js-sys",
  "num-traits",
+ "wasm-bindgen",
+ "windows-link",
 ]
 
 [[package]]
@@ -767,6 +791,30 @@ dependencies = [
 ]
 
 [[package]]
+name = "iana-time-zone"
+version = "0.1.65"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e31bc9ad994ba00e440a8aa5c9ef0ec67d5cb5e5cb0cc7f8b744a35b389cc470"
+dependencies = [
+ "android_system_properties",
+ "core-foundation-sys",
+ "iana-time-zone-haiku",
+ "js-sys",
+ "log",
+ "wasm-bindgen",
+ "windows-core",
+]
+
+[[package]]
+name = "iana-time-zone-haiku"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f31827a206f56af32e590ba56d5d2d085f558508192593743f16b2306495269f"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "icu_collections"
 version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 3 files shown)


rajatarya:
## Review Notes

Reviewed the full diff (+1,824 / -661 across 42 files). The architecture is clean — separating transfer bytes (network) from decompressed bytes (disk) through independent call sites is the right approach, and extracting progress from `RetryWrapper` internals is a nice simplification.

### Items worth discussing

1. **Relaxed atomics ordering** — All counters use `Relaxed`. This is fine for monotonic counters, but `report_bytes_written` has `debug_assert_le!(bytes_completed, item_total_bytes)` where `item_total_bytes` is loaded with `Relaxed` from a different atomic that may be updated concurrently by `update_item_size`. Under weak memory ordering, the assert could theoretically see a stale `item_total_bytes` while seeing a current `bytes_completed`. Unlikely to fire in practice, but worth considering `Acquire`/`Release` on the total-size stores/loads if the asserts matter.

2. **Fire-and-forget `tokio::spawn` for progress updates** — `DownloadTaskUpdater` sends all progress via `tokio::spawn(async move { inner.register_updates(...).await })`. If the runtime is shutting down or under pressure, these can be silently dropped. Fine for UI progress bars, but worth a comment documenting this is intentional.

3. **`cfg!(debug_assertions)` default tracker in `FileReconstructor::new()`** — In debug builds, every `FileReconstructor` creates a `correctness_verification_tracker` even when no progress tracking is needed, and `assert_complete()` runs at the end. This means debug-mode downloads are strictly validated (good for catching bugs), but it's a behavioral difference from release that's easy to miss. Consider a comment explaining the intent.

4. **`determine_size_if_possible` returning `None`** — When the DP can't chain chunk ranges, `remote_client.rs` falls back to `Vec::with_capacity(0)`, meaning no pre-allocation for the decompressed buffer. Is this the common or rare case? If common, might be worth a follow-up to handle it.

5. **`merge_in` logic change in `progress_info.rs`** — The merge now conditionally adds `other.total_bytes_increment` when `other.total_bytes == 0`, instead of always taking `max`. This subtly changes behavior when both sides have non-zero totals with increments — worth verifying this matches the intended semantics.

6. **Unrelated changes** — git-lfs availability guards in install/uninstall tests, `sort_by` → `sort_by_key(Reverse(...))` in `shard_file_manager.rs` and `logging.rs`, and wasm formatting changes are mixed in. Minor, but could be split out.

### What looks good

- `StreamProgressReporter` with `fetch_max` high-water-mark for retry dedup is elegant
- `RetryWrapper` simplification (`Fn(Option<Arc<dyn Fn(f64, u64)>>)` → `Fn()`) reduces cognitive load across ~15 call sites
- `cas_client` no longer depends on `progress_tracking` — good decoupling
- Test coverage is solid: ~600 new test lines covering the DP algorithm (16 cases), `DownloadTaskUpdater` flows (14 cases), stream progress (4 cases), and end-to-end `FileReconstructor` progress (4 cases)
- `is_final` semantics on `update_item_size` cleanly handle both the data_client (size-known-upfront) and reconstruction (incremental-discovery) paths

rajatarya:
(I'm still working through Claude-assisted review of this PR - will update with more comments/questions as I get through it. But this summary raises some good suggestions around adding comments.)

hoytak:
Good feedback; updated in response to a couple things there (the atomic ordering, the merge_in change, and the unrelated changes).

progress_tracking/src/download_tracking.rs
```diff
@@ -0,0 +1,571 @@
...
+            item_bytes: 0.into(),
+            item_transfer_bytes: 0.into(),
+            item_size_finalized: false.into(),
+            bytes_completed: 0.into(),
+            transfer_bytes_completed: 0.into(),
+            tracker,
+        }
+    }
+
+    pub fn correctness_verification_tracker() -> Arc<Self> {
```
rajatarya, line 72: As a debug only checker, can this function be annotated that way?

file_reconstruction/src/reconstruction_terms/xorb_block.rs
```diff
@@ -18,12 +20,22 @@ pub struct XorbBlockData {
...
+    pub uncompressed_size: usize,
+}
+
 /// A downloadable xorb block identified by hash and chunk range, with cached data.
 /// Multiple file terms may reference the same xorb block.
 pub struct XorbBlock {
     pub xorb_hash: MerkleHash,
     pub chunk_range: ChunkRange,
     pub xorb_block_index: usize,
+    /// All file-term chunk ranges covered by this xorb block, sorted by range start.
```
rajatarya, line 36: Needs a comment for the pub struct here.

file_reconstruction/src/file_reconstructor.rs
```diff
@@ -47,7 +47,13 @@ impl FileReconstructor {
             file_hash,
             byte_range: None,
             output,
-            progress_updater: None,
+            progress_updater: {
+                if cfg!(debug_assertions) {
+                    Some(DownloadTaskUpdater::correctness_verification_tracker())
```
rajatarya, line 44: nit: I'd love it if there was a comment explaining that this is a test hook for unit-tests and is compiled out.

file_reconstruction/src/file_reconstructor.rs
```diff
@@ -184,7 +196,13 @@ impl FileReconstructor {
                     })?;
 
                 // Get the data future that will retrieve the term's data.
-                let data_future = file_term.get_data_task(client.clone()).await?;
+                let data_future = file_term.get_data_task(client.clone(), progress_updater.clone()).await?;
+
+                #[cfg(debug_assertions)]
+                {
+                    let refs = &file_term.xorb_block.references;
```
rajatarya, line 172: It would be good if there was a comment describing this assert. Not sure if folks down the road will remember why these debug assertions exist.

file_reconstruction/src/file_reconstructor.rs
```diff
@@ -495,6 +525,131 @@ mod tests {
         reconstruct_and_verify_full(&client, &file_contents, test_config()).await;
     }
 
+    // ==================== Progress tracker tests ====================
```
rajatarya, line 499: Love seeing all the new tests!

cas_client/src/retry_wrapper.rs
```diff
@@ -226,35 +226,23 @@ impl RetryWrapper {
         let self_ = Arc::new(self);
         let try_count = AtomicUsize::new(0);
 
-        // Extract the partial progress reporting function from the connection permit if it exists
```
rajatarya, line 229: Love seeing this progress reporting getting pulled out of retry wrapping. Definitely cleaner abstractions this way!

cas_client/src/remote_client.rs
```diff
@@ -329,19 +341,45 @@ impl Client for RemoteClient {
...
+                        let result = cas_object::deserialize_async::deserialize_chunks_to_writer_from_stream(
+                            incoming_stream,
+                            &mut writer,
+                        )
+                        .await;
+
+                        match result {
+                            Ok((_compressed_len, chunk_byte_indices)) => {
+                                if let Some(expected) = uncompressed_size_if_known {
+                                    debug_assert_eq!(
```
rajatarya, line 365: I think this should log `error!()` or `warn!()` if the expected bytes are not returned.

This code still returns Ok() for that case, does that mess things later on downstream?

cas_client/src/remote_client.rs
```diff
@@ -492,29 +532,12 @@ impl Client for RemoteClient {
...
-                            }
-                        };
-
-                        let upload_stream = upload_stream.clone_with_reset_and_new_callback(progress_callback);
+                    .run_and_extract_json(move || {
+                        let upload_stream = UploadProgressStream::wrap_bytes_as_stream(
+                            serialized_data.clone(),
+                            block_size,
+                            upload_reporter.clone(),
+                        );
```
rajatarya, line 543: Love seeing the progress reporting have its own clean abstraction and having Upload and Download have their own ProgressStream implementations.

rajatarya (approved):
Love seeing these changes. Did as thorough a review as I could with all the file changes - and overall I LOVE these changes. Minor comments on adding docstrings and comments.

cas_client/src/remote_client.rs
hoytak: In this case, it could mean that the remote data is corrupted somehow, so I'm not sure what to do at that point.  It's one of the things that should never happen.  Practically, however, it will still use the remote data as reference and proceed as normal...

PR #648, opened 2026-02-11
Rework simulation pipeline for adaptive concurrency and connection resiliency. 

hoytak:
This PR replaces the previous collection of scripts around setting up docker containers with a much more nimble and lightweight set of rust scripts and a simple, reusable proxy that can limit bandwidth and congestion simulations.   The previous scripts are rewritten to be more nimble and use more reusable components. 

New tools: 
- cas_client/src/simulation/network_simulation: A lightweight, in-process network congestion simulation proxy that lives between the LocalServer instance and the RemoteClient instance, allowing simulation tests to run on a network with realistic congestion conditions and a gated bandwidth.   This can be controlled dynamically through a LocalTestServer instance.
- simulation/: A new package for collecting simulation scripts and analyzing the results.  

To run the new simulation scripts for the adaptive concurrency on upload, compile in release mode and run one of the scripts in `simulation/src/adaptive_concurrency/scripts/`.  Docker is no longer needed to run any of the simulations. 

The old `cas_client/tests/adaptive_concurrency/` paths were removed.

### DIFF
--- a/cas_client/src/simulation/local_server/server.rs
+++ b/cas_client/src/simulation/local_server/server.rs
@@ -396,6 +396,7 @@ impl LocalTestServer {
     }
 
     /// Returns the deletion-capable client if the server backend supports it.
+    #[allow(unused)]
     pub fn deletion_client(&self) -> Option<&Arc<dyn DeletionControlableClient>> {
         self.deletion_client.as_ref()
     }


cas_client/src/simulation/local_server/handlers.rs
```diff
@@ -32,9 +32,23 @@ use futures_util::StreamExt;
...
+#[derive(Clone)]
+pub struct ServerState {
+    pub client: Arc<dyn DirectAccessClient>,
+    pub latency_simulation: Arc<LatencySimulation>,
+}
+
+// ServerState is Send + Sync because:
+// - Arc<dyn DirectAccessClient> is Send + Sync (DirectAccessClient requires Send + Sync)
+// - Arc<LatencySimulation> is Send + Sync (LatencySimulation contains only Send + Sync types)
+unsafe impl Send for ServerState {}
```
rajatarya: `unsafe impl Send for ServerState {}` and `unsafe impl Sync for ServerState {}` appear unnecessary. `ServerState` contains `Arc<dyn DirectAccessClient>` (trait requires `Send + Sync`) and `Arc<LatencySimulation>` (all inner types are `Send+Sync`). The compiler should derive these automatically. Remove the `unsafe` impls — if they're actually needed, it indicates a deeper problem that should be investigated.

simulation/src/bin/run_upload_scenario.rs
```diff
@@ -0,0 +1,215 @@
...
+// ── Utilities ────────────────────────────────────────────────────────────────
+
+/// Parses a duration string (duration_str crate: e.g. "10ms", "1s").
+fn parse_duration(s: &str) -> ScenarioResult<Duration> {
+    duration_str::parse(s).map_err(|e| ScenarioError::Scenario(format!("invalid duration {:?}: {}", s, e)))
+}
+
+fn setup_logging(out_dir: &Path) {
+    let log_dest = format!("{}/", out_dir.display());
+    unsafe { std::env::set_var("HF_XET_LOG_DEST", &log_dest) };
```
rajatarya, line 138: `unsafe { std::env::set_var(...) }` is unsound in a multi-threaded context. `set_var` was made `unsafe` in Rust 1.83+ precisely because it is not thread-safe. This is called after `main()` starts, so other threads may already be running. Either set the env var before spawning any threads or use a different mechanism to configure the tracing subscriber directly.

cas_client/src/simulation/local_server/handlers.rs
```diff
@@ -481,6 +522,191 @@ async fn collect_body(body: Body) -> Result<Bytes, String> {
...
+    if min > max {
+        return Err(format!("Min duration ({:?}) cannot be greater than max ({:?})", min, max));
+    }
+
+    Ok(min..max)
+}
+
+pub async fn set_config(State(state): State<ServerState>, uri: axum::http::Uri, body: Body) -> Response {
+    // Try to parse as JSON body first
+    let body_bytes = match axum::body::to_bytes(body, usize::MAX).await {
```
rajatarya: `axum::body::to_bytes(body, usize::MAX)` has no body size limit. A misbehaving client could send an arbitrarily large body causing OOM. Even in simulation-only context, a reasonable limit (e.g., 1MB) would be more defensive.

cas_client/src/simulation/network_simulation/bandwidth_limit_router.rs
```diff
@@ -0,0 +1,365 @@
...
+                }
+            }
+        });
+    }
+
+    /// Call from a spawned task. Accepts connections and spawns a handler for each; exits when shutdown_flag is set.
+    pub async fn run_accept_loop(self: Arc<Self>) {
+        let listener = {
+            let mut guard = self.listener.lock().await;
+            guard.take().expect("accept loop already started or listener taken")
```
rajatarya: `guard.take().expect("accept loop already started or listener taken")` will panic if `run_accept_loop` is called twice. Since this is infrastructure code (not just tests), consider returning a `Result` instead.

cas_client/src/simulation/network_simulation/bandwidth_limit_router.rs
```diff
@@ -0,0 +1,365 @@
...
+/// Each semaphore permit represents this many bytes (1 KiB). Permits are acquired/issued in these units to stay within
+/// u32.
+const BASE_BANDWIDTH_PERMIT_SIZE: u64 = 1024;
+
+/// Depth of the latency simulation pipe — limits how many chunks can be "in flight"
+/// between the bandwidth-limited reader and the latency-delayed writer.
+const LATENCY_PIPE_DEPTH: usize = 16;
+
+/// Single struct for the network simulation proxy. Methods take `Arc<Self>` for use in async tasks.
+pub struct NetworkSimulationProxy {
```
rajatarya, line 35: All fields of `NetworkSimulationProxy` are `pub(crate)`, exposing internal state (semaphores, atomics, shutdown flags, listener) to the entire crate. These should be private — the struct already has proper public methods for all needed operations.

cas_client/src/remote_client.rs
```diff
@@ -114,6 +114,10 @@ impl RemoteClient {
         &self.endpoint
     }
 
+    pub fn http_client(&self) -> Arc<ClientWithMiddleware> {
```
rajatarya: New `pub fn http_client(&self) -> Arc<ClientWithMiddleware>` exposes the internal HTTP client with its middleware stack (retries, etc.) publicly. This increases the API surface and allows arbitrary HTTP calls through the client's middleware. Consider whether a more targeted API (e.g., exposing just the base URL) would suffice.

cas_client/src/simulation/network_simulation/bandwidth_limit_router.rs
```diff
@@ -0,0 +1,365 @@
...
+            if let Some(ref lim) = limiter {
+                let permits = (n as u64).div_ceil(BASE_BANDWIDTH_PERMIT_SIZE).min(u32::MAX as u64) as u32;
+                let permit = lim.acquire_many(permits).await.map_err(|_| {
+                    std::io::Error::new(std::io::ErrorKind::ConnectionReset, "bandwidth limiter closed")
+                })?;
+                permit.forget();
+            }
+            let delay_secs = base_secs + rand::rng().random_range(-jitter_secs..=jitter_secs);
+            let delivery_time = Instant::now() + Duration::from_secs_f64(delay_secs.max(0.0));
+            if tx.send((delivery_time, buf[..n].to_vec())).await.is_err() {
```
rajatarya: Nit: `buf[..n].to_vec()` allocates a new `Vec` for every 64KB chunk in the latency pipeline hot path. Consider using `Bytes::copy_from_slice` or a buffer pool to reduce allocation pressure at high throughput.

cas_client/src/simulation/direct_access_client.rs
```diff
@@ -28,7 +28,7 @@ use crate::interface::Client;
 /// local server to work with either backend.
 #[cfg_attr(not(target_family = "wasm"), async_trait)]
 #[cfg_attr(target_family = "wasm", async_trait(?Send))]
-pub trait DirectAccessClient: Client + Send + Sync {
+pub trait DirectAccessClient: Client + Send + Sync + 'static {
```
rajatarya: Adding `'static` to `DirectAccessClient: Client + Send + Sync + 'static` is a breaking trait change — any downstream implementor holding non-`'static` references will no longer compile. Worth a comment explaining why this is needed (likely for `Arc<dyn DirectAccessClient>` in async contexts).

rajatarya (commented):
## Code Review: Rework simulation pipeline for adaptive concurrency

Clean architectural improvement — replacing Docker-based simulation with an in-process Rust framework significantly improves developer ergonomics and CI portability. The separation of concerns (bandwidth proxy, server-side latency simulation, scenario orchestration) is well-designed. The `CancellationToken` parent/child pattern and RAII guards for connection tracking are done correctly. Comprehensive test coverage across unit and integration tests.

The deleted Docker scripts had hardcoded paths (`/Users/hoytak/workspace/...`), confirming they were developer-local tooling — nothing of value is lost.

### Issues

- **`unsafe impl Send/Sync for ServerState`**: unnecessary — should be derived automatically
- **`unsafe { std::env::set_var(...) }`**: unsound in multi-threaded context (Rust 1.83+)
- **`to_bytes(body, usize::MAX)`**: no body size limit on `set_config` handler
- **`guard.take().expect(...)` in `run_accept_loop`**: panics on double-call instead of returning `Result`
- **`NetworkSimulationProxy` fields all `pub(crate)`**: internal state unnecessarily exposed
- **`pub fn http_client()`**: exposes internal HTTP client with middleware stack publicly
- **`DirectAccessClient: 'static`**: breaking trait change, should be documented
- **`buf[..n].to_vec()` in hot path**: per-chunk allocation in bandwidth-limited copies
- **Test semantics changed**: `test_session_resume` switched from in-memory to ephemeral disk server

hoytak:
Addressed the PR feedback -- should be better now.

cas_client/src/simulation/direct_access_client.rs
hoytak: removed; is unneeded.

cas_client/src/simulation/network_simulation/bandwidth_limit_router.rs
hoytak: Switched to using Bytes.

cas_client/src/remote_client.rs
hoytak: Doesn't need to be exposed here -- dropped.

cas_client/src/simulation/network_simulation/bandwidth_limit_router.rs
hoytak: It's test code, but did fix this.

cas_client/src/simulation/local_server/handlers.rs
hoytak: Fixed.

simulation/src/bin/run_upload_scenario.rs
hoytak: This is in a single threaded context here, so leaving it.  Also, the simulation is all test code, so it's less critical...

cas_client/src/simulation/local_server/handlers.rs
hoytak: Removed.

rajatarya:
## Code Review Summary

Solid PR — replacing Docker-based simulation with pure Rust tooling is a clear win for developer experience and accessibility. The architecture is clean with good use of builder patterns, cancellation tokens, and separation of concerns.

Given this is **internal simulation tooling**, most findings are minor style/hygiene items that can be deferred. Three items are worth addressing since they affect the tool's correctness or usability:

1. **`_duration_sec` parameter ignored in `run_added_uploads`** — parameter is accepted but unused, with hardcoded 600s sleep durations instead. Will confuse future callers.
2. **Unlimited bandwidth reports 0% utilization** — when `bytes_per_sec` is `None`, the utilization calculation produces 0%, generating misleading reports from the tool whose purpose is accurate reporting.
3. **Double `elapsed()` capture in stats loop** — `start_instant.lock().unwrap().elapsed()` called twice per iteration can skew the measurements this tool is designed to produce.

See inline comments for exact locations.

simulation/src/bin/run_upload_scenario.rs
```diff
@@ -0,0 +1,216 @@
...
+async fn run_gitxet_upload_burst(mut scenario: SimulationScenario, duration_sec: u64) -> ScenarioResult<()> {
+    for _ in 0..10 {
+        add_upload_client_task(&mut scenario);
+    }
+    sleep(Duration::from_secs(duration_sec)).await;
+    scenario.finish().await
+}
+
+/// Staggered client starts: launch clients in waves with pauses between.
+async fn run_added_uploads(mut scenario: SimulationScenario, _duration_sec: u64) -> ScenarioResult<()> {
```
rajatarya: **`_duration_sec` is ignored.** This parameter is accepted but never used — the function hardcodes sleep durations totaling 600s. Someone will pass a value expecting it to control the run duration and be confused when nothing changes.

Either wire it through or remove it from the signature.

cas_client/src/simulation/network_simulation/bandwidth_limit_router.rs
```diff
@@ -0,0 +1,372 @@
...
+                let elapsed_since_start = now.duration_since(self.start_instant);
+                let profile = self.network_profile_provider.current_config(elapsed_since_start);
+                *self.current_network_profile.lock().await = profile.clone();
+
+                let mut last_guard = self.last_refill_instant.lock().await;
+                let elapsed_since_last = now.duration_since(*last_guard);
+                *last_guard = now;
+                drop(last_guard);
+
+                let bytes_per_sec = profile.bandwidth_bytes_per_sec.unwrap_or(0);
```
rajatarya: **Unlimited bandwidth → 0% utilization in reports.** When `bandwidth_bytes_per_sec` is `None` (unlimited), `unwrap_or(0)` makes `bytes_per_sec = 0`, so `total_upload_possible` is always 0. The utilization calculation then yields 0% (or NaN), producing misleading simulation reports.

Since the purpose of this tool is accurate reporting, consider returning `None` or `f64::INFINITY` for the unlimited case instead of 0.

simulation/src/upload_concurrency/upload_simulation_client.rs
```diff
@@ -0,0 +1,445 @@
...
+            }
+            let current_bytes = counters.bytes_sent.load(Ordering::Relaxed);
+            let interval_bytes = current_bytes.saturating_sub(last_reported_bytes);
+            let interval_sec = STATS_INTERVAL_MS as f64 / 1000.0;
+            let interval_throughput_bps = if interval_sec > 0.0 {
+                interval_bytes as f64 / interval_sec
+            } else {
+                0.0
+            };
+            let elapsed_sec = start_instant.lock().unwrap().elapsed().as_secs_f64();
```
rajatarya: **`elapsed()` captured multiple times per stats iteration.** `start_instant.lock().unwrap().elapsed()` is called here (line 134) and again at line 157, yielding slightly different values within the same stats snapshot. For a measurement tool, this can skew results.

Capture `elapsed` once at the top of the iteration and reuse it:
```rust
let elapsed = start_instant.lock().unwrap().elapsed();
let elapsed_sec = elapsed.as_secs_f64();
```

rajatarya (commented):
Inline comments on items worth addressing.

hoytak:
All issues should be fixed.

PR #651, opened 2026-02-12
Make specifying the file size at the beginning of an upload optional. 

hoytak:
Currently, the progress and dependency tracking in the upload path requires that the total size of a file be specified at the start.  This PR changes this so that in cases where the upload is streamed and the total size is not known, it's updated as soon as new data is processed.   Both routes now work and correctly track the file sizes.


hoytak:
@kszucs -- FYI.

PR #652, opened 2026-02-12
FileDownloadGroup

hoytak:
This PR adds a FileDownloadSession struct that parallels the FileUploadSession struct, replacing the FileDownloader.  It's an intermediate step in preparation for a session-based API that integrates well with interfaces other than the python interface in hf_xet.


hoytak:
> Looks good to me, we should expose this in python too?

The session interface should expose a simple wrapper around this.

PR #656, opened 2026-02-13
Streaming data writer

hoytak:
This PR adds an integrated API for streaming downloads, exposing a DownloadStream object that is integrated with the file reconstructor.  It also uses the same memory management buffer limiting process to work with the stream object.  

It also introduces cancellation support to the FileReconstructor to ensure that tasks waiting on a long running download or semaphore wait don't cause things to hang when an error is reported or the user drops the stream.


data/src/file_download_session.rs
```diff
@@ -75,21 +75,22 @@ impl FileDownloadSession {
...
+
+        Ok(n_bytes)
     }
 
     /// Downloads a byte range of a file and writes it to the provided writer.
     ///
     /// The provided `source_range` is interpreted against the original file; output
     /// starts at the writer's current position.
+    ///
+    /// This path does not acquire the session-level file download semaphore.
```
seanses, line 99: Any reason why not?

data/src/file_download_session.rs
```diff
@@ -132,7 +131,31 @@ impl FileDownloadSession {
             task
         });
 
-        self.run_download(file_info, output, None, per_file_tracker).await
+        let file_id = file_info.merkle_hash()?;
```
seanses: Can we fold this block into the `setup_reconstructor` function? Maybe explicitly pass in a `progress_updater: Option<Arc<dyn TrackingProgressUpdater>>`

data/src/file_download_session.rs
```diff
@@ -147,32 +170,37 @@ impl FileDownloadSession {
...
-        output: DataOutput,
         range: Option<FileRange>,
-        progress_updater: Option<Arc<DownloadTaskUpdater>>,
-    ) -> Result<u64> {
-        let semaphore = XetRuntime::current().common().file_download_semaphore.clone();
-        let _permit = semaphore.acquire().await?;
-
+        tracking_id: Option<&str>,
+        write_path: Option<&Path>,
+        explicit_name: Option<Arc<str>>,
```
seanses, line 206: It seems `explicit_name` is never set except at one place in `download_to_writer`
```
let reconstructor = self.setup_reconstructor(file_info, Some(range), None, None, tracking_name)?;
``` 
where it `tracking_name` should be placed in the parameter field of `tracking_id`.

So let's remove this `explicit_name` parameter to make it more clear. 

Also the `tracking_name` parameter of `download_to_writer` function has type `Option<Arc<str>>`, which is incompatible and different from `tracking_id: Option<&str>` of `download_file`. Let's unify them to `Option<&str`.

data/src/file_download_session.rs
```diff
@@ -412,4 +440,408 @@ mod tests {
...
+                .await
+                .unwrap();
+
+                assert_eq!(collected, original_data);
+            })
+            .unwrap();
+    }
+
+    #[test]
+    fn test_download_stream_auto_start_blocking() {
```
seanses: So this test is identical to the above `test_download_stream_blocking` except for the test name..

data/src/file_download_session.rs
```diff
@@ -412,4 +440,408 @@ mod tests {
...
+                let temp = tempdir().unwrap();
+                let cas_path = temp.path().join("cas");
+                let original_data = b"Stream with tracking";
+
+                let xfi = upload_data(&cas_path, original_data).await;
+
+                let config = TranslatorConfig::local_config(&cas_path).unwrap();
+                let session = FileDownloadSession::new(config.into(), None).await.unwrap();
+
+                let mut stream = session.download_stream(&xfi, Some("my-stream-tracking")).unwrap();
```
seanses: This test is identical to `test_download_stream_async` except a tracking id is specified here for `download_stream`, but there's no check on progress using the tracking id, what are we testing?

data/src/file_download_session.rs
```diff
@@ -412,4 +440,408 @@ mod tests {
...
+                let xfi = upload_data(&cas_path, original_data).await;
+
+                let config = TranslatorConfig::local_config(&cas_path).unwrap();
+                let session = FileDownloadSession::new(config.into(), None).await.unwrap();
+
+                // Create the stream, then drop without reading any chunks.
+                let stream = session.download_stream(&xfi, None).unwrap();
+                drop(stream);
+
+                // Yield to let the runtime process the cancellation.
```
seanses: This `test_drop_stream_started_no_reads_then_download` test is identical to `test_drop_stream_never_started_then_download` except for adding a yield here. What makes them different?

data/src/file_download_session.rs
seanses: Actually I have a different thought to change the `tracking_id` to a Ulid: see this [PR](https://github.com/huggingface/xet-core/pull/667)

data/src/file_download_session.rs
hoytak: I thought about it, and I'm willing to change this.  First, currently our Box<dyn Write> interface doesn't use it -- which may be an oversight.   Second, it strikes me that the current parallel limit is mainly due to how resources have been previously handled, and those are controlled using other methods now.  Inefficient disk access, however, is a valid reason, so it makes some sense in the file downloads, especially since the typical behavior there is to give it a large group of files and have those files be downloaded as group as efficiently as possible.  

With streams, it's usually that the user is controlling the group behavior above this API, so if someone is streaming something then the max parallel limit is just another unneeded constraint and would likely just serve to delay the start of the download stream.  

However, I am definitely open to changing this.

data/src/file_download_session.rs
hoytak: Good catch!    The perils of putting in a PR and then updating it a week later ;-)

data/src/file_download_session.rs
hoytak: Agreed (per discussion) but let's update all of that systemwide in a later PR.

data/src/file_download_session.rs
seanses: Ok this makes sense -- but also there is no way for this function to know if the `write: W` is writing to a file or a stream. So I'm actually thinking in another direction: how about we move the all semaphore controls (introduced to `download_file` and `download_file_with_updater`) to a higher level? The reason is that the new XetSession can report accurately if a task is in a queued state or a running state by managing the semaphore control itself.

data/src/file_download_session.rs
seanses: sounds good

data/src/file_download_session.rs
seanses: Not a blocking issue to merge this PR

PR #657, opened 2026-02-13
XetSession API

seanses:
Derived from [this plan](https://docs.google.com/document/d/1A470wZKbM0H0s6AF0WP989e4GoFB0FbC7_HFmQWjTdw/edit?usp=sharing)

This PR introduces a new `xet_session` crate that provides a session-based hierarchical API: Users create a XetSession to manage runtime and configuration, then batch uploads into UploadCommit objects and downloads into DownloadGroup objects — each of which runs transfers in the background by the inner XetRuntime.

All pub functions are exposed as sync functions - making them easy to use in other languages, e.g. Python, C, ...

Will expose another interface for Rust async native environment.

### Example

```
let session = XetSession::new(XetConfig::new(...));

let upload_commit = session.new_upload_commit();

let handle = upload_commit.upload_file_from_path(file_path); // return immediately
let handle = upload_commit.upload_bytes(bytes); // return immediately

let progress = upload_commit.get_progress(); // return a collection of upload progress

let result = upload_commit.commit(); // blocking wait: finalizing upload, pushes all data to remote, return a collection of file metadata (hash and size)

let download_group = session.new_download_group();
let handle = download_group.download_file_to_path(file_hash, dest_path); // return immediately

let progress = download_group.get_progress(); // return a collection of download progress

let result = download_group.finish() // blocking wait: return a collection of download results (metadata)
```

### Next steps

- [ ] Make progress reporting pull-based not push-based: introduces a `get_process()` function to FileUploadSession and FileDownloadSession so XetSession can pull.
- [ ] Plug into Xet migration service to test.

### DIFF
--- a/xet_session/Cargo.toml
+++ b/xet_session/Cargo.toml
@@ -5,8 +5,8 @@ edition = "2024"
 description = "Session-based file upload/download API for XetHub"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/xetdata/xet-core"
-keywords = ["xet", "upload", "download", "deduplication", "session"]
-categories = ["network-programming", "api-bindings"]
+keywords = ["xet", "upload", "download", "deduplication", "session", "huggingface"]
+categories = ["api-bindings"]
 
 [lib]
 name = "xet_session"
--- a/xet_session/examples/example.rs
+++ b/xet_session/examples/example.rs
@@ -7,7 +7,7 @@ use std::time::Duration;
 
 use anyhow::Result;
 use clap::{Parser, Subcommand};
-use xet_session::{FileMetadata, TaskHandle, TaskStatus, XetFileInfo, XetSession};
+use xet_session::{FileMetadata, TaskHandle, TaskStatus, XetFileInfo, XetSessionBuilder};
 
 #[derive(Parser)]
 #[clap(name = "session-demo", about = "XetSession API demo")]
@@ -35,22 +35,25 @@ enum Command {
     },
 }
 
-#[tokio::main]
-async fn main() -> Result<()> {
+fn main() -> Result<()> {
     tracing_subscriber::fmt::init();
     let cli = Cli::parse();
     match cli.command {
-        Command::Upload { files, endpoint } => upload_files(files, endpoint).await,
+        Command::Upload { files, endpoint } => upload_files(files, endpoint),
         Command::Download {
             metadata_file,
             output_dir,
             endpoint,
-        } => download_files(metadata_file, output_dir, endpoint).await,
+        } => download_files(metadata_file, output_dir, endpoint),
     }
 }
 
-async fn upload_files(files: Vec<PathBuf>, endpoint: Option<String>) -> Result<()> {
-    let session = XetSession::new(endpoint, None, None, None)?;
+fn upload_files(files: Vec<PathBuf>, endpoint: Option<String>) -> Result<()> {
+    let mut builder = XetSessionBuilder::new();
+    if let Some(ep) = endpoint {
+        builder = builder.with_endpoint(ep);
+    }
+    let session = builder.build()?;
     let commit = session.new_upload_commit()?;
 
     // Enqueue all uploads; each starts immediately in the background.
@@ -62,7 +65,7 @@ async fn upload_files(files: Vec<PathBuf>, endpoint: Option<String>) -> Result<(
 
     // Spawn a task to print progress; the main thread blocks in commit() below.
     let commit_for_progress = commit.clone();
-    let progress_task = tokio::spawn(async move {
+    std::thread::spawn(move || {
         loop {
             if let Ok(snapshot) = commit_for_progress.get_progress() {
                 let p = snapshot.total();
@@ -72,13 +75,12 @@ async fn upload_files(files: Vec<PathBuf>, endpoint: Option<String>) -> Result<(
                     .count();
                 println!("{}/{} files | {}/{} bytes", done, n_files, p.total_bytes_completed, p.total_bytes);
             }
-            tokio::time::sleep(Duration::from_millis(100)).await;
+            std::thread::sleep(Duration::from_millis(100));
         }
     });
 
     // Block until all uploads finish and metadata is finalized.
-    let metadata = commit.commit()?;
-    progress_task.abort();
+    let metadata: Vec<_> = commit.commit()?.into_iter().filter_map(|m| m.ok()).collect();
 
     for m in &metadata {
         println!("  {} -> {} ({} bytes)", m.tracking_name.as_deref().unwrap_or("?"), m.hash, m.file_size);
@@ -90,11 +92,15 @@ async fn upload_files(files: Vec<PathBuf>, endpoint: Option<String>) -> Result<(
     Ok(())
 }
 
-async fn download_files(metadata_file: PathBuf, output_dir: PathBuf, endpoint: Option<String>) -> Result<()> {
+fn download_files(metadata_file: PathBuf, output_dir: PathBuf, endpoint: Option<String>) -> Result<()> {
     let metadata: Vec<FileMetadata> = serde_json::from_str(&std::fs::read_to_string(metadata_file)?)?;
     std::fs::create_dir_all(&output_dir)?;
 
-    let session = XetSession::new(endpoint, None, None, None)?;
+    let mut builder = XetSessionBuilder::new();
+    if let Some(ep) = endpoint {
+        builder = builder.with_endpoint(ep);
+    }
+    let session = builder.build()?;
     let group = session.new_download_group()?;
 
     // Enqueue all downloads; each starts immediately in the background.
@@ -115,7 +121,7 @@ async fn download_files(metadata_file: PathBuf, output_dir: PathBuf, endpoint: O
 
     // Spawn a task to print progress; the main thread blocks in finish() below.
     let group_for_progress = group.clone();
-    let progress_task = tokio::spawn(async move {
+    std::thread::spawn(move || {
         loop {
             if let Ok(snapshot) = group_for_progress.get_progress() {
                 let p = snapshot.total();
@@ -125,13 +131,12 @@ async fn download_files(metadata_file: PathBuf, output_dir: PathBuf, endpoint: O
                     .count();
                 println!("{}/{} files | {}/{} bytes", done, n_files, p.total_bytes_completed, p.total_bytes);
             }
-            tokio::time::sleep(Duration::from_millis(100)).await;
+            std::thread::sleep(Duration::from_millis(100));
         }
     });
 
     // Block until all downloads finish.
-    let results = group.finish()?;
-    progress_task.abort();
+    let results: Vec<_> = group.finish()?.into_iter().filter_map(|m| m.ok()).collect();
 
     for r in &results {
         println!("  {} ({} bytes)", r.dest_path.display(), r.file_info.file_size);
--- a/xet_session/src/download_group.rs
+++ b/xet_session/src/download_group.rs
@@ -2,7 +2,7 @@
 
 use std::collections::HashMap;
 use std::path::PathBuf;
-use std::sync::{Arc, Mutex, RwLock};
+use std::sync::{Arc, Mutex, MutexGuard, RwLock};
 
 use data::{FileDownloadSession, XetFileInfo};
 use tokio::task::JoinHandle;
@@ -116,12 +116,6 @@ impl DownloadGroup {
     }
 
     /// Return a snapshot of progress for every queued download.
-    ///
-    /// Per-file byte counts come from [`GroupProgress`], which is updated by
-    /// the [`TrackingProgressUpdater`] implementation as data flows through
-    /// `FileDownloadSession`.  The method is safe to call frequently from Python:
-    /// integer counters are read atomically and the per-file map requires only
-    /// a brief lock.
     pub fn get_progress(&self) -> Result<ProgressSnapshot, SessionError> {
         self.progress.snapshot()
     }
@@ -134,7 +128,7 @@ impl DownloadGroup {
     /// Consumes `self` — subsequent calls on any clone will return
     /// [`SessionError::AlreadyFinished`] (or a channel-closed error if the
     /// background worker has already exited).
-    pub fn finish(self) -> Result<Vec<DownloadResult>, SessionError> {
+    pub fn finish(self) -> Result<Vec<Result<DownloadResult, SessionError>>, SessionError> {
         let inner = self.inner.clone();
         self.session
             .runtime
@@ -173,8 +167,8 @@ impl DownloadGroupInner {
     // ===== State helpers =====
 
     /// Check whether the group is still accepting new tasks.
-    fn check_accepting_tasks(&self) -> Result<(), SessionError> {
-        match *self.state.lock()? {
+    fn check_accepting_tasks(state: &MutexGuard<GroupState>) -> Result<(), SessionError> {
+        match **state {
             GroupState::Finished => Err(SessionError::AlreadyFinished),
             GroupState::Aborted => Err(SessionError::Aborted),
             GroupState::Alive => Ok(()),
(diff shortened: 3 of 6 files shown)


xet_session/Cargo.toml
```diff
@@ -0,0 +1,39 @@
+[package]
+name = "xet-session"
+version = "0.1.0"
+edition = "2024"
+description = "Session-based file upload/download API for XetHub"
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/xetdata/xet-core"
+keywords = ["xet", "upload", "download", "deduplication", "session"]
+categories = ["network-programming", "api-bindings"]
```
assafvayner: I'm not sure I would categorize xet_session as network-programming.
maybe also add "huggingface" to the keywords?

assafvayner: If this is our main final exposed crate, I may also suggest we take over `xet` instead of `xet-session`

xet_session/src/session.rs
```diff
@@ -0,0 +1,329 @@
...
+    Alive,
+    Aborted,
+}
+
+/// All shared state for a session.
+/// Lives behind `Arc<XetSessionInner>` — do not use this type directly.
+#[doc(hidden)]
+pub struct XetSessionInner {
+    // Independently cloned by background tasks, so needs its own Arc.
+    pub runtime: Arc<XetRuntime>,
```
assafvayner: should we have these to be pub struct and fields? (I only ask since we do limit the other fields.

Maybe for access to the runtime, on the XetSession we impl `fn runtime(&self) -> Arc<XetRuntime> { self.inner.runtime.clone() }`

Do we use the Inner struct anywhere outside this file/crate? is it wrong to use it outside here?
^ In this file there's no functions on Inner, so I think it does not need to be pub

xet_session/src/session.rs
```diff
@@ -0,0 +1,329 @@
...
+    ///
+    /// Like [`XetSession::new`] but also accepts a fully-populated
+    /// [`XetConfig`].
+    pub fn new_with_config(
+        config: XetConfig,
+        endpoint: Option<String>,
+        token_info: Option<(String, u64)>,
+        token_refresher: Option<Arc<dyn TokenRefresher>>,
+        custom_headers: Option<Arc<HeaderMap>>,
+    ) -> Result<Self, SessionError> {
```
assafvayner: I think we should introduce builder pattern for XetSession given all the Option<> params here.

xet_session/src/session.rs
```diff
@@ -0,0 +1,329 @@
...
+                id: session_id,
+            }),
+        })
+    }
+
+    /// Create a new [`UploadCommit`] that groups related file uploads.
+    ///
+    /// Returns `Err(SessionError::Aborted)` if the session has been aborted.
+    pub fn new_upload_commit(&self) -> Result<UploadCommit, SessionError> {
+        let state = self.state.lock()?;
```
assafvayner, line 214: here and elsewhere, is there impact to using std::sync::Mutex over tokio::sync::Mutexes if we're in a tokio runtime? std::sync::Mutex::lock() blocks a system thread right? Or are we even in a tokio runtime if these functions are supposed to get called from a sync context.

I raise this as a point, but I think I can answer it myself, it doesn't matter much here, a user won't be calling upload_new_commit/new_donwload_group so frequently that we expect much impact from contention/stalling.

Consider resolved

xet_session/src/download_group.rs
```diff
@@ -0,0 +1,591 @@
...
+                return Err(SessionError::AlreadyFinished);
+            }
+            *state_guard = GroupState::Aborted; // stop new tasks while draining
+        }
+
+        // Wait for all downloads to complete
+        let handles: Vec<_> = {
+            let mut tasks = self.active_tasks.write()?;
+            tasks.drain().collect()
+        };
```
assafvayner: you don't need the extra scopes, I would even suggest this be a 1-liner:
`let handles = self.active_tasks.write()?.drain().collect();`

I also think `.drain().collect()` can be replaced with some kind of `.take()` maybe? basically, take vec/map and replace with default/empty?

xet_session/src/download_group.rs
```diff
@@ -0,0 +1,591 @@
...
+    ///
+    /// Returns a [`TaskHandle`] that can be used to poll status and per-file
+    /// progress without taking the GIL.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`SessionError::Aborted`] if the session has been aborted, or
+    /// [`SessionError::AlreadyFinished`] if [`finish`](Self::finish) has already
+    /// been called.
+    pub fn download_file_to_path(
```
assafvayner, line 100: can punt for later, but maybe document, if we're going to expose a non-filepath based download interface, i.e. to a Write/AsyncWrite in rust

xet_session/src/errors.rs
```diff
@@ -0,0 +1,89 @@
...
+    Io(#[from] std::io::Error),
+
+    /// A catch-all for errors that don't fit a more specific variant.
+    #[error("Other error: {0}")]
+    Other(String),
+}
+
+impl SessionError {
+    /// Create an [`Other`](SessionError::Other) variant from any [`Display`](std::fmt::Display) value.
+    pub fn other(msg: impl std::fmt::Display) -> Self {
```
assafvayner, line 65: minor nit to use `impl ToString` rather than Display, sometimes things are ToString and not Display I think

assafvayner (approved):
Some comments.

I do have 2 additional requests for this revision, then more content can go in in the future.

- Add a README.md in xet_session, should be easy just generate and review. This should describe the general interface exposed and the motivation.
- ^ should fix this, If you run cargo doc on this crate, the first page is pretty empty, so we should fill something in there, doesn't have to be final/release version, just something.

xet_session/Cargo.toml
hoytak: I don't think we should publish this yet -- but yes, to that.   Let's iterate and make sure we're exposing the correct bits with internal testing for a bit.

xet_session/src/session.rs
hoytak: pub(crate) should be enough...

xet_session/src/session.rs
seanses: Yeah, for locks over almost instantaneous operations like these, a sync Mutex is better than a tokio Mutex unless you have to hold the lock across an await

xet_session/src/download_group.rs
seanses: Yeah will expose a streaming download interface later

xet_session/src/errors.rs
seanses: That's not what the official ToString [doc](https://doc.rust-lang.org/std/string/trait.ToString.html) recommends:

> This trait is automatically implemented for any type which implements the [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) trait. As such, ToString shouldn’t be implemented directly: [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) should be implemented instead, and you get the ToString implementation for free.

Meaning `Display` is a more general trait than `ToString`.

rajatarya:
## Code Review Summary

Well-structured session API with a clean hierarchical design (XetSession → UploadCommit/DownloadGroup). The sync-over-async approach is appropriate for FFI consumers.

Existing review feedback already covers: inner struct visibility/`Deref` pattern, builder pattern for `XetSession`, `std::sync::Mutex` usage, and README/docs. These findings are omitted below.

### P0 — Must Fix
1. **Example will deadlock** — `#[tokio::main]` + `commit()` (which calls `external_run_async_task` → `block_on`) deadlocks because you can't `block_on` inside a Tokio runtime.
2. **`handle_commit`/`handle_finish` abandon remaining tasks on first error** — `??` propagates immediately, dropping un-awaited `JoinHandle`s. Remaining tasks continue running after `commit()`/`finish()` returns.

### P1 — Should Fix
3. **`HashMap::drain()` returns results in non-deterministic order** — `commit()`/`finish()` return `Vec<FileMetadata>`/`Vec<DownloadResult>` in arbitrary order. Use `BTreeMap<Ulid>` (time-ordered) or document.
4. **`abort()` fails partway on `PoisonError`** — If `status.lock()?` fails mid-iteration, remaining tasks are left un-cancelled. `XetSession::abort()` also returns early on the first failing sub-abort.
5. **`file_size` field assigned wrong value in download** — `result?` is `n_bytes` downloaded, not the logical file size. Should use `file_info.file_size`.

### P2 — Nice to Fix
6. **`crate-type` includes `cdylib`/`staticlib`** — triples compile time unless FFI is needed now
7. **Double `Arc` wrapping of `XetRuntime`** — `XetRuntime` is already internally `Arc`-wrapped
8. **Silent progress drops on poisoned locks** in `register_updates`
9. **Download concurrency semaphore moved to caller** with only a comment as the contract

See inline comments for exact locations of P0/P1 issues.

xet_session/examples/example.rs
```diff
@@ -0,0 +1,141 @@
...
+    Download {
+        metadata_file: PathBuf,
+        #[clap(short, long, default_value = "./downloads")]
+        output_dir: PathBuf,
+        #[clap(long)]
+        endpoint: Option<String>,
+    },
+}
+
+#[tokio::main]
```
rajatarya: **P0: This example will deadlock.** `#[tokio::main]` runs `main()` on a Tokio runtime. When `commit.commit()` is called (line ~78), it internally calls `external_run_async_task` which does `handle.block_on(...)`. Calling `block_on` from within a Tokio async context panics/deadlocks.

Either:
- Remove `#[tokio::main]` and use a plain `fn main()` (the whole point of the sync API), or
- Document that `XetSession` methods must never be called from within a Tokio runtime.

xet_session/src/upload_commit.rs
```diff
@@ -0,0 +1,711 @@
...
+
+        // Wait for all uploads to complete
+        let handles: Vec<_> = {
+            let mut tasks = self.active_tasks.write()?;
+            tasks.drain().collect()
+        };
+
+        let mut results = Vec::new();
+        for (_task_id, handle) in handles {
+            let file_info = handle.join_handle.await.map_err(SessionError::TaskJoinError)??;
```
rajatarya: **P0: First task error abandons remaining join handles.** The `??` propagates immediately, and the remaining `JoinHandle`s in `handles` are dropped without being awaited. Those tasks continue running in the background after `commit()` returns an error.

Consider draining all handles first, collecting results, then returning the combined error:
```rust
let mut results = Vec::new();
let mut first_error = None;
for (_task_id, handle) in handles {
    match handle.join_handle.await {
        Ok(Ok(info)) => results.push(FileMetadata { ... }),
        Ok(Err(e)) if first_error.is_none() => first_error = Some(e),
        Err(e) if first_error.is_none() => first_error = Some(SessionError::TaskJoinError(e)),
        _ => {}
    }
}
if let Some(e) = first_error { return Err(e); }
```

xet_session/src/download_group.rs
```diff
@@ -0,0 +1,591 @@
...
+
+        // Wait for all downloads to complete
+        let handles: Vec<_> = {
+            let mut tasks = self.active_tasks.write()?;
+            tasks.drain().collect()
+        };
+
+        let mut results = Vec::new();
+        for (_task_id, handle) in handles {
+            let file_info = handle.join_handle.await.map_err(SessionError::TaskJoinError)??;
```
rajatarya: **P0: Same issue as `handle_commit`** — first task error abandons remaining join handles. Same fix applies: drain all handles, collect errors, return after all are awaited.

xet_session/src/upload_commit.rs
```diff
@@ -0,0 +1,711 @@
...
+            if *state_guard == GroupState::Finished {
+                return Err(SessionError::AlreadyCommitted);
+            }
+            *state_guard = GroupState::Aborted; // stop new tasks while draining
+        }
+
+        // Wait for all uploads to complete
+        let handles: Vec<_> = {
+            let mut tasks = self.active_tasks.write()?;
+            tasks.drain().collect()
```
rajatarya: **P1: `HashMap::drain()` returns results in arbitrary order.** The `Vec<FileMetadata>` returned by `commit()` will not correspond to the submission order of uploads. Since `tracking_id` is a `Ulid` (time-ordered), switching to `BTreeMap<Ulid, InnerTaskHandle>` would give deterministic chronological ordering for free.

xet_session/src/upload_commit.rs
```diff
@@ -0,0 +1,711 @@
...
+        &self.session.runtime
+    }
+
+    /// Cancle all tasks and set task status to "Cancelled"
+    fn abort(&self) -> Result<(), SessionError> {
+        *self.state.lock()? = GroupState::Aborted;
+        let mut active_tasks = self.active_tasks.write()?;
+        for (_tracking_id, inner_task_handle) in active_tasks.drain() {
+            inner_task_handle.join_handle.abort();
+            *inner_task_handle.status.lock()? = TaskStatus::Cancelled;
```
rajatarya: **P1: `abort()` fails partway on `PoisonError`.** If `status.lock()?` fails for any task, the `?` returns early and remaining tasks in the iterator are never cancelled. Consider using `let _ = inner_task_handle.status.lock().map(|mut s| *s = TaskStatus::Cancelled);` to best-effort cancel all tasks regardless of individual lock failures.

Same issue in `XetSession::abort()` which returns on the first failing sub-abort.

xet_session/src/download_group.rs
```diff
@@ -0,0 +1,591 @@
...
+            let new_status = if result.is_ok() {
+                TaskStatus::Completed
+            } else {
+                TaskStatus::Failed
+            };
+            *status.lock()? = new_status;
+
+            Ok(XetFileInfo {
+                hash: file_info.hash,
+                file_size: result?,
```
rajatarya, line 208: **P1: `file_size` is assigned the wrong value.** `result?` is the return value of `download_file()`, which is `n_bytes` (bytes written/smudged), not the logical file size. For non-range downloads these happen to be equal, but the semantics are wrong.

Should be:
```rust
file_size: file_info.file_size,
```
(and `let _ = result?;` to still propagate errors).

xet_session/Cargo.toml
seanses: Dropped "network-programming" from `categories`. Added "huggingface" to `keywords`.

xet_session/src/session.rs
seanses: Changed to `pub(crate)`

xet_session/src/errors.rs
assafvayner: Whoops I may need to go and fix that some places

xet_session/src/session.rs
seanses: Added `XetSessionBuilder`

xet_session/src/download_group.rs
seanses: Good suggestion! Replaced with
`let active_tasks = std::mem::take(&mut *self.active_tasks.write()?);`

xet_session/examples/example.rs
seanses: Fixed

xet_session/src/upload_commit.rs
seanses: Updated both UploadCommit and DownloadGroup to handle this.

xet_session/src/download_group.rs
seanses: fixed.

xet_session/src/upload_commit.rs
seanses: Replaced with `let active_tasks = std::mem::take(&mut *self.active_tasks.write()?);`. The order is likely not a concern here.

xet_session/src/upload_commit.rs
seanses: Switched to this

xet_session/src/download_group.rs
seanses: won't fix, this is expected behavior.

PR #659, opened 2026-02-17
Removing pyproject.toml from repo root

### LINKED ISSUE #658: hf-xet source builds fail due to python-source/module path mismatch
### Summary
`hf-xet` source builds fail because `pyproject.toml` sets:

```toml
python-source = "hf_xet/python"
```

but that path only contains `.gitkeep` and not the expected Python module directory (`hf_xet/python/hf_xet`).

### Repro
```bash
pip wheel --no-binary hf-xet hf-xet==1.2.0
```

### Error
```
python-source is set to .../hf_xet/python, but the python module at .../hf_xet/python/hf_xet does not exist.
```

### Packaging impact
Homebrew had to patch around this during `pocket-tts` packaging:
- https://github.com/Homebrew/homebrew-core/pull/267834

Could you either:
1. Ship the expected Python module path under `hf_xet/python/hf_xet`, or
2. Remove/adjust `python-source` in `pyproject.toml` for source distributions?

Thanks.

rajatarya:
Not being used to build hf-xet package anyway this is confusing the pip wheel command.

Fixes #658


rajatarya:
This doesn't fully work, needed to update the `hf_xet/pyproject.toml` as well (see latest commit).

This is the correct way to get build the wheel:
`pip wheel --no-binary hf-xet "hf-xet @ git+https://github.com/huggingface/xet-core@fix-gh-658-remove-pyproject-toml-root#subdirectory=hf_xet"`

because `hf-xet` is not defined at the root of the repo.

PR #660, opened 2026-02-19
Removed incorrectly included AGENTS.md.

hoytak:
The AGENTS.md file was incorrectly checked into the repository (part of a claude process to prepare and check a diff for PR).  This PR removes that.


rajatarya (approved):
My suggestion is to put your AGENTS.md in `.git/info/exclude` file (it acts like .gitignore but never in the repo tracking files).

PR #662, opened 2026-02-20
Add Windows ARM64 build support

rajatarya:
## Summary

Closes #588

- Add `win11-arm` runner with `aarch64-pc-windows-msvc` target to the hf-xet Python wheel release pipeline
- Add `win11-arm` runner with `aarch64` target to the git-xet CLI release pipeline, parameterizing the WiX installer `-arch` flag

## Test plan

- [x] Trigger a workflow_dispatch run of the Release workflow and verify `windows` matrix includes both `x64` and `aarch64` entries
- [x] Verify ARM64 wheels and .pdb debug symbols are built and uploaded
- [ ] Trigger a workflow_dispatch run of the git-xet Release workflow and verify ARM64 binary and MSI installer are produced

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #661, opened 2026-02-20
Feat: optional `request_headers` on hf_xet API calls

bpronan:
Adding support for setting an optional `request_header` map on the hf_xet upload and download API calls. This map is augmented with the hf_xet user agent string and is passed along with the requests to xetcas.

This PR also adds some unit tests for testing the map merging behavior to `hf_xet/lib.rs` and adds support for running these with cargo test and in github actions CI step.

PR #663, opened 2026-02-23
Don't build shellexpand in wasm

seanses:
Crate `shellexpand` seems to be not compatible with the nightly toolchain: https://github.com/huggingface/xet-core/actions/runs/22319509143/job/64573656769
This should fix the WASM build error on the main branch and in https://github.com/huggingface/xet-core/pull/661 and https://github.com/huggingface/xet-core/pull/662

Will revisit whether to drop the nightly toolchain requirement from WASM build in the future.


bpronan:
Thanks @seanses! I've applied this change directly to https://github.com/huggingface/xet-core/pull/661, so feel free and close this out.

PR #666, opened 2026-02-25
 Scale download buffer memory limit by number of active downloads

hoytak:
Currently, the maximum number of downloaded files is fixed, regardless of the number of downloads currently in flight.  However, as the number of downloads increases, a fixed size total could lead to waiting on individual segments that download out-of-order or don't have enough turnaround time to saturate the output.  While writing to disk or the download itself often becomes the bottleneck before these effects, planned features such as streaming files and caching could be affected by this limit.  The default formula for the download buffer size now is (2GB + 512MB * number of concurrent downloads) up to a maximum of 8GB (these are adjustable).  

This PR alleviates this by allocating an additional 512MB buffer allocation per file, prioritized to the specific download, releasing that capacity when the file finishes downloading.  This is done using the AdjustableSemaphore class, first introduced for the concurrent scaling, which allows the number of total permits in a semaphore to be incremented or decremented; on decrement, permits are discarded upon return until the total permits is at the target number.


rajatarya:
(just need to point out this is PR 666 in xet-core.)

utils/src/adjustable_semaphore.rs
```diff
@@ -1,148 +1,346 @@
...
+                if let Some(p) = permit.split(decreases_resolved as usize) {
+                    p.forget();
+                } else {
+                    debug_assert!(false, "Failed to split permit; mismatch in self.num_permits.");
+                }
+            }
+        } else {
+            // Virtual permit (from increment_total_permits): release non-consumed
+            // permits into the semaphore.
+            let to_return = (num_permits - decreases_resolved) as usize;
```
rajatarya, line 109: Potential underflow if `decreases_resolved > num_permits`. While `attempt_sub` should clamp, a `debug_assert!(decreases_resolved <= num_permits)` here would make the invariant explicit and catch bugs early.

utils/src/adjustable_semaphore.rs
```diff
@@ -1,148 +1,346 @@
...
+        Some(removed * self.basis)
+    }
+
+    /// Increment the total number of permits by up to `n` (logical) up to
+    /// the maximum bound.
+    ///
+    /// Returns a permit holding the newly added capacity, or `None`
+    /// if no permits could be added (already at max). The permits enter the
+    /// semaphore when the returned permit is dropped.  This allows a user to
+    /// acquire a permit immediately that bipasses the FIFO queue so the caller
```
rajatarya: Typo: "bipasses" → "bypasses"

utils/src/adjustable_semaphore.rs
```diff
@@ -1,148 +1,346 @@
...
+    /// Split `n` logical permits off this permit into a new permit.
+    /// Returns `None` if the scaled physical count exceeds what this permit holds.
+    pub fn split(&mut self, n: u64) -> Option<AdjustableSemaphorePermit> {
+        let physical_n = n.div_ceil(self.parent.basis);
+        if physical_n > self.num_permits as u64 {
+            return None;
+        }
+        let physical_n = physical_n as u32;
+
+        self.num_permits -= physical_n;
```
rajatarya: When `n == 0`, this returns `Some(AdjustableSemaphorePermit { permit: None, num_permits: 0, .. })` — a zero-permit virtual permit. On drop it harmlessly calls `attempt_sub(0)` + `add_permits(0)`. Consider returning `None` for `n == 0` to avoid the odd semantics.

cas_client/src/adaptive_concurrency/controller.rs
```diff
@@ -491,7 +497,7 @@ impl AdaptiveConcurrencyController {
             // permits too quickly.
             if state_lg.last_adjustment_time.elapsed() > self.min_concurrency_decrease_delay {
                 let old_concurrency = self.concurrency_semaphore.total_permits();
-                self.concurrency_semaphore.decrement_total_permits();
+                let _ = self.concurrency_semaphore.decrement_total_permits(1);
```
rajatarya, line 503: Nit: the old code also silently ignored the bool return, but now that it returns `Option<u64>` with richer info, a brief `// Already at min; ok to ignore` comment would help readers.

file_reconstruction/src/file_reconstructor.rs
```diff
@@ -113,11 +115,39 @@ impl FileReconstructor {
...
+        // downloads: target = (base + n * perfile).min(limit). On start we increment to
+        // the new target, possibly getting back a virtual permit that lets this download begin
+        // immediately without queuing behind existing acquires. On exit, the ClosureGuard
+        // recomputes the target for the reduced download count and shrinks back if needed.
+        let mut seed_buffer_permit;
+        let _download_count_decrement_guard;
+
+        if !is_custom {
+            let active_downloads = XetRuntime::current().common().active_downloads.clone();
+            let n = active_downloads.fetch_add(1, Ordering::Relaxed) + 1;
```
rajatarya, line 132: With `Relaxed` ordering, two near-simultaneous download starts can both see `n=1`, both compute the same target, and only one actually gets permits added. The second gets `None` from `increment_permits_to_target` and no seed permit. This is safe (falls through to shared acquire) but means the priority-buffer optimization is lost for concurrent starters. Worth a comment documenting this trade-off, or consider `SeqCst`/`AcqRel` if the seed permit is important.

xet_config/src/groups/reconstruction.rs
```diff
@@ -65,10 +75,7 @@ crate::config_group!({
...
     /// When false, standard sequential writes are used.
     ///
-    /// Vectorized writes can improve performance for writers that benefit from batching,
-    /// such as network sockets or buffered file writers.
-    ///
-    /// The default value is false.
+    /// The default value is true.
     ///
     /// Use the environment variable `HF_XET_RECONSTRUCTION_USE_VECTORED_WRITE` to set this value.
     ref use_vectored_write: bool = true;
```
rajatarya, line 81: The default for `use_vectored_write` changed from `false` to `true` — this is a behavioral change unrelated to buffer scaling. Consider splitting into a separate PR or at minimum calling it out in the PR description.

rajatarya: Was this change intentional? If so, then probably worth calling out in PR description.

rajatarya:
I had Claude Code actually to the PR review and then add the comments as if I would have - by adding them to line numbers in files as separate comments. This way we can discuss more easily in thread in the PR. Most of these comments are minor nits/typos, but I think the relaxed ordering comment is worth considering.

file_reconstruction/src/file_reconstructor.rs
hoytak: I think this is actually wrong.  With atomics -- even with relaxed ordering -- these operations always succeed correctly.  The ordering argument only affects how other atomics around this one are scheduled; Relaxed means the cpu can schedule things how it's needed.   So if you have 

```
let n1 = my_atomic.fetch_add(1, Relaxed); 
let n2 = my_atomic.fetch_add(1, Relaxed); 
```

With relaxed ordering, you might actually get `n1 = 2` and `n1 = 1`, but you would never get `n1 = 1` and `n2 = 1`.

xet_config/src/groups/reconstruction.rs
hoytak: What changed was updating the comment -- the comment said it was false by default, but it's actually true by default.  That can be a separate PR if need be.  But the behavior hasn't changed.

hoytak:
> I had Claude Code actually to the PR review and then add the comments as if I would have - by adding them to line numbers in files as separate comments. This way we can discuss more easily in thread in the PR. Most of these comments are minor nits/typos, but I think the relaxed ordering comment is worth considering.

Fixed the small stuff.  I think the Relaxed ordering comment is actually wrong -- you won't see that behavior with atomics, even with Relaxed ordering.  See my comment above.

file_reconstruction/src/file_reconstructor.rs
rajatarya: Oh wow I didn't know that. I appreciate the explanation. I assumed relaxed ordering meant that there wasn't a wait on the lock so simultaneous reads can independently increment and then one of them wins on the writing side - this makes a lot more sense in multi-core systems for the definition of relaxed ordering. Appreciate it, resolving this comment!

xet_config/src/groups/reconstruction.rs
rajatarya: Ah, let's keep it as-is then. Silly LLM.

rajatarya (approved):
Thanks for quickly addressing the feedback! And for explaining how relaxed atomics work to me.

PR #667, opened 2026-02-26
Change progress tracking ids from Arc<str> to Ulid

seanses:
Built on top of [Streaming data writer #656](https://github.com/huggingface/xet-core/pull/656)

Using `file_name: Arc<str>` as the progress tracking ids is problematic:
1. When file name is not specified (e.g. by `start_clean`), this tracking id becomes an empty String after `.unwrap_or_default()`: https://github.com/huggingface/xet-core/blob/0d0f4883ad41220b0fe3aa7b22adfe4f2f26f5e8/data/src/file_upload_session.rs#L265, and it's impossible to find out which a particular `ItemProgressUpdate` belongs to because the only key is the `file_name: Arc<str>`: https://github.com/huggingface/xet-core/blob/0d0f4883ad41220b0fe3aa7b22adfe4f2f26f5e8/progress_tracking/src/progress_info.rs#L7

2. The conversion from a `PathBuf` to a `str` is not 1-1 because of `to_string_lossy`: https://github.com/huggingface/xet-core/blob/0d0f4883ad41220b0fe3aa7b22adfe4f2f26f5e8/data/src/file_upload_session.rs#L151

3. With the new XetSession interface, users may issue multiple calls to the `FileUploadSession::upload_files()` function, with identical file names but with different current working directory context so actually representing different files. 

This PR changes the tracking id to a Ulid (unique 128 bit integer) to avoid the above problems, and also prepares for XetSession progress reporting API.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1016,6 +1016,7 @@ dependencies = [
  "error_printer",
  "file_reconstruction",
  "hub_client",
+ "itertools 0.14.0",
  "lazy_static",
  "mdb_shard",
  "merklehash",
@@ -1378,6 +1379,7 @@ dependencies = [
  "tokio",
  "tokio-util",
  "tracing",
+ "ulid",
  "utils",
  "xet_config",
  "xet_runtime",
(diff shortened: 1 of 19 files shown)


seanses:
All merged into https://github.com/huggingface/xet-core/pull/657

PR #668, opened 2026-02-26
fix: wrap TrackingProgressUpdater in AggregatingProgressUpdater

Hugoch:
## Summary

Wrap download progress updaters in `AggregatingProgressUpdater` to eliminate GIL contention when Python callers provide per-file progress callbacks.

The upload path has had this aggregation since v1.1.3 (PR #340), but the download path was missed. Without aggregation, each XORB chunk triggers a `spawn_blocking` + `Python::with_gil()` callback. With many concurrent file downloads, this causes severe GIL contention — measured as a **4x throughput reduction** (3000 MB/s → 750 MB/s on a 25 Gbps link).

The fix wraps the caller-provided `TrackingProgressUpdater` in an `AggregatingProgressUpdater` (200ms flush interval) inside `download_file_with_updater()`, matching the pattern already used by `FileUploadSession`. This reduces Python callback frequency from thousands/sec to ~5/sec per file while preserving progress bar feedback.

## Root cause

When `huggingface_hub` calls `hf_xet.download_files()`, it passes a per-file Python callback for progress bar updates. On the Rust side, each callback invocation goes through:

```
report_bytes_written() / report_transfer_progress()
  → tokio::spawn(register_updates())
    → spawn_blocking(Python::with_gil(callback))
```

With the detailed download progress tracking added in PR #645 (hf-xet v1.3.0), both `report_bytes_written` and `report_transfer_progress` fire per chunk, roughly doubling callback frequency. With 8+ concurrent file downloads, each spawning dozens of concurrent XORB streams, the GIL becomes a severe bottleneck.

## History

The problem has existed since xet download support was introduced, but worsened over time:

| Version | Date | Impact |
|---------|------|--------|
| `huggingface_hub v0.30.0` / `hf-xet 0.1.x` | Mar 2025 | Moderate — synchronous `with_gil()` per chunk, but hf_xet was an optional extra |
| `huggingface_hub v0.31.0` / `hf-xet >=1.1.0` | May 2025 | Moderate — hf-xet became a hard dependency on x86_64/arm64 |
| `hf-xet v1.1.3` | Jun 2025 | Upload path fixed with `AggregatingProgressUpdater` (PR #340); download path left unprotected |
| `hf-xet v1.3.0` | Feb 2026 | **Severe** — PR #645 added detailed per-chunk progress tracking to downloads, doubling callback frequency without aggregation |

PR #340 explicitly noted: *"each [update] has to acquire a global GIL lock. This negatively affects the upload speed on fast connections"* — the same problem, but only the upload side was addressed.

## Benchmarks

Downloading 3 safetensors files (16.1 GB total) from `Qwen/Qwen3.5-35B-A3B` on a 25 Gbps machine:

| Test | Before | After |
|------|--------|-------|
| `download_files()` with `progress_updater=None` (baseline) | 3119 MB/s | 3119 MB/s |
| `download_files()` with per-file Python callbacks | **746 MB/s** | **1789 MB/s** |
| `snapshot_download()` (full Python CLI path with tqdm) | ~750 MB/s | **2395 MB/s** |

Progress callback overhead drops from **4x slowdown to <1%**.

### DIFF
--- a/data/src/file_download_session.rs
+++ b/data/src/file_download_session.rs
@@ -12,8 +12,7 @@ use progress_tracking::aggregator::AggregatingProgressUpdater;
 use progress_tracking::download_tracking::{DownloadProgressTracker, DownloadTaskUpdater};
 use tracing::instrument;
 use ulid::Ulid;
-use xet_runtime::XetRuntime;
-use xet_runtime::xet_config;
+use xet_runtime::{XetRuntime, xet_config};
 
 use crate::configurations::TranslatorConfig;
 use crate::errors::*;


hoytak (approved):
This looks great to me as a quick fix.  Let's get it in. 

We're currently working actively on a session based API, and our plan there is to overhaul the progress update process to move away from a push-based model and to a polling model; objects using the session can poll for the progress.   Then, we can have hf_xet spawn a python thread on session start that polls this on intervals and reports it back.  This will avoid the part about acquiring the GIL completely.  However, that's part of a bigger change, so getting this in sooner would be good.

data/src/file_download_session.rs
```diff
@@ -125,14 +126,39 @@ impl FileDownloadSession {
...
-        let per_file_tracker = progress_updater.map(|updater| {
+        // Wrap the progress updater in an aggregator to batch updates and reduce
+        // callback overhead. Without this, each XORB chunk triggers a Rust→Python
+        // callback that acquires the GIL, causing severe contention with many
+        // concurrent file downloads.
+        let (updater_for_tracker, aggregator) = match progress_updater {
+            Some(updater) => {
+                let flush_interval = xet_config().data.progress_update_interval;
+                let sampling_window = xet_config().data.progress_update_speed_sampling_window;
+                if !flush_interval.is_zero() {
```
rajatarya, line 137: The upload path (`file_upload_session.rs:98`) gates aggregation on `config.progress_config.aggregate` in addition to `!flush_interval.is_zero()`. This method doesn't check that flag. In practice `aggregate` is always `true`, but for consistency and to respect `disable_progress_aggregation()` in tests, consider adding the same check. This would require threading `config` into this method (it's not currently available here).

rajatarya (commented):
## Code Review: fix GIL contention on download progress callbacks

Excellent, well-scoped fix. The PR description is thorough with clear root cause analysis, history, and benchmarks. The change mirrors the existing upload-side pattern from PR #340 and the 2.4x throughput improvement speaks for itself.

### Suggestions

- **Missing `config.progress_config.aggregate` check**: upload path gates aggregation on this flag; download path skips it
- **`download_file` and `download_to_writer` are not wrapped**: only `download_file_with_updater` gets aggregation — the session-level tracker in other methods may have the same problem

data/src/file_download_session.rs
rajatarya: This is more of some tech debt that we can address after this PR, not necessary for now.

rajatarya:
@Hugoch : I'll go ahead and merge and we will release this first thing in the morning (Seattle time). Thanks for fixing this! I am pretty sure this is being hit by a customer that reached out saying 1.3.0 is slower for them (they have a fast network and I bet they are using many concurrent downloads).

PR #670, opened 2026-02-27
feat: add V2 reconstruction response types for multi-range fetching

XciD:
## Summary

- Add `cas_types` V2 reconstruction response types optimized for multi-range fetching (RFC 7233)
- Update OpenAPI spec with separate V1 and V2 reconstruction endpoints
- New types: `QueryReconstructionResponseV2`, `XorbMultiRangeFetch`, `XorbRangeDescriptor`
- Implement V2 multi-range reconstruction in `cas_client` (client trait, remote client, multipart parser, file_term integration)

### How it works

1. Client sends `GET /v2/reconstructions/{file_id}` (separate endpoint from V1)
2. CAS returns V2 response with 1 signed URL per xorb (instead of N URLs per range)
3. Client sends a single multi-range HTTP GET per xorb with the full `Range` header matching the signed `X-Xet-Signed-Range`
4. Transfer service returns `multipart/byteranges` response (RFC 7233)
5. Client parses the multipart response, deserializes each part as a CAS object, and pre-populates XorbBlocks

Falls back to V1 transparently if the server doesn't support V2 (404 or 501).

### V2 response format

Instead of N presigned URLs per fragmented xorb, V2 returns 1 signed URL per xorb with all byte ranges listed separately:

```json
{
  "xorbs": {
    "abc...": [
      {
        "url": "https://transfer.xethub.hf.co/xorbs/default/abc...?X-Xet-Signed-Range=bytes%3D0-131071,500000-600000&Signature=...",
        "ranges": [
          {"chunks": {"start": 0, "end": 4}, "bytes": {"start": 0, "end": 131071}},
          {"chunks": {"start": 10, "end": 12}, "bytes": {"start": 500000, "end": 600000}}
        ]
      }
    ]
  }
}
```

### Benchmarks (24GB GGUF — `unsloth/Qwen3.5-35B-A3B-Experiments-GGUF`)

| Metric | V1 | V2 | Reduction |
|---|---|---|---|
| Reconstruction response size | 63.6 MB | 24.0 MB | **62%** |
| HTTP requests for xorb data | 60,928 | 1,226 | **98%** |
| Reconstruction response time (median, local) | ~12s | ~7s | **~40%** |
| Terms | 117,664 | 117,664 | — |
| Xorbs | 1,226 | 1,226 | — |

Tested end-to-end with `hf_xet` against a local CAS server using production data. Multi-range requests return `206 Partial Content` with `multipart/byteranges` and are parsed + deserialized correctly.

### Key implementation detail

The signed URL encodes the authorized byte ranges via the `X-Xet-Signed-Range` query parameter. The transfer service validates that the `Range` header matches — individual sub-range requests get 403. The client must send the **full** multi-range in a single GET per fetch entry, which is what the eager download path in `retrieve_from_v2` does.

### Latest changes (review followup)

- Removed `XorbFetchDescriptor` wrapper struct — `xorbs` map now uses `Vec<XorbMultiRangeFetch>` directly (simpler JSON, no `{ "fetch": [...] }` wrapper)
- Made doc comments backend-agnostic (removed CloudFront/S3 references)
- Added URL length limit estimate to doc comments (~8 KiB, ~500 ranges)
- Split OpenAPI spec into separate V1 and V2 reconstruction endpoints
- Added 501 response code to V2 endpoint

### Related

- CAS server implementation: https://github.com/huggingface-internal/xetcas/pull/496
- HfCdn multi-range validation: https://github.com/huggingface-internal/xetshared/pull/6
- Xet spec documentation: https://github.com/huggingface/hub-docs/pull/2262

### Dependencies

> **Note**: xetcas needs to be updated to point to this branch before merging. The xetcas changes (API signature updates for `usize` → `u64`, removal of `mdb_shard`/`cas_object` direct imports) depend on this PR.


cas_types/src/lib.rs
```diff
@@ -217,6 +217,53 @@ pub struct QueryReconstructionResponse {
...
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct QueryReconstructionResponseV2 {
+    pub offset_into_first_range: u64,
+    pub terms: Vec<CASReconstructionTerm>,
+    /// Map from xorb hash -> xorb fetch descriptor with single URL + all ranges.
+    pub xorbs: HashMap<HexMerkleHash, XorbFetchDescriptor>,
+}
+
+/// Fetch descriptor for a single xorb: one or more signed URLs covering all byte ranges.
+/// Usually 1 URL covers all ranges. If the URL would exceed CloudFront's 8,192 byte limit,
```
assafvayner: How many different ranges do we need to have in a URL to have to reach this limit? Add this info to the comment here.

Arguably we do not need this reason here, we do need this reasoning on the server side change

cas_types/src/lib.rs
```diff
@@ -217,6 +217,53 @@ pub struct QueryReconstructionResponse {
     pub fetch_info: HashMap<HexMerkleHash, Vec<CASReconstructionFetchInfo>>,
 }
 
+/// V2 reconstruction response - optimized for multi-range fetching.
+/// Instead of N presigned URLs per fragmented xorb, provides 1 signed
+/// CloudFront URL per xorb + all byte ranges listed separately.
```
assafvayner: Since for the S3 case we still provide urls for different ranges, I think the phrasing here should specify that the response "can/may" provide N presigned URL per xorb (and remove the ref to fragmentation).

Generally also in xet-core we should not be mentioning/needing-to-mention the S3/cloudfront backends. This client/the protocol should be agnostic to that.

cas_types/src/lib.rs
```diff
@@ -217,6 +217,53 @@ pub struct QueryReconstructionResponse {
...
+
+/// Fetch descriptor for a single xorb: one or more signed URLs covering all byte ranges.
+/// Usually 1 URL covers all ranges. If the URL would exceed CloudFront's 8,192 byte limit,
+/// ranges are split across multiple URLs.
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct XorbFetchDescriptor {
+    /// List of fetch entries. Each entry has a signed URL + the ranges it covers.
+    /// Typically 1 entry. Multiple entries only when URL length limit forces a split.
+    pub fetch: Vec<XorbMultiRangeFetch>,
+}
```
assafvayner: I think we can get rid of this as a struct and then just use Vec<XorbMultiRangeFetch> everywhere, in JSON this simplified away a bunch of `{ fetch: [array] }` into just the array.

cas_types/src/lib.rs
XciD: Done. Added (~8 KiB, roughly ~500 ranges) to the doc comment. Also added the reasoning on the server side (xetcas).

cas_types/src/lib.rs
XciD: Done. removed all CloudFront/S3 references, rephrased to: 

> "Typically 1 entry per xorb. Multiple entries only when the URL length limit forces a split." 

The client is now fully agnostic to the backend.

cas_types/src/lib.rs
XciD: Nice one. removed XorbFetchDescriptor entirely. xorbs is now HashMap<HexMerkleHash, Vec<XorbMultiRangeFetch>> everywhere (cas_types, cas_client, file_reconstruction, xetcas, OpenAPI spec)

XciD:
> requesting more documentation/fixes to docs
> 
> Also we need a new version for xet specification that includes this change: https://huggingface.co/docs/xet/index

Thank Assaf for the review. Doc updated here: https://github.com/huggingface/hub-docs/pull/2262

PR #673, opened 2026-02-27
Allow deduplication to be turned off completely for a given file.

hoytak:
This PR adds in the ability to turn of deduplication entirely using a config setting.  Default behavior hasn't changed.

### DIFF
--- a/data/src/file_cleaner.rs
+++ b/data/src/file_cleaner.rs
@@ -50,7 +50,11 @@ impl SingleFileCleaner {
         sha256: Option<Sha256>,
         session: Arc<FileUploadSession>,
     ) -> Self {
-        let deduper = FileDeduper::new(UploadSessionDataManager::new(session.clone()), file_id);
+        let deduper = FileDeduper::new(
+            UploadSessionDataManager::new(session.clone()),
+            file_id,
+            xet_config().deduplication.clone(),
+        );
 
         Self {
             file_name,
--- a/deduplication/src/file_deduplication.rs
+++ b/deduplication/src/file_deduplication.rs
@@ -8,6 +8,7 @@ use merklehash::{MerkleHash, file_hash};
 use more_asserts::{debug_assert_le, debug_assert_lt};
 use progress_tracking::upload_tracking::FileXorbDependency;
 use utils::MerkleHashMap;
+use xet_config::DeduplicationConfig;
 
 use crate::Chunk;
 use crate::constants::{MAX_XORB_BYTES, MAX_XORB_CHUNKS};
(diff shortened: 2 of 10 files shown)


data/src/file_cleaner.rs
```diff
@@ -50,7 +50,11 @@ impl SingleFileCleaner {
         sha256: Option<Sha256>,
         session: Arc<FileUploadSession>,
     ) -> Self {
-        let deduper = FileDeduper::new(UploadSessionDataManager::new(session.clone()), file_id);
+        let deduper = FileDeduper::new(
+            UploadSessionDataManager::new(session.clone()),
+            file_id,
+            xet_config().deduplication.clone(),
```
seanses, line 56: This is very likely not going to work well with the Xet migration service: the migrator is a long running process working on batch requests from a message queue and it's impossible to toggle `HF_XET_DEDUPLICATION_ENABLE` on/off to control which files are to be migrated with dedup on/off.

I think it needs a fine-grained control all the way down from `migrate_with_external_runtime`

xet_config/src/groups/deduplication.rs
```diff
@@ -1,4 +1,9 @@
 crate::config_group!({
+    /// Enable deduplication; if false, then deduplication is disabled.  Note that
+    /// this may result in significantly higher data and memory usage.
+    ref enable_deduplication: bool = true;
```
seanses, line 4: This already has the prefix `HF_XET_DEDUPLICATION_`. So better just name it `enabled`.

data/src/file_cleaner.rs
seanses: So maybe punt on this and wait to see how much the other approach alleviates the issue. And if we move on with this we'll need to coordinate with @sirahd to see how to plumb through the information.

deduplication/src/file_deduplication.rs
```diff
@@ -96,66 +101,68 @@ impl<DataInterfaceType: DeduplicationDataInterface> FileDeduper<DataInterfaceTyp
...
-                    // Check for global deduplication.
-                    if
-                    // Only do this query on the first pass.
-                    first_pass
-                        // The first hash of every file and those matching a pattern are eligible. 
-                        && (global_chunk_index == 0
-                            || hash_is_global_dedup_eligible(&chunk_hashes[local_chunk_index]))
-                        // Limit by enforcing at least 4MB between chunk queries.
-                        && global_chunk_index >= self.next_chunk_index_eligible_for_global_dedup_query
+        if self.config.enable_deduplication {
```
assafvayner, line 104: can we early return something instead of indent everything?

hoytak:
Will be taking a different route with this.

PR #674, opened 2026-02-28
V2 Reconstruction info integration in FileReconstructor; extensive tests. 

hoytak:
This PR: 
- updates the file reconstruction integration for the V2 download path.  This essentially means making multirange terms the default assumption in a XorbBlock, as the previous V1 case is a subset of that.  This logic is plumbed through all the structures consistently.
- Integrates full testing through the LocalTestServers and ensures full testing of the different paths.
- Cleans up the get_reconstruction in the Client trait to automatically attempt to query the v2 endpoint, with fallback to v1.

### DIFF
--- a/data/src/test_utils.rs
+++ b/data/src/test_utils.rs
@@ -228,7 +228,7 @@ impl HydrateDehydrateTest {
             },
             HydrationMode::ServerV1Fallback => {
                 self.ensure_server_created().await;
-                self.test_server.as_ref().unwrap().disable_v2_reconstruction();
+                self.test_server.as_ref().unwrap().disable_v2_reconstruction(404);
             },
             HydrationMode::ServerMaxRanges2 => {
                 self.ensure_server_created().await;
--- a/file_reconstruction/src/file_reconstructor.rs
+++ b/file_reconstruction/src/file_reconstructor.rs
@@ -1323,7 +1323,7 @@ mod tests {
             .unwrap();
 
         // Disable V2 so the remote client falls back to V1 + conversion.
-        server.disable_v2_reconstruction();
+        server.disable_v2_reconstruction(404);
 
         let config = test_config();
         let result = reconstruct_via_server(&server, file_contents.file_hash, None, &config)
@@ -1341,7 +1341,7 @@ mod tests {
             .await
             .unwrap();
 
-        server.disable_v2_reconstruction();
+        server.disable_v2_reconstruction(404);
 
         let file_len = file_contents.data.len() as u64;
         let range = FileRange::new(file_len / 4, file_len * 3 / 4);
@@ -1362,7 +1362,7 @@ mod tests {
             .await
             .unwrap();
 
-        server.disable_v2_reconstruction();
+        server.disable_v2_reconstruction(404);
 
         let config = test_config();
         let result = reconstruct_via_server(&server, file_contents.file_hash, None, &config)
@@ -1380,7 +1380,7 @@ mod tests {
             .await
             .unwrap();
 
-        server.disable_v2_reconstruction();
+        server.disable_v2_reconstruction(404);
 
         let config = test_config();
         let result = reconstruct_via_server(&server, file_contents.file_hash, None, &config)
@@ -1399,7 +1399,7 @@ mod tests {
             .await
             .unwrap();
 
-        server.disable_v2_reconstruction();
+        server.disable_v2_reconstruction(404);
 
         let config = test_config();
         let result = reconstruct_via_server(&server, file_contents.file_hash, None, &config)


cas_client/src/remote_client.rs
```diff
@@ -340,129 +303,20 @@ impl Client for RemoteClient {
...
-        let mut transfer_reporter = StreamProgressReporter::new(0)
-            .with_adaptive_concurrency_reporter(download_permit.get_partial_completion_reporting_function());
-        if let Some(cb) = progress_callback {
-            transfer_reporter = transfer_reporter.with_progress_callback(cb);
+        file_id: &MerkleHash,
+        bytes_range: Option<FileRange>,
+    ) -> Result<Option<QueryReconstructionResponseV2>> {
+        match self.get_reconstruction_v2(file_id, bytes_range).await {
+            Ok(result) => Ok(result),
+            Err(_) => Ok(self.get_reconstruction_v1(file_id, bytes_range).await?.map(Into::into)),
```
rajatarya: The V2-to-V1 fallback catches `Err(_)` — swallowing all V2 errors including transient network failures, 500s, auth errors, and corrupted responses. Only 404/feature-not-supported should trigger fallback; other errors should propagate. At minimum, log the V2 error at `warn` level so production failures are visible:
```rust
Err(e) => {
    tracing::warn!("V2 reconstruction failed, falling back to V1: {e}");
    Ok(self.get_reconstruction_v1(file_id, bytes_range).await?.map(Into::into))
}
```

cas_client/src/remote_client.rs
```diff
@@ -560,38 +420,109 @@ impl Client for RemoteClient {
...
+
+                                let (data, mut chunk_indices) =
+                                    cas_object::deserialize_chunks(&mut std::io::Cursor::new(part.data.as_ref()))
+                                        .map_err(|e| {
+                                            RetryableReqwestError::RetryableError(CasClientError::CasObjectError(e))
+                                        })?;
+
+                                let base_offset = all_decompressed.len() as u32;
+                                if !all_chunk_indices.is_empty() {
+                                    chunk_indices = chunk_indices.iter().skip(1).map(|&o| o + base_offset).collect();
```
rajatarya: The `skip(1)` logic assumes `deserialize_chunks` always returns a leading `0` in the chunk indices array — an undocumented invariant. More importantly, this exact chunk-merging block is copy-pasted across `remote_client.rs`, `local_client.rs`, and `memory_client.rs`. A bug fix in one must be mirrored in three places. Consider extracting this into a shared utility in `xorb_utils` alongside `compute_reconstruction_ranges`.

cas_client/src/remote_client.rs
```diff
@@ -560,38 +420,109 @@ impl Client for RemoteClient {
...
-                                        expected,
+                        let is_multipart = content_type.contains("multipart/byteranges");
+
+                        if is_multipart {
+                            // Multi-range response: collect the full body, parse the multipart
+                            // parts, and deserialize each CAS object independently. We bypass
+                            // the streaming progress wrapper since the multipart MIME overhead
+                            // would inflate the byte count beyond what the progress tracker expects.
+                            let body = resp
+                                .bytes()
```
rajatarya, line 471: For multipart responses, `resp.bytes().await` collects the entire body without `DownloadProgressStream`, so the adaptive concurrency controller receives no intermediate progress signals during what could be a large multi-range download. The single `report_progress(total_compressed_bytes)` at the end means throughput estimates are bursty. The comment acknowledges the MIME byte count issue — could the streaming wrapper be used with MIME overhead subtracted afterward instead?

rajatarya (commented):
## Code Review: V2 Reconstruction info integration in FileReconstructor

Solid refactoring that cleanly integrates V2 multirange terms as the default path while keeping V1 as a fallback. Good design choices: `compute_reconstruction_ranges` extraction into `xorb_utils`, the gap-bridging DP in `determine_size_if_possible`, `From<V1> for V2` response unification, and simplifying the `Client` trait by hiding V1/V2 internals. Test coverage with `HydrationMode` variants is thorough.

### Issues

- **V2-to-V1 fallback swallows all errors**: `Err(_)` catches network errors, 500s, etc. — not just 404/unsupported
- **Chunk merging `skip(1)` logic duplicated 3x**: fragile implicit invariant copy-pasted across remote/local/memory clients
- **Progress bypassed for multipart responses**: adaptive concurrency gets no streaming signals during large multi-range downloads
- **`compute_block_start_index` returns past-end index on miss**: silent bad index instead of explicit error
- **`reconstruction_api_version` config added but never read**: dead config or incomplete feature
- **`assert!(!terms.is_empty())` in production code**: panics instead of returning `Result`
- **`xtool` only queries V1**: no way to debug V2 endpoint issues
- **No `enable_v2_reconstruction()` counterpart**: `disable_v2_reconstruction()` is one-way, test state leaks possible

file_reconstruction/src/reconstruction_terms/file_term.rs
```diff
@@ -75,6 +83,39 @@ impl FileTerm {
...
+    offset_into_first_range: u64,
+    /// Index into the `xorb_blocks` / `xorb_block_retrieval_urls` vectors.
+    xorb_block_index: usize,
+    /// Flattened index into the xorb block's `chunk_offsets` for this term's start chunk.
+    xorb_block_start_index: usize,
+}
+
+/// Computes the index into chunk_offsets for a given chunk start within a set of chunk ranges.
+/// Returns the flattened index: sum of chunks in earlier ranges + offset within the containing range.
+fn compute_block_start_index(chunk_ranges: &[ChunkRange], chunk_start: u32) -> usize {
```
rajatarya: `compute_block_start_index` returns `idx` (past-end index) when `chunk_start` isn't found in any `chunk_ranges`. This would cause a later out-of-bounds panic when indexing into `chunk_offsets` in `extract_bytes`, producing an opaque error. Consider returning `Result` or panicking with a descriptive message:
```rust
if found { idx } else { panic!("chunk_start {chunk_start} not found in any chunk range") }
```

xet_config/src/groups/client.rs
```diff
@@ -242,4 +242,13 @@ crate::config_group!({
...
     ref unix_socket_path: Option<String> = None;
 
+    /// The reconstruction API version to request from the CAS server.
+    /// Version 1 returns per-range presigned URLs; version 2 returns
+    /// per-xorb multi-range fetch descriptors.
+    ///
+    /// The default value is 2.
+    ///
+    /// Use the environment variable `HF_XET_CLIENT_RECONSTRUCTION_API_VERSION` to set this value.
+    ref reconstruction_api_version: u32 = 2;
```
rajatarya: `reconstruction_api_version` is declared with default `2` and env var `HF_XET_CLIENT_RECONSTRUCTION_API_VERSION`, but nothing in the codebase reads this value. `RemoteClient::get_reconstruction` always tries V2 first regardless. Either wire this up to control the fallback behavior, or remove it to avoid dead config that misleads users.

cas_client/src/simulation/xorb_utils.rs
```diff
@@ -0,0 +1,513 @@
...
+        fetch_info_map.entry(segment.cas_hash).or_default().push(FetchInfoIntermediate {
+            chunk_range,
+            byte_range,
+        });
+
+        cumulative_bytes += segment.unpacked_segment_bytes as u64;
+        s_idx += 1;
+    }
+
+    assert!(!terms.is_empty());
```
rajatarya: `assert!(!terms.is_empty())` will panic in production if a well-formed file somehow produces zero matching segments. Since this is in a library function (not test code), return a `Result::Err` with a descriptive message instead.

xet_config/src/groups/client.rs
rajatarya: This probably should be removed as dead code as v2 is always attempted.

cas_client/src/simulation/local_server/handlers.rs
```diff
@@ -284,13 +379,58 @@ pub async fn fetch_term(
...
     };
 
-    // Get total length of the raw XORB data for Range header handling
-    let total_length = match state.xorb_raw_length(&xorb_hash).await {
+    if !decoded.byte_ranges.is_empty() {
+        if decoded.byte_ranges.len() == 1 {
+            // Single range: return the raw bytes directly (standard 206 single-range).
+            let range = &decoded.byte_ranges[0];
+            return match state.get_xorb_raw_bytes(&decoded.hash, Some(*range)).await {
+                Ok(data) => (StatusCode::OK, data).into_response(),
```
XciD: nit: I think its a 206 here for range requests.

xet_config/src/groups/client.rs
hoytak: I think it could be useful to be able to force v1 for testing or benchmarking purposes, at least for a short bit.  I'll have the get_reconstruction respect this for now.

cas_client/src/simulation/xorb_utils.rs
hoytak: It's code used only for testing, so it's not critical to really do the error handling -- debug_assert should be enough here.

cas_client/src/remote_client.rs
hoytak: I implemented this, and also have it fall back to V1 for 501 Not implemented errors as well.  I can imagine scenarios where the fetch urls don't support multirange reads, so this allows us to cleaning determine that on the server end.

cas_client/src/remote_client.rs
hoytak: I switched it to report after each segment instead.  This could still be a delay, but the streaming would be pretty hard to incorporate.  I'll try, but let's do that on a subsequent PR.

file_reconstruction/src/reconstruction_terms/file_term.rs
hoytak: It's guaranteed by the logic outside this function, but this doesn't need to be a helper function so I integrated it back in to the code where the correctness is more obvious. .

XciD:
@hoytak @rajatarya going to merge to avoid conflicts as I'm fixing Assaf comments.

PR #675, opened 2026-03-01
feat: expose CAS client factory and optional chunk cache

XciD:
## Summary

- **`data::create_remote_client()` made public**: Allows downstream consumers to create a CAS client and wrap it with custom caching layers before passing to `FileDownloadSession::from_client()`.

- **Re-exports from `data` crate**: `CasClient`, `ChunkCache`, `CacheConfig`, `get_cache`, and `create_remote_client` are publicly accessible, so downstream crates don't need to depend on `cas_client`/`chunk_cache` directly.

- **Optional `ChunkCache` in download path**: `FileDownloadSession`, `FileReconstructor`, and `XorbBlock` accept an optional `ChunkCache`. When provided, xorb blocks are looked up in the on-disk cache before making HTTP requests, and stored after download. This enables cross-file deduplication for mount-style workloads where the same xorbs are accessed repeatedly.

- **Lazy permit acquisition**: Download permits are acquired only on chunk cache miss, so cache hits don't consume network concurrency slots.

- **Pass file size to `FileReconstructor`**: For full-file downloads, `FileReconstructor` now uses `FileRange::new(0, file_size)` instead of `FileRange::full()` (0..u64::MAX). This prevents `ReconstructionTermManager` from speculatively prefetching blocks beyond EOF, which caused the CAS server to return 416 Range Not Satisfiable for small files.

- **`start_clean` accepts `Option<u64>` for size**: Changed the `size` parameter from `u64` to `Option<u64>`. Passing `None` signals that the final size is unknown (streaming uploads), which prevents `debug_assert` panics when `completed_bytes` exceeds the initially declared `total_bytes=0`.

Note: this is a breaking change to `FileDownloadSession::new`, `from_client`, and `start_clean` signatures. The only known consumers are `hf_xet` (in this repo) and `hf-mount`.

### DIFF
--- a/xet_data/src/processing/bin/example.rs
+++ b/xet_data/src/processing/bin/example.rs
@@ -92,7 +92,9 @@ async fn clean(mut reader: impl Read, mut writer: impl Write, size: u64) -> Resu
         FileUploadSession::new(TranslatorConfig::local_config(std::env::current_dir()?)?.into(), None).await?;
 
     let mut size_read = 0;
-    let mut handle = translator.start_clean(None, size, Sha256Policy::Compute, Ulid::new()).await;
+    let mut handle = translator
+        .start_clean(None, Some(size), Sha256Policy::Compute, Ulid::new())
+        .await;
 
     loop {
         let bytes = reader.read(&mut read_buf)?;
--- a/xet_data/src/processing/data_client.rs
+++ b/xet_data/src/processing/data_client.rs
@@ -265,7 +265,7 @@ pub async fn clean_bytes(
     #[allow(clippy::unwrap_or_default)] // Ulid::default is Ulid::nil
     let tracking_id = tracking_id.unwrap_or_else(Ulid::new);
     let mut handle = processor
-        .start_clean(None, bytes.len() as u64, Sha256Policy::Compute, tracking_id)
+        .start_clean(None, Some(bytes.len() as u64), Sha256Policy::Compute, tracking_id)
         .await;
     handle.add_data(&bytes).await?;
     handle.finish().await
@@ -292,7 +292,7 @@ pub async fn clean_file(
     let mut handle = processor
         .start_clean(
             Some(filename.as_ref().to_string_lossy().into()),
-            filesize,
+            Some(filesize),
             Sha256::from_hex(sha256.as_ref()).ok().into(),
             tracking_id,
         )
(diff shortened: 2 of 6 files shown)


XciD:
Addressed all review comments:

**#1 Breaking change**: Updated PR description to explicitly note the breaking signature change. Only known consumers are `hf_xet` (already updated), so wrapper methods would be unnecessary surface.

**#2 `create_client` vs `create_remote_client`**: Fixed — we removed the `create_client` wrapper and made `create_remote_client` public directly. PR description updated to match.

**#3 Empty prefix**: Fixed in cf336ae — now using `"default"` to match `PREFIX_DEFAULT` / the rest of the codebase.

**#4 Tests**: Acknowledged, will add in a follow-up.

**#5 Permit on cache hit**: Fixed in cf336ae — `acquire_download_permit()` moved from `get_data_task` into `retrieve_data`, after the chunk cache check. Cache hits no longer block on the network concurrency semaphore.

rajatarya:
@XciD : Can you share a bit about the context for these changes? Maybe I haven't read them carefully enough - is the goal to enable the chunk_cache on the new adaptive concurrency download/upload paths?

XciD:
My comment have been removed, re-adding it:

I'm pocing a Fuse/NFS mount with xet for hf buckets, and those are change I need to enable xorb cache level on it.

rajatarya:
> My comment have been removed, re-adding it:
> 
> I'm pocing a Fuse/NFS mount with xet for hf buckets, and those are change I need to enable xorb cache level on it.

Ah - that makes sense! @assafvayner didn't you have a branch with this somewhere? I remember there being a demo after the offsite last year. Might make sense to resurrect that branch and build from it instead.

assafvayner:
this is my mount branch, but it's a outdates as far as as the new download interface. It does set up an nfs mount daemon though. https://github.com/huggingface/xet-core/tree/assaf/mount

It uses a fork (https://github.com/huggingface/nfsserve) of the original xetdata/nfsserve library that provides scaffolding for setting up nfs v3

XciD:
Thanks for sharing, did you share it somewhere? I can't find anything on Slack.
I started this weekend as a side project with a "S3 Fuse Mount"-like for HF Buckets, and as it grew, I decided to impl both NFS + FUSE inside the same project: https://github.com/huggingface-internal/hf-mount

Would love your thoughts on the implementation. Was done with Claude, and re-read carefully. I think I will post on Slack later this week if we can start to test it over science cluster for example.

XciD:
Will split this one in multiple small PR

#730 #731 #732

PR #678, opened 2026-03-03
feat: accept pre-computed SHA-256 in upload_files()

XciD:
## Summary

- Add optional `sha256s` keyword parameter to the Python-exposed `upload_files()` function
- Forward it to `data_client::upload_async()` which already supports it

### Double computation today

`huggingface_hub` computes SHA-256 on every file during `CommitOperationAdd.__post_init__()` for LFS batch negotiation, then `hf_xet` recomputes it internally because `upload_files()` doesn't accept pre-computed hashes.

### Performance impact

This change eliminates the redundant computation entirely.

### Backward compatibility

- `sha256s` is a keyword-only parameter with default `None` — no change for existing callers
- `data_client::upload_async()` already accepts `sha256s: Option<Vec<String>>` since day one
- When provided, `SingleFileCleaner` uses `ShaGenerator::ProvidedValue` and skips internal recomputation

Companion PR: huggingface/huggingface_hub#3876

### DIFF
--- a/hf_xet/src/lib.rs
+++ b/hf_xet/src/lib.rs
@@ -141,14 +141,14 @@ pub fn upload_files(
     request_headers: Option<HashMap<String, String>>,
     sha256s: Option<Vec<String>>,
 ) -> PyResult<Vec<PyXetUploadInfo>> {
-    if let Some(ref s) = sha256s {
-        if s.len() != file_paths.len() {
-            return Err(PyRuntimeError::new_err(format!(
-                "sha256s length ({}) must match file_paths length ({})",
-                s.len(),
-                file_paths.len()
-            )));
-        }
+    if let Some(ref s) = sha256s
+        && s.len() != file_paths.len()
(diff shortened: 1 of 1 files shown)


Wauplin:
Really nice finding! I reviewed the `huggingface_hub` related PR which looks good to me. We are planning a release by the end of the week so if it can be included in it it's good (otherwise no problem^^).

(orthogonal but https://github.com/huggingface/xet-core/pull/679 would also be nice to have for Buckets upload :smiley: )

hoytak (approved):
LGTM!

XciD:
> It would be good if we have some unit-tests for this case, but otherwise looks good.
> 
> We should have some validation when provided that there is a 1:1 match in lengths between file_paths and sha256s.

Yeah you right, catching the array length is easy check. Also added a test. 
testing that sha is not compute is not trivial

seanses (approved):

PR #679, opened 2026-03-03
feat: add skip_sha256 option to SingleFileCleaner

XciD:
## Summary

- Add `ShaGenerator::Skip` variant that skips SHA-256 computation entirely
- `ShaGenerator::finalize()` now returns `Option<Sha256>` (None when skipped)
- `SingleFileCleaner::new()` and `FileUploadSession::start_clean()` accept a `skip_sha256` boolean
- When skipped, no `FileMetadataExt` is included in the shard

## Context

Bucket uploads don't need SHA-256 in the shard metadata — the `sha_index` GSI is only used for LFS pointer resolution, which doesn't apply to buckets. Skipping SHA-256 for bucket uploads removes the main CPU bottleneck in the upload pipeline on non-SHA-NI instances.

## Alternative: dummy SHA-256

Instead of skipping entirely, the client could send a zeroed/dummy `FileMetadataExt`. The server would still store it but queries would never match. This avoids the server-side schema change (xetcas PR) but pollutes the GSI with dummy entries.

Companion PRs:
- xetcas: huggingface-internal/xetcas#498 (make `FileIdItem.sha256` optional server-side)

### DIFF
--- a/data/src/file_cleaner.rs
+++ b/data/src/file_cleaner.rs
@@ -15,7 +15,26 @@ use crate::XetFileInfo;
 use crate::deduplication_interface::UploadSessionDataManager;
 use crate::errors::Result;
 use crate::file_upload_session::FileUploadSession;
-use crate::sha256::ShaGenerator;
+use crate::sha256::Sha256Generator;
+
+/// Controls how SHA-256 is handled during file cleaning.
+pub enum Sha256Policy {
+    /// Compute SHA-256 from the file data.
+    Compute,
+    /// Use a pre-computed SHA-256 value.
+    Provided(Sha256),
+    /// Skip SHA-256 entirely; no metadata_ext is written to the shard.
+    Skip,
+}
+
+impl From<Option<Sha256>> for Sha256Policy {
+    fn from(sha256: Option<Sha256>) -> Self {
+        match sha256 {
+            Some(hash) => Self::Provided(hash),
+            None => Self::Compute,
+        }
+    }
+}
 
 /// A class that encapsulates the clean and data task around a single file.
 pub struct SingleFileCleaner {
@@ -35,22 +54,30 @@ pub struct SingleFileCleaner {
     // on await so that we can background this part.
     dedup_manager_fut: Pin<Box<dyn Future<Output = Result<FileDeduper<UploadSessionDataManager>>> + Send + 'static>>,
 
-    // Generating the sha256 hash
-    sha_generator: ShaGenerator,
+    // SHA-256 generator, present only when computing from file data.
+    sha_generator: Option<Sha256Generator>,
+
+    // Pre-computed or finalized SHA-256 value.
+    provided_sha256: Option<Sha256>,
 
     // Start time
     start_time: DateTime<Utc>,
 }
 
 impl SingleFileCleaner {
-    // If a sha256 value is given in the parameter, the cleaner avoids computing the sha256 again internally.
     pub(crate) fn new(
         file_name: Option<Arc<str>>,
         file_id: CompletionTrackerFileId,
-        sha256: Option<Sha256>,
+        sha256: Sha256Policy,
         session: Arc<FileUploadSession>,
     ) -> Self {
         let deduper = FileDeduper::new(UploadSessionDataManager::new(session.clone()), file_id);
+
+        let (sha_generator, provided_sha256) = match sha256 {
+            Sha256Policy::Compute => (Some(Sha256Generator::default()), None),
+            Sha256Policy::Provided(hash) => (None, Some(hash)),
+            Sha256Policy::Skip => (None, None),
+        };
 
         Self {
             file_name,
@@ -58,7 +85,8 @@ impl SingleFileCleaner {
             dedup_manager_fut: Box::pin(async move { Ok(deduper) }),
             session,
             chunker: deduplication::Chunker::default(),
-            sha_generator: sha256.map(ShaGenerator::ProvidedValue).unwrap_or_else(ShaGenerator::generate),
+            sha_generator,
+            provided_sha256,
             start_time: Utc::now(),
         }
     }
@@ -123,7 +151,9 @@ impl SingleFileCleaner {
         };
 
         // Update the sha256 hasher, which hands this off to be done in the background.
-        self.sha_generator.update(data.clone()).await?;
+        if let Some(ref mut generator) = self.sha_generator {
+            generator.update(data.clone()).await?;
+        }
 
         // Get the chunk data and start processing it.
         let (chunks, chunker) = chunk_data_jh.await?;
@@ -142,7 +172,7 @@ impl SingleFileCleaner {
         Ok(())
     }
 
-    /// Ensures all current background work is completed.  
+    /// Ensures all current background work is completed.
     pub async fn checkpoint(&mut self) -> Result<()> {
         // Flush the background process by sending it a dummy bit of data.
         self.deduper_process_chunks(Arc::new([])).await
@@ -157,12 +187,16 @@ impl SingleFileCleaner {
             self.deduper_process_chunks(data).await?;
         }
 
-        // Finalize the sha256 hashing and create the metadata extension
-        let sha256: Sha256 = self.sha_generator.finalize().await?;
-        let metadata_ext = FileMetadataExt::new(sha256);
+        // Resolve the SHA-256: computed, provided, or skipped.
+        let sha256 = if let Some(generator) = self.sha_generator.take() {
+            Some(generator.finalize().await?)
+        } else {
+            self.provided_sha256
+        };
+        let metadata_ext = sha256.map(FileMetadataExt::new);
 
         let (file_hash, remaining_file_data, deduplication_metrics) =
-            self.dedup_manager_fut.await?.finalize(Some(metadata_ext));
+            self.dedup_manager_fut.await?.finalize(metadata_ext);
 
         let file_info = XetFileInfo::new(file_hash.hex(), deduplication_metrics.total_bytes);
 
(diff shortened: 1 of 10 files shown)


data/src/file_cleaner.rs
```diff
@@ -44,21 +44,29 @@ pub struct SingleFileCleaner {
...
     pub(crate) fn new(
         file_name: Option<Arc<str>>,
         file_id: CompletionTrackerFileId,
         sha256: Option<Sha256>,
+        skip_sha256: bool,
         session: Arc<FileUploadSession>,
     ) -> Self {
         let deduper = FileDeduper::new(UploadSessionDataManager::new(session.clone()), file_id);
 
+        let sha_generator = if skip_sha256 {
```
rajatarya: The `skip_sha256` flag silently wins when both `sha256 = Some(hash)` and `skip_sha256 = true` are passed — the provided hash is discarded with no warning.

At minimum, a `debug_assert!` would catch misuse:
```rust
debug_assert!(!(skip_sha256 && sha256.is_some()), "skip_sha256 and sha256 are mutually exclusive");
```

Better yet, consider collapsing `sha256: Option<Sha256>` and `skip_sha256: bool` into a single enum:
```rust
pub enum Sha256Policy {
    Compute,
    Provided(Sha256),
    Skip,
}
```
This eliminates the invalid state at the type level and makes callsites self-documenting (`Sha256Policy::Skip` vs an opaque `false`).

data/src/file_upload_session.rs
```diff
@@ -252,20 +252,22 @@ impl FileUploadSession {
...
-    /// avoid redundant computation.
+    /// avoid redundant computation. If skip_sha256 is true, SHA-256 computation is
+    /// skipped entirely and no metadata_ext is included in the shard.
     pub async fn start_clean(
         self: &Arc<Self>,
         file_name: Option<Arc<str>>,
         size: u64,
         sha256: Option<Sha256>,
+        skip_sha256: bool,
     ) -> SingleFileCleaner {
```
rajatarya, line 263: With this signature, callsites now read `start_clean(None, size, None, false)` — the trailing `false` is opaque without jumping to the definition.

If the enum approach from the other comment is adopted, this collapses into a single `sha256: Sha256Policy` param replacing both `sha256` and `skip_sha256`.

data/src/sha256.rs
```diff
@@ -6,20 +6,23 @@ use xet_runtime::XetRuntime;
 pub enum ShaGenerator {
     Generate(Sha256Generator),
     ProvidedValue(Sha256),
+    /// Skip SHA-256 computation entirely. `finalize()` returns `None`.
+    Skip,
```
rajatarya: The `Skip` variant has no test coverage. Worth adding at least:
```rust
#[tokio::test]
async fn test_sha_skip() {
    let mut gen = ShaGenerator::Skip;
    gen.update(b"some data").await.unwrap();
    assert!(gen.finalize().await.unwrap().is_none());
}
```

data/src/sha256.rs
```diff
@@ -6,20 +6,23 @@ use xet_runtime::XetRuntime;
...
     }
 
-    pub async fn finalize(self) -> Result<Sha256, JoinError> {
+    pub async fn finalize(self) -> Result<Option<Sha256>, JoinError> {
         match self {
-            Self::Generate(generator) => generator.finalize().await,
-            Self::ProvidedValue(hash) => Ok(hash),
+            Self::Generate(generator) => generator.finalize().await.map(Some),
+            Self::ProvidedValue(hash) => Ok(Some(hash)),
+            Self::Skip => Ok(None),
```
rajatarya: nit: Since `finalize` now returns `Option<Sha256>`, the existing tests (`test_sha_generation_builder`, etc.) that go through `ShaGenerator` will need to unwrap the `Option` — otherwise they won't compile. (They may already have been updated — just flagging in case.)

data/src/file_cleaner.rs
XciD: love the enum idea. on it

data/src/file_upload_session.rs
XciD: solved with enum

XciD:
Addressed all comments:
- Replaced `(sha256: Option<Sha256>, skip_sha256: bool)` with a `Sha256Policy` enum (`Compute`, `Provided`, `Skip`)
- Added test for `ShaGenerator::Skip`
- Added a comment documenting the WASM divergence

Out of curiosity — who actually uses the WASM upload path today?

seanses:
> Out of curiosity — who actually uses the WASM upload path today?

Not by any official clients, but some guy showed up recently and fixed our wasm build script -- apparently he was experimenting with it.

data/src/lib.rs
```diff
@@ -4,7 +4,7 @@ mod deduplication_interface;
 pub mod errors;
 mod file_cleaner;
 mod file_download_session;
-mod file_upload_session;
+pub mod file_upload_session;
```
hoytak: Instead of making file_upload_session be a pub mod, let's just add `pub use Sha256Policy;` here.

hoytak:
It's a weird abstraction that the ShaGenerator class actually handles the logic of whether or not it gets computed.  I think we should leave it as it is and move all of that logic to the SingleFileCleaner class, making the `sha_generator : ShaGenerator` there be `sha_generator : Option<ShaGenerator>` with `provided_sha : Option<...>` and making the computation logic around what happens explicit.  It'll just mean adding a few `if let Some(...) ` but it'll make it cleaner and more readable.  

Then move the Sha256Policy to file_cleaner.rs.

rajatarya:
## Cross-Cutting Review: xet-core#679 + xetcas#498

These two PRs form a paired change: the client (this PR) stops sending SHA-256 in shard metadata for bucket uploads, and the server (xetcas#498) makes `FileIdItem.sha256` optional so it can accept those shards. Here's an analysis of the concerns that span both PRs.

---

### 1. Deployment Ordering — Hard Dependency

**Risk: HIGH**

The server **must** be deployed before any client running this code. The current flow:

- **Client (this PR):** When `skip_sha256 = true`, `FileMetadataExt` is omitted from the shard binary. The `FileDataSequenceHeader.contains_metadata_ext()` flag is `false`, so the shard doesn't contain the SHA section at all.
- **Server (xetcas#498):** `read_file_info_ranges()` already returns `Option<MerkleHash>` — it respects the `contains_metadata_ext()` flag. **However**, the old server code at `database.rs:294-296` treats a missing SHA as a 400 error:
  ```rust
  let sha256 = sha.ok_or_else(|| {
      CasError::bad_request(format!("invalid shard: sha for file: {file_id:x} is missing"))
  })?;
  ```
  So a new client sending SHA-less shards to the old server will get rejected.

**Recommendation:** Both PRs should document the deployment order explicitly. Consider a feature flag or version negotiation so the client only skips SHA-256 when it knows the server supports it. At minimum, the rollout plan should be:
1. Deploy xetcas#498 to all server instances
2. Only then release a client version with this PR

---

### 2. Wire Format Compatibility — Already Sound

**Risk: LOW**

The shard binary format is already compatible. `FileDataSequenceHeader` has a `contains_metadata_ext()` bit flag that controls whether the `FileMetadataExt` section is present. The server's `read_file_info_ranges()` (in `mdb_shard/src/shard_format.rs:917`) already checks this flag and returns `None` when the section is absent. This is not a new format — it's an existing capability that was previously unused because the client always sent metadata.

---

### 3. No End-to-End Integration Test

**Risk: MEDIUM**

Neither PR includes a test that exercises the full path: client creates a shard without SHA-256 → server parses it → DynamoDB item has `sha256: None` → GSI doesn't index it → LFS pointer resolution still works for other files.

The closest tests are:
- xet-core#679: No test with `skip_sha256 = true` at all
- xetcas#498: Tests update existing `Some(sha256)` paths but don't test `sha256: None` items against the GSI

**Recommendation:** At minimum, add a server-side test in xetcas that constructs a shard binary without `FileMetadataExt`, calls `sync_shard_impl`, and verifies the resulting `FileIdItem` has `sha256: None` and doesn't appear in GSI queries.

---

### 4. Semantic Gap in `sizzle_sync`

**Risk: MEDIUM**

In xetcas#498, `sizzle_sync/src/commands/dynamodb_sync.rs` handles missing SHA-256 with `unwrap_or_default()`, converting `None` to `""`. But the client-side never sends an empty string — it either sends a valid SHA-256 hex string or omits the field entirely. The empty string `""` is a third state that doesn't exist in the source data.

If `sizzle_sync` downstream code does string comparisons, joins, or lookups using the `sha256` field, `""` will behave differently than a proper `None`/absent value. This is a cross-repo semantic leak.

---

### 5. No Caller of `skip_sha256 = true` in This PR

**Risk: LOW (intentional)**

This PR adds the `skip_sha256` parameter but every callsite passes `false`. The actual bucket-upload integration that passes `true` is presumably in a follow-up PR. This is fine for incremental merging, but means the feature is untestable in production until that follow-up lands. Consider noting the follow-up PR in the description so reviewers know where the feature gets activated.

---

### 6. WASM Client Divergence

**Risk: LOW-MEDIUM**

The WASM client (`hf_xet_wasm/src/wasm_file_cleaner.rs`) has its own `SingleFileCleaner` that always computes SHA-256 and always passes `Some(metadata_ext)` to `finalize()`. If bucket uploads ever go through the WASM path, they'll still compute SHA-256 unnecessarily. The two `SingleFileCleaner` implementations are diverging — the native one now has `skip_sha256` but the WASM one doesn't. This should be documented, and if WASM bucket uploads are planned, the WASM cleaner should be updated in a follow-up.

---

### Summary Table

| Concern | Risk | Mitigation |
|---------|------|------------|
| Deployment ordering | HIGH | Document order; consider feature flag |
| Wire format compat | LOW | Already supported by shard format |
| No E2E integration test | MEDIUM | Add server-side test with SHA-less shard |
| sizzle_sync empty string | MEDIUM | Make field `Option<String>` or document why `""` is safe |
| No caller passes `true` yet | LOW | Note follow-up PR in description |
| WASM divergence | LOW-MEDIUM | Document; update WASM if needed |

PR #685, opened 2026-03-06
fix: per-request timeout for shard uploads (XET-885)

rajatarya:
Fixes [XET-885](https://linear.app/xet/issue/XET-885/investigate-unsloth-upload-failure-shard-upload-timeout-on-cas)

## Summary

Shard uploads to CAS can take a long time due to server-side processing (DynamoDB writes scale with file entry count). The default `read_timeout(120s)` on the reqwest client kills these uploads.

**Key insight:** reqwest's per-request `RequestBuilder::timeout()` does NOT override the client-level `read_timeout()` — they are independent mechanisms polled as separate futures. So the original approach of using per-request timeouts was ineffective.

**Fix:** Create a dedicated `shard_upload_http_client` on `RemoteClient` with **no `read_timeout`**, built once at construction time and reused for all shard uploads. All other settings (connect timeout, pool config, auth middleware) are identical to the standard client.

### `cas_client/src/http_client.rs`
- Added `reqwest_client_no_read_timeout()` — creates a reqwest client with no `read_timeout`
- Added `build_auth_http_client_no_read_timeout()` — public API wrapping it with middleware
- 4 unit tests for the new builder

### `cas_client/src/remote_client.rs`
- Added `shard_upload_http_client` field to `RemoteClient` (cfg'd out on wasm)
- `upload_shard()` uses the pre-built no-timeout client instead of building one per request

### `cas_client/tests/test_shard_upload_timeout.rs`
- Updated: slow server test now asserts **success** (shard uploads should wait as long as needed)

### `xet_config/src/groups/client.rs`
- Removed `shard_read_timeout` config field (no longer needed)

## Verification

\`\`\`bash
# All tests pass
cargo test -p cas_client  # 87 lib + 2 integration + 2 doc tests

# Clippy clean
cargo clippy -p cas_client -- -D warnings

# Format clean
cargo +nightly fmt --check
\`\`\`

### DIFF
--- a/cas_client/src/remote_client.rs
+++ b/cas_client/src/remote_client.rs
@@ -1,7 +1,5 @@
 use std::sync::Arc;
 use std::sync::atomic::{AtomicU64, Ordering};
-#[cfg(not(target_family = "wasm"))]
-use std::time::Duration;
 
 use bytes::Bytes;
 use cas_types::{
@@ -44,12 +42,12 @@ pub struct RemoteClient {
     dry_run: bool,
     http_client: Arc<ClientWithMiddleware>,
     authenticated_http_client: Arc<ClientWithMiddleware>,
+    /// Authenticated client with no read_timeout, used for shard uploads where server-side
+    /// processing time scales with file entry count and can exceed the global read_timeout.
+    #[cfg(not(target_family = "wasm"))]
+    shard_upload_http_client: Arc<ClientWithMiddleware>,
     upload_concurrency_controller: Arc<AdaptiveConcurrencyController>,
     download_concurrency_controller: Arc<AdaptiveConcurrencyController>,
-    auth_config: Option<AuthConfig>,
-    session_id: String,
-    unix_socket_path: Option<String>,
-    custom_headers: Option<Arc<HeaderMap>>,
 }
 
 impl RemoteClient {
@@ -80,12 +78,13 @@ impl RemoteClient {
             http_client: Arc::new(
                 http_client::build_http_client(session_id, unix_socket_path, custom_headers.clone()).unwrap(),
             ),
+            #[cfg(not(target_family = "wasm"))]
+            shard_upload_http_client: Arc::new(
+                http_client::build_auth_http_client_no_read_timeout(auth, session_id, unix_socket_path, custom_headers)
+                    .unwrap(),
+            ),
             upload_concurrency_controller: AdaptiveConcurrencyController::new_upload("upload"),
             download_concurrency_controller: AdaptiveConcurrencyController::new_download("download"),
-            auth_config: auth.clone(),
-            session_id: session_id.to_string(),
-            unix_socket_path: unix_socket_path.map(|s| s.to_string()),
-            custom_headers,
         })
     }
 
@@ -417,30 +416,12 @@ impl Client for RemoteClient {
         let api_tag = "cas::upload_shard";
         let url = Url::parse(&format!("{}/shards", self.endpoint))?;
 
-        // Build a dedicated HTTP client for this shard upload with an appropriate read_timeout.
-        // reqwest's per-request timeout() does NOT override the client-level read_timeout(),
-        // so we must create a separate client with the correct read_timeout for shard uploads.
-        // Server-side shard processing scales linearly with file entry count and can exceed
-        // the global read_timeout (120s) for large shards.
+        // Use the no-read-timeout client for shard uploads. reqwest's per-request timeout()
+        // does NOT override the client-level read_timeout(), so we use a separate client
+        // with no read_timeout. Server-side shard processing scales linearly with file entry
+        // count and can exceed the global read_timeout (120s) for large shards.
         #[cfg(not(target_family = "wasm"))]
-        let client = {
-            let shard_read_timeout = {
-                let configured = xet_config().client.shard_read_timeout;
-                if configured.is_zero() {
-                    Duration::from_secs((n_upload_bytes / 1024).clamp(120, 600) as u64)
-                } else {
-                    configured
-                }
-            };
-
-            Arc::new(http_client::build_auth_http_client_with_read_timeout(
-                &self.auth_config,
-                &self.session_id,
-                self.unix_socket_path.as_deref(),
-                self.custom_headers.clone(),
-                shard_read_timeout,
-            )?)
-        };
+        let client = self.shard_upload_http_client.clone();
 
         #[cfg(target_family = "wasm")]
         let client = self.authenticated_http_client.clone();
--- a/cas_client/src/http_client.rs
+++ b/cas_client/src/http_client.rs
@@ -108,14 +108,14 @@ fn reqwest_client(
     Ok(client)
 }
 
-/// Creates a reqwest client with a custom read_timeout. Not cached, since the timeout
-/// varies per request (e.g., based on shard size).
+/// Creates a reqwest client with no read_timeout. Used for shard uploads where server-side
+/// processing time is unbounded and would otherwise be killed by the global read_timeout.
+/// Not cached — uses a separate connection pool from the main client.
 #[cfg(not(target_family = "wasm"))]
 #[allow(unused_variables)]
-fn reqwest_client_with_read_timeout(
+fn reqwest_client_no_read_timeout(
     unix_socket_path: Option<&str>,
     custom_headers: Option<Arc<HeaderMap>>,
-    read_timeout: std::time::Duration,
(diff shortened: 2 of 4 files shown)


rajatarya:
@hoytak Do you think I should add a config override for this so users can set an env variable to configure this?

danielhanchen:
Nice work thanks! Would be nice if there was a patch pypi release for this soon!

hoytak:
If this fixes it, then LGTM! 

A config var would make sense -- the max timeout -- e.g. large shard timeout?  There are situations where 5 min may not be enough.

rajatarya:
> If this fixes it, then LGTM!
> 
> A config var would make sense -- the max timeout -- e.g. large shard timeout? There are situations where 5 min may not be enough.

This PR has the max time be 10m (clamped at 600s) - which I imagine should be enough for 128MB shard - which I believe is the max shard size.

I am building out some tests for this and then will merge.

rajatarya:
Fortunately @XciD did some actual testing with this patch and learned that the per-request timeout **does not take precedence** over the overall read_timeout timing. If overall read_timeout is set then per-request being longer doesn't matter.

Need to rethink this fix. Also, the timeout is silently failing - need to add testing to confirm that when timeout occurs we need to actually fail the upload with a reasonable explanation.

More in internal Slack thread: https://huggingface.slack.com/archives/C087TU2FE3G/p1772992380446739?thread_ts=1772768281.424279&cid=C087TU2FE3G

rajatarya:
## Update: Switching from per-request timeout to dedicated shard upload client

### Problem
The current approach uses `RequestBuilder::timeout()` to set per-request timeouts on shard uploads. However, this **does not override** `ClientBuilder::read_timeout()` — they are independent mechanisms in reqwest.

### Evidence
- reqwest's source constructs `total_timeout` and `read_timeout_fut` as **independent futures** — both are polled simultaneously
- `RequestBuilder::timeout()` only sets the `TotalTimeout` extension, which overrides `ClientBuilder::timeout()` — it has **no effect** on `read_timeout`
- The [reqwest docs](https://docs.rs/reqwest/latest/reqwest/struct.RequestBuilder.html#method.timeout) explicitly state it overrides `ClientBuilder::timeout()` — no mention of `read_timeout`
- @XciD confirmed this empirically: the client's `read_timeout(120s)` still fires when the CAS server stalls during shard processing, killing the upload regardless of the per-request `timeout(300s)`

### Plan
Create a **separate reqwest client** for shard uploads with the dynamic timeout as its `read_timeout`:

1. **`http_client.rs`** — Add `build_auth_http_client_with_read_timeout()` that creates a reqwest client with a custom `read_timeout`. This client is **not cached** since the timeout varies per shard size.
2. **`remote_client.rs`** — Store construction parameters (auth, session_id, socket path, headers) in `RemoteClient` so shard-specific clients can be built on demand. In `upload_shard()`, compute the dynamic timeout, build a new `ClientWithMiddleware` with that `read_timeout`, and use it for the shard POST.
3. **Remove the per-request `timeout()` call** — it's ineffective.
4. **Update tests** — Add unit tests for the new client builder function covering boundary cases.

cas_client/src/remote_client.rs
```diff
@@ -44,6 +46,10 @@ pub struct RemoteClient {
     authenticated_http_client: Arc<ClientWithMiddleware>,
     upload_concurrency_controller: Arc<AdaptiveConcurrencyController>,
     download_concurrency_controller: Arc<AdaptiveConcurrencyController>,
+    auth_config: Option<AuthConfig>,
```
rajatarya: @seanses @hoytak : any concerns with adding this to the RemoteClient struct?

Since a new RemoteClient is created on each shard upload request this is recording these - any concerns with that approach?

seanses: Hmm.. I guess we don't want to create a new client for each request. Can we hold another authenticated_http_client_with_no_timeout in this RemoteClient struct?

XciD:
did you find why the error was silenced ? worth investigating if we are silenting some other errors.
Could have been a retry policy failling over and over in my case, but worth checking

cas_client/src/remote_client.rs
rajatarya: Done in latest commit.

rajatarya:
> did you find why the error was silenced ? worth investigating if we are silenting some other errors. Could have been a retry policy failling over and over in my case, but worth checking

From the analysis it should have reported an ERROR message in the logs after all retries were exhausted. Each retry should be in the logs at INFO level. What is less clear is what makes its way back to Python - but for the testing you were doing that was in Rust, right?

Here is the writeup for the swallowed errors from Claude:
```
  Why Timeout Errors Appear Silent

  The error path is: reqwest timeout → RetryWrapper → upload_shard → shard_interface → file_upload_session::finalize → upload_commit → user.

  The error does propagate — it's not silently discarded in code. But here's why it appeared silent:

  Root Cause: Timeouts are classified as Retryable::Transient

  In on_request_failure (line 473):
  if error.is_timeout() || is_connect {
      Some(Retryable::Transient)
  }

  This means:

  1. Each timeout is logged at info level, not error — process_error_response logs transient errors with log_as_info = true (line 115). If the operator or user doesn't have info-level logging enabled, they see nothing during the retry attempts.
  2. The retry loop silently retries 3 times (default retry_max_attempts). With the old 120s read_timeout, that's up to 6 minutes of silent retries before failing. Each retry hits the same timeout.
  3. The final "No more retries; aborting" IS logged at error level (line 320), but only after exhausting all retries.
  4. From the user's perspective: the upload hangs for ~6 minutes (3 × 120s timeouts), then fails with a generic CasClientError that wraps the reqwest timeout. The error message may not clearly indicate "shard upload timed out" — it's wrapped in layers of error types.

  Contributing factors:

  - No progress indication during shard upload — unlike xorb uploads which have progress callbacks, shard upload has no intermediate progress reporting. The user sees the upload "freeze" with no explanation.
  - Default log level may hide the info-level retry messages — if INFORMATION_LOG_LEVEL is DEBUG (which it is by default, line 28-29 of lib.rs), the retry messages are at debug level, invisible to most users.
  - The error message itself may be unclear — a reqwest timeout wraps into CasClientError::from(reqwest_middleware::Error) which may not say "shard upload timed out after 120s" in a user-friendly way.
```

XciD:
> > did you find why the error was silenced ? worth investigating if we are silenting some other errors. Could have been a retry policy failling over and over in my case, but worth checking
> 
> From the analysis it should have reported an ERROR message in the logs after all retries were exhausted. Each retry should be in the logs at INFO level. What is less clear is what makes its way back to Python - but for the testing you were doing that was in Rust, right?
> 
> Here is the writeup for the swallowed errors from Claude:
> 
> ```
>   Why Timeout Errors Appear Silent
> 
>   The error path is: reqwest timeout → RetryWrapper → upload_shard → shard_interface → file_upload_session::finalize → upload_commit → user.
> 
>   The error does propagate — it's not silently discarded in code. But here's why it appeared silent:
> 
>   Root Cause: Timeouts are classified as Retryable::Transient
> 
>   In on_request_failure (line 473):
>   if error.is_timeout() || is_connect {
>       Some(Retryable::Transient)
>   }
> 
>   This means:
> 
>   1. Each timeout is logged at info level, not error — process_error_response logs transient errors with log_as_info = true (line 115). If the operator or user doesn't have info-level logging enabled, they see nothing during the retry attempts.
>   2. The retry loop silently retries 3 times (default retry_max_attempts). With the old 120s read_timeout, that's up to 6 minutes of silent retries before failing. Each retry hits the same timeout.
>   3. The final "No more retries; aborting" IS logged at error level (line 320), but only after exhausting all retries.
>   4. From the user's perspective: the upload hangs for ~6 minutes (3 × 120s timeouts), then fails with a generic CasClientError that wraps the reqwest timeout. The error message may not clearly indicate "shard upload timed out" — it's wrapped in layers of error types.
> 
>   Contributing factors:
> 
>   - No progress indication during shard upload — unlike xorb uploads which have progress callbacks, shard upload has no intermediate progress reporting. The user sees the upload "freeze" with no explanation.
>   - Default log level may hide the info-level retry messages — if INFORMATION_LOG_LEVEL is DEBUG (which it is by default, line 28-29 of lib.rs), the retry messages are at debug level, invisible to most users.
>   - The error message itself may be unclear — a reqwest timeout wraps into CasClientError::from(reqwest_middleware::Error) which may not say "shard upload timed out after 120s" in a user-friendly way.
> ```

It was rust, but in a e2e test were I didnt had the logs. So I was maybe block on a retry loop

PR #686, opened 2026-03-08
feat: merge hf-mount-integration into main (download_stream_from_offset, create_remote_client)

XciD:
Merges `feat/hf-mount-integration` into `main`, bringing in the APIs needed by hf-mount:

- `create_remote_client`: public factory for creating a CAS client without going through the full XetSession setup
- `download_stream_from_offset`: stream file content from an arbitrary byte offset, enabling parallel range downloads (used by hf-mount's multi-slot prefetch)
- `SingleFileCleaner`: exposed from `data` crate for use in hf-mount's flush pipeline

Conflicts resolved by keeping `main`'s `start_clean` signature (with `size: u64` and `Ulid` tracking ID) and integrating the new download APIs from the integration branch.


XciD:
Redundant with #675 — feat/hf-mount-integration has been updated directly.

PR #688, opened 2026-03-09
feat: add download_to_buffer for direct memory writes

XciD:
## Summary

Add a zero-copy download path that writes CAS data directly into a caller-provided memory buffer, bypassing disk I/O entirely. This enables downloading model weights straight into CUDA pinned memory for DMA transfer to GPU.

### Changes

- **`file_reconstruction`**: add `reconstruct_to_buffer` that resolves all file terms in a single CAS roundtrip and writes xorb blocks in parallel to non-overlapping buffer regions via `JoinSet`
- **`data`**: add `FileDownloadSession::download_to_buffer` with optional byte range support
- **`hf_xet`**: expose `download_to_buffer(hash, file_size, buf_ptr, buf_len, endpoint, token_info, ...)` to Python
- Optimize decompression path: use `decompress_from_reader` to write directly into the target writer, eliminating one intermediate `Vec` per chunk

### Python usage

```python
import torch
import hf_xet

# Allocate CUDA pinned memory
buf = torch.empty(file_size, dtype=torch.uint8, pin_memory=True)

# Download directly into pinned buffer (no disk I/O)
hf_xet.download_to_buffer(
    hash=xet_hash,
    file_size=file_size,
    buf_ptr=buf.data_ptr(),
    buf_len=file_size,
    endpoint=cas_url,
    token_info=(cas_token, cas_exp),
    token_refresher=None,
)

# DMA transfer to GPU
tensor = torch.frombuffer(buf.numpy(), dtype=torch.bfloat16)
gpu_tensor = tensor.cuda(non_blocking=True)  # ~0.26s for 3.4 GB
```

## Benchmark (SmolLM2-1.7B, 3.4 GB, g5.xlarge A10G GPU)

| Path | Median TTFT | Speedup |
|---|---|---|
| Xet (CAS) -> disk -> load -> GPU | 12.00s | 1x |
| **Xet zero-copy -> pinned mem -> DMA GPU** | **5.75s** | **2.1x** |

TTFT = time to first token (download + load + single token generation).


XciD:
example of usage: https://github.com/huggingface/vllm-xet-loader

hoytak:
@XciD -- Does the unordered stream API in https://github.com/huggingface/xet-core/pull/729 fulfill this use case?

XciD:
Yes perfect thanks

PR #690, opened 2026-03-09
Rework the interface for session task to get result from registered upload

seanses:
This PR updates the interface for retrieving per-task results after UploadCommit::commit() or DownloadGroup::finish(). The problem with the previous interface is that commit() and finish() return a vector of FileMetadata or DownloadResult, making it difficult for users to associate each result with a specific task.

The new interface uses `task_id` as a strong binding bridge:

## Upload per-task result access patterns
After commit() completes, there are two equivalent ways to retrieve a per-task FileMetadata result:

1. Lookup in the global result map:
```
let commit = session.new_upload_commit()?;
let handle = commit.upload_from_path(src)?;
let results = commit.commit()?;
let result = results.get(&handle.task_id)
```

2. Direct access from the handle:
```
let commit = session.new_upload_commit()?;
let handle = commit.upload_from_path(src)?;
commit.commit()?;
// handle.result() is populated by commit() via the shared Arc.
let result = handle.result()
```

## Download per-task result access patterns
The pattern is similar to the above.

## Why not put results in a vector in the same order as tasks are registered to the commit instance?
After a commit instance is created, it can be cloned (since it is itself an Arc wrapping an internal struct) and sent to different threads. When multiple threads are registering tasks, there is no static registration order that a program can observe upfront.

Fix XET-890

### DIFF
--- a/xet_pkg/examples/example.rs
+++ b/xet_pkg/examples/example.rs
@@ -7,7 +7,9 @@ use std::time::Duration;
 
 use anyhow::Result;
 use clap::{Parser, Subcommand};
-use xet::xet_session::{FileMetadata, TaskStatus, XetFileInfo, XetSessionBuilder};
+use xet::xet_session::{
+    DownloadTaskHandle, FileMetadata, TaskStatus, UploadTaskHandle, XetFileInfo, XetSessionBuilder,
+};
 
 #[derive(Parser)]
 #[clap(name = "session-demo", about = "XetSession API demo")]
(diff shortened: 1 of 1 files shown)


xet_pkg/examples/example.rs
```diff
@@ -58,7 +58,7 @@ fn upload_files(files: Vec<PathBuf>, endpoint: Option<String>) -> Result<()> {
 
     // Enqueue all uploads; each starts immediately in the background.
     let n_files = files.len();
-    let handles: Vec<TaskHandle> = files
+    let handles: Vec<_> = files
```
assafvayner: for the sake of examples, can we avoid omitting types, this makes me have to go to the code to figure out all the return types I need to use, making the example less useful (still useful, but more useful with no explicit type omissions.

seanses: Yes for sure! Inserted types

PR #693, opened 2026-03-10
Code reorganization towards release of xet cargo package

hoytak:
This PR is a massive rearrangement of the code base into 5 packages intended for release on cargo. The directories and corresponding packages are:

1. xet_runtime/ — compiles into the xet-runtime package.  Contains the runtime, config, and logging management. 
2. xet_core_structures/ — compiles into the xet-core-structures package.  Contains core data structures for hashing, shards, and xorbs as well as internal data structures that depend on these. 
3. xet_client/ — compiles into the xet-client package, contains client code for remotely connecting to the Hugging Face servers.
4. xet_data/ — compiles into the xet-data package, contains the data processing pipeline: chunking/deduplication, file reconstruction, clean/smudge operations, and progress tracking. 
5. xet_pkg/ — compiles into the hf-xet package, provides the top-level session-based API for file upload and download with user-facing error categorization.  This is the primary package downstream dependencies would use.   This also contains a single summary error type, XetError, that translates cleanly into python error types. 

In addition, the other tools are: 

- git_xet/ — the git_xet CLI binary crate (location preserved). 
- hf_xet/ -- the hf_xet python package (location preserved).
- simulation/ — the simulation crate for upload scenario benchmarking.
- wasm/ -- the wasm objects. 

The full description — and information for an AI agent to use to update downstream dependencies — is at api_changes/update_260309_package_restructure.md.

Summary of moves:

- xet_runtime: became xet_runtime::core inside xet_runtime/.
- utils: became xet_runtime::utils inside xet_runtime/.
- xet_config: became xet_runtime::config inside xet_runtime/.
- xet_logging: became xet_runtime::logging inside xet_runtime/.
- error_printer: became xet_runtime::error_printer inside xet_runtime/.
- file_utils: became xet_runtime::file_utils inside xet_runtime/.
- merklehash: became xet_core_structures::merklehash inside xet_core_structures/.
- mdb_shard: became xet_core_structures::metadata_shard inside xet_core_structures/.
- xorb_object: became xet_core_structures::xorb_object inside xet_core_structures/.
- cas_client: became xet_client::cas_client inside xet_client/.
- hub_client: became xet_client::hub_client inside xet_client/.
- cas_types: became xet_client::cas_types inside xet_client/.
- chunk_cache: became xet_client::chunk_cache inside xet_client/.
- data: became xet_data::processing inside xet_data/.
- deduplication: became xet_data::deduplication inside xet_data/.
- file_reconstruction: became xet_data::file_reconstruction inside xet_data/.
- progress_tracking: became xet_data::progress_tracking inside xet_data/.
- xet_session: became xet::xet_session inside xet_pkg/.

- Wasm packages (hf_xet_wasm, hf_xet_thin_wasm): moved from top-level into wasm/; internal imports updated, public APIs unchanged.

### DIFF
--- a/git_xet/src/utils/ssh_connect.rs
+++ b/git_xet/src/utils/ssh_connect.rs
@@ -443,7 +443,7 @@ mod tests {
             let _env = EnvVarGuard::set("GIT_SSH_COMMAND", "ssh -i ~/.ssh/id_rsa");
             let (cmd, args) = get_sshcmd_and_args(&meta, &repo)?;
             assert_eq!(cmd, "sh");
-            assert_eq!(args, vec!["-c", "ssh -i ~/.ssh/id_rsa git@hf.co auth org/repo upload"]);
+            assert_eq!(args, vec!["-c", "ssh -i '~/.ssh/id_rsa' git@hf.co auth org/repo upload"]);
         }
 
         // Test with GIT_SSH


.github/workflows/release.yml
```diff
@@ -358,7 +358,7 @@ jobs:
           working-directory: hf_xet
       - name: Patch sdist
         run: |
-          pushd hf_xet/dist
+          pushd bindings/hf_xet/dist
```
rajatarya: The `working-directory` 3 lines above (line 358) still says `hf_xet` instead of `bindings/hf_xet`. Same issue on all 5 maturin actions in this file — lines 67, 163, 246, 311, and 358.

Since `hf_xet/` was moved to `bindings/hf_xet/`, maturin won't find `Cargo.toml`/`pyproject.toml` and every release build (linux, musllinux, windows, macos, sdist) will fail.

All other path references (`sed`, `cp`, `pushd`, artifact paths) in this file were correctly updated — only the `working-directory` parameter was missed.

Fix: change `working-directory: hf_xet` → `working-directory: bindings/hf_xet` at lines 67, 163, 246, 311, 358.

rajatarya:
Not sure any way to avoid this, but by moving `hf_xet` from root to `bindings/hf_xet` will break all the *nix distro build scripts (Arch/Fedora/etc) - they are all expecting to use $repo_root/hf_xet for Python package root to build from.

Maybe we can use git symlinks to keep them alive for one release and mark them deprecated? Not sure it is worth it since they will likely not notice _until_ it breaks.

hoytak:
> Not sure any way to avoid this, but by moving `hf_xet` from root to `bindings/hf_xet` will break all the *nix distro build scripts (Arch/Fedora/etc) - they are all expecting to use $repo_root/hf_xet for Python package root to build from.
> 
> Maybe we can use git symlinks to keep them alive for one release and mark them deprecated? Not sure it is worth it since they will likely not notice _until_ it breaks.

We can move them back for now...  That's less critical.

xet_client/src/cas_client/simulation/mod.rs
jgodlew, line 1: Is there a reason why we aren't moving the `cas_client/simulation` module into the `simulation` crate? It probably makes sense to move this code there.

hoytak: That is the part of the simulation code -- e.g. LocalClient -- that is heavily used by the test suites for the client code and all of the data processing code.  The simulation crate has all the tools for simulating connections, benchmarking, etc.   Moving cas_client/simulation/ to simulation/ would create a circular dependency, which isn't really resolvable.

seanses:
IIUC to publish `libxet` we still need to publish `runtime`, `core_structures`, `client` and `data`, just want to double check if they are published with name as the directory (in that case the crate may be already taken) or under any other name.

seanses:
> Not sure any way to avoid this, but by moving `hf_xet` from root to `bindings/hf_xet` will break all the *nix distro build scripts (Arch/Fedora/etc) - they are all expecting to use $repo_root/hf_xet for Python package root to build from.
> 
> Maybe we can use git symlinks to keep them alive for one release and mark them deprecated? Not sure it is worth it since they will likely not notice _until_ it breaks.

If possible let's also keep git_xet under the repo root for now. There are a lot of hub docs linking to here and a moon-landing proxy rule leading to this location.

rajatarya (approved):
Don't block on me for this - but the `git-xet` comment from @seanses is worth doing ahead of time.

hoytak:
> IIUC to publish `libxet` we still need to publish `runtime`, `core_structures`, `client` and `data`, just want to double check if they are published with name as the directory (in that case the crate may be already taken) or under any other name.

The packages published are hf-xet, xet-runtime, xet-core-structures, xet-client, and xet-data.

PR #694, opened 2026-03-10
Support XetSession in async context

seanses:
`XetSession` always created its own tokio runtime via `XetRuntime::new_with_config`, and calling `external_run_async_task` panics when already inside a tokio context. This blocked embedding the session in async Rust frameworks.

Core strategy:

 - `RuntimeMode` enum — 
 `Owned` (session created its own thread pool via `XetSessionBuilder::build` or `XetSessionBuilder::build_async` when outside tokio context. Both `_blocking` and async methods are supported. Async methods use an internal `bridge_to_owned` bridge that routes futures onto the owned thread pool, so they work from any executor (tokio, smol, async-std)) 
vs
 `External` (session wraps a caller-supplied tokio handle via `XetSessionBuilder::with_tokio_handle` or `XetSessionBuilder::build_async` when inside qualified tokio context. Only async methods may be called; `_blocking` methods return `SessionError::WrongRuntimeMode`. No second thread pool is created). 
- `XetRuntime::bridge_to_owned` — a new bridge that routes a future onto the owned tokio thread pool from any executor (smol, async-std, futures::executor, non-qualified tokio runtime) by delivering the result via a `tokio::sync::oneshot` channel that can be polled by any async executor.
 - Async public API — `UploadCommit` and `DownloadGroup` methods (`upload_from_path`, `upload_bytes`, `upload_file`, `commit`, `finish`) are now async fn. Factory methods `XetSession::new_upload_commit` and `new_download_group` are async.
Example:
```
let session = XetSessionBuilder::new().build_async().await?;
// Upload
 let commit = session.new_upload_commit().await?;
 let handle = commit.upload_from_path("file.bin".into()).await?;
 let results = commit.commit().await?;

 // Download
 let group = session.new_download_group().await?;
 let info = XetFileInfo {
     hash: ...,
     file_size: ...,
 };
 let dl_handle = group.download_file_to_path(info, "out/file.bin".into())?;
 let finish_results = group.finish().await?;
```

 -  Sync wrappers — New `UploadCommitSync` / `DownloadGroupSync` in `xet_session/sync/` expose a fully blocking API for sync Rust and Python (PyO3) callers. Returned by `new_upload_commit_blocking()` and `new_download_group_blocking()`.
Example:
```
let session = XetSessionBuilder::new().build()?;
// Upload
let commit = session.new_upload_commit_blocking()?;
 let handle = commit.upload_from_path("file.bin".into())?;
 let results = commit.commit()?;
 let m = results.values().next().unwrap().as_ref().as_ref().unwrap();

// Download
 let group = session.new_download_group_blocking()?;
 let info = XetFileInfo {
     hash: ...,
     file_size: ...,
 };
 let dl_handle = group.download_file_to_path(info, "out/file.bin".into())?;
 let finish_results = group.finish()?;
```

Additional fixes: `download_file_to_path` and `upload_from_path` now canonicalize paths with `std::path::absolute` before enqueuing; task status is only overwritten when still `Running`, preventing a race with concurrent abort().

Fix XET-891

### DIFF
--- a/xet_pkg/src/xet_session/session.rs
+++ b/xet_pkg/src/xet_session/session.rs
@@ -7,6 +7,7 @@ use std::sync::{Arc, Mutex};
 use std::task::{Context, Waker};
 
 use http::HeaderMap;
+use tracing::info;
 use ulid::Ulid;
 use xet_client::cas_client::auth::TokenRefresher;
 use xet_runtime::RuntimeError;
@@ -228,8 +229,12 @@ impl XetSessionBuilder {
     /// Use [`build_async`](Self::build_async) as a convenient alternative when building from
     /// within a tokio async context.
     pub fn with_tokio_handle(self, handle: tokio::runtime::Handle) -> Self {
+        let accept = handle_meets_session_requirements(&handle);
+        if !accept {
+            info!("supplied tokio handle rejected (missing drivers or wrong flavor); falling back to Owned mode");
+        }
         Self {
-            tokio_handle: handle_meets_session_requirements(&handle).then_some(handle),
+            tokio_handle: accept.then_some(handle),
             ..self
         }
     }
--- a/xet_pkg/src/xet_session/sync/upload_commit_sync.rs
+++ b/xet_pkg/src/xet_session/sync/upload_commit_sync.rs
@@ -44,10 +44,12 @@ impl UploadCommitSync {
 
     /// Queue a file for upload. See [`UploadCommit::upload_from_path`] for full documentation.
     pub fn upload_from_path(&self, file_path: PathBuf) -> Result<UploadTaskHandle, SessionError> {
-        let commit = self.inner.clone();
+        self.inner.session.check_alive()?;
+
+        let commit_inner = self.inner.inner.clone();
         self.inner
             .runtime()
-            .external_run_async_task(async move { commit.upload_from_path(file_path).await })?
+            .external_run_async_task(async move { commit_inner.start_upload_file_from_path(file_path).await })?
     }
 
     /// Queue raw bytes for upload. See [`UploadCommit::upload_bytes`] for full documentation.
@@ -56,10 +58,12 @@ impl UploadCommitSync {
         bytes: Vec<u8>,
         tracking_name: Option<String>,
     ) -> Result<UploadTaskHandle, SessionError> {
-        let commit = self.inner.clone();
+        self.inner.session.check_alive()?;
+
+        let commit_inner = self.inner.inner.clone();
         self.inner
             .runtime()
-            .external_run_async_task(async move { commit.upload_bytes(bytes, tracking_name).await })?
+            .external_run_async_task(async move { commit_inner.start_upload_bytes(bytes, tracking_name).await })?
     }
 
     /// Begin an incremental file upload, returning a [`SingleFileCleaner`] to stream bytes.
@@ -79,10 +83,12 @@ impl UploadCommitSync {
         file_name: Option<String>,
         file_size: u64,
     ) -> Result<(TaskHandle, SingleFileCleaner), SessionError> {
-        let commit = self.inner.clone();
+        self.inner.session.check_alive()?;
+
+        let commit_inner = self.inner.clone();
         self.inner
             .runtime()
-            .external_run_async_task(async move { commit.upload_file(file_name, file_size).await })?
+            .external_run_async_task(async move { commit_inner.start_upload_file(file_name, file_size).await })?
     }
 
     /// Return a snapshot of progress for every queued upload.
--- a/xet_pkg/src/xet_session/upload_commit.rs
+++ b/xet_pkg/src/xet_session/upload_commit.rs
@@ -41,7 +41,7 @@ use super::session::XetSession;
 /// [`UploadCommitSync`]: crate::xet_session::sync::UploadCommitSync
 #[derive(Clone)]
 pub struct UploadCommit {
-    inner: Arc<UploadCommitInner>,
+    pub(super) inner: Arc<UploadCommitInner>,
 }
 
 impl std::ops::Deref for UploadCommit {
@@ -218,7 +218,7 @@ struct InnerUploadTaskHandle {
 #[doc(hidden)]
 pub struct UploadCommitInner {
     commit_id: Ulid,
-    session: XetSession,
+    pub(super) session: XetSession,
 
     // Active upload tasks for this commit
     active_tasks: RwLock<HashMap<Ulid, InnerUploadTaskHandle>>,
@@ -317,7 +317,10 @@ impl UploadCommitInner {
         })
     }
 
-    async fn start_upload_file_from_path(&self, file_path: PathBuf) -> Result<UploadTaskHandle, SessionError> {
+    pub(super) async fn start_upload_file_from_path(
+        &self,
+        file_path: PathBuf,
+    ) -> Result<UploadTaskHandle, SessionError> {
         // Hold the state lock for the duration of this function so commit() will not run
         // when an upload task is registering.
         let state = self.state.lock().await;
@@ -360,7 +363,7 @@ impl UploadCommitInner {
     /// The state lock is held across `start_clean(...).await` so that a concurrent
     /// `handle_commit` cannot finalise the upload session between the state check and the
     /// creation of the cleaner.
-    async fn start_upload_file(
+    pub(super) async fn start_upload_file(
         &self,
         tracking_name: Option<String>,
         file_size: u64,
@@ -389,7 +392,7 @@ impl UploadCommitInner {
     }
 
     /// Enqueue a bytes upload task, spawning it on the runtime immediately.
-    async fn start_upload_bytes(
+    pub(super) async fn start_upload_bytes(
         &self,
         bytes: Vec<u8>,
         tracking_name: Option<String>,
--- a/xet_runtime/src/core/runtime.rs
+++ b/xet_runtime/src/core/runtime.rs
@@ -11,7 +11,7 @@ use reqwest::Client;
 use tokio::runtime::{Builder as TokioRuntimeBuilder, Handle as TokioRuntimeHandle, Runtime as TokioRuntime};
 use tokio::sync::oneshot;
 use tokio::task::JoinHandle;
-use tracing::{debug, info};
+use tracing::{debug, info, warn};
 
 use super::XetCommon;
 use crate::config::XetConfig;
@@ -324,6 +324,8 @@ impl XetRuntime {
         });
         if let Ok(mut reg) = EXTERNAL_RUNTIME_REGISTRY.write() {
             reg.insert(id, Arc::downgrade(&rt));
+        } else {
+            warn!("EXTERNAL_RUNTIME_REGISTRY poisoned; skipping registration");
         }
         rt
     }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1885,6 +1885,7 @@ dependencies = [
  "tempfile",
  "thiserror 2.0.18",
  "tokio",
+ "tracing",
  "tracing-subscriber",
  "ulid",
  "xet-client",
--- a/xet_pkg/Cargo.toml
+++ b/xet_pkg/Cargo.toml
@@ -23,6 +23,7 @@ http = { workspace = true }
 tokio = { workspace = true, features = ["net"] }
 ulid = { workspace = true }
 thiserror = { workspace = true }
+tracing = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 
 [dev-dependencies]


xet_runtime/src/core/runtime.rs
```diff
@@ -422,6 +489,28 @@ impl XetRuntime {
...
+    ///
+    /// Unlike [`external_run_async_task`](Self::external_run_async_task) which **blocks**
+    /// the calling thread, this method returns a future that any executor can poll.
+    /// The result is delivered via a `tokio::sync::oneshot` channel whose receiver only
+    /// requires a `std::task::Waker`, making it compatible with every standard executor.
+    ///
+    /// Returns `Err(RuntimeError::TaskCanceled)` if the runtime shuts down before
+    /// the task result can be delivered.
+    pub async fn bridge_to_owned<T, F>(&self, task_name: &'static str, fut: F) -> Result<T, RuntimeError>
+    where
```
hoytak, line 508: `bridge_to_owned` spawns the future via `self.spawn()` but discards the `JoinHandle`. If the bridged future panics, the panic is silently swallowed and the caller gets a generic `TaskCanceled`. Consider awaiting the `JoinHandle` to distinguish cancellation from panic and surface a more diagnostic error.

xet_runtime/src/core/runtime.rs
```diff
@@ -148,6 +155,15 @@ thread_local! {
     static THREAD_RUNTIME_REF: RefCell<Option<(u32, Arc<XetRuntime>)>> = const { RefCell::new(None) };
 }
 
+// Registry for External-mode runtimes created via from_external_with_config.
+// Keyed by tokio runtime ID so current_if_exists() can find the right XetRuntime
+// (with the correct XetConfig and XetCommon) when called from the caller's tokio threads,
+// where THREAD_RUNTIME_REF is never set.
+lazy_static! {
+    static ref EXTERNAL_RUNTIME_REGISTRY: std::sync::RwLock<HashMap<tokio::runtime::Id, Weak<XetRuntime>>> =
+        std::sync::RwLock::new(HashMap::new());
```
hoytak: `EXTERNAL_RUNTIME_REGISTRY` uses `lazy_static!` with `std::sync::RwLock`. `std::sync::LazyLock` is stable since Rust 1.80 and would remove the `lazy_static` dependency. At minimum, add a comment clarifying why this uses `std::sync` rather than `tokio::sync` (must be accessible from non-async contexts like `Drop`).

xet_pkg/src/xet_session/upload_commit.rs
```diff
@@ -475,9 +512,25 @@ impl UploadCommitInner {
...
+    /// the runtime shutdown cancels the in-flight task regardless.
+    ///
+    /// Clearing `upload_session` prevents future `start_upload_file` calls from
+    /// obtaining a session and prevents `handle_commit` from calling `finalize`.
+    /// It does not invalidate any `SingleFileCleaner` already in the caller's hands,
+    /// since the cleaner holds its own `Arc` to the session.
     fn abort(&self) -> Result<(), SessionError> {
-        *self.state.lock()? = GroupState::Aborted;
+        if let Ok(mut guard) = self.state.try_lock() {
+            *guard = GroupState::Aborted;
```
hoytak, line 524: `abort()` uses `try_lock()` on the tokio `Mutex`. If a registration method currently holds the lock, the state is never set to `Aborted`, leaving the commit in `Alive` state after `abort()` returns — a real race if aborting for application-level reasons (not process shutdown). Consider `lock().await` in an async abort variant, or at minimum document that the state flag update is best-effort.

xet_pkg/src/xet_session/upload_commit.rs
```diff
@@ -46,41 +52,41 @@ impl std::ops::Deref for UploadCommit {
...
 
         let inner = Arc::new(UploadCommitInner {
             commit_id,
             session,
             active_tasks: RwLock::new(HashMap::new()),
             progress,
             upload_session: Mutex::new(Some(upload_session)),
-            state: Mutex::new(GroupState::Alive),
+            state: tokio::sync::Mutex::new(GroupState::Alive),
         });
```
hoytak, line 71: `UploadCommitInner` uses `tokio::sync::Mutex` while `DownloadGroupInner` uses `std::sync::Mutex`. The rationale (holding the lock across an `.await`) is sound but undocumented. Add a comment explaining the asymmetry so readers don't mistake it for an oversight.

xet_pkg/src/xet_session/upload_commit.rs
```diff
@@ -92,13 +98,21 @@ impl UploadCommit {
...
-        self.inner.start_upload_file_from_path(absolute_path)
+
+        if matches!(self.session.runtime_mode, RuntimeMode::External) {
+            return self.inner.start_upload_file_from_path(absolute_path).await;
+        }
+        let inner = self.inner.clone();
+        self.session
+            .runtime
+            .bridge_to_owned("upload_from_path", async move { inner.start_upload_file_from_path(absolute_path).await })
+            .await?
```
hoytak, line 111: The `RuntimeMode` dispatch pattern (branch on External vs bridge through `bridge_to_owned`) is repeated verbatim across `upload_from_path`, `upload_bytes`, `upload_file`, `commit`, `finish`, and factory methods. Consider extracting a `run_on_runtime<T>` helper to reduce boilerplate and centralize the mode logic.

xet_pkg/src/xet_session/session.rs
```diff
@@ -19,49 +22,124 @@ enum SessionState {
...
+/// 3. **IO driver** — required for all network I/O via `reqwest`/`hyper`.
+///
+/// Driver availability is probed by entering the handle's context and polling a
+/// driver-dependent future once inside `catch_unwind`.  Tokio panics synchronously
+/// on the first poll when a driver is absent, so the result is immediate — no
+/// spawning or blocking required.
+///
+/// Returns `true` if all requirements are met, `false` otherwise.
+fn handle_meets_session_requirements(handle: &tokio::runtime::Handle) -> bool {
+    // Non-WASM: require a multi-threaded runtime.
```
hoytak, line 93: `handle_meets_session_requirements` probes driver availability by catching panics from tokio futures inside `catch_unwind`. This relies on tokio's undocumented internal behavior of panicking synchronously on first poll when a driver is absent. A future tokio version could return an error instead, silently accepting an unsuitable runtime. Add a comment noting this assumption and the tokio version it was validated against.

xet_pkg/src/xet_session/session.rs
```diff
@@ -125,15 +205,63 @@ impl XetSessionBuilder {
...
+    /// Build and automatically attach to the current runtime.
+    ///
+    /// - **Tokio context** with a suitable runtime (multi-thread, time + IO drivers): wraps the caller's handle via
+    ///   [`with_tokio_handle`](Self::with_tokio_handle) — External mode.
+    /// - **Tokio context** with an unsuitable runtime (e.g. `current_thread`): handle is discarded by
+    ///   `with_tokio_handle`; falls back to an owned thread pool — Owned mode.
+    /// - **Non-tokio context** (smol, async-std, etc.): creates an owned thread pool — Owned mode; async methods use an
+    ///   internal bridge compatible with any executor.
+    pub async fn build_async(self) -> Result<XetSession, SessionError> {
+        match tokio::runtime::Handle::try_current() {
```
hoytak, line 256: `build_async` is declared `async fn` but contains no `.await` points — it's synchronous under the hood. This is fine ergonomically, but could confuse readers expecting actual async work. Worth a short comment noting the future resolves immediately.

xet_pkg/src/xet_session/upload_commit.rs
seanses: Yeah the main obstacle was that I wanted to make the higher level `XetSession::abort()` sync so can be called in a Ctrl-C handler. This function is private and only called from the the `XetSession::abort()` after performing the runtime sigint shutdown, so the whether the state variable is left as "Aborted" or not is not strictly a race.

xet_pkg/src/xet_session/session.rs
seanses: The unit tests below should catch if a new tokio version changes this behavior.

xet_pkg/src/xet_session/sync/upload_commit_sync.rs
```diff
@@ -0,0 +1,309 @@
...
+        let commit = session.runtime.external_run_async_task(UploadCommit::new(session.clone()))??;
+        Ok(Self { inner: commit })
+    }
+
+    /// Queue a file for upload. See [`UploadCommit::upload_from_path`] for full documentation.
+    pub fn upload_from_path(&self, file_path: PathBuf) -> Result<UploadTaskHandle, SessionError> {
+        let commit = self.inner.clone();
+        self.inner
+            .runtime()
+            .external_run_async_task(async move { commit.upload_from_path(file_path).await })?
```
rajatarya: In Owned mode this becomes: `external_run_async_task` (blocks calling thread) → `upload_from_path` → `dispatch()` → `bridge_to_owned` (spawns task + oneshot channel). The `bridge_to_owned` hop is redundant here since `external_run_async_task` already runs the future on the owned runtime's thread pool.

Not a correctness issue, but for upload-heavy workloads the extra spawn + channel per operation could add up. One option: have the sync wrappers call `UploadCommitInner` methods directly (bypassing `dispatch`), since they already know they're in Owned mode. That would also avoid the `inner.clone()` on every call.

Happy to leave this as-is if you'd rather keep the sync wrappers thin — just wanted to flag the overhead.

xet_runtime/src/core/runtime.rs
```diff
@@ -263,6 +291,46 @@ impl XetRuntime {
...
+                        config.system_monitor.sample_interval,
+                        config.system_monitor.log_path.clone(),
+                    )
+                    .ok()
+                })
+                .flatten(),
+            config: Arc::new(config),
+            external_handle_id: Some(id),
+        });
+        if let Ok(mut reg) = EXTERNAL_RUNTIME_REGISTRY.write() {
```
rajatarya, line 325: nit: if the `write()` fails (poisoned lock), the runtime is silently unregistered from the global lookup. `current_if_exists()` will then return a default-config wrapper instead of this instance's config. Consider adding a `tracing::warn` here and in the `Drop` impl so this doesn't become a silent config mismatch in production:

```rust
if let Ok(mut reg) = EXTERNAL_RUNTIME_REGISTRY.write() {
    reg.insert(id, Arc::downgrade(&rt));
} else {
    tracing::warn!("EXTERNAL_RUNTIME_REGISTRY poisoned; skipping registration");
}
```

xet_pkg/src/xet_session/session.rs
```diff
@@ -125,15 +214,68 @@ impl XetSessionBuilder {
...
+    ///
+    /// If the handle does **not** meet requirements (e.g. `current_thread` flavor or missing
+    /// drivers), it is silently ignored and [`build`](Self::build) will fall back to creating
+    /// an owned thread pool (Owned mode) instead.
+    ///
+    /// Use [`build_async`](Self::build_async) as a convenient alternative when building from
+    /// within a tokio async context.
+    pub fn with_tokio_handle(self, handle: tokio::runtime::Handle) -> Self {
+        Self {
+            tokio_handle: handle_meets_session_requirements(&handle).then_some(handle),
```
rajatarya: Nice that an unsuitable handle is silently discarded here rather than erroring. One question though — if a caller explicitly passes a handle via `with_tokio_handle()` and it gets silently dropped (e.g. missing IO driver), they might be surprised to end up in Owned mode with a second thread pool. Would it be worth logging a `tracing::info` or `debug` when the handle is rejected, so callers can diagnose why they ended up in Owned mode?

Something like:
```rust
if !handle_meets_session_requirements(&handle) {
    tracing::debug!("supplied tokio handle rejected (missing drivers or wrong flavor); falling back to Owned mode");
}
```

xet_pkg/src/xet_session/upload_commit.rs
```diff
@@ -46,41 +52,41 @@ impl std::ops::Deref for UploadCommit {
...
-            let progress_updater = progress_clone as Arc<dyn xet_data::progress_tracking::TrackingProgressUpdater>;
-            FileUploadSession::new(Arc::new(config), Some(progress_updater)).await
-        })??;
+        let progress_updater = progress.clone() as Arc<dyn xet_data::progress_tracking::TrackingProgressUpdater>;
+        let upload_session = FileUploadSession::new(Arc::new(config), Some(progress_updater)).await?;
 
         let inner = Arc::new(UploadCommitInner {
             commit_id,
             session,
             active_tasks: RwLock::new(HashMap::new()),
```
rajatarya, line 67: Good call switching to `tokio::sync::Mutex` here and documenting why — the asymmetry with `DownloadGroupInner` is now well-explained. The comment is helpful.

xet_pkg/src/xet_session/upload_commit.rs
```diff
@@ -258,7 +272,11 @@ impl UploadCommitInner {
             } else {
```
rajatarya, line 272: Good defensive change — preventing abort()'s `Cancelled` from being overwritten by a concurrent task completion. Applies correctly in all three registration paths.

PR #702, opened 2026-03-12
Update config groups to handle more of the data management values. 

hoytak:
This PR moves some config values that were part of the data configuration into XetConfig, specifically the compression_policy, staging_subdir, session_dir_name, and global_dedup_query_enabled.   This also consolidates the remaining values into a single struct with endpoint and authentication information.

### DIFF
--- a/xet_runtime/tests/config_env_test.rs
+++ b/xet_runtime/tests/config_env_test.rs
@@ -63,6 +63,16 @@ fn test_environment_variable_aliases() {
         assert_eq!(XetConfig::new().data.session_xorb_metadata_flush_max_count, 128);
     }
 
+    // Xorb aliases (old HF_XET_DATA_XORB_* names)
+    {
+        let _guard = EnvVarGuard::set("HF_XET_DATA_XORB_COMPRESSION_SCHEME_RETEST_INTERVAL", "64");
+        assert_eq!(XetConfig::new().xorb.compression_scheme_retest_interval, 64);
+    }
+    {
+        let _guard = EnvVarGuard::set("HF_XET_DATA_XORB_COMPRESSION_POLICY", "lz4");
+        assert_eq!(XetConfig::new().xorb.compression_policy.as_str(), "lz4");
+    }
+
     // Shard aliases
     {
         let _guard = EnvVarGuard::set("HF_XET_MDB_SHARD_CACHE_SIZE_LIMIT", "24gb");
--- a/xet_core_structures/src/xorb_object/compression_scheme.rs
+++ b/xet_core_structures/src/xorb_object/compression_scheme.rs
@@ -26,7 +26,6 @@ pub enum CompressionScheme {
     #[default]
     Auto = 99,
 }
-pub const NUM_COMPRESSION_SCHEMES: usize = 4;
 
 impl Display for CompressionScheme {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
--- a/xet_core_structures/src/xorb_object/xorb_object_format.rs
+++ b/xet_core_structures/src/xorb_object/xorb_object_format.rs
@@ -2542,4 +2542,12 @@ mod tests {
         let deserialized = XorbObject::deserialize(&mut reader).unwrap();
         assert_eq!(deserialized.info.num_chunks, 1);
     }
+
+    #[test]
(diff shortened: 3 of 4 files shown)


xet_runtime/src/config/groups/deduplication.rs
```diff
@@ -23,4 +23,13 @@ crate::config_group!({
...
     ref min_n_chunks_per_range: f32 = 8.0;
+
+    /// Whether to enable global deduplication queries to the server.
+    /// When enabled, the system will query the server for deduplication shards
+    /// based on chunk hashes to enable cross-repository deduplication.
+    ///
+    /// The default value is true.
+    ///
+    /// Use the environment variable `HF_XET_DEDUPLICATION_GLOBAL_DEDUP_QUERY_ENABLED` to set this value.
+    ref global_dedup_query_enabled: bool = true;
```
seanses, line 34: Not used? Maybe not add now

xet_runtime/src/config/groups/session.rs
```diff
@@ -0,0 +1,9 @@
+crate::config_group!({
+
+    /// Subdirectory name for shard session data within the staging directory.
+    ///
+    /// The default value is "shard-session".
+    ///
+    /// Use the environment variable `HF_XET_SESSION_SESSION_DIR_NAME` to set this value.
```
seanses: `HF_XET_SESSION_SESSION_DIR_NAME` is not an elegant name: repetitive `SESSION`

xet_data/src/processing/bin/xtool.rs
```diff
@@ -238,6 +223,19 @@ async fn query_reconstruction(
...
     let cli = XCommand::parse();
+
+    if let Command::Dedup(ref arg) = cli.command
+        && let Some(c) = arg.compression
+    {
+        let scheme = CompressionScheme::try_from(c).map_err(|_| {
+            anyhow::anyhow!("Invalid compression value {c}; expected one of: 0 (none), 1 (lz4), 2 (bg4-lz4)")
+        })?;
+        let scheme_str: &str = scheme.into();
+        unsafe {
```
seanses: It feels a bit awkward that this have to set an env var to configure itself.. Can we explicitly create a XetConfig and pass it into the XetRuntime construction below?

xet_runtime/src/config/groups/xorb.rs
```diff
@@ -9,4 +9,13 @@ crate::config_group!({
...
     ref compression_scheme_retest_interval : usize = 32;
+
+    /// Compression policy for xorb data.
+    /// Valid values: "" or "auto" for automatic detection, "none", "lz4", "bg4-lz4".
+    /// When set to "" or "auto", the best compression scheme is chosen based on data analysis.
+    ///
+    /// The default value is "auto" (auto-detect).
+    ///
+    /// Use the environment variable `HF_XET_XORB_COMPRESSION_POLICY` to set this value.
+    ref compression_policy: String = "auto".to_string();
```
seanses: Define this type as `CompressionScheme` so to evaluate and catch error early.

xet_runtime/tests/config_env_test.rs
```diff
@@ -74,6 +78,31 @@ fn test_environment_variable_aliases() {
     }
 }
 
+#[test]
+#[serial(config_env)]
+fn test_new_config_env_overrides() {
```
seanses, line 93: This file is called "alias_test" but this doesn't look like aliases. Where's the place that test the"old" config env overrides (or maybe they were assumed correct because built on top of the `config_group!` tool)?

xet_runtime/src/config/groups/xorb.rs
hoytak: I considered doing that, but then that puts a circular dependency here -- we have to define the setting locally here instead of in the xet_data area.  Instead, I added a new type, ConfigEnum, that forces a string value to be one of a few specific values, so we should have the same functionality here...

xet_runtime/tests/config_env_test.rs
hoytak: These are all the aliases.  Note that most of them simply cut the group name off...

xet_runtime/src/config/groups/deduplication.rs
hoytak: It is actually used....

PR #703, opened 2026-03-12
V2 reconstruction with client-side optional single range splitting

hoytak:
This PR introduces V2 multirange URL fetching for xorbs, but optionally splits the multirange requests into multiple single-range requests that can be executed in parallel.  This allows the reconstruction process to generate full multirange presigned URLs, but the client effectively perform the retrieval stage as a sequence of parallel single-range queries.   

The config variable `client.prefer_multirange_fetching` controls this behavior; by default it is set to false due to the current observed slowness of fetching multiranged URLs.

### DIFF
--- a/xet_data/src/file_reconstruction/reconstruction_terms/file_term.rs
+++ b/xet_data/src/file_reconstruction/reconstruction_terms/file_term.rs
@@ -265,7 +265,11 @@ pub async fn retrieve_file_term_block(
                 }
                 idx += (range.end - range.start) as usize;
             }
-            debug_assert!(found, "chunk_start {chunk_start} not found in chunk_ranges {chunk_ranges:?}");
+            if !found {
+                return Err(FileReconstructionError::CorruptedReconstruction(format!(
+                    "chunk_start {chunk_start} not found in chunk_ranges {chunk_ranges:?} for file term {local_term_index}"
+                )));
+            }
             idx
         };
 


xet_client/src/cas_client/remote_client.rs
```diff
@@ -168,58 +171,125 @@ impl RemoteClient {
...
+        api_version: &str,
+    ) -> Result<Option<T>>
+    where
+        T: serde::de::DeserializeOwned + 'static,
+    {
         let call_id = FN_CALL_ID.fetch_add(1, Ordering::Relaxed);
-        let url = Url::parse(&format!("{}/v1/reconstructions/{}", self.endpoint, file_id.hex()))?;
+        let url = Url::parse(&format!("{}/{api_version}/reconstructions/{}", self.endpoint, file_id.hex()))?;
+        let api_tag_owned = format!("cas::get_reconstruction_{api_version}");
+        let api_tag: &'static str = api_tag_owned.leak();
```
XciD: this leak on each call no ? might be harmless

xet_data/src/file_reconstruction/reconstruction_terms/xorb_block.rs
```diff
@@ -51,87 +74,134 @@ impl Eq for XorbBlock {}
...
-                    uncompressed_size,
-                    data,
-                }))
-            })
-            .await
-            .cloned()
+        // Fast path: another task may have already downloaded this block.
+        if let Some(ref xorb_block_data) = *self.data.read().await {
+            return Ok(xorb_block_data.clone());
+        }
```
XciD: should we keep the OneCell here ?

xet_data/src/file_reconstruction/reconstruction_terms/file_term.rs
```diff
@@ -77,170 +108,229 @@ pub async fn retrieve_file_term_block(
...
+            let mut found = false;
+            for range in chunk_ranges {
+                if chunk_start >= range.start && chunk_start < range.end {
+                    idx += (chunk_start - range.start) as usize;
+                    found = true;
+                    break;
+                }
+                idx += (range.end - range.start) as usize;
+            }
+            debug_assert!(found, "chunk_start {chunk_start} not found in chunk_ranges {chunk_ranges:?}");
```
XciD: Should we raise in release ?

hoytak: This should be a data corruption issues; let me see.

PR #706, opened 2026-03-12
Optimize config struct for direct access in python

hoytak:
This PR adds in a feature flag, "python" to the xet_runtime package such that when compiled, the XetConfig struct is built to have python getters and setters.  This integrates the handling of the config struct directly into the XetConfig struct and the macros used to register the config values, making the handling of values in the python bindings seamless.

### DIFF
--- a/wasm/hf_xet_thin_wasm/Cargo.lock
+++ b/wasm/hf_xet_thin_wasm/Cargo.lock
@@ -4030,6 +4030,7 @@ dependencies = [
  "tokio-util",
  "tracing",
  "ulid",
+ "url",
  "walkdir",
  "xet-client",
  "xet-core-structures",
--- a/wasm/hf_xet_wasm/Cargo.lock
+++ b/wasm/hf_xet_wasm/Cargo.lock
@@ -4202,6 +4202,7 @@ dependencies = [
  "tokio-util",
  "tracing",
  "ulid",
+ "url",
  "walkdir",
  "xet-client",
  "xet-core-structures",


xet_runtime/src/utils/configuration_utils.rs
```diff
@@ -87,14 +92,28 @@ impl ParsableConfigValue for bool {
...
+    fn to_config_string(&self) -> String {
+        if *self { "true" } else { "false" }.to_owned()
+    }
 }
 
 /// Enable Option<T> to allow the default value to be None if nothing is set and appear as
 /// Some(Value) if the user specifies the value.
 impl<T: ParsableConfigValue> ParsableConfigValue for Option<T> {
     fn parse_user_value(value: &str) -> Option<Self> {
+        if value.trim().is_empty() {
```
rajatarya, line 105: This is a behavior change that affects non-python builds too. Previously, `bool::parse_user_value("")` returned `Some(false)` (via the `"" => false` arm in the old code), so `HF_XET_SOME_BOOL=""` would set the field to `false`.

Now for `Option<T>`, empty/whitespace strings short-circuit to `Some(None)` here. And for plain `bool`, the empty string no longer matches any arm and returns `None`, which triggers the default.

I think this is arguably more correct (empty env var = unset), but it's a subtle change for anyone who relied on `export HF_XET_FOO=` to mean `false`. Worth calling out in the PR description?

Cargo.toml
```diff
@@ -75,6 +75,7 @@ once_cell = "1.21"
 oneshot = "0.1"
 paste = "1.0"
 pin-project = "1"
+pyo3 = { version = "0.26", features = ["abi3-py37", "multiple-pymethods"] }
```
rajatarya, line 79: `abi3-py37` targets Python 3.7 which has been EOL since June 2023. Do we need this for compatibility with something specific? If not, we should consider bumping to `abi3-py38` or even `abi3-py39` — newer stable ABI versions enable PyO3 to use more efficient C API paths for datetime conversions (relevant since we expose `Duration` as `timedelta`).

xet_runtime/src/config/macros.rs
```diff
@@ -101,9 +163,125 @@ macro_rules! config_group {
...
+            pub fn inner(&self) -> &ConfigValueGroup {
+                &self.inner
+            }
+        }
+
+        #[pyo3::pymethods]
+        impl PyConfigValueGroup {
+            /// Get a field's string representation by name.
+            #[pyo3(name = "get")]
+            fn py_get(&self, name: &str) -> pyo3::PyResult<String> {
```
rajatarya: nit: `py_get` returns a `String` for all field types, while the property getters return native Python types. Consider adding a brief docstring to `get()` noting that it's the string representation — users wanting typed access should use the property getters instead. Something like:

```python
def get(name: str) -> str:
    """Return the string representation of a config field. Use property access for typed values."""
```

xet_runtime/src/config/xet_config.rs
```diff
@@ -78,4 +121,275 @@ impl XetConfig {
 
         self
     }
+
+    /// Return a new config with the value at the given dotted path updated
+    /// (e.g. "data.max_concurrent_file_ingestion").
+    pub fn with_config(mut self, path: &str, value: impl ToString) -> Result<Self, ConfigError> {
```
rajatarya, line 187: Nice API. The immutable builder pattern (`with_config` returns a new config) is clean and Pythonic. The dict overload on the Python side is a nice touch too.

xet_runtime/src/config/xet_config.rs
```diff
@@ -78,4 +121,275 @@ impl XetConfig {
...
+            .unwrap()
+            .with_config("log.dest", "")
+            .unwrap();
+        assert_eq!(config.log.dest, None);
+    }
+}
+
+/// Generates `#[pymethods]` getters on `PyXetConfig` for each config group.
+/// Defined here (before the `py_xet_config` child module) so it is visible inside it.
+#[allow(unused_macros)]
```
rajatarya: nit: the macro definition itself isn't gated behind `#[cfg(feature = "python")]`, only the invocation at line 394. So this macro is compiled but unused in non-python builds (hence the `allow(unused_macros)`). Could gate the definition too for slightly cleaner separation, but not a big deal.

xet_runtime/src/utils/configuration_utils.rs
hoytak: I don't think this is an issue, since the values where this is affected were pretty esoteric.  But worth calling out.

Cargo.toml
hoytak: I think this can happen, but it should probably be a bump to 1.5.  And 3.8 is a reasonable target IMO.

xet_runtime/src/config/xet_config.rs
hoytak: Updated.

xet_runtime/src/config/macros.rs
hoytak: Fixed.

PR #708, opened 2026-03-12
Adjust RTT prediction determining concurrency by transmission size.

hoytak:
Currently, the condition for increasing connection concurrency is gated on the model predicting that a 64MB transmission will complete within 90 seconds.  However, when the transmissions are primarily composed of small packets, this can drastically overestimate the round trip, artificially suppressing the connection concurrency.  

This PR fixes this issue by also modeling the average predicted packet size, using the 95% quantile of that (bounded by two config variables) to predict the round trip time when considering a concurrency increase.


XciD:
nice !

PR #711, opened 2026-03-13
Implement scaled connection concurrency

hoytak:
Currently, small transfers count the same way in modeling and adapting the concurrency as do larger transfers.  This PR introduces a scaling model that permits smaller connections to count less towards a concurrency total, as the bulk is spent waiting for the server latency.


hoytak:
Closing this PR; with benchmarks, it appears to have too small an impact and possible side effects.

PR #714, opened 2026-03-13
Xorb download URL debug logs

bpronan:
It's a bit annoying to try to ensure our CDN routing is correct. Logging the URL domain for the first fetch term download to the debug logs.

Please don't hesitate to recommend alternative approaches.

### DIFF
--- a/xet_data/src/file_reconstruction/reconstruction_terms/manager.rs
+++ b/xet_data/src/file_reconstruction/reconstruction_terms/manager.rs
@@ -36,7 +36,6 @@ pub struct ReconstructionTermManager {
     progress_updater: Option<Arc<DownloadTaskUpdater>>,
     total_bytes_reported: u64,
     total_transfer_bytes_reported: u64,
-    logged_domain: bool,
 }
 
 impl ReconstructionTermManager {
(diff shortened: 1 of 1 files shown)


xet_data/src/file_reconstruction/reconstruction_terms/manager.rs
```diff
@@ -36,6 +36,7 @@ pub struct ReconstructionTermManager {
     progress_updater: Option<Arc<DownloadTaskUpdater>>,
     total_bytes_reported: u64,
     total_transfer_bytes_reported: u64,
+    logged_domain: bool,
```
seanses: This is deep down from the high level API, where / how are you going to set it? Maybe consider place the flag inside `xet_runtime/src/config/groups/reconstruction.rs` with name `log_domain` so you can set with env var `HF_XET_RECONSTRUCTION_LOG_DOMAIN` and access using `xet_config().reconstruction.log_domain`

bpronan: I've restructured this chance to avoid this field. As @hoytak recommended, we're now logging the domain for each file term block.

I was originally concerned about the noise this creates in the logs, but it's we're already logging every block anyway. I can solve the noise issue at the log parsing level instead of introducing this new field.

PR #716, opened 2026-03-14
fix: bound file reconstruction range using file_size to prevent 416 errors

XciD:
## Summary

Bounds the file reconstruction range using the known file size instead of requesting beyond EOF. This prevents spurious 416 Range Not Satisfiable errors from the CAS server for small files.

When downloading a full file without an explicit byte range, `FileReconstructor` used `FileRange::full()` (0..u64::MAX). This caused `ReconstructionTermManager` to speculatively prefetch blocks beyond EOF, making the CAS server return 416 for small files. The large start values (~1.8e19) visible in production logs confirm unsigned arithmetic wrapping around u64::MAX.

## Impact (production, last 24h)

- **~32k errors/24h**: u64 overflow in range computation (`start: 18446744073709...`, `end: 15xxx`) directly caused by `FileRange::full()`
- **~29M errors/24h** (~1M/h): `range start is greater than file length` (416 responses). Likely a mix of this bug (speculative prefetch beyond EOF) and legitimate client-side range errors

### DIFF
--- a/xet_data/src/file_reconstruction/file_reconstructor.rs
+++ b/xet_data/src/file_reconstruction/file_reconstructor.rs
@@ -37,10 +37,6 @@ pub struct FileReconstructor {
     /// When cancelled, reconstruction stops at its next check point. Long waits
     /// (such as semaphore acquisition) use `tokio::select!` so they abort promptly.
     cancellation_token: CancellationToken,
-
-    /// Known file size used to bound the reconstruction range for full-file downloads.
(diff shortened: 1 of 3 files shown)


rajatarya:
Probably not worth blocking this going in but I would love it if we used the LocalClient/LocalTestServer to write some tests to exercise this behavior and confirm no 416 returned.

rajatarya (approved):

XciD:
went with a cleaner version + tests

hoytak (approved):

PR #717, opened 2026-03-16
feat: range aware file write

XciD:
## Summary

APIs for range-aware file writes: instead of re-uploading an entire file when only part of it changed, compose a new CAS file from stable segments + re-chunked dirty windows. Supports resize edits (insert / delete / arbitrary replace) in addition to in-place rewrites.

### API: `upload_ranges`

```rust
pub async fn upload_ranges(
    config: Arc<TranslatorConfig>,
    cas_client: Arc<dyn Client>,
    original_hash: MerkleHash,
    original_size: u64,
    dirty_inputs: Vec<DirtyInput>,
) -> Result<XetFileInfo>
```

```rust
/// A single edit applied to the original file: replace `original_range` with
/// `new_length` bytes from `reader`. Edits are expressed in original-file coordinates.
pub struct DirtyInput {
    pub original_range: Range<u64>,
    pub reader: Pin<Box<dyn AsyncRead + Send>>,
    pub new_length: u64,
}
```

The output file size is **derived** from the inputs (no `total_size` parameter): `original_size - removed + added`.

### Edit shapes (all expressible with the same struct)

| Operation | `original_range` | `new_length` |
|---|---|---|
| In-place edit | `a..b` | `b - a` |
| Resize replace | `a..b` | any |
| Pure insert | `p..p` | `> 0` |
| Pure delete | `a..b` | `0` |
| Append | `original_size..original_size` | `> 0` |
| Truncate to N | `N..original_size` | `0` |
| No-op | empty `dirty_inputs` | — |

Motivating example:

```text
abc + upload_ranges([0..1), "foo", 3) = foobc
abc + upload_ranges([0..0), "foo", 3) = fooabc
abc + upload_ranges([0..1), "",    0) = bc
```

**Per-range `AsyncRead` instead of `ReadSeek` over the staging file.** The earlier prototype took `dirty_ranges: &[(u64, u64)] + dirty_source: &mut dyn ReadSeek`. That had a subtle bug: for truncation we silently extended the dirty set with a boundary chunk and read those bytes from the staging file, but if the file was never opened for write the staging file contains zeros at those positions (real bytes are in CAS) → silent corruption on the truncation boundary chunk. Pairing each edit with its own reader makes that structurally impossible: any byte not provided by the caller is fetched from CAS.

<details>
<summary>How it works</summary>

### High level

```
                           upload_ranges
   +----------------------+   |   +----------------------+
   |  original file (CAS) |---+-->|  composed file (CAS) |
   +----------------------+       +----------------------+
   only the dirty windows are re-uploaded; everything else
   is reused as whole CAS segments.
```

### Step 1 — coalesce + snap edits to segment boundaries

Edits are user-coordinates (byte ranges). We snap each edit's `original_range` to the **enclosing CAS segments** so composition can swap whole segments instead of truncating one mid-chunk. Adjacent / overlapping snapped ranges are then coalesced.

Pure inserts (`start == end`) snap to the segment that owns `start`; an insert at `original_size` snaps to the last segment.

### Step 2 — server returns windows + gap subtrees

Single CAS call: `GET /v2/file-chunk-hashes/{file_id}` with the segment-aligned ranges in an `X-Range-Dirty: bytes=A-B,C-D` header. Response shape (xetcas#987):

```rust
struct FileChunkHashesResponse {
    windows:      Vec<ChunkWindow>,         // one per dirty range
    hash_ranges:  Vec<Option<MerkleHashSubtree>>, // N+1 entries: [gap0, gap1, ..., gapN]
}
```

`windows[i].chunks` carries the chunk hashes the server actually owns for that window (we re-upload these bytes). `hash_ranges[i]` is the **MerkleHashSubtree** for the i-th unmodified gap, or `None` when there is no gap there. This is the key to composing the final file hash without touching unmodified bytes.

### Step 3 — for each window, stream `[CAS prefix | edits | CAS suffix]` through a fresh cleaner

```
window = [w_start ............................................. w_end]
edits in this window:        [edit_a]    [edit_b]
                                ^           ^
streamed input to the cleaner:
  CAS bytes [w_start, edit_a.start)
  reader bytes for edit_a (new_length bytes)
  CAS bytes [edit_a.end, edit_b.start)
  reader bytes for edit_b
  CAS bytes [edit_b.end, w_end)
```

Pure inserts contribute zero original bytes but still emit `new_length` reader bytes. Pure deletes contribute zero reader bytes. The cleaner produces a new `MDBFileInfo` per window and a `ChunkHashList`.

### Step 4 — compose the file hash via `MerkleHashSubtree::merge`

```text
merge_seq = [gap0, w0, gap1, w1, ..., wN, gapN]   // skip None gaps

merged          = MerkleHashSubtree::merge(merge_seq)
aggregated_hash = merged.final_hash()
combined_hash   = aggregated_hash.hmac(zero)      // matches cleaner's file_hash
```

Special-case: if `total_size == 0` (e.g. truncate to empty) the result is `MerkleHash::default()` *without* HMAC, mirroring `file_hash([])`.

### Step 5 — splice segments + register

Walk the original `MDBFileInfo.segments` and replace any segment that falls inside a window with that window's freshly-uploaded segments. Verification entries follow segment-for-segment when present. `metadata_ext = None` (no SHA-256, see Limitations). Then `register_composed_file` + `finalize`.

### Multi-window example

Two edits: replace `[50MB, 51MB)` and `[150MB, 151MB)` on a 200MB file:

```
+-----------+-------+------------+-------+-----------+
|  GAP 0    |  W0   |   GAP 1    |  W1   |  GAP 2    |
|  reused   |upload |  reused    |upload |  reused   |
| (subtree) | ~1MB  | (subtree)  | ~1MB  | (subtree) |
+-----------+-------+------------+-------+-----------+

Wire transfer: ~2MB upload + a few hundred KB of CAS reads for window
boundary chunks. Old approach: 200MB download + 200MB upload.
```

### Empty original short-circuit

When `original_size == 0` there is nothing to compose against — every edit's `original_range` must be `0..0` (validated). We just stream the new bytes through a fresh cleaner (`upload_fresh_file`).

</details>

### Reviewer note: `chunk_window_builder` is a re-implementation of xetcas

`xet_client/src/cas_client/chunk_window_builder.rs` is a port of the same window-building state machine that already lives in xetcas — it's only used by the local / in-memory simulation clients (`local_client`, `memory_client`) so the mock CAS server returns the same shape as the real one in tests. **No need to re-review it as part of this PR**: it mirrors logic already reviewed and merged in xetcas#987. A follow-up xetcas PR will deduplicate by removing the server-side copy and pulling this one in (or vice versa); the duplication is intentional and temporary.

### Limitations

- **No SHA-256 metadata**: composed files have `metadata_ext = None` since recomputing SHA-256 would require reading the full file. Only suitable for contexts that don't require SHA-256 verification (HF buckets, xet-native repos), not for Git LFS-backed repos.
- **Memory**: for very large files, the per-window in-memory state (chunk hash list + composed segments) is bounded by the dirty regions, not the whole file. The chunk-hashes response is paginated by the server-defined window granularity.

### Tests (27)

Covering all edit shapes + edge cases. Notable:

| Test | Purpose |
|---|---|
| `test_resize_edits_abc` | The 3 motivating FUSE examples |
| `test_resize_large_replace_grows_file` | Replace `[a..b)` with much more data |
| `test_resize_large_replace_shrinks_file` | Replace `[a..b)` with much less data |
| `test_resize_mid_file_insert` | Pure insert in the middle |
| `test_resize_mid_file_delete` | Pure delete in the middle |
| `test_resize_multi_edit_mix` | Insert + replace + delete in one call |
| `test_resize_insert_at_segment_boundary` | Snapping correctness for inserts |
| `test_upload_ranges_mid_file_edit` | In-place edit |
| `test_upload_ranges_truncation` | Pure truncate (sub-segment) |
| `test_upload_ranges_truncation_empty_staging` | Truncate when staging is all-zero (boundary read from CAS) |
| `test_upload_ranges_truncation_with_overlapping_dirty` | Truncate + dirty range overlapping the boundary |
| `test_truncate_to_empty_matches_clean_empty` | Truncating to 0 hashes to `MerkleHash::default()` (matches a fresh empty cleaner) |
| `test_upload_ranges_append` | Pure append |
| `test_append_with_gap_before_dirty_range` | Append where reader covers a sparse gap too |
| `test_append_sparse_staging_file` | Append on a sparse staging file |
| `test_mid_edit_plus_append` | Mid-file edit *and* append in one call (P1 codex regression) |
| `test_empty_original_append` | `original_size == 0` + append falls into the fresh-file path (P2 codex regression) |
| `test_empty_original_validates_ranges` | `original_size == 0` still runs validation (reviewer regression) |
| `test_upload_ranges_at_file_start` | Edit at offset 0 (no stable prefix) |
| `test_upload_ranges_multiple_regions` | Two non-adjacent dirty windows with stable gap |
| `test_single_input_spanning_many_chunks` | One edit covering many CDC chunks |
| `test_data_integrity_scenarios` | 5 sub-scenarios covering composition correctness |
| `test_noop_returns_original_hash` | Empty `dirty_inputs` → no CAS call, original hash returned |
| `test_rejects_dirty_range_past_total_size` | Validation: range past `original_size` |
| `test_rejects_overlapping_dirty_ranges` | Validation: overlapping edits |
| `test_rejects_unsorted_dirty_ranges` | Validation: unsorted edits |
| `test_upload_ranges_small_file_mid_edit` | Small files (single segment) |

### Dependencies

- xetcas: `GET /v2/file-chunk-hashes/{file_id}` with `windows[] + hash_ranges[]` response shape — huggingface-internal/xetcas#987 (merged).
- Consumer: huggingface-internal/hf-mount#41.

---

> [!NOTE]
> **High Risk**
> High risk because it adds a new partial-upload composition path that splices CAS segments and recomputes file hashes from window subtrees, touching core data integrity and client/server chunk-boundary logic.
> 
> **Overview**
> Adds range-aware file writes via new `upload_ranges`, letting callers apply insert/delete/replace edits and upload only re-chunked dirty windows while reusing stable CAS segments.
> 
> Introduces a new CAS API `get_file_chunk_hashes` (`GET /v2/file-chunk-hashes/{file_id}` with `X-Range-Dirty`) plus response types (`FileChunkHashesResponse`, `ChunkWindow`) and simulation support (`chunk_window_builder`) that extends dirty ranges to *stable* chunk boundaries and returns gap `MerkleHashSubtree` summaries + stable-segment verification.
> 
> Refactors dedup/cleaning plumbing to expose per-chunk hash lists (`ChunkHashList`), adds detached cleaner/session completion and `register_composed_file` to avoid orphan shard entries, and moves/re-exports `next_stable_chunk_boundary` into `xet_core_structures` for shared stable-window computations.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f4cee46df01ae526d37f55a614768bec059b6e0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1571 @@
...
+    for region in raw {
+        if let Some(last) = merged.last_mut()
+            && region.first_chunk <= last.last_chunk
+        {
+            last.dirty_end = last.dirty_end.max(region.dirty_end);
+            last.last_chunk = last.last_chunk.max(region.last_chunk);
+            continue;
+        }
+        merged.push(region);
+    }
```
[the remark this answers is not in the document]
XciD: False positive. The append backup always produces `first_chunk = num_chunks - 1`, and any dirty range in the last chunk also computes `first_chunk = num_chunks - 1` via `partition_point`. They're always equal when merging, so `min()` would be a no-op.

Verified by writing a unit test with controlled chunk offsets: the backed-up `first_chunk` never ends up smaller than a preceding region's `first_chunk`.

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1571 @@
...
+        let clamped_end = dirty_end.min(original_size);
+        let last_chunk = (0..num_chunks)
+            .rev()
+            .find(|&i| chunk_offsets[i] < clamped_end)
+            .map(|i| i + 1)
+            .ok_or_else(|| {
+                DataProcessingError::InternalError(format!(
+                    "no chunk starts before clamped_end ({clamped_end}), chunks may be inconsistent"
+                ))
+            })?;
```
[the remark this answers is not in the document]
XciD: Good catch, fixed. Replaced with `partition_point` to match `first_chunk`.

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1568 @@
...
+    seg_cursor: &mut usize,
+) -> (Vec<FileDataSequenceEntry>, Vec<FileVerificationEntry>) {
+    let mut segments = Vec::new();
+    let mut verification = Vec::new();
+
+    // Compute the chunk-level cursor from the segment cursor.
+    let mut chunk_cursor: usize = original_mdb.segments[..*seg_cursor]
+        .iter()
+        .map(|s| (s.chunk_index_end - s.chunk_index_start) as usize)
+        .sum();
```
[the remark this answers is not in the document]
XciD: Fixed. `extract_segments` now takes a `seg_chunk_cursor` tracked by the caller instead of recomputing from segments.

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1566 @@
...
+    // Appended bytes must be covered by dirty_inputs (CAS has no data beyond original_size).
+    if total_size > original_size {
+        let last_input_end = dirty_ranges.last().map_or(0, |&(_, e)| e);
+        if last_input_end < total_size {
+            return Err(DataError::InternalError(format!(
+                "total_size ({total_size}) > original_size ({original_size}) but dirty_inputs \
+                 only cover up to byte {last_input_end} (must reach total_size)"
+            )));
+        }
+    }
```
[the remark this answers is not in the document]
XciD: valid one, fixed [79a1ef3](https://github.com/huggingface/xet-core/pull/717/commits/79a1ef332df4ba01d1fa22a6f91f9a324d722d6e)

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1623 @@
...
+        }
+
+        let (info, chunks, _metrics) = cleaner.finish().await?;
+        uploaded_regions.push(UploadedRegion { region, info, chunks });
+    }
+
+    // Checkpoint: flush xorbs without consuming the session, then retrieve MDBFileInfos.
+    // TODO: the middle files are registered in the shard as real files, but nobody will
+    // ever reference them. Check if GC cleans up unreferenced file entries, or find a way
+    // to retrieve segments from the session without persisting them to the shard.
```
rajatarya: This is the one thing I'd want resolved (or tracked as an issue) before merge. If the shard GC doesn't clean up unreferenced file entries, this creates shard bloat proportional to the number of dirty regions across all `upload_ranges` calls. Do you know if GC handles this today?

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1623 @@
...
+            let overlap = &original_chunks[overlap_start..overlap_end];
+            let mut bytes = 0u64;
+            let mut hashes = Vec::with_capacity(overlap.len());
+            for &(hash, size) in overlap {
+                bytes += size;
+                hashes.push(hash);
+            }
+            // u32 cast: unpacked_segment_bytes is u32 in the shard format.
+            // Safe because CDC parameters prevent segments from exceeding u32::MAX.
+            truncated.unpacked_segment_bytes = bytes as u32;
```
rajatarya: nit: The comment explains why this is safe with current CDC parameters (128KB max chunk * 8192 chunks/xorb = 1GB, fits u32). A `debug_assert!(bytes <= u32::MAX as u64)` would be cheap insurance against future parameter changes.

xet_client/src/cas_client/remote_client.rs
```diff
@@ -735,6 +736,45 @@ impl Client for RemoteClient {
...
+        let client = self.authenticated_http_client.clone();
+
+        let response: FileChunkHashesResponse = RetryWrapper::new(api_tag)
+            .run_and_extract_json(move || client.get(url.clone()).with_extension(Api(api_tag)).send())
+            .await?;
+
+        let chunks = response.chunks.into_iter().map(|entry| (entry.hash, entry.size)).collect();
+
+        Ok(chunks)
+    }
```
rajatarya, line 793: nit: Just confirming -- `camelCase` here matches the actual response format from xetcas #501? I see the field is just `chunks` (single word, no case transformation needed), but wanted to make sure the `ChunkHashEntry` fields (`hash`, `size`) also come back camelCase from the server.

rajatarya (commented):
Great PR -- one of the best-documented I've seen. The per-range `AsyncRead` redesign is a textbook "make illegal states unrepresentable" fix, the three-region decomposition is clean, and the test coverage is excellent (round-trip verification on every content-modifying test is the gold standard).

**Should address:**
1. **Orphaned middle file entries (the TODO at line 375)** -- re-chunked middle regions get registered as real files in the shard that nobody references. Worth confirming GC handles these, or creating a tracking issue if not.
2. **Unchecked u32 cast in `extract_segments`** -- `unpacked_segment_bytes = bytes as u32` is safe with current CDC parameters but a `debug_assert!` is cheap insurance.

**Nice to have:**
3. nit: `camelCase` serde -- just verifying this matches the actual xetcas #501 response format.
4. nit: `extract_segments` assumes verification entries are 1:1 with segments -- a brief comment confirming this invariant would help future readers.

Overall: no blockers, code quality is high, tests are comprehensive.

rajatarya:
My understanding is GC will not catch this - it is driven by files marked as deleted in the Xet Storage backend. @sirahd please confirm.

Can we avoid registering the shard for partial region and only register it for the completed file?

rajatarya (approved):
My main concern is the GC question about registering the partial file shard - I think we should avoid that because we aren't designing v1 GC to find these shards.

rajatarya:
To assist other HF reviewers - I have a writeup generated across all the PRs related to Partial File Editing in Xet Storage - please see the writeup here: https://github.com/huggingface-internal/xetcas/pull/501#issuecomment-4139242784

(Unfortunately Notion doesn't like importing this flavor of Markdown)

xet_data/src/processing/range_upload.rs
XciD: pinging @sirahd for this

sirahd: Yes, the current design for GC should already handle the scenarios presented by this API change. I created an issue to add more details and track its validation here: https://github.com/huggingface-internal/xet-garbage-collection/issues/2

lhoestq (commented):
I'm playing a bit with it now :)

Btw would it be hard to support mid-file edits with a range of different size ?

e.g. given a file content `abc` support this kind of edits:

- `abc` + `upload_ranges([0, 1), "foo")` = `foobc`
- `abc` + `upload_ranges([0, 0), "foo")` = `fooabc`
- `abc` + `upload_ranges([0, 1), "")` = `bc`

lhoestq:
Forgot to share earlier but for those who want to play with it in python I made a simple integration in `huggingface_hub`:

```python
""" 
This requires 

- https://github.com/lhoestq/xet-core/tree/lhoestq/py-upload_ranges
- https://github.com/huggingface/huggingface_hub/tree/upload-ranges
""" 

from huggingface_hub.hf_api import HfApi, FileRange, hffs

hffs.write_text("buckets/username/my-bucket/data.txt", "Hello world")

ranges = [FileRange(start=6, end=11, content=b"you !")]
HfApi()._batch_bucket_file_ranges("username/my-bucket", "data.txt", ranges=ranges)

print(hffs.read_text("buckets/username/my-bucket/data.txt"))
# Hello you !
```

Also sharing some ideas of what we can have in the future, since this requires the ability for new ranges to be smaller or longer than the overwritten range:

```python
"""
(Brainstorming)
Example of filesystem API we could have:
- similar interface as python lists operations
- open(..., "m") is for for mutate/modify
- replace/insert/remove bytes in the middle of the file
"""
from huggingface_hub import hffs

path = "buckets/username/my-bucket/data.txt"
hffs.write_text(path, "Hello world")

with hffs.open(path, "m") as f:
   # do stuff here

# List-like API:
f[6:] = "all"
f[:0] = "Welcome. ")
del f[:5]

# Regular methods
# f[6:] = "all"
f.seek(6)
f.write("all")
f.truncate()
# f[:0] = "Welcome. ")
f.seek(0)
f.insert("Welcome. ")
# del f[:5]
f.seek(0)
f.delete(5)

# One-liners for convenience
# f[6:] = "all"
f.write("all", index=6)
f.truncate()
# f[:0] = "Welcome. ")
f.insert("Welcome. ", index=0)
# del f[:5]
f.delete(5, index=0)
```

cc @Wauplin for viz and feedbacks

xet_client/src/cas_client/remote_client.rs
```diff
@@ -735,6 +736,45 @@ impl Client for RemoteClient {
...
+}
+
+/// Response from `GET /v2/file-chunk-hashes/{file_id}`.
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
```
seanses: Let's move these two structs (`FileChunkHashesResponse` & `ChunkHashEntry`) to under "xet_client/src/cas_types/mod.rs" where CAS query response types are defined.

xet_client/src/cas_client/remote_client.rs
```diff
@@ -735,6 +736,45 @@ impl Client for RemoteClient {
...
+struct FileChunkHashesResponse {
+    chunks: Vec<ChunkHashEntry>,
+}
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct ChunkHashEntry {
+    #[serde(deserialize_with = "deserialize_merkle_hash")]
+    hash: MerkleHash,
```
seanses: Change this type to `hash: HexMerkleHash`. `HexMerkleHash` already has its deserialization from str function defined so you don't need to implement one again below. See examples using `HexMerkleHash` in other structs under "xet_client/src/cas_types/mod.rs".

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1623 @@
...
+) -> Result<XetFileInfo> {
+    if dirty_inputs.is_empty() && total_size == original_size {
+        return Ok(XetFileInfo::new(original_hash.hex(), original_size));
+    }
+
+    // Extract ranges for validation and build_dirty_regions.
+    let dirty_ranges: Vec<(u64, u64)> = dirty_inputs.iter().map(|d| (d.range.start, d.range.end)).collect();
+
+    if !dirty_ranges.windows(2).all(|w| w[0].1 <= w[1].0) {
+        return Err(DataError::InternalError(format!(
```
seanses: Better use `ParameterError` (same below)

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1623 @@
...
+                "total_size ({total_size}) > original_size ({original_size}) but dirty_inputs \
+                 only cover up to byte {last_input_end} (must reach total_size)"
+            )));
+        }
+
+        // Verify no gaps beyond original_size between inputs. Walk the append region
+        // [original_size, total_size) and ensure it is fully covered.
+        let mut covered_up_to = original_size;
+        for &(start, end) in &dirty_ranges {
+            if start > covered_up_to && covered_up_to >= original_size {
```
seanses: `covered_up_to >= original_size`: this condition is always true because initially `covered_up_to := original_size` and only grows (`.max(end)`). Remove for clarity

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1623 @@
...
+                return Err(DataError::InternalError(format!(
+                    "gap in append region: bytes [{covered_up_to}, {start}) are beyond \
+                     original_size ({original_size}) and not covered by any dirty input"
+                )));
+            }
+            covered_up_to = covered_up_to.max(end);
+        }
+    }
+
```
seanses, line 127: Let's try to reduce the size of this function, consider extract the above checks into a "validate_parameters" function

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1623 @@
...
+    // 1. Fetch chunk hashes and reconstruction info in parallel.
+    let (original_chunks, recon_result) = tokio::try_join!(
+        cas_client.get_file_chunk_hashes(&original_hash),
+        cas_client.get_file_reconstruction_info(&original_hash),
+    )?;
+    let original_mdb = recon_result
+        .map(|(mdb, _)| mdb)
+        .ok_or_else(|| DataError::InternalError("no reconstruction info for original file".into()))?;
```
seanses: `ParameterError("file not found")`

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1623 @@
...
+    for &(dirty_start, dirty_end) in dirty_ranges {
+        // Find the first chunk whose end offset exceeds dirty_start.
+        let mut first_chunk = chunk_offsets[1..].partition_point(|&o| o <= dirty_start);
+        debug_assert!(first_chunk <= num_chunks, "first_chunk {first_chunk} out of bounds ({num_chunks} chunks)");
+
+        // For append regions (dirty_start >= original_size), include the last original
+        // chunk so it gets re-chunked with the appended data. The last chunk was
+        // terminated by EOF (not by the rolling hash), so its boundary is artificial.
+        // The boundary prefix mechanism will download its bytes from CAS.
+        if total_size > original_size && dirty_start >= original_size && first_chunk > 0 {
```
seanses: `total_size > original_size`: this condition is implied by `dirty_start >= original_size` because we validated the parameters, consider remove for clarity.

`first_chunk > 0`: this condition is also implied by `dirty_start >= original_size`, because in this case a chunk whose end offset exceeds `dirty_start` doesn't exist, so `first_chunk = num_chunks > 0`.

Wauplin:
Hey! Very excited about this feature! To avoid polluting the PR with discussions about the Python API (see @lhoestq 's https://github.com/huggingface/xet-core/pull/717#issuecomment-4253676022), I've moved the discussion to a new issue https://github.com/huggingface/huggingface_hub/issues/4153. Anyone's welcome to share their feedback as well

xet_data/src/processing/range_upload.rs
XciD: Followed up by sirahd above and tracked in https://github.com/huggingface-internal/xet-garbage-collection/issues/2. The TODO comment is also gone now: the per-region MDB extraction was rewritten in https://github.com/huggingface/xet-core/pull/717/commits/a9d9cfe6 to consume the per-window MDBs straight from the session checkpoint, so there are no more standalone middle file entries to worry about.

xet_data/src/processing/range_upload.rs
XciD: The `extract_segments` helper that did the `unpacked_segment_bytes as u32` cast was removed in https://github.com/huggingface/xet-core/pull/717/commits/a9d9cfe6. Composition now swaps whole segments straight from the original MDB, so we never recompute that field on the client; no insurance needed.

xet_client/src/cas_client/remote_client.rs
XciD: The flat `chunks` / `ChunkHashEntry` response was replaced by xetcas#987 with a windows + opaque `MerkleHashSubtree` gap shape. The corresponding client types now live in `xet_client::cas_types` (`FileChunkHashesResponse`, `ChunkWindow`) with `#[serde(rename_all = "camelCase")]` and mirror the server structs 1:1.

xet_client/src/cas_client/remote_client.rs
XciD: Done in https://github.com/huggingface/xet-core/pull/717/commits/78de6886. `FileChunkHashesResponse`, `ChunkWindow`, and the `X_RANGE_DIRTY_HEADER` constant now live in `xet_client/src/cas_types/mod.rs`. `ChunkHashEntry` is gone (xetcas#987 dropped per-chunk hashes from the response in favour of opaque gap subtrees).

xet_client/src/cas_client/remote_client.rs
XciD: Obsolete: `ChunkHashEntry` and the local `deserialize_merkle_hash` helper are both gone. The new response only carries `Vec<Option<MerkleHashSubtree>>`, which has its own serde derive.

xet_data/src/processing/range_upload.rs
XciD: Done in https://github.com/huggingface/xet-core/pull/717/commits/2b572dc1: caller-input violations now return `DataError::ParameterError`. Remaining `InternalError`s are reserved for genuine internal failures (`MerkleHashSubtree::merge` mismatch, missing per-window MDB after checkpoint, etc.).

xet_data/src/processing/range_upload.rs
XciD: Good catch, removed in https://github.com/huggingface/xet-core/pull/717/commits/2b572dc1.

xet_data/src/processing/range_upload.rs
XciD: Done in https://github.com/huggingface/xet-core/pull/717/commits/2b572dc1: validation lives in `validate_dirty_ranges()`, called from both `upload_ranges` and `upload_fresh_file`.

xet_data/src/processing/range_upload.rs
XciD: Done in https://github.com/huggingface/xet-core/pull/717/commits/2b572dc1: now returns `DataError::ParameterError("file not found")` when reconstruction info is missing for the requested `original_hash`.

xet_data/src/processing/range_upload.rs
XciD: The whole `build_dirty_regions` helper was removed in https://github.com/huggingface/xet-core/pull/717/commits/a9d9cfe6 in favour of segment-level snapping driven by the new windows endpoint, so neither condition exists anymore.

XciD:
@lhoestq I've modified the api to answer to your need. We can't test right know over the prod as the code is not deployed yet.

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1775 @@
...
+
+            cursor = edit_end;
+        }
+        input_idx = edits_end;
+
+        if cursor < w_end {
+            stream_cas_range(&ctx, &cas_client, original_hash, cursor, w_end, &mut cleaner).await?;
+        }
+
+        let (info, chunks, _metrics) = cleaner.finish_with_chunks().await?;
```
rajatarya: This per-window `finish_with_chunks` ends up calling `register_single_file_clean_completion`, which (via `process_aggregated_data_as_xorb` → `add_file_reconstruction_info`) registers the window's MDB in the session shard. Combined with `register_composed_file` later (line 383), a single `upload_ranges` call writes N+1 reconstruction entries — N orphans + 1 composed. Same concern as my March pass; this PR's redesign didn't change the registration path. Worth confirming the v1 plan: ship as-is and rely on a follow-up to suppress the per-window entries, or expose an `unregistered finish` variant on the cleaner so composition flows can opt out.

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1775 @@
...
+    }
+    Ok(())
+}
+
+/// Compute the resulting file size: `original_size` minus bytes removed plus bytes added.
+fn compute_total_size(original_size: u64, dirty_inputs: &[DirtyInput]) -> u64 {
+    let (removed, added) = dirty_inputs
+        .iter()
+        .fold((0u64, 0u64), |(r, a), d| (r + (d.original_range.end - d.original_range.start), a + d.new_length));
+    original_size + added - removed
```
rajatarya: `original_size + added - removed` can overflow when callers pass large `new_length` values (`added` is just `sum(d.new_length)` with no upper bound). Since `validate_dirty_ranges` already runs and returns `ParameterError`, a `checked_add(added).and_then(|s| s.checked_sub(removed))` here with a `ParameterError` on `None` would keep the validation idiom consistent. Not exploitable in normal use — but a buggy caller currently silently wraps.

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1775 @@
...
+
+    session.checkpoint().await?;
+    let mdb_list = session.file_info_list().await?;
+    let mdb_by_hash: HashMap<MerkleHash, MDBFileInfo> =
+        mdb_list.into_iter().map(|m| (m.metadata.file_hash, m)).collect();
+
+    // Merge sequence: [gap0, w0, gap1, w1, ..., gapN]. Empty gaps (`None`) are skipped.
+    let mut hash_ranges = response.hash_ranges;
+    let trailing_gap = hash_ranges.pop().flatten();
+    let first_window_at_start = hash_ranges.first().is_some_and(Option::is_none);
```
rajatarya: nit: this reads obliquely on first pass — could you add a short comment? Something like `// leading gap is empty ⇒ first window starts at byte 0`. The `is_some_and(Option::is_none)` on `Vec<Option<_>>` is the kind of thing that's easy to misread as a typo.

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1775 @@
...
+    let composed_mdb = MDBFileInfo {
+        metadata: FileDataSequenceHeader::new(combined_hash, all_segments.len(), with_verification, false),
+        segments: all_segments,
+        verification: all_verification,
+        // SHA-256 is intentionally omitted: the file content changed, and recomputing it
+        // would require reading the full file.
+        metadata_ext: None,
+    };
+
+    session.register_composed_file(composed_mdb).await?;
```
rajatarya: nit: the `mdb_by_hash` lookup path (line 279) assumes window content-hashes are unique within the session. Two windows producing identical content would dedupe in the HashMap, but the splice loop would then reuse the same `middle_mdb` for both — which is actually correct because the content is identical. Worth a one-line comment noting that hash-collision-by-content is fine here, so future readers don't add a defensive check.

rajatarya (commented):
Re-reviewing this since the design has shifted substantially since my March approval (per-edit `AsyncRead`, resize-edit support, `MerkleHashSubtree::merge` composition, multi-range API adoption). The redesign is a clear improvement — particularly:

- **Per-edit `AsyncRead` + reader ownership**: structurally fixes the prior truncation-corruption hazard (callers literally cannot supply zero bytes for a region not in their edit set).
- **Defensive contract checks** on the server response (`response.windows.len() != n_windows`, `hash_ranges.len() != n_windows + 1`, plus the post-loop `input_idx != dirty_inputs.len()` check) — these turn what would otherwise be silent file-hash corruption into loud errors.
- **Validation runs before the empty-original short-circuit** — the regression test `test_empty_original_validates_ranges` is exactly the right shape.
- **Test matrix is excellent.** Round-trip + clean-upload-hash equivalence on every content-modifying test is the gold standard, and the codex regressions are captured (`test_mid_edit_plus_append`, `test_empty_original_append`, `test_truncate_to_empty_matches_clean_empty`).

My approval still stands; logging a few items here that came up on this pass:

1. **Orphan window file entries (still present from my March review).** Each window's `cleaner.finish_with_chunks()` ultimately calls `register_single_file_clean_completion` → `add_file_reconstruction_info(fi)` (file_upload_session.rs:494), and then `register_composed_file` registers the composed MDB on top. So a single `upload_ranges` call writes N+1 file reconstruction entries to the session shard: one per window plus the composed one. Functionally fine — segments are reachable via the composed file — but the per-window entries are unreferenced and pollute the catalog, and v1 GC isn't designed to spot them. Worth confirming this is intentional for v1 and tracking the cleanup, or filing an issue if not yet logged.
2. **`compute_total_size` (range_upload.rs:437)** does plain arithmetic on caller-supplied `new_length`. Since validation only checks `original_range.end <= original_size` (not totals), a malicious or buggy caller passing huge `new_length` values overflows `original_size + added`. A `checked_add` (+ `ParameterError`) would close that.
3. nit: `hash_ranges.first().is_some_and(Option::is_none)` (range_upload.rs:285) reads obliquely — a single-line comment like `// leading gap is empty ⇒ first window starts at byte 0` would help future readers parse the double-Option intent.
4. nit: `register_composed_file`'s lookup-by-hash path (collecting `mdb_by_hash` from `session.file_info_list()`) implicitly assumes window content-hashes are unique within the session. If two windows happened to produce identical content (unlikely but possible with synthetic data), the HashMap would dedupe them and the `.get(&middle_hash)` would still resolve, but the segment splice would reuse the same `middle_mdb` twice — which is actually correct since the content is identical. Worth a brief comment that this is intentional.

Nothing here changes my approve verdict — items 2-4 are small. Item 1 is the same flag from March; happy to proceed if the call is to ship and clean up later.

xet_data/src/processing/range_upload.rs
XciD: Fixed in 46e1be3f. `compute_total_size` now uses `checked_add`/`checked_sub` and returns `ParameterError` on overflow.

xet_data/src/processing/range_upload.rs
XciD: Done in 46e1be3f.

xet_data/src/processing/range_upload.rs
XciD: Done in 46e1be3f.

xet_data/src/processing/range_upload.rs
XciD: Fixed in c41c195c. `upload_ranges` now uses a new `finish_with_chunks_detached()` that uploads the xorb data but returns the MDBFileInfo directly instead of registering it in the session shard. Only the final composed file gets registered, eliminating the N orphan per-window entries.

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1807 @@
...
+            response.windows.len()
+        )));
+    }
+    if response.hash_ranges.len() != n_windows + 1 {
+        return Err(DataError::InternalError(format!(
+            "server returned {} hash_ranges, expected {} (n_windows + 1)",
+            response.hash_ranges.len(),
+            n_windows + 1
+        )));
+    }
```
rajatarya, line 183: These three guards (windows count, hash_ranges count at L182, edits-not-assigned-to-window at L269, and the `gap_verification` length checks in `compose_mdb`) are exactly the right thing to do. A subtle composition algorithm where `zip` silently truncates the merge sequence is precisely where you want loud bails over quiet corruption. The error messages even include expected vs actual counts, which makes server contract drift trivial to diagnose.

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1807 @@
...
+
+        // Find the slice of edits that land in this window. A pure insert at exactly
+        // `w_end` belongs here only when `w_end == original_size` (no later window can
+        // take it); anywhere else it belongs to the next window starting at that byte.
+        let edits_end = dirty_inputs[input_idx..]
+            .iter()
+            .take_while(|d| {
+                let r = &d.original_range;
+                if r.start == r.end {
+                    r.start < w_end || (r.start == w_end && w_end == original_size)
```
rajatarya, line 205: The boundary case for pure inserts at exactly `w_end` is genuinely tricky — an insert at byte `w_end` belongs to *the next* window unless `w_end == original_size` (in which case there is no next window). The comment at L201-203 captures the why. Worth referencing `test_resize_insert_at_segment_boundary` and `test_mid_edit_plus_append` here so a future reader can find the exercising tests quickly. nit.

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1807 @@
...
+
+            cursor = edit_end;
+        }
+        input_idx = edits_end;
+
+        if cursor < w_end {
+            stream_cas_range(&ctx, &cas_client, original_hash, cursor, w_end, &mut cleaner).await?;
+        }
+
+        let (_info, chunks, mdb, _metrics) = cleaner.finish_with_chunks_detached().await?;
```
rajatarya, line 253: The `_detached` split is a real improvement — this resolves the GC concern I had on the prior pass. Window uploads now contribute their xorbs without registering an `MDBFileInfo` in the session shard, so only the final composed file shows up there. Combined with sirahd's tracking issue, the data lifecycle here is much cleaner than what I originally approved.

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1807 @@
...
+            dirty_inputs.len() - input_idx,
+            dirty_inputs.len()
+        )));
+    }
+
+    // Merge sequence: [gap0, w0, gap1, w1, ..., gapN]. Empty gaps (`None`) are skipped.
+    let mut hash_ranges = response.hash_ranges;
+    let trailing_gap = hash_ranges.pop().flatten();
+    // Leading gap is empty (None) => first window starts at byte 0.
+    let first_window_at_start = hash_ranges.first().is_some_and(Option::is_none);
```
rajatarya: nit: `hash_ranges.first().is_some_and(Option::is_none)` is concise but a bit cryptic — the reader has to remember `Vec<Option<T>>::first()` returns `Option<&Option<T>>` and reason about both layers. `matches!(hash_ranges.first(), Some(None))` reads more directly. Same below for `last_window_at_end` if you want symmetry.

xet_data/src/processing/range_upload.rs
```diff
@@ -0,0 +1,1807 @@
...
+                DataError::InternalError(format!(
+                    "ran out of gap_verification entries at stable segment {seg_idx}; \
+                     server response is inconsistent with the segment layout"
+                ))
+            })?;
+            all_verification.push(FileVerificationEntry::new(entry.into()));
+            gap_idx += 1;
+            seg_idx += 1;
+        }
+        let original_window_end = w.end.min(original_size);
```
rajatarya: Defensive question, not a fix: under the current server contract, can `w.end` ever exceed `original_size`? The server returns chunk-aligned ranges within the original file, and the snap logic on the client also caps at `original_size`. If `w.end > original_size` is structurally impossible, this `min` is dead defense and a `debug_assert!(w.end <= original_size)` would document the invariant more loudly. If it *can* happen (e.g. some forward-compat shape), worth a one-line comment about which case justifies it.

xet_client/src/cas_client/chunk_window_builder.rs
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    let (windows, hash_ranges) = builder.finish();
+    if windows.is_empty() {
+        return Err(ClientError::Other("dirty ranges do not overlap any chunks".into()));
+    }
+
+    // Emit one range hash per stable segment (= no overlap with any window).
+    // Segments and windows are both monotonic, so a two-pointer walk is O(S+W).
+    let gap_verification = if file_info.verification.len() == file_info.segments.len() {
```
rajatarya: The branch at L172 silently emits an empty `gap_verification` when `file_info.verification.len() != file_info.segments.len()`. Defensible for legacy shards that genuinely have no verification entries — but the failure mode for a *partially-populated* file_info would be a confusing runtime error in `compose_mdb` ("ran out of gap_verification entries at stable segment N") rather than something pointed at the actual cause. Two reasonable options:

- Tighten: emit empty only if `file_info.verification.is_empty()`, otherwise return `ClientError::Other(...)` here with a clearer message.
- Or document this branch's contract explicitly so the runtime error in `compose_mdb` is interpretable.

Not blocking — calling it out since the simulation-only path is what tests run against.

xet_client/src/cas_client/chunk_window_builder.rs
```diff
@@ -0,0 +1,205 @@
+//! Server-side state machine for `GET /v2/file-chunk-hashes/{file_id}` (mirrored from
+//! xetcas PR #987) plus a small driver helper used by simulation clients to produce a
+//! [`FileChunkHashesResponse`] without routing through HTTP.
```
rajatarya, line 3: The doc comment is honest about the duplication-with-xetcas#987 situation, which I appreciate. Worth filing a follow-up issue to track the eventual dedup (either xetcas pulling this in, or this getting moved to a shared crate) so it doesn't quietly become a permanent fork. Duplications like this almost always do without a forcing function.

xet_client/src/cas_types/mod.rs
```diff
@@ -311,6 +311,42 @@ pub struct QueryChunkResponse {
     pub shard: MerkleHash,
 }
 
+/// HTTP header carrying the dirty byte ranges to feed to `GET /v2/file-chunk-hashes/{file_id}`.
+///
+/// Distinct from the standard `Range` header (which scopes the response body): this header tags
+/// regions that the client intends to re-chunk, and the response covers the whole file (windows +
+/// gap subtrees). Value uses the same `bytes=A-B,C-D` syntax as `Range`.
+pub const X_RANGE_DIRTY_HEADER: &str = "x-range-dirty";
```
rajatarya: nit: `"x-range-dirty"` lowercase is unusual when most header constants in this crate (e.g. `SESSION_ID_HEADER` at L17 = `"X-Xet-Session-Id"`, `REQUEST_ID_HEADER` at L19 = `"X-Request-Id"`) use the `X-Foo-Bar` capitalized form. HTTP headers are case-insensitive on the wire so this is purely cosmetic, but matching the surrounding style would be a small consistency win.

rajatarya (commented):
## Re-review (post xetcas#987 merge)

This PR has come a long way since my approval on March 27. I went back and re-read the whole thing end-to-end against the new wire shape, and the rework is genuinely good — what shipped is *cleaner* than what I approved.

### What's new since last pass

1. **Wire shape pivot.** Composition is now driven by `windows[] + hash_ranges[]` of opaque `MerkleHashSubtree` summaries (xetcas#987), not flat per-chunk hashes. Per-chunk hashes never cross the wire, so the response stays O(windows + gaps) regardless of file size. Final hash is rebuilt with `MerkleHashSubtree::merge` — no `ChunkHashList`-glue layer, no duplicated tree-folding logic on the client.
2. **Resize edits.** `DirtyInput { original_range, reader, new_length }` collapses in-place edit / replace / pure-insert / pure-delete / append / truncate into a single shape. The `original_range.len() != new_length` branch falls out cleanly thanks to per-edit readers (no `ReadSeek` over a sparse staging file — that bug class is structurally gone).
3. **My GC concern is addressed at the client layer.** The new `register_single_file_clean_completion_detached` + `register_composed_file` split means window-uploads no longer pollute the session shard with orphan `MDBFileInfo` entries — only the final composed file is registered. Combined with sirahd's GC tracking issue (huggingface-internal/xet-garbage-collection#2), the design now does the right thing: window xorbs are findable through the composed file, and there are no dangling shard references.
4. **Defensive runtime invariants** (commit 87b90611): server-contract violations (`windows.len() != n_windows`, `hash_ranges.len() != n_windows + 1`, leftover edits not assigned to any window, `gap_verification` length mismatch with stable-segment count) are now hard runtime errors instead of silent corruption. Exactly the right call for a composition algorithm where `zip` would otherwise drop windows quietly.

### What `upload_ranges` offers, visually

**End-to-end flow:**

```mermaid
sequenceDiagram
    autonumber
    participant C as Caller (e.g. hf-mount)
    participant U as upload_ranges
    participant S as CAS server
    participant D as Cleaner / dedup

    C->>U: dirty_inputs[]<br/>(original_range, reader, new_length)
    Note over U: validate + coalesce<br/>snap to segment boundaries
    U->>S: GET /v2/file-chunk-hashes/{id}<br/>X-Range-Dirty: bytes=A-B,C-D
    S-->>U: windows[N] + hash_ranges[N+1]<br/>+ gap_verification[stable_segs]
    loop one cleaner per window
        U->>S: stream CAS prefix (boundary chunks)
        U->>D: feed prefix bytes
        U->>D: feed reader bytes (the edit)
        U->>S: stream CAS suffix
        U->>D: feed suffix bytes
        D-->>U: chunks + new MDBFileInfo (detached)
    end
    Note over U: merge_seq = [gap0, w0, gap1, ..., gapN]<br/>combined_hash = merge(seq).hmac(0)
    U->>S: register_composed_file(MDBFileInfo)
    U->>S: finalize session
    U-->>C: XetFileInfo { hash, size }
```

**Why this is a big deal — the wire savings:**

```mermaid
flowchart LR
    subgraph old ["Old: full re-upload"]
        OF["200MB original"] --> OU["download 200MB"]
        OU --> OE["edit 1MB"]
        OE --> OUP["upload 200MB"]
    end
    subgraph new ["upload_ranges: 2MB edit"]
        NF["200MB original"] --> NS["snap edits to<br/>segment boundaries"]
        NS --> NW["server returns<br/>2 windows + 3 gap subtrees"]
        NW --> NB["download ~few KB<br/>boundary chunks"]
        NB --> NU["upload ~2MB<br/>(2 dirty windows)"]
    end
    style old fill:#ffe5e5
    style new fill:#e5ffe5
```

**Hash composition without re-reading the file:**

```mermaid
flowchart TB
    F["Original file: chunks 0..M"] --> G0["gap0 subtree<br/>(server-stored)"] & G1["gap1 subtree<br/>(server-stored)"] & G2["gap2 subtree<br/>(server-stored)"]
    R["DirtyInput readers"] --> CL1["cleaner: window 0"] --> W0["w0 subtree<br/>(client-computed)"]
    R --> CL2["cleaner: window 1"] --> W1["w1 subtree<br/>(client-computed)"]
    G0 --> M{{"MerkleHashSubtree::merge<br/>[gap0, w0, gap1, w1, gap2]"}}
    W0 --> M
    G1 --> M
    W1 --> M
    G2 --> M
    M --> H["final_hash().hmac(0)<br/>== file_hash of edited file"]
```

**Edit-shape coverage (one struct, six operations):**

```mermaid
flowchart LR
    DI["DirtyInput<br/>{ original_range: a..b, new_length: n }"] --> Q1{"a == b?"}
    Q1 -->|yes| Q2{"n > 0?"}
    Q1 -->|no| Q3{"n == 0?"}
    Q2 -->|yes| INS["pure insert at a"]
    Q2 -->|no| NOOP["degenerate no-op"]
    Q3 -->|yes| DEL["pure delete a..b"]
    Q3 -->|no| Q4{"b - a == n?"}
    Q4 -->|yes| IPE["in-place edit"]
    Q4 -->|no| RES["resize replace"]
    INS -.->|special: a == original_size| APP["append"]
    DEL -.->|special: b == original_size| TRU["truncate to a"]
```

### Items to think about (none blocking)

1. **`chunk_window_builder.rs` duplication.** PR description flags this as intentional and temporary, mirroring xetcas#987 server logic so simulation clients answer `get_file_chunk_hashes` without HTTP. Worth filing a follow-up issue (or noting #717 in xetcas) to track the eventual dedup so the duplication doesn't quietly stick around as duplications always do without a forcing function.
2. **Empty `gap_verification` on partial-verification shards.** When `file_info.verification.len() != file_info.segments.len()`, the simulation builder silently emits an empty `gap_verification`, and then `compose_mdb` errors out at runtime with a less-targeted message. Inline below — either error up-front or document the silent path.
3. **One readability nit + one defensive question.** Inline.

Nothing here blocks the merge — most of these are conversation starters. Approving again on the strength of the current state. Nice work.

seanses (approved):

xet_client/src/cas_types/mod.rs
XciD: Done in 954ba6f4. Updated to `"X-Range-Dirty"` to match `SESSION_ID_HEADER` / `REQUEST_ID_HEADER` style. The wire is case-insensitive (and HTTP/2 lowercases on the wire regardless) so no server-side impact.

xet_client/src/cas_client/chunk_window_builder.rs
XciD: Tightened in 954ba6f4. The branch now distinguishes three cases explicitly: `verification.is_empty()` → empty `gap_verification` (legacy / test files), 1:1 match → walk segments, partial mismatch → `ClientError::Other` with a message that includes both counts. The comment block above documents the contract so the runtime expectation is loud at the source rather than deferred to `compose_mdb`.

xet_data/src/processing/range_upload.rs
XciD: It's structurally impossible: xetcas `core::get_file_chunk_hashes` clamps `dirty_byte_range.end` to `file_size` before returning windows. Replaced the `min` with a `debug_assert!` in 954ba6f4 and added a comment pointing at the server-side guarantee, so any future drift in the contract will fire loudly in tests.

xet_data/src/processing/range_upload.rs
XciD: Done in 954ba6f4. Switched to `matches!(hash_ranges.first(), Some(None))` — agree it reads more directly. Left `trailing_gap.is_none()` as-is since it doesn't have the same double-Option indirection.

xet_client/src/cas_client/chunk_window_builder.rs
XciD: Filed https://github.com/huggingface-internal/xetcas/issues/1059 on the xetcas side to track the eventual dedup. Listed three options (shared crate / xetcas depending on xet-core type / single source of truth) so it doesn't get forgotten.

xet_data/src/processing/range_upload.rs
XciD: Thanks — that was the intent. The composition has enough silent-truncation hazards (zip semantics, partial gap_verification consumption) that loud bails with expected-vs-actual counts felt much safer than trying to be clever about "recovering" from server contract drift.

xet_data/src/processing/range_upload.rs
XciD: Done in 2ebcfcc1 — added a one-line pointer to `test_resize_insert_at_segment_boundary` and `test_mid_edit_plus_append` right above the take_while.

xet_data/src/processing/range_upload.rs
XciD: Thanks for the second look. Splitting out `finish_with_chunks_detached` was the cleaner shape once it was clear the windows are scaffolding, not user-visible files — only the composed result should land in the session shard.

xet_data/src/processing/file_cleaner.rs
```diff
@@ -201,59 +202,82 @@ impl SingleFileCleaner {
...
+    /// session shard. Returns the MDBFileInfo directly so the caller can compose it into a
+    /// larger file without creating orphan shard entries.
+    pub async fn finish_with_chunks_detached(
+        self,
+    ) -> Result<(XetFileInfo, ChunkHashList, MDBFileInfo, DeduplicationMetrics)> {
+        Self::finish_inner(self, false).await
+    }
+
+    async fn finish_inner(
+        mut this: Self,
```
hoytak: Any reason `self` is changed to `this`

xet_data/src/processing/file_upload_session.rs
```diff
@@ -297,7 +298,8 @@ impl FileUploadSession {
             }
             cleaner.add_data(&buffer[..n]).await?;
         }
-        cleaner.finish().await
+        let (file_info, metrics) = cleaner.finish().await?;
+        Ok((file_info, metrics))
```
hoytak: Did the error type change here?  Confused why it made this change.

xet_data/src/processing/file_cleaner.rs
XciD: Done in 624ba7c4, no good reason, switched back to `mut self`.

xet_data/src/processing/file_upload_session.rs
XciD: Good catch, the destructure+rewrap was unnecessary noise. `cleaner.finish()` returns `(XetFileInfo, DeduplicationMetrics)` which matches the function signature, so I reverted it back to `cleaner.finish().await` (and same in the spawn block above) in 3fd592eb.

PR #718, opened 2026-03-16
Add optional Sha256 hash propegation through XetFileInfo object.

hoytak:
Currently, the SHA-256 hash of uploaded file content is computed internally during the upload pipeline but not surfaced to callers. Downstream consumers — e.g. OpenDAL's Hugging Face backend — need the SHA-256 to commit files to the Hub API.

This PR adds an optional sha256 field to XetFileInfo, the session-layer FileMetadata, and the Python-exposed PyXetUploadInfo. The field is populated from the already-computed hash when Sha256Policy::Compute or Sha256Policy::Provided is used, and left None for downloads and when Sha256Policy::Skip is used. Serde attributes (default, skip_serializing_if) ensure backward-compatible serialisation — existing serialised data without the field deserialises cleanly.

Needed for the functionality in https://github.com/huggingface/xet-core/pull/642.

### DIFF
--- a/xet_pkg/src/xet_session/download_group.rs
+++ b/xet_pkg/src/xet_session/download_group.rs
@@ -847,6 +847,7 @@ mod tests {
         Ok(XetFileInfo {
             hash: meta.hash.clone(),
             file_size: meta.file_size,
+            sha256: meta.sha256.clone(),
         })
     }
 
@@ -884,6 +885,7 @@ mod tests {
             XetFileInfo {
                 hash: meta.hash.clone(),
                 file_size: meta.file_size,
+                sha256: meta.sha256.clone(),
             }
         };
 


hf_xet/src/lib.rs
```diff
@@ -404,21 +404,27 @@ pub struct PyXetUploadInfo {
     pub hash: String,
     #[pyo3(get)]
     pub file_size: u64,
+    #[pyo3(get)]
```
seanses, line 407: Does hf-xet need this? I think huggingface_hub actually passes in the `sha256` to hf-xet, rather than getting it back from hf-xet: https://github.com/huggingface/huggingface_hub/pull/3876

xet_data/src/processing/xet_file.rs
```diff
@@ -20,7 +24,20 @@ impl XetFileInfo {
...
-        Self { hash, file_size }
+        Self {
+            hash,
+            file_size,
+            sha256: None,
+        }
+    }
+
+    /// Creates a new `XetFileInfo` instance with a SHA-256 hash.
+    pub fn with_sha256(hash: String, file_size: u64, sha256: String) -> Self {
```
seanses: Typically `with_xxx` is a function like `with_xxx(mut self, xxx) -> Self`. So maybe call it `new_with_sha256`.

hf_xet/src/lib.rs
hoytak: The buckets work needs this -- see the requirement in https://github.com/huggingface/xet-core/pull/642.

PR #719, opened 2026-03-16
Unify sync and async download/upload groups in session interface.

hoytak:
Currently, `UploadCommitSync` and `DownloadGroupSync` are thin wrappers around `UploadCommit` and `DownloadGroup` that delegate every method through `external_run_async_task`.  This means two types, two sets of doc comments, and two test suites covering the same underlying behavior.

This PR removes the separate sync types and adds `_blocking` suffixed methods directly on `UploadCommit` and `DownloadGroup`.  The session factory methods `new_upload_commit_blocking()` and `new_download_group_blocking()` now return the same types as their async counterparts, and the entire `xet_session::sync` module is deleted (~680 lines removed).

This also fixes a minor bug: `UploadCommitSync::upload_from_path` did not call `std::path::absolute()` on the file path before dispatching, unlike the async version.  The new `upload_from_path_blocking` includes the `std::path::absolute()` call, matching the async version's behavior.


xet_pkg/src/xet_session/session.rs
```diff
@@ -387,9 +386,13 @@ impl XetSession {
 
     /// Create a new [`UploadCommit`] from a **sync** (non-async) context.
     ///
+    /// The returned [`UploadCommit`] supports both async methods (`upload_from_path`,
```
seanses, line 389: I think this is a personal choice/preference: after an user creates a upload commit instance using `new_upload_commit_blocking` I want them only be able to use the blocking version APIs, and the thin wrapper `UploadCommitSync` does this by hiding the async version APIs behind. It feels a bit awkward to me that users can still call async APIs of UploadCommit after a blocking construction.

hoytak: The main reason is that we will need to pass back a streaming data object for streaming downloads, which has both a blocking and async interface on it; it seemed weird to have that returned but have a mixed interface there.

seanses (approved):
Communicated offline -- ACK to support streaming download.

api_changes/update_260316_unify_upload_download_sync_async.md
```diff
@@ -0,0 +1,123 @@
+# API Update: Unify UploadCommit and DownloadGroup async/sync types (2026-03-16)
```
rajatarya, line 1: Nice — appreciate the thorough migration guide with before/after examples. This makes the breaking change much easier for downstream consumers to adopt.

PR #720, opened 2026-03-16
Fix race condition in task status transitions on abort.

hoytak:
Currently, the Queued → Running status transition in spawned upload tasks is unconditional — it overwrites whatever the current status is, including Cancelled set by a concurrent abort() call. This creates a race window: if abort() sets Cancelled between the semaphore acquisition and the status write, the task overwrites it with Running, then the completion guard (if matches!(*s, TaskStatus::Running)) passes and sets Completed. The result is a task that was aborted but reports Completed.

This PR makes the Queued → Running transition conditional, matching the already-guarded Running → Completed/Failed transition. If the status is no longer Queued when the task starts, it bails early with SessionError::Aborted. This closes the race window — all three status transitions are now properly guarded against concurrent abort().

This was observed as a flaky test failure on Windows CI (test_abort_while_state_lock_held_skips_state_update_but_drains_tasks).

### DIFF
--- a/xet_pkg/src/xet_session/download_group.rs
+++ b/xet_pkg/src/xet_session/download_group.rs
@@ -209,10 +209,16 @@ impl DownloadGroupInner {
     ) -> JoinHandle<Result<XetFileInfo, SessionError>> {
         let semaphore = self.runtime().common().file_download_semaphore.clone();
         self.runtime().spawn(async move {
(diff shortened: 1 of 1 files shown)


seanses:
Great catch! Can you also update for `DownloadGroup::spawn_download_task`?

PR #743, opened 2026-03-20
fix: gate whoami to exclude iOS/tvOS/watchOS — fixes aarch64-apple-ios linker error

ArthurZucker:
## Problem

Building any crate that depends on `xet-core` for `aarch64-apple-ios` (or the iOS simulator) fails at link time:

```
Undefined symbols for architecture arm64:
  "_SCDynamicStoreCopyComputerName", referenced from:
      objc2_system_configuration in libapp.a (whoami)
ld: symbol(s) not found for architecture arm64
```

### Root cause

`whoami v2` declares `objc2-system-configuration` as a dependency for **all** `target_vendor = "apple"` targets — including iOS, tvOS, and watchOS. That crate wraps `SystemConfiguration.framework`, which exposes `SCDynamicStoreCopyComputerName`. The symbol **exists on macOS** but is **not available on iOS/tvOS/watchOS**, so the linker fails.

Dependency chain: `data` → `cas_client` → `file_utils` → `xet-runtime` → `whoami` → `objc2-system-configuration` → `SCDynamicStoreCopyComputerName` 💥

### Why the fix is safe

The only use of `whoami` in xet-runtime is in `privilege_context.rs::permission_warning()` — a diagnostic helper that prints a `chown` hint when a permission-denied error occurs. iOS, tvOS, and watchOS have no POSIX multi-user model, so:
1. `sudo chown` is meaningless on those platforms
2. The diagnostic message is purely informational, not load-bearing

## Fix

**`xet_runtime/Cargo.toml`** — narrow the `whoami` target cfg from `unix` to exclude Apple mobile targets:

```toml
# before
[target.'cfg(unix)'.dependencies]
whoami = { workspace = true }

# after
[target.'cfg(all(unix, not(target_os = "ios"), not(target_os = "tvos"), not(target_os = "watchos")))'.dependencies]
whoami = { workspace = true }
```

**`privilege_context.rs`** — mirror the cfg gate on the call site, with a simplified fallback message for mobile targets.

## Testing

```bash
# Must compile clean (no SCDynamicStoreCopyComputerName undefined symbol)
cargo build -p data --target aarch64-apple-ios
cargo build -p data --target aarch64-apple-ios-sim

# macOS behaviour unchanged
cargo test -p xet-runtime
```

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are limited to target-specific dependency gating and a diagnostic-only warning message, plus a new CI `cargo check` for `aarch64-apple-ios`. Main potential impact is missing the `chown` hint on Apple mobile targets, where it is not applicable.
> 
> **Overview**
> Fixes an iOS-family linker failure by tightening `xet-runtime`’s `whoami` dependency and call sites to **exclude** `ios`/`tvos`/`watchos` targets, avoiding the transitive `SystemConfiguration` symbol.
> 
> Adds a small regression test around `permission_warning()` and introduces a new CI job that runs `cargo check -p xet-runtime --target aarch64-apple-ios` on macOS runners to catch future target-specific build issues.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 912eb81ed76a7e8965304959a5a9e77ee5448805. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,6 +104,23 @@ jobs:
       - name: Check benchmarks compile
         run: |
           cargo bench --no-run
+  build-ios:
+    name: Check iOS target compiles (aarch64-apple-ios)
+    runs-on: macos-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+      - name: Install Rust 1.89 + iOS target
+        uses: dtolnay/rust-toolchain@1.89.0
+        with:
+          targets: aarch64-apple-ios
+      - uses: ./.github/actions/cache-rust-build
+      - name: Cargo check for aarch64-apple-ios
+        # `cargo check` resolves all symbols without needing an iOS SDK for linking.
+        # This is sufficient to catch the SCDynamicStoreCopyComputerName linker error
+        # that previously occurred when whoami pulled in objc2-system-configuration.
+        run: cargo check -p xet-runtime --target aarch64-apple-ios
+
   build_and_test-wasm:
     name: Build WASM
     runs-on: ubuntu-latest
--- a/xet_runtime/Cargo.toml
+++ b/xet_runtime/Cargo.toml
@@ -45,7 +45,7 @@ pyo3 = { workspace = true, optional = true }
 [target.'cfg(windows)'.dependencies]
 winapi = { workspace = true }
 
-[target.'cfg(unix)'.dependencies]
+[target.'cfg(all(unix, not(target_os = "ios"), not(target_os = "tvos"), not(target_os = "watchos")))'.dependencies]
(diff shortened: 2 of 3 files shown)


julien-c:
do you intend to download models from watchOS?

ArthurZucker:
👀 maybe maybe

ArthurZucker:
It was failing on ios!

rajatarya:
Following up here, can you retry this with the latest `hf-xet` crate (https://crates.io/crates/hf-xet)? We're now releasing an 'official' hf-xet crate for Rust. We discourage directly depending on xet-core repo.

PR #730, opened 2026-03-18
feat: expose CAS client factory and chunk cache re-exports

XciD:
## Context

These changes support the hf-mount project, which needs direct access to CAS client types.

## Summary

- Changed `create_remote_client` visibility from `pub(crate)` to `pub`
- Added re-exports for `CasClient`, `ChunkCache`, `CacheConfig`, and `get_cache` in `xet_data::processing`

---

> [!NOTE]
> **Low Risk**
> Low risk: this change only widens public API surface (function visibility and re-exports) without altering the underlying client creation logic or data handling paths.
> 
> **Overview**
> Makes `create_remote_client` publicly accessible (was `pub(crate)`) and re-exports it from `xet_data::processing`.
> 
> Also re-exports CAS client and chunk cache types/utilities (`CasClient`, `ChunkCache`, `CacheConfig`, `get_cache`) to provide a stable, convenient API surface for downstream projects.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b094f21cd05c287a1b7edc42496d4420296aed96. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #754, opened 2026-03-25
feat: make start_clean size parameter optional for streaming uploads

XciD:
## Summary
Make the `size` parameter of `FileUploadSession::start_clean()` optional by changing from `u64` to `Option<u64>`. This allows FUSE streaming uploads where the final file size is unknown, preventing debug assertion panics when `completed_bytes` exceeds the initially declared `total_bytes=0`.

## Changes
- `start_clean()` signature now takes `Option<u64>` for size
- All existing callers updated to wrap size argument in `Some(...)`
- Updated public API `upload_file()` to use `Option<u64>`
- Fixed test calls to use `Some()` wrapper

## Breaking Changes
- `FileUploadSession::start_clean()` signature changed from `size: u64` to `size: Option<u64>`

---

> [!NOTE]
> **Medium Risk**
> This is a breaking API change across upload/cleaning entry points (`start_clean`/`upload_file`) that affects progress tracking and completion accounting; incorrect `None` handling could cause misreported progress or assertions at finalize. Most call sites/tests are updated mechanically, lowering risk to runtime behavior beyond the new unknown-size path.
> 
> **Overview**
> Makes streaming upload size optional by changing `FileUploadSession::start_clean` (and `UploadCommit::upload_file` / `upload_file_blocking`) to take `Option<u64>` so callers can stream when the final size is unknown.
> 
> Updates all internal call sites, examples, and integration/unit tests to pass `Some(size)` where sizes are known, and wires the optional size through to the completion/progress tracker registration.
> 
> Includes a few small cleanup tweaks unrelated to the API change (minor error mapping simplification in python config bindings, test assertion style improvements, and replacing a float literal with `std::f32::consts::PI`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 16447dae52e28c7d234f49b3c932a4cd6410aa9a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/xet_data/src/processing/bin/example.rs
+++ b/xet_data/src/processing/bin/example.rs
@@ -90,7 +90,7 @@ async fn clean(mut reader: impl Read, mut writer: impl Write, size: u64) -> Resu
     let translator = FileUploadSession::new(TranslatorConfig::local_config(std::env::current_dir()?)?.into()).await?;
 
     let mut size_read = 0;
-    let (_id, mut handle) = translator.start_clean(None, size, Sha256Policy::Compute)?;
+    let (_id, mut handle) = translator.start_clean(None, Some(size), Sha256Policy::Compute)?;
 
     loop {
         let bytes = reader.read(&mut read_buf)?;
--- a/xet_data/src/processing/data_client.rs
+++ b/xet_data/src/processing/data_client.rs
@@ -44,7 +44,7 @@ pub async fn clean_bytes(
     bytes: Vec<u8>,
     sha256_policy: Sha256Policy,
 ) -> Result<(XetFileInfo, DeduplicationMetrics)> {
-    let (_id, mut handle) = processor.start_clean(None, bytes.len() as u64, sha256_policy)?;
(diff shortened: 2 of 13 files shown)


rajatarya (approved):
Solid change. `Option<u64>` is the right type for "size may not be known" — fixes the debug assertion panic on FUSE streaming uploads cleanly.

All callers updated consistently, docs updated, and the clippy bonus fixes (`assert!(false, ...)` → `panic!()`, removing `assert!(true)`, closure → function ref, `3.14f32` → `PI`) are all welcome.

One note: `upload_files` (the batch API at `file_upload_session.rs:114`) still calls `register_new_file(updater, Some(file_size))` directly without going through `start_clean`. Not a bug since it reads actual size from `fs::metadata`, but worth keeping in mind if this path ever gets refactored — there are now two separate registration paths.

Since this is a breaking change to `upload_file()` on `UploadCommit`, worth coordinating merge order with #730 (hf-mount will be a consumer of this API).

seanses (commented):
I think the current main already has this?

PR #759, opened 2026-03-27
Update xtool to add GC simulation functionality 

hoytak:
This PR adds new functionality to the xtool binary for directly uploading, downloading, and inspecting files against a CAS endpoint or remote repo — useful for development, debugging, and scripting without going through git-xet or huggingface_hub.

The binary adds `xtool file` and exposes four new subcommands:

upload — upload one or more files (or stdin) and emit file metadata (hash, size, sha256).
download — download by xet hash to a file or stdout, with optional source and write byte ranges.
scan — dry-run dedup/compression analysis without uploading data.
dump-reconstruction — fetch and display reconstruction metadata as JSON.

Endpoint resolution currently follows the same conventions as session API, but using arguments passed in.  --endpoint overrides HF_ENDPOINT, which defaults to https://huggingface.co; endpoint can also be a local directory in which case a LocalClient is used.   Token resolution uses --token then HF_TOKEN. 

All config values can be overridden with -c KEY=VALUE.

```
# Upload a single file and see its hash/size/sha256
xtool file upload mydata.bin

# Upload multiple files at once
xtool file upload file1.bin file2.bin file3.bin

# Upload from stdin
cat model.safetensors | xtool file upload -

# Upload and write results to JSON
xtool file upload --output results.json *.bin

# Download a file by hash
xtool file download abc123...def -o restored.bin 

# Download a byte range to stdout
xtool file download abc123...def --source-range 0..4096

# Download a range and write it into a specific offset of an existing file
xtool file download abc123...def -o target.bin --source-range 1024..2048 --write-range 0..1024

# Dry-run scan to check dedup/compression ratio without uploading
xtool file scan --recursive ./dataset/

# Inspect reconstruction metadata for a file hash
xtool file dump-reconstruction abc123...def

# Use a local CAS directory instead of a remote endpoint
xtool --endpoint /tmp/local-cas file upload data.bin
```

---

> [!NOTE]
> **Medium Risk**
> Large new CLI surface touches CAS upload/download, Hub JWT auth, and endpoint policy; mistakes could affect local dev workflows or auth handling, but remote upload is explicitly guarded.
> 
> **Overview**
> **Moves and restructures the `xtool` developer CLI** from `xet_data` into `hf-xet` (`xet_pkg`) behind the `internal-tools` feature, with a modular command tree and broad test coverage.
> 
> The CLI now exposes **`file upload` / `file download`**, top-level **`dedup`** (dry-run or `--migrate`), and **`query`** (reconstruction JSON). Shared **`EndpointConfig`** resolves either **direct CAS** (`--endpoint`, `local://` paths, tokens) or **Hub mode** (`--repo-type` / `--repo-id` → CAS JWT + token refresh). Uploads are **restricted to local or loopback CAS** and **blocked in Hub mode** so data is not committed outside Hub APIs.
> 
> **`file download`** adds optional **`--source-range`**, **`--write-range`**, and **`--size`**, including bounded fetches when only a write window is specified. **`dedup`** supports recursive directory walks and a direct-CAS path when not using Hub.
> 
> **Wasm builds** no longer pull in file logging (`tracing-appender` gated; file logging returns unsupported on wasm). Lockfiles pick up minor dependency bumps. The old monolithic `xet_data` `xtool` binary is removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7de4fe1dbc0d926ff3a7cca0079d4b85a961d5d0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rajatarya:
Won't these changes make garbage-collection harder @sirahd since xorbs/shards aren't known to Xet backend?

I understand the benefits (just read the description) maybe the right approach is that this is only allowed for localhost CAS?

hoytak:
Gotcha, I'm open to that adjustment and to not allowing uploads in this way.  Essentially, the problem would be that all files uploaded this way would be ephemeral, i.e. they would get garbage collected quickly, so another option would be to simply warn that this is the case.  

For now, though, I can make uploads error out to remote CAS.

hoytak:
Updated this PR to disallow uploads to huggingface endpoints; downloads and queries still work as expected.

xet_pkg/src/bin/xtool/main.rs
```diff
@@ -0,0 +1,283 @@
...
+}
+
+#[derive(Subcommand)]
+pub enum FileCommands {
+    /// Upload one or more files (or stdin) to the CAS endpoint.
+    Upload(UploadArgs),
+    /// Download a file by its xet hash.
+    Download(DownloadArgs),
+    /// Dry-run dedup and compression analysis (no upload).
+    Scan(ScanArgs),
```
seanses: It seems this is a duplicate of the "dedup" command above.

xet_pkg/src/bin/xtool/main.rs
```diff
@@ -0,0 +1,283 @@
...
+    /// Show reconstruction metadata for a file hash (JSON).
+    DumpReconstruction(DumpReconstructionArgs),
```
seanses: And this is a duplicate of the "query" command above

xet_pkg/src/bin/xtool/upload.rs
```diff
@@ -0,0 +1,249 @@
...
+    let session = super::session::build_xet_session(ctx)?;
+    let results =
+        run_upload(&session, &ep.cas_endpoint, args, ep.token_info(), ep.token_refresh(Operation::Upload)?).await?;
+
+    if let Some(ref output_path) = args.output {
+        let json = serde_json::to_string_pretty(&results)?;
+        std::fs::write(output_path, json)?;
+    } else if !cli.quiet {
+        for meta in &results {
+            eprintln!(
```
rajatarya: Stdout vs stderr policy worth nailing down at the tool level. Cursor flagged `hub_query.rs:33` for routing structured output to stderr; here `upload` writes its human-readable per-file summary to stderr; `stats.rs:36-39` writes to stdout. A reader trying to script around `xtool` won't have a consistent rule.

A reasonable convention for a dev tool like this:
- `download` to stdout when `-o` is omitted → file payload goes to stdout (already correct)
- `--output FILE` → JSON results to that file, nothing to stdout
- Otherwise → human-readable summary to **stdout** (so `xtool file upload mydata.bin | jq '...'` works on the JSON-y form), status/progress to stderr

If you go that route, this `eprintln!` would become a `println!`, and `download.rs:136` would stay on stderr (status line). What do you think?

xet_pkg/src/bin/xtool/upload.rs
```diff
@@ -0,0 +1,249 @@
...
+                output.push(meta);
+            },
+            Err(e) => {
+                eprintln!("ERROR: {name}: {e}");
+                had_error = true;
+            },
+        }
+    }
+
+    if had_error {
```
rajatarya, line 125: What does `commit.abort()` do to xorbs that successful files (the N-1 that finalized before file N failed) have already pushed up? If they're left in CAS, then a user who uploads 100 files, sees 99 success lines, hits one error, gets a non-zero exit, and has no obvious cleanup story for the partial state. If `abort` deletes them, then the user might be surprised that their successful uploads were rolled back.

Either answer is fine, but it's worth being explicit in the docs / `--help`. A `--continue-on-error` flag (or `--all-or-nothing` as the explicit opt-in) would also let scripts choose the semantics they want.

rajatarya (commented):
Big +1 on bringing xtool into xet_pkg with first-class direct/hub mode resolution and a real test surface (upload/download/scan/query roundtrips backed by a `LocalClient`). Tests exercising both single-file and multi-file paths against `local://` end up doing real value as smoke tests for the session/commit machinery. The `EndpointConfig::resolve` split (direct vs Hub-JWT) reads cleanly and the `operation_for_command` mapping is the right shape.

@seanses already caught the top-level `Dedup` / `Query` vs `file scan` / `file dump-reconstruction` duplication on `main.rs:91`/`:93` — that's the biggest UX wart and worth resolving before merge. My instinct would be to keep the top-level forms and drop the `file` aliases (since `xtool dedup` and `xtool query` are how the existing scripting probably calls in), but either direction is fine as long as one is picked.

A few items to consider beyond what cursor + @seanses flagged:

1. **Stdout vs stderr policy is inconsistent** across commands — `stats` uses `println!`, `upload`/`download` use `eprintln!`, `hub_query` was flagged by cursor for `eprintln!` debug output. Worth nailing down a one-paragraph rule (e.g. "binary file payload to stdout, structured JSON via `--output`, status/info to stderr") and applying it uniformly. See inline on `upload.rs:43`.

2. **`commit.abort()` semantics on partial upload failure** — when one of N uploads fails, the current path aborts the commit but the user has already seen success messages for the other N-1. Are those uploaded xorbs cleaned up by `abort`, or does the user end up with orphan content in CAS? See inline on `upload.rs:117`.

3. **`--write-range` without `--source-range` still streams the full file** from CAS and discards the bytes outside the window. For a 100GB file with `--write-range 0..1024`, that's 100GB of network for 1KB of writes. See inline on `download.rs:108`.

Once the duplicate-command structure is decided + the high-severity cursor findings (global `-c` collision on `dedup.rs:44`, scan requesting upload JWT on `main.rs:159`, upload guard not blocking direct-mode remote on `main.rs:211`) are addressed, I think this is in good shape.

xet_pkg/src/bin/xtool/download.rs
```diff
@@ -0,0 +1,246 @@
...
+                    .write(true)
+                    .create(true)
+                    .truncate(false)
+                    .open(output_path)
+                    .with_context(|| format!("failed to open {}", output_path.display()))?
+            } else {
+                File::create(output_path).with_context(|| format!("failed to create {}", output_path.display()))?
+            };
+
+            if let Some(ref wr) = write_range {
```
rajatarya, line 116: When `--write-range 0..1024` is set without `--source-range`, the stream still pulls every byte from the source and we just discard everything past `write_limit`. For a 100 GB file with a 1 KB write window, that's a lot of wasted bandwidth.

Would it make sense to default `source_range = Some(write_range)` when `--source-range` isn't given? Or at minimum, document that callers who want a bounded fetch need to pass both. (The `--write-range` doc says "Seeks to the start offset before writing" which implies it's purely a sink-side concern, but the bandwidth implication is the part that bites.)

api_changes/update_260423_xtool_cli_in_hf_xet.md
```diff
@@ -0,0 +1,50 @@
+# API Update: `xtool` moved to `hf-xet` and expanded command surface (2026-04-23)
```
seanses: We can omit this file entirely. This is only an internal tool and not even an API change.

seanses (approved):
Looks good!

PR #792, opened 2026-04-09
Expose XetSession APIs to Python

seanses:
## Summary

Replaces the old `upload_files` / `download_files` / `hash_files` Python functions with a new object-oriented API that exposes `XetSession` and its child objects directly as PyO3 classes. This gives Python callers full control over session lifecycle, connection pooling, and progress reporting.

The previous module-level functions are kept under `hf_xet/src/legacy/` and remain importable as `from hf_xet import upload_files` etc., but now emit `DeprecationWarning`.

## New Python API

```python
import hf_xet

# Optional: create a custom config (immutable; use .with_config() to derive updates)
config = hf_xet.XetConfig().with_config("data.max_concurrent_file_ingestion", 8)

# Create session; config is optional (defaults to XetConfig() with HF_XET_* env overrides)
session = hf_xet.XetSession(config=config)

# Upload — multiple files, bytes, and streaming within one commit
with session.new_upload_commit(
        endpoint="https://cas.xethub.hf.co",
        token="jwt", token_expiry_unix_secs=9999999999,
        token_refresh_url="https://…/xet-write-token/main",
        token_refresh_headers={"Authorization": "Bearer hf_…"},
    ) as commit:
    h1 = commit.start_upload_file("/path/to/model.bin")
    h2 = commit.start_upload_file("/path/to/tokenizer.json", sha256="f2358d9a…")
    h3 = commit.start_upload_bytes(b"...", name="config.json")

    stream = commit.start_upload_stream(name="big.bin")
    for chunk in produce_chunks():
        stream.write(chunk)
    stream.finish()  # must be called before the with-block exits
# on normal exit: wait_to_finish() is called automatically
# on exception:   abort() is called automatically

# SHA-256 sentinels
commit.start_upload_file("/path/to/model.bin", sha256=hf_xet.COMPUTE_SHA256)  # default
commit.start_upload_file("/path/to/model.bin", sha256=hf_xet.SKIP_SHA256)     # skip

# Progress callback — receives (GroupProgressReport, dict[UniqueID, ItemProgressReport])
def on_progress(group, items):
    bar.n = group.total_bytes_completed
    bar.refresh()

with session.new_upload_commit(
        token_refresh_url="https://…/xet-write-token/main",
        token_refresh_headers={"Authorization": "Bearer hf_…"},
        progress_callback=on_progress,
        progress_interval_ms=100,
    ) as commit:
    commit.start_upload_file("/path/to/model.bin")

# File download — multiple files within one group (downloads run concurrently)
file_info_a = hf_xet.XetFileInfo(hash_a, size_a)
file_info_b = hf_xet.XetFileInfo(hash_b, size_b)
with session.new_file_download_group(
        token_refresh_url="https://…/xet-read-token/main",
        token_refresh_headers={"Authorization": "Bearer hf_…"},
    ) as group:
    group.start_download_file(file_info_a, dest_path_a)
    group.start_download_file(file_info_b, dest_path_b)

# Streaming download — start/end are both optional (open-ended ranges supported)
group = session.new_download_stream_group(
    token_refresh_url="https://…/xet-read-token/main",
    token_refresh_headers={"Authorization": "Bearer hf_…"},
)
for chunk in group.download_stream(file_info):                    # whole file
    f.write(chunk)
for chunk in group.download_stream(file_info, start=1024):        # 1024 .. EOF
    f.write(chunk)
for chunk in group.download_stream(file_info, start=0, end=4096): # 0 .. 4096
    f.write(chunk)
for offset, chunk in group.download_unordered_stream(file_info):
    buf[offset:offset+len(chunk)] = chunk

# Ctrl-C handling
try:
    with session.new_upload_commit(...) as commit:
        commit.start_upload_file("model.bin")
except KeyboardInterrupt:
    session.sigint_abort()
    raise
```

### New files
- `hf_xet/src/py_xet_session.rs` — `XetSession` PyO3 class
- `hf_xet/src/py_upload_commit.rs` — `XetUploadCommit`, SHA-256 sentinels (`COMPUTE_SHA256`, `SKIP_SHA256`), report types
- `hf_xet/src/py_file_upload_handle.rs` — `XetFileUpload`
- `hf_xet/src/py_stream_upload_handle.rs` — `XetStreamUpload`
- `hf_xet/src/py_file_download_group.rs` — `XetFileDownloadGroup`
- `hf_xet/src/py_file_download_handle.rs` — `XetFileDownload`
- `hf_xet/src/py_download_stream_group.rs` — `XetDownloadStreamGroup`
- `hf_xet/src/py_download_stream_handle.rs` — `XetDownloadStream`, `XetUnorderedDownloadStream`
- `hf_xet/src/config.rs` — `XetConfig` Python class (immutable, `with_config()`, `get()`, `items()`, `keys()`, `__getitem__`)
- `hf_xet/src/headers.rs` — `build_headers_with_user_agent` helper
- `hf_xet/src/legacy/mod.rs` — re-exports all legacy symbols
- `hf_xet/src/legacy/types.rs` — `PyXetDownloadInfo`, `PyXetUploadInfo`, `PyPointerFile`
- `hf_xet/src/legacy/functions.rs` — deprecated `upload_bytes`, `upload_files`, `download_files`, `force_sigint_shutdown`; `hash_files` retained without deprecation
- `hf_xet/src/legacy/progress_update.rs` — `PyItemProgressUpdate`, `PyTotalProgressUpdate`, `WrappedProgressUpdater`
- `hf_xet/src/legacy/runtime.rs` — async runtime + SIGINT handler (used by legacy functions)
- `hf_xet/src/legacy/token_refresh.rs` — Python callback token refresher (used by legacy functions)
- `hf_xet/tests/conftest.py` — shared fixtures and upload helpers
- `hf_xet/tests/test_upload_commit.py` — upload tests (file, bytes, stream, SHA-256 policy, progress, abort)
- `hf_xet/tests/test_file_download.py` — file download tests (handles, round-trips, progress, cancel)
- `hf_xet/tests/test_stream_download.py` — ordered and unordered streaming download tests with range variants
- `hf_xet/tests/test_progress.py` — progress callback argument types and field verification
- `hf_xet/tests/test_session.py` — `XetSession` lifecycle and group/commit creation tests
- `hf_xet/tests/test_config.py` — `XetConfig` construction, `with_config`, `get`, `items`, `keys`

### Modified files
- `hf_xet/src/lib.rs` — module declarations; `XetTaskState` Python enum; `blocking_call_with_signal_check` utility; legacy module registered at top level for backward compatibility
- `hf_xet/src/logging.rs` — calls `xet_pkg::init_logging()` instead of `xet_runtime` directly
- `hf_xet/Cargo.toml` — added `xet-runtime`, `xet-client` deps (for legacy module); feature flags route through `xet-pkg`
- `xet_pkg/src/xet_session/file_download_group.rs` — exposes `XetDownloadGroupReport` as a Python class (`pyclass(get_all)`, `__repr__`)
- `xet_data/src/processing/xet_file.rs` — `#[new]` Python constructor for `XetFileInfo`
- `xet_pkg/Cargo.toml` — added `no-default-cache`, `tokio-console`, `elevated_information_level` features
- `xet_pkg/src/lib.rs` — added `init_logging()` wrapper
- `xet_runtime/src/core/runtime.rs` — fork-safe `Drop`: detect child process via stored PID, discard runtime instead of blocking shutdown
- `.github/workflows/ci.yml` — added Python integration test step (maturin + pytest) to Linux, Windows, macOS jobs

## Test Plan

- [x] Rust unit tests: `cargo test --verbose --no-fail-fast` in `hf_xet/`
- [x] Python integration tests: `maturin develop && pytest hf_xet/tests/ -v`
- [x] Verified legacy APIs work with huggingface_hub: [Test huggingface_hub xet tests / hub-python-tests (pull_request)](https://github.com/huggingface/xet-core/actions/runs/24497085281/job/71594463047?pr=792), also see the deprecation messages
- [x] End-to-end test with `huggingface_hub` upload/download flows against staging: "Xet" test in https://github.com/huggingface/huggingface_hub/pull/4116; manually tested progress update and Ctrl-C handling

## Design Notes

- **API style**: groups and commits are created with keyword arguments directly on the factory method (`session.new_upload_commit(endpoint=..., token_refresh_url=..., progress_callback=...)`) rather than a builder chain. The Rust builder is used internally and never surfaces in the Python API.

- **Token refresh**: the old API required Python to pass a token-refresh callable that Rust invoked across the GIL boundary. The new API uses `token_refresh_url` + `token_refresh_headers` — Rust refreshes autonomously via HTTP, removing GIL re-entry on the hot path. `WrappedTokenRefresher` is kept only in `legacy/`.

- **SHA-256 policy**: `start_upload_file`, `start_upload_bytes`, and `start_upload_stream` accept a `sha256` argument that is either a pre-computed hex string, `COMPUTE_SHA256` (default), or `SKIP_SHA256`. Sentinel objects have descriptive `__repr__` for debugging.

- **XetConfig**: immutable Python class wrapping `XetConfig`. `with_config(name, value)` or `with_config({...})` returns a new config with updates applied. Supports dotted-path access (`config["data.max_concurrent_file_ingestion"]`, `config.get("data.max_concurrent_file_ingestion")`), `items()`, `keys()`, and iteration.

- **XetTaskState**: a Python class with variants `Running`, `Finalizing`, `Completed`, `UserCancelled`. The `Error` variant of the internal Rust enum surfaces as a raised Python exception rather than an enum value (PyO3 0.26 does not support mixed unit/complex enum variants in `#[pyclass]`).

- **Progress callbacks**: `progress_callback` spawns a background thread that delivers `(GroupProgressReport, dict[UniqueID, ItemProgressReport])` to the Python callable every `progress_interval_ms` milliseconds (default 100). The same signature covers both upload and download groups.

- **GIL release and Ctrl-C**: queue operations (`start_upload_file`, `start_upload_bytes`, `start_download_file`) use `py.detach()` and return quickly. Long-wait operations (`wait_to_finish()`) run the blocking call on a background thread while the calling thread releases the GIL for 100 ms windows and polls `py.check_signals()` — Ctrl-C raises `KeyboardInterrupt` within one interval without starving other Python threads. `XetError::KeyboardInterrupt` maps to `PyKeyboardInterrupt`. The recommended caller pattern is `except KeyboardInterrupt: session.sigint_abort(); raise`, which is idiomatic Python: `sigint_abort()` flags the runtime so the background thread exits cleanly at its next checkpoint.

- **Context managers and concurrency**: `XetUploadCommit` and `XetFileDownloadGroup` implement `__enter__`/`__exit__`; `__exit__` delegates to `wait_to_finish()` on success and `abort()` on exception. Multiple `start_upload_file` / `start_download_file` calls within a `with` block run concurrently; the block exit waits for all to complete. `XetDownloadStreamGroup` is not a context manager — streams are opened with `download_stream()` / `download_unordered_stream()` and iterated directly.

- **Streaming uploads**: `commit.start_upload_stream()` returns a `XetStreamUpload` handle for incremental writes (`.write(bytes)`, then `.finish()` before `wait_to_finish()`).

- **Streaming downloads**: `download_stream` and `download_unordered_stream` accept optional `start` / `end` byte offsets; either may be omitted independently. `download_unordered_stream` yields `(offset, bytes)` tuples in completion order.

- **Fork-safe runtime drop**: `XetRuntime` records its creating PID; if `drop` fires in a child process after `fork`, the parent's Tokio threads don't exist so `shutdown_timeout()` would block. The runtime is discarded via `mem::forget` instead.

- **Backward compatibility**: all pre-1.x functions (`upload_bytes`, `upload_files`, `hash_files`, `download_files`, `force_sigint_shutdown`) and types (`PyXetDownloadInfo`, `PyXetUploadInfo`, `PyPointerFile`, `PyItemProgressUpdate`, `PyTotalProgressUpdate`) remain importable from the top-level `hf_xet` module. Deprecated functions emit `DeprecationWarning` at `stacklevel=2`. `hash_files` is not deprecated.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it reworks the Python-facing API surface and adds new concurrency/progress-callback behavior around uploads/downloads, which could introduce behavioral regressions despite extensive tests.
> 
> **Overview**
> Adds a new object-oriented Python API around `XetSession` (plus upload commits, file-download groups, and ordered/unordered download streams), including progress callbacks, SHA-256 policy sentinels, and signal-friendly blocking waits.
> 
> Moves the pre-1.x module-level functions/types into a `legacy` module, keeps them importable for compatibility, and makes the main entry points emit `DeprecationWarning` while sharing header/User-Agent handling and centralized error conversion.
> 
> Extends `xet_pkg`/`xet_data` with Python-facing `pyclass` report/value types and minor API tweaks (e.g., optional stream progress, cloneable handles, active task info), bumps `hf_xet` to `1.5.0`, and updates CI to build the extension via `maturin` and run new pytest integration tests on Linux/Windows/macOS.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc478ab801e3c43dc65a015c7c396e7635fd0105. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hf_xet/src/py_upload_commit.rs
```diff
@@ -0,0 +1,515 @@
...
+    ///     stream.write(chunk)
+    /// result = stream.finish()   # must be called before commit()
+    /// print(result.xet_info.hash, result.xet_info.file_size)
+    /// ```
+    #[pyo3(signature = (name=None, sha256=None))]
+    pub fn upload_stream(&self, name: Option<String>, sha256: Option<PySha256Policy>) -> PyResult<PyXetStreamUpload> {
+        let policy = sha256.map(|p| p.inner).unwrap_or(Sha256Policy::Compute);
+        let handle = self.inner.upload_stream_blocking(name, policy).map_err(convert_xet_error)?;
+        Ok(PyXetStreamUpload { inner: handle })
+    }
```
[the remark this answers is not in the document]
seanses: This is intentional

.github/workflows/ci.yml
```diff
@@ -68,6 +68,22 @@ jobs:
...
+        uses: PyO3/maturin-action@04ac600d27cdf7a9a280dadf7147097c42b757ad  # v1
+        with:
+          command: develop
+          sccache: 'true'
+          working-directory: hf_xet
+      - name: Python integration tests (hf_xet)
+        run: |
+          source .venv/bin/activate
+          pip install pytest
+          pytest hf_xet/tests/ -v
```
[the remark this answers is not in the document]
seanses: This is incorrect.

hoytak (approved):
Largely looks good to me, especially as a first pass.  I think there are some things that we could tighten up: 

- I think we should drop the builder pattern as python supports optional keyword parameters, so the API for a creating things could just replace the current with_XXX calls in the builder struct with optional keyword arguments to the main constructor.  E.g.  `session.new_download_group().with_endpoint("my endpoint").build()` becomes  `session.new_download_group(endpoint = "my endpoint")` instead, with all the arguments being optional. 

- The __repr__ methods are really minimal, especially with download and progress reporting; normally __str__ is meant to be the quick readable version and __repr__ is meant to give a detailed and more complete representation.   I think we should have both, filling in more fields in the __repr__ stuff. 

- The process changing on PID stuff can be tightened up.  Now that it's session based, I think the best way to do this is just raise a RuntimeError inside all of the XetRuntime bridge methods if the pid changed.   This can be handled just in XetRuntime and that should be sufficient for all our needs.

hf_xet/src/py_upload_commit.rs
```diff
@@ -0,0 +1,515 @@
...
+    /// stream = commit.upload_stream(name="model.bin")
+    /// for chunk in produce_chunks():
+    ///     stream.write(chunk)
+    /// result = stream.finish()   # must be called before commit()
+    /// print(result.xet_info.hash, result.xet_info.file_size)
+    /// ```
+    #[pyo3(signature = (name=None, sha256=None))]
+    pub fn upload_stream(&self, name: Option<String>, sha256: Option<PySha256Policy>) -> PyResult<PyXetStreamUpload> {
+        let policy = sha256.map(|p| p.inner).unwrap_or(Sha256Policy::Compute);
+        let handle = self.inner.upload_stream_blocking(name, policy).map_err(convert_xet_error)?;
```
rajatarya: `upload_file` (L272) and `upload_bytes` (L296) both wrap `*_blocking` in `py.detach()` to release the GIL for the queue handshake. `upload_stream` here doesn't — `self.inner.upload_stream_blocking(...)` runs while holding the GIL. If there's any blocking work inside (e.g. waiting on a semaphore / acquiring a slot), other Python threads stall and the progress thread can't fire its callback. Either mirror the `py.detach()` pattern here or leave a comment explaining why streaming setup doesn't need it.

hf_xet/src/py_upload_commit.rs
```diff
@@ -0,0 +1,515 @@
...
+                    std::thread::sleep(interval);
+                    let group_report = inner.progress();
+                    let item_reports: HashMap<UniqueID, ItemProgressReport> = handles_for_thread
+                        .read()
+                        .map(|g| g.iter().filter_map(|h| h.progress().map(|p| (h.task_id(), p))).collect())
+                        .unwrap_or_default();
+                    let is_terminal =
+                        !matches!(inner.status(), Ok(XetTaskState::Running) | Ok(XetTaskState::Finalizing));
+                    let result = Python::attach(|py| callback.call1(py, (group_report, item_reports)));
+                    if result.is_err() || is_terminal {
```
rajatarya: Once `is_terminal` flips true, the loop `break`s after the `call1` for *this* tick — but that tick saw the snapshot from *before* the terminal transition. A tqdm-style bar (literally the example in the `with_progress_callback` docstring) ends up showing 99.x% forever. Would be worth one final `callback.call1(py, (inner.progress(), ...))` after the break, so callers always get a 100%-complete signal at the end. Same pattern in `py_file_download_group.rs:178`.

hf_xet/src/py_upload_commit.rs
```diff
@@ -0,0 +1,515 @@
...
+        exc_type: Bound<'_, pyo3::PyAny>,
+        _exc_val: Bound<'_, pyo3::PyAny>,
+        _exc_tb: Bound<'_, pyo3::PyAny>,
+    ) -> PyResult<bool> {
+        if exc_type.is_none() {
+            // Normal exit: commit all uploads (signal-interruptible).
+            self.commit(py)?;
+        } else {
+            // Exception: cancel uploads.
+            let _ = self.inner.abort();
```
rajatarya: `let _ = self.inner.abort()` silently discards abort failures on the exception path. If abort itself errors (say, because the runtime is already shutting down), the original exception propagates fine, but we lose any trace of the secondary failure. A `tracing::warn!` here (and at `py_file_download_group.rs:237`) would make post-mortems easier without changing user-visible behaviour.

hf_xet/src/py_download_stream_handle.rs
```diff
@@ -0,0 +1,121 @@
...
+        slf
+    }
+
+    /// Return the next ``bytes`` chunk, or raise ``StopIteration`` when done.
+    ///
+    /// Note: the GIL is held while waiting for the next chunk.
+    /// ``XetDownloadStream`` is not ``Clone``, so ``py.detach()`` cannot be
+    /// used here.  In practice chunks arrive quickly from the background task,
+    /// so this is not expected to cause significant contention.
+    fn __next__(&mut self, py: Python<'_>) -> PyResult<Option<Py<PyBytes>>> {
```
rajatarya, line 52: The GIL-held `__next__` is the one spot in the new API where behaviour will feel un-Pythonic on a slow network. Two concrete issues beyond contention:

1. If another Python thread is iterating this stream, the progress-callback thread on the *same* session can't acquire the GIL to fire its callback, so progress freezes during streaming.
2. Ctrl-C won't be observed until the next chunk lands — a stalled TCP read becomes a completely un-interruptible Python call.

The blocker per the doc comment is that `XetDownloadStream` isn't `Clone`. Is making it clonable (wrapping the inner in `Arc`) on the table in `xet_pkg`? That'd let this mirror the `blocking_call_with_signal_check` pattern used for `commit()` / `finish()` and fix both issues at once. Same concern at `__next__` on `PyXetUnorderedDownloadStream` below.

hf_xet/src/py_xet_session.rs
```diff
@@ -0,0 +1,117 @@
...
+        Ok(PyXetDownloadStreamGroupBuilder {
+            inner: Some(builder),
+            custom_headers: None,
+        })
+    }
+
+    /// Current task state: ``"Running"``, ``"Finalizing"``, ``"Completed"``, or
+    /// ``"UserCancelled"``.  Raises on error state.
+    pub fn status(&self) -> PyResult<&'static str> {
+        task_state_to_str(self.inner.status().map_err(convert_xet_error)?)
```
rajatarya: Returning `&'static str` here (and `status()` on every handle / commit / group) makes the state part of the public API stringly-typed. Two things this costs us:

- From Python, `if session.status() == "Running"` has no autocomplete, typos fail silently, and IDE refactors can't follow it.
- The `Error` variant's message is folded into an exception, so callers that want to *inspect* state without raising can't see error states at all — `status()` becomes `status_or_raise()`.

Consider either a PyO3 enum (`XetTaskState.RUNNING`, etc.) or at least a frozen `str` constant set on the module. Flagging this explicitly as a public-API concern rather than a nit.

hf_xet/src/py_xet_session.rs
```diff
@@ -0,0 +1,117 @@
...
+    /// Cancel all active operations on this session.
+    ///
+    /// The session remains usable after abort — new commits and groups can be
+    /// created.
+    pub fn abort(&self) -> PyResult<()> {
+        self.inner.abort().map_err(convert_xet_error)
+    }
+
+    /// SIGINT-style abort: shuts down the runtime and cancels all tasks.
+    pub fn sigint_abort(&self) -> PyResult<()> {
```
rajatarya, line 278: The doc split between `abort()` above ("session remains usable after abort") and `sigint_abort()` here ("shuts down the runtime") is accurate but subtle — from a Python caller's POV the names don't telegraph the difference. Worth explicitly contrasting them in each docstring ("unlike `sigint_abort`, the session remains usable for new commits/groups" / "destroys the runtime; session must be recreated afterwards") so callers in an `except KeyboardInterrupt:` block don't reach for the wrong one.

hf_xet/src/py_upload_commit.rs
```diff
@@ -0,0 +1,515 @@
...
+        Ok(slf)
+    }
+
+    /// Register a Python callable to receive periodic progress updates.
+    ///
+    /// The callable is invoked on a configurable interval for the lifetime of
+    /// the commit with two positional arguments:
+    ///
+    /// 1. :class:`GroupProgressReport` — aggregate bytes processed / transferred.
+    /// 2. ``dict[UniqueId, ItemProgressReport]`` — one entry per queued file, keyed by task ID.
```
rajatarya: nit: the docstring says `dict[UniqueId, ItemProgressReport]` (lowercase 'd') but the registered Python class is `UniqueID` (uppercase 'D' — see `lib.rs` module registration and `xet_runtime/src/utils/unique_id.rs`). The struct in `unique_id.rs` is actually named `UniqueId` in Rust — worth a quick grep to pick one casing consistently in code, docstrings, and the Python name. Same typo at `py_file_download_group.rs:114`.

rajatarya (commented):
Really nice to see the object-oriented API land — the `with session.new_upload_commit()...build() as commit:` flow is a big ergonomic upgrade over the old flat functions, and the split between `XetUploadCommit`/`XetFileDownloadGroup`/`XetDownloadStreamGroup` maps cleanly onto how callers actually think about batching. The PyO3 class layering, legacy module shim with `DeprecationWarning`, and the fork-safe `XetRuntime::drop` fix are all solid. Tests look thorough against the `local://` endpoint. I won't re-litigate things hoytak (builder-pattern, `__repr__` richness, PID handling in `XetRuntime` rather than per-session) or the cursor bot (fd limit removal, stream-upload progress visibility, panic propagation in `blocking_call_with_signal_check`) already flagged — I agree with those.

A handful of things I'd like to see addressed or discussed before this merges:

**Things to address**

1. **`upload_stream` doesn't release the GIL** (py_upload_commit.rs:326). `upload_file` and `upload_bytes` both wrap `*_blocking` in `py.detach()`; `upload_stream` calls `self.inner.upload_stream_blocking(...)` directly. If this method does any I/O or awaits a queue slot, it'll block the interpreter for the duration. Either make it consistent with siblings or document why streaming setup is cheap enough to skip.

2. **Progress callback misses the terminal tick** (py_upload_commit.rs:187-193, py_file_download_group.rs:175-181). Once `is_terminal` is true, the loop breaks without delivering one last progress snapshot. Callers that drive a progress bar (the docstring example with tqdm) will freeze at whatever sub-terminal value the previous tick saw, not at 100%. Consider calling `callback.call1(py, ...)` one final time with the terminal state before breaking.

3. **Streaming download `__next__` holds the GIL while waiting** (py_download_stream_handle.rs:47, 105). The inline comment is honest about it, but this has two real consequences worth reconsidering: (a) the progress thread registered on a `XetUploadCommit` / `XetFileDownloadGroup` can't run if another Python thread is looping on an iterator from the same session (common in a `ThreadPoolExecutor`-style consumer); (b) Ctrl-C won't be observed until the next chunk arrives, so a stalled download becomes uninterruptible from Python. `XetDownloadStream` not being `Clone` is a `xet_pkg` thing — could the handle be made clonable (or expose a `blocking_next` that takes `&Arc<Self>`) so `py.detach()` is possible? This is the one part of the API where a large download hang will feel non-Pythonic.

4. **`status()` returns `&'static str`** (py_xet_session.rs:77, and the same pattern on every handle/commit/group). This is a public API. Stringly-typed states are hard to match on from Python (`if session.status() == "Running"` with no autocomplete, no typos caught), and the `Error` variant's message is silently lifted into an exception — callers that want to inspect state *without* raising can't. I'd really like this to be either a PyO3 enum or a small `XetTaskState` class with factory constants. This felt like a public-API issue so not marking as a nit.

5. **`__exit__` silently drops `abort()` errors on the exception path** (py_upload_commit.rs:250, py_file_download_group.rs:237). If `abort()` itself errors while an exception is already in flight, we swallow it with `let _ =`. If abort-on-exception ever fails for a real reason (e.g. runtime already shut down), the original exception surfaces fine but we lose the observability. A `tracing::warn!` around the discarded error would at least leave a breadcrumb.

6. **`XetSession::abort()` vs `sigint_abort()` aren't distinguished clearly** (py_xet_session.rs:84, 89). The docstrings say "Cancel all active operations on this session. The session remains usable after abort" vs "SIGINT-style abort: shuts down the runtime and cancels all tasks." It's not obvious from Python when to reach for which — worth a sentence on each explicitly contrasting the other ("unlike `sigint_abort`, the session remains usable" / "destroys the runtime; session is unusable afterwards").

**Questions**

- `with_token_refresh_url(url, headers)` bakes in the `{accessToken, exp, casUrl}` response shape. That's fine for the Hub, but hf_xet's Python surface is becoming a public-ish API (pipe for `huggingface_hub`, but also surfaced via `import hf_xet`). Worth renaming to something Hub-flavoured (`with_hf_token_refresh_url`?) or documenting that the JSON schema is part of the contract?
- `XetSession()` has no constructor arguments. Is there a design intent that all tunables live on the commit/group builder, or is session-level config (cache dir, concurrency limits) just not plumbed yet?
- `hash_files` is the only legacy function kept without `DeprecationWarning`. Is that because `huggingface_hub` still needs it, or does it have a place in the new surface too?

Overall this is a good shape for a 1.x story — once the above are resolved I think this is ready to ship.

Wauplin:
Thanks for exposing the XetSession! This is awesome and will likely unlock very nice use cases (I especially look forward the "upload bytes from stream" feature).

Small nit regarding the interface, would it be possible to have a context manager for the upload stream? i.e. instead of  (or in addition of)

```py
    stream = commit.upload_stream(name="big.bin")
    for chunk in produce_chunks():
        stream.write(chunk)
    stream.finish()  # must be called before the with-block exits
```

be able to do

```py
    with commit.upload_stream(name="big.bin") as stream:
        for chunk in produce_chunks():
            stream.write(chunk)
```

=> no need for explicit `stream.finish()` call

Wauplin:
Made some other comments in https://github.com/huggingface/huggingface_hub/pull/4116 :)

hf_xet/src/py_xet_session.rs
seanses: `XetTaskState` enum now exposed directly to Python.

hf_xet/src/py_download_stream_handle.rs
seanses: Will address this in a future PR (as an optimization)

hf_xet/src/py_xet_session.rs
seanses: "abort()" removed from the Python API, it's not very different from "sigint_abort()" and doesn't let users resume after calling it, will consider change this API from Rust XetSession API entirely.

hf_xet/src/py_upload_commit.rs
seanses: Now left a warning message.

hf_xet/src/py_upload_commit.rs
seanses: There was an unnecessary alias under "xet_data/src/progress_tracking/mod.rs": 
```
pub use xet_runtime::utils::UniqueId as UniqueID;
```
Will drop this alias in another PR as that will touch too many files outside hf_xet

hf_xet/src/py_upload_commit.rs
seanses: Now releases the GIL

hf_xet/src/py_upload_commit.rs
seanses: Moved termination check to before getting progress, now it won't miss the final update.

hf_xet/src/py_upload_commit.rs
seanses: Addressed in https://github.com/huggingface/xet-core/pull/824

seanses:
> with_token_refresh_url(url, headers) bakes in the {accessToken, exp, casUrl} response shape. That's fine for the Hub, but hf_xet's Python surface is becoming a public-ish API (pipe for huggingface_hub, but also surfaced via import hf_xet). Worth renaming to something Hub-flavoured (with_hf_token_refresh_url?) or documenting that the JSON schema is part of the contract?

Documented that the JSON schema is part of the contract.

> XetSession() has no constructor arguments. Is there a design intent that all tunables live on the commit/group builder, or is session-level config (cache dir, concurrency limits) just not plumbed yet?

Session level config plumbed through in commit [0676fd4](https://github.com/huggingface/xet-core/pull/792/commits/0676fd42ab4505633dedfe3a9af057854476007f)

> hash_files is the only legacy function kept without DeprecationWarning. Is that because huggingface_hub still needs it, or does it have a place in the new surface too?

`hash_files` is a function involving only CPU and local disk read, there's no plan to move it into the new interface yet. Will keep it there until there's the necessity.

hf_xet/src/py_upload_commit.rs
```diff
@@ -0,0 +1,447 @@
...
+    progress_callback: Option<Py<PyAny>>,
+    progress_interval_ms: u64,
+) -> PyResult<PyXetUploadCommit> {
+    let mut builder = session.new_upload_commit().map_err(convert_xet_error)?;
+    if let Some(ep) = endpoint {
+        builder = builder.with_endpoint(ep);
+    }
+    if let (Some(tok), Some(exp)) = (token, token_expiry_unix_secs) {
+        builder = builder.with_token_info(tok, exp);
+    }
```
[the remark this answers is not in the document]
seanses: This is the documented and expected behavior.

seanses:
> Small nit regarding the interface, would it be possible to have a context manager for the upload stream?

Added in https://github.com/huggingface/xet-core/pull/825

rajatarya (approved):
Thanks for the careful pass — looking through the changes since `5d3aa3c6`, every actionable item from my prior review is addressed:

| Prior item | Resolution |
|---|---|
| **#1** `upload_stream` not releasing GIL | Now wrapped in `py.detach(...)` (py_upload_commit.rs:298), consistent with `upload_file` / `upload_bytes`. |
| **#2** Missing terminal-tick progress callback | Progress thread now checks `is_terminal` *after* firing the callback for the snapshot, so the final 100% tick lands (py_upload_commit.rs:117-130). |
| **#3** `__next__` holds GIL on stream download | Acknowledged and deferred to a follow-up as an optimization. The constraint (`XetDownloadStream` not being `Clone` in `xet_pkg`) is real, so this is reasonable to land separately. |
| **#4** `status()` returning `&'static str` | Replaced with `PyXetTaskState` enum (lib.rs:46-53). PyO3's mixed-variant restriction is documented inline, and the `Error` variant is correctly surfaced as a raised exception via `task_state_to_pystate`. |
| **#5** `__exit__` swallowing `abort()` errors | `tracing::warn!` added on the exception path (py_upload_commit.rs:208-210). |
| **#6** `abort()` vs `sigint_abort()` confusion | Resolved by removing `XetSession::abort()` from the Python surface entirely; `sigint_abort` docstring now contrasts with the per-handle `abort` (py_xet_session.rs:269-280). |
| **nit** `UniqueID`/`UniqueId` mismatch | Split into #824. |

A few additional things I particularly liked in this iteration:

- **`PyXetConfig` API** — the `with_config(name, value)` / `with_config({...})` overload, dict-like `__getitem__` / `keys()` / `items()` / iteration, and the immutable-update-returning-new-instance pattern is exactly the right Python ergonomics. Nicely isolated in `hf_xet/src/config.rs` so `XetConfig`'s Rust shape can evolve without churning the Python surface.
- **Sha256 sentinels** (`COMPUTE_SHA256` / `SKIP_SHA256` as zero-data `#[pyclass(frozen)]` types + module-level constants) is much more Pythonic than the prior policy enum, and `parse_sha256` accepts `None | sentinel | str` cleanly.
- **Rename to `XetSession.new_upload_commit` / `new_file_download_group` / `new_download_stream_group`** reads better than the older builder-pattern naming.
- **`__repr__` on `PyXetSession`** including config dump is useful for debugging — and the test suite's `test_session.py` / `test_config.py` give good shape coverage.
- **The legacy module shim with `DeprecationWarning`** keeps `huggingface_hub`'s pinned `>=1.4.3,<2.0.0` range valid, which was the right call coming out of the Slack discussion.

The doc-string `:meth:`/`:class:` cross-refs are consistent throughout, the kwargs-only signatures (`#[pyo3(signature = (...))]`) match the Python convention, and the GIL discipline (everywhere except the deferred download-stream `__next__`) is now uniform.

This is a big surface and a clean landing. Ship it after #824 and #825 stack in. 🚀

hf_xet/src/py_upload_commit.rs
```diff
@@ -0,0 +1,447 @@
...
+                let result = Python::attach(|py| callback.call1(py, (group_report, item_reports)));
+                if let Err(e) = result {
+                    Python::attach(|py| e.print(py));
+                    break;
+                }
+                if is_terminal {
+                    break;
+                }
+            }
+        });
```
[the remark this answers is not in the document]
seanses: Not a concern.

PR #801, opened 2026-04-17
Move XetRuntime model away from thread-local statics

hoytak:
This PR moves the XetRuntime model away from using thread-local statics and decouples the XetConfig and XetCommon structs from a single runtime.  It introduces a struct XetContext that gives the runtime context for operations: 

```
struct XetContext { 
    pub runtime : Arc<XetRuntime>,  // The current tokio runtime wrapper, minus the config and common objects..
    pub common : Arc<XetCommon>, // The common cache objects, semaphores, rate trackers, etc.
    pub config : Arc<XetConfig> // The config 
 }
 ```
 
Now, instead of using functions like `xet_runtime()` and `xet_config()` that examine the thread-local storage, we now explicitly passing through a XetContext instance from the session creation that gets stored in each major processing struct.  

This allows decoupling between the runtime, config, and common caches, especially: 
- Running multiple config settings and/or endpoints within the same pre-existing tokio runtime.
- Running multiple runtimes that share the same XetCommon object.

---

> [!NOTE]
> **Medium Risk**
> Touches core runtime/context plumbing across multiple crates, including how Tokio runtimes are selected and how retry/concurrency behavior reads config; mistakes could surface as subtle async/runtime or performance regressions.
> 
> **Overview**
> This PR **splits the old monolithic `XetRuntime` into an execution-only runtime plus a new clonable `XetContext`** that carries `{ runtime, config, common }`, and updates call sites to pass `XetContext` explicitly instead of relying on thread-local/process-global accessors.
> 
> Client/session construction and helpers are refactored accordingly: `RetryWrapper`, `AdaptiveConcurrencyController`, `RemoteClient`/simulation clients, token refreshers, chunk cache initialization, and WASM/Python bindings now take/propagate `XetContext` (and use `ctx.config` for tuning such as retry/backoff and upload progress block sizes). Logging setup switches to `LoggingConfig::from_directory(&XetConfig, ...)`, and the local CAS simulation path tightens resource/lifetime management (e.g., keep-alive to prevent FD leaks and safer redb DB cache/drop semantics).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 13c8e309da52e6ae0ae378160bcfe7da744f6d92. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### CODE REVIEW

File: xet_pkg/src/xet_session/common.rs [hunk 1]
```diff
@@ -37,13 +37,11 @@ pub(super) async fn create_translator_config(
...
 
-    // Build token refresher
     let token_refresher: Option<Arc<dyn TokenRefresher>> = if let Some((url, token_refresh_headers)) = token_refresh {
-        let client = build_http_client(&session_id, None, Some(Arc::new(token_refresh_headers)))?;
-        let direct_route_refresher = DirectRefreshRouteTokenRefresher::new(url, client, None);
+        let client = build_http_client(&session.inner.ctx, &session_id, None, Some(Arc::new(token_refresh_headers)))?;
+        let direct_route_refresher =
+            DirectRefreshRouteTokenRefresher::new(session.inner.ctx.clone(), url, client, None);
 
-        // CAS endpoint is not provided but CAS token refresh endpoint is provided, we
```
seanses, line 45: don't remove this comment

hf_xet/src/runtime.rs
```diff
@@ -8,13 +8,13 @@ use pyo3::prelude::*;
...
 use xet_runtime::RuntimeError;
-use xet_runtime::core::XetRuntime;
+use xet_runtime::core::XetContext;
 use xet_runtime::core::sync_primatives::spawn_os_thread;
 
 lazy_static! {
     static ref SIGINT_DETECTED: Arc<AtomicBool> = Arc::new(AtomicBool::new(false));
     static ref SIGINT_HANDLER_INSTALL_PID: (AtomicU32, Mutex<()>) = (AtomicU32::new(0), Mutex::new(()));
-    static ref MULTITHREADED_RUNTIME: RwLock<Option<(u32, Arc<XetRuntime>)>> = RwLock::new(None);
+    static ref MULTITHREADED_RUNTIME: RwLock<Option<(u32, Arc<XetContext>)>> = RwLock::new(None);
```
seanses, line 17: **[High Risk] Process-global SIGINT static**

If two `XetContext` instances are created (e.g. a session is rebuilt), the second silently replaces the first in this static and the first's context is dropped. If the old context is still live during SIGINT, the handler fires against a stale (or wrong) runtime.

The PID guard on lines ~101-144 doesn't prevent a race between the PID check and the actual `perform_sigint_shutdown()` call, and doesn't verify that the cached context matches the current one.

Consider using a `OnceLock` + explicit deregistration, or document explicitly that only one `XetContext` may be alive per Python process at a time.

xet_runtime/src/core/context.rs
```diff
@@ -0,0 +1,83 @@
...
+#[derive(Clone)]
+pub struct XetContext {
+    pub runtime: Arc<XetRuntime>,
+    pub config: Arc<XetConfig>,
+    pub common: Arc<XetCommon>,
+}
+
+impl XetContext {
+    /// Creates a context from a pre-built thread pool and configuration.
+    pub fn new(config: XetConfig, runtime: Arc<XetRuntime>) -> Self {
```
seanses, line 24: **[Medium Risk] Public constructor allows mismatched config/runtime**

Callers can construct `XetContext::new(config1, runtime_built_from_config2)` and the mismatch won't be caught until runtime — e.g. `system_monitor_config` or retry settings in the runtime differ from those in the context's `config`.

Consider making this `pub(crate)` and only exposing `default()`, `from_external()`, or other validated constructors as the public API surface.

xet_runtime/src/core/context.rs
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    /// Creates a context with default configuration and an auto-detected thread pool.
+    ///
+    /// If called from an owned runtime worker thread, reuses that owned [`XetRuntime`].
+    /// Otherwise, if called from within an existing tokio runtime, wraps that runtime.
+    /// If neither is available, spins up a new owned tokio thread pool.
+    #[allow(clippy::should_implement_trait)]
+    pub fn default() -> Result<Self, RuntimeError> {
```
seanses, line 40: **[Medium Risk] Silent runtime mode switching**

This function picks Owned vs. External mode via `Handle::try_current()` without any observable signal. Callers have no way to know which mode they received, and the behaviour can differ subtly (e.g. External mode doesn't own thread pool lifecycle).

The API docs do warn about this, but consider adding an `info!()` log line when External mode is chosen ("detected existing Tokio runtime, using external handle") so it shows up in production logs when the mode switch is unexpected.

seanses (commented):
Overall this is a well-motivated and largely clean refactoring. The explicit context threading through the networking stack is thorough and the `XetCommon::cache_get_or_create()` pattern nicely replaces global statics. Main concerns are below — mostly around the remaining global state in Python bindings, the public constructor allowing mismatched config/runtime, and a few rough edges in the migration.

xet_client/src/cas_client/retry_wrapper.rs
```diff
@@ -27,6 +27,8 @@ struct ConnectionPermitInfo {
 }
 
 pub struct RetryWrapper {
+    #[allow(dead_code)]
```
seanses, line 30: **[Low] Dead field**

`ctx` is captured at construction time only to read `ctx.config.client.retry_max_attempts` (line ~43) and then never used again. The stored field is misleading — it implies per-request config access is possible via `self.ctx`, but it's actually fixed at construction.

If the intent is future-proofing, add a comment to that effect. Otherwise remove the field and the `#[allow(dead_code)]`.

xet_client/src/cas_client/adaptive_concurrency/controller.rs
```diff
@@ -73,12 +74,13 @@ struct ConcurrencyControllerState {
 }
 
 impl ConcurrencyControllerState {
-    fn new() -> Self {
-        let config = xet_config();
+    fn new(ctx: XetContext) -> Self {
```
seanses, line 77: **[Style] Inconsistent ownership conventions for XetContext**

Across the new APIs, constructors use three different ownership conventions with no apparent rule:
- `build_http_client(&ctx: &XetContext)` — borrowed
- `AdaptiveConcurrencyController::new(ctx: XetContext)` — owned (here)
- `WrappedProgressUpdater::new(..., ctx: Arc<XetContext>)` — Arc

This makes it hard to know when a clone happens. Since `XetContext` is cheap to clone (`Arc` fields), a single convention — e.g. take `&XetContext` and `.clone()` internally when storing — would be cleaner and explicit about cost.

xet_client/src/cas_client/adaptive_concurrency/controller.rs
```diff
@@ -764,8 +770,10 @@ impl ConcurrencyControllerState {
 #[cfg(test)]
 impl AdaptiveConcurrencyController {
     pub fn new_testing(concurrency: usize, concurrency_bounds: (usize, usize)) -> Arc<Self> {
+        let ctx = XetContext::default().expect("test runtime");
```
seanses, line 773: **[Low] Panicking test setup hides initialization errors**

`XetContext::default().expect("test runtime")` will panic with a generic message if runtime initialization fails (e.g. config parse error, thread pool exhaustion). The panic is uncaught and appears as a test crash rather than a test failure with context.

Prefer `XetContext::default().unwrap_or_else(|e| panic!("failed to create test context: {e}"))` or a shared `test_ctx()` helper that returns `Result` so `?` can propagate through `#[tokio::test]` tests.

wasm/hf_xet_wasm/src/wasm_file_upload_session.rs
```diff
@@ -82,8 +89,8 @@ impl FileUploadSession {
             let mut current_session_data = self.current_session_data.lock().await;
 
             // Do we need to cut one of these to a xorb?
-            if current_session_data.num_bytes() + file_data.num_bytes() > *XORB_CUT_THRESHOLD_BYTES
-                || current_session_data.num_chunks() + file_data.num_chunks() > *XORB_CUT_THRESHOLD_CHUNKS
+            if current_session_data.num_bytes() + file_data.num_bytes() > *MAX_XORB_BYTES
```
seanses, line 92: **[Low] Partial migration: WASM still uses compile-time constants instead of ctx config**

The rest of the codebase was migrated to read XORB thresholds from `ctx.config.xorb.*`, but this file still imports `MAX_XORB_BYTES`/`MAX_XORB_CHUNKS` from `xet_core_structures::xorb_object::constants` (line 11) and uses them as fixed values here.

If the WASM path is intentionally not configurable, add a comment explaining why. Otherwise migrate to `self.ctx.config.xorb.max_bytes` / `.max_chunks` (or equivalent) to keep behaviour consistent.

hf_xet/src/logging.rs
```diff
@@ -36,7 +36,7 @@ pub fn init_logging(py: Python) {
     let xet_cache_directory = xet_runtime::core::xet_cache_root();
     let log_dir = xet_cache_directory.join("logs");
 
-    let cfg = LoggingConfig::default_to_directory(version_info, log_dir);
+    let cfg = LoggingConfig::from_directory(&xet_runtime::config::XetConfig::new(), version_info, log_dir);
```
seanses, line 39: **[Low] Logging setup creates a fresh XetConfig outside of context**

`XetConfig::new()` constructs a standalone config independent of any live `XetContext`. This is inconsistent with the PR's goal of explicit context threading — if logging options are ever added to `XetConfig`, this call site will silently use defaults rather than the session's actual config.

If a context is available at this call site, prefer `&ctx.config`. If not (e.g. called before session creation), document that this is intentional early-init logging with default config.

seanses (commented):
LGTM!

seanses (approved):

PR #819, opened 2026-04-26
feat(chunk_cache): coalesce overlapping/adjacent puts into one item

XciD:
## Summary

Make `DiskCache::put_impl` coalesce a freshly-fetched range with any existing items it overlaps or is adjacent to, so the cache converges to one item per touched portion of a xorb instead of accumulating sibling fragments that lookups can't compose.

## Why

`CacheState::find_match` only returns a hit when a single cached item fully contains the requested range:

```rust
for item in items.iter() {
    if item.range.start <= range.start && range.end <= item.range.end {
        return Some(item.clone());
    }
}
```

So when distinct byte-range queries hit the same xorb (which happens whenever a consumer issues range-bounded reads at different offsets), `put_impl` writes them as separate items. A subsequent get spanning two siblings misses even though every byte is on disk: each miss fires a fresh CAS round-trip and writes yet another fragmented record. Caches grow without bound and never converge.

Concrete trigger from one downstream user (hf-mount FUSE): a single `transformers.from_pretrained` call ended up issuing ~800 redundant CAS fetches against shards whose data was already cached, simply because each FUSE read produced a slightly different `ChunkRange` per xorb.

## Approach

Two existing put-side checks already cover the easy cases:
- the new range is fully contained in an existing item → `validate_match` no-ops
- a concurrent put already inserted a containing item → re-check under the state write lock and abort

This PR adds the missing case: the new range overlaps or is adjacent to one or more existing items. Under the read lock we snapshot the touching items, load each one's `(header, chunk_data)` from disk via `read_and_verify_cache_file`, and verify the checksum (corrupt items are dropped, not merged). The merge builds a single contiguous record whose chunks are taken from the new range when available and from the existing items otherwise — the union of the touching ranges is provably gap-free, so every chunk in the merged range is sourced from one of the inputs without an extra CAS fetch.

When the merged record would exceed `capacity` we fall back to writing just the new range; the original range may still fit even if the union with existing siblings doesn't. We keep `read_and_verify_cache_file` as a small extracted helper shared with `validate_match`.

Read path is unchanged — `find_match` still does single-item containment, which is correct against the merged record. On-disk file format and per-item file naming are unchanged. No migration.

## Trade-offs vs. compose-on-read

I prototyped a compose-on-read alternative (have `get_impl` stitch a query from multiple covering items). On a 15.7 GB scattered-read workload, coalesce-on-put consistently re-downloaded ~6× less than compose-on-read on the warm phase. The reason isn't compose itself — it is that cache.get hits via compose end up *skipping* fetches that would otherwise have populated the cache, leaving the cold state sparser and breaking later warm queries. Coalesce-on-put doesn't change get semantics, so cold puts always run, the cache state stays dense, and `find_match` strict reliably hits warm queries. The on-disk and put-time cost is bounded (read N + write 1 + delete N per overlap, no extra CAS).

## Tests

- `coalesce_overlapping_ranges_into_one_item`
- `coalesce_adjacent_ranges_into_one_item`
- `coalesce_three_overlapping_ranges`
- `disjoint_ranges_are_not_merged`
- `merged_too_large_falls_back_to_new_range_only`
- `coalesce_preserves_total_bytes_accounting`

All 20 pre-existing `chunk_cache` tests still pass (26/26 total).

## Real-world impact

Validated on `google/gemma-3n-E4B-it` (15.7 GB, 4 safetensors shards), 32 scattered 256 KiB reads (8 per shard), kernel page cache dropped between cold and warm. Cold reads at MB offsets, warm reads at the same offsets shifted by +384 KiB. Three fresh trials per binary, cache on EBS (not tmpfs). Average warm Δ re-download:

|                | UNPATCHED | this PR  |
|----------------|-----------|----------|
| trial 1        | +1015 MB  | +2.6 MB  |
| trial 2        | +983 MB   | +109 MB  |
| trial 3        | +1.9 GB   | +41 MB   |
| **average**    | **~1.3 GB** | **~50 MB** |

~26× less re-downloaded on the warm phase. Cold-phase items in the cache drop from ~190 to ~100 because coalescing produces wider items.

### DIFF
--- a/xet_client/src/chunk_cache/disk.rs
+++ b/xet_client/src/chunk_cache/disk.rs
@@ -61,6 +61,17 @@ impl CacheState {
         None
     }
 
+    /// Returns every cache item whose range overlaps or is adjacent to `range`.
+    /// Used by `put_impl` to coalesce fragments of the same xorb into one item
+    /// instead of accumulating sibling fragments that `find_match` can't
+    /// compose at lookup time.
+    fn find_touching(&self, key: &Key, range: &ChunkRange) -> Vec<VerificationCell<CacheItem>> {
+        let Some(items) = self.inner.get(key) else {
+            return Vec::new();
+        };
+        items.iter().filter(|item| ranges_touch(&item.range, range)).cloned().collect()
+    }
+
     /// removed items from the cache (including deleting from file system)
     /// until at least to_remove number of bytes have been removed
     ///
@@ -388,10 +399,48 @@ impl DiskCache {
             }
         }
 
-        let header = CacheFileHeader::new(chunk_byte_indices);
+        // Coalesce with overlapping/adjacent siblings so the cache converges
+        // to one item per touched portion of the xorb. Without this,
+        // distinct byte-range queries against the same xorb produce sibling
+        // fragments that `find_match` cannot compose at lookup time.
+        let touching = self.state.read().await.find_touching(key, range);
+        let mut loaded = Vec::with_capacity(touching.len());
+        let mut replaced_items: Vec<VerificationCell<CacheItem>> = Vec::with_capacity(touching.len());
+        for item in &touching {
+            if let Some(li) = self.load_item_for_merge(key, item).await? {
+                loaded.push(li);
+                replaced_items.push(item.clone());
+            } else {
+                self.remove_item(key, item).await?;
+            }
+        }
+
+        let new_only = || MergedRecord {
+            range: *range,
+            chunk_byte_indices: chunk_byte_indices.to_vec(),
+            data: data.to_vec(),
+        };
+        let mut merged = if loaded.is_empty() {
+            new_only()
+        } else {
+            merge_chunks(range, chunk_byte_indices, data, &loaded)?
+        };
+
+        let mut header = CacheFileHeader::new(merged.chunk_byte_indices.clone());
         let mut header_buf = Vec::with_capacity(header.header_len());
         header.serialize(&mut header_buf)?;
-        let len = (header_buf.len() + data.len()) as u64;
+        let mut len = (header_buf.len() + merged.data.len()) as u64;
+        // If coalescing pushed the merged record past capacity, fall back to
+        // storing the new range on its own — the original range may still
+        // fit, even if the union with existing siblings doesn't.
+        if len > self.capacity && !loaded.is_empty() {
+            merged = new_only();
+            replaced_items.clear();
+            header = CacheFileHeader::new(merged.chunk_byte_indices.clone());
+            header_buf.clear();
+            header.serialize(&mut header_buf)?;
+            len = (header_buf.len() + merged.data.len()) as u64;
+        }
         if len > self.capacity {
             // refusing to add this item as it is too large for the cache with configured capacity
             return Ok(());
@@ -400,12 +449,12 @@ impl DiskCache {
         let checksum = {
             let mut hasher = crc32fast::Hasher::new();
             hasher.update(&header_buf);
-            hasher.update(data);
+            hasher.update(&merged.data);
             hasher.finalize()
         };
 
         let cache_item = CacheItem {
-            range: *range,
+            range: merged.range,
             len,
             checksum,
         };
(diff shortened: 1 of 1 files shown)


XciD:
End-to-end validation against `google/gemma-3n-E4B-it` (15.7 GB, 4 safetensors shards) on EC2 with hf-mount, comparing `56faae12` (unpatched) against this branch (`e3453793`). Workload: 32 scattered 256 KiB reads (8 per shard at 1/13/27/41/55/69/83/95% of shard size), fresh cache each run, kernel page cache dropped before measurement.

|                  | UNPATCHED | PATCHED | Δ |
|------------------|-----------|---------|---|
| total items      | 105       | 56      | **-47%** |
| xorbs touched    | 54        | 47      | -13% |
| avg items/xorb   | 1.94      | 1.19    | -39% |
| cache size       | 1.5 GB    | 1003 MB | **-33%** |
| wall-clock time  | 6 s       | 4 s     | -33% |
| worst-fragmented xorb | 19 items | 8 items | -58% |

Items-per-xorb distribution:

```
UNPATCHED                 PATCHED
  1 xorb  with 19 items     1 xorb  with 8 items
  1 xorb  with 4  items     2 xorbs with 2 items
  3 xorbs with 3  items    46 xorbs with 1 item
 25 xorbs with 2  items
 24 xorbs with 1  item
```

The 33% cache-size drop equals real CAS bandwidth saved on cold reads — bytes the unpatched version was downloading twice (once per ChunkRange variant). Smaller-scale validation on `google/gemma-3-270m` (536 MB, 21 reads) showed the same trend: 28→12 items, 3.50→1.50 avg, 2.9 s→1.9 s.

Smoke-tested in hf-mount via the local crate override; no behavioral regressions on the 259 hf-mount unit tests or the 25 chunk_cache tests in this PR.

XciD:
Re-ran the gemma-3n-E4B-it workload (15.7 GB, 4 safetensors shards) with this revision. Cold phase: 32 scattered 256 KiB reads (8 per shard at 1/13/27/41/55/69/83/95% of shard size). Warm phase: same offsets shifted by +384 KiB so the resulting `ChunkRange`s differ from cold but overlap with the stored fragments. Kernel page cache dropped between cold and warm.

|                  | UNPATCHED  | this PR (compose-on-read) |
|------------------|------------|---------------------------|
| cold time        | 9 s        | 7 s                       |
| cold items       | 160        | 178                       |
| cold cache size  | 3.8 GB     | 4.6 GB                    |
| warm time        | 5 s        | 2 s                       |
| warm Δitems      | +73        | **+6**                    |
| warm Δ cache size| **+1.2 GB**| **+87 MB**                |

The warm Δ is the headline: the unpatched build re-downloaded 1.2 GB on the second pass because `find_match` missed for chunk_ranges that didn't fall fully inside any single stored item. With `find_covering` + compose, the same shifted-offset reads compose from the existing fragments and only ~87 MB of genuinely new bytes are fetched (~14× less re-download).

Cold-phase totals look slightly worse on disk for this PR (4.6 GB vs 3.8 GB / 178 vs 160 items), but that's run-to-run variance in xet-core's prefetch+block scheduling — not a regression from this change. Cold-phase put behavior is unchanged.

(For comparison, an earlier coalesce-on-put prototype of this fix shrank the on-disk size on cold by merging fragments at write time but achieved a smaller warm re-download reduction, +561 MB. The compose-on-read approach does no extra work on put and produces the better warm read result on this workload.)

XciD:
Update: I force-pushed back to the original coalesce-on-put design after benchmark data showed it consistently outperforms the compose-on-read alternative I had pivoted to. Three trials on the gemma-3n-E4B-it scattered-read workload (offsets shifted +384 KiB between cold and warm):

| | UNPATCHED warm Δ | compose-on-read warm Δ | coalesce-on-put warm Δ (this PR) |
|---|---|---|---|
| trial 1 | +1015 MB | +234 MB | +2.6 MB |
| trial 2 | +983 MB | +489 MB | +109 MB |
| trial 3 | +1.9 GB | +231 MB | +41 MB |

Compose-on-read helps over unpatched but not nearly as much as coalesce-on-put. The reason isn't compose itself: under compose-on-read, `cache.get` hits via composition cause xet-core to *skip* fetches it would otherwise issue, leaving the cold cache sparser and breaking later warm queries. Coalesce-on-put doesn't change get semantics, so cold puts always run and the state stays dense.

Earlier benchmark numbers I posted were also influenced by tmpfs cache_dir filling up under load — re-running on EBS is what surfaced the real difference. Apologies for the churn.

PR #825, opened 2026-04-30
Add context manager for upload stream

seanses:
## Summary

Addresses the review [comment](https://github.com/huggingface/xet-core/pull/792#issuecomment-4305213150) on #792: adds a context manager to `XetStreamUpload` so callers no longer need an explicit `finish()` call.

```python
# Before
stream = commit.start_upload_stream(name="big.bin")
for chunk in produce_chunks():
    stream.write(chunk)
stream.finish()  # must be called explicitly

# After
with commit.start_upload_stream(name="big.bin") as stream:
    for chunk in produce_chunks():
        stream.write(chunk)
# finish() called automatically on normal exit
# abort() called automatically on exception
```

## Changes

- **`hf_xet/src/py_stream_upload_handle.rs`** — add `__enter__` / `__exit__` to `PyXetStreamUpload`:
  - Normal exit: calls `finish()` (GIL-releasing via `py.detach()`), idempotent — skips if `finish()` was already called inside the `with`-block
  - Exception: calls `abort()` (infallible)
  - Updated class docstring with usage example
- **`hf_xet/tests/test_upload_commit.py`** — three new tests in `TestUploadStream`:
  - `test_context_manager_finishes_on_normal_exit` — verifies `try_finish()` is non-`None` after the block
  - `test_context_manager_aborts_on_exception` — verifies exception propagates and `try_finish()` is `None`
  - `test_context_manager_idempotent_when_finish_called_inside` — verifies calling `finish()` explicitly inside the block does not cause a double-finish error on exit

---

> [!NOTE]
> **Low Risk**
> Adds Python context-manager semantics around stream finalization/abort; behavior is largely additive but could affect edge cases around finishing/aborting if callers rely on exceptions during cleanup.
> 
> **Overview**
> `XetStreamUpload` can now be used as a Python context manager: `__exit__` auto-calls `finish()` on normal exit and `abort()` when exiting due to an exception, without suppressing errors.
> 
> The implementation is explicitly *idempotent* by skipping `finish()` in `__exit__` if the stream already reports finished, and docs/tests were updated to cover normal-exit finishing, exception-triggered abort, and the “finish inside with-block” case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 34c6f67272f5681437ea1d2c1a3df7a1cddeafbe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/hf_xet/src/py_stream_upload_handle.rs
+++ b/hf_xet/src/py_stream_upload_handle.rs
@@ -9,8 +9,20 @@ use crate::{PyXetTaskState, convert_xet_error};
 /// Handle for a streaming upload within an :class:`XetUploadCommit`.
 ///
 /// Returned by :meth:`XetUploadCommit.start_upload_stream`.  Feed data incrementally
-/// with :meth:`write`, then call :meth:`finish` to finalise ingestion.
-/// **:meth:`finish` must be called before** :meth:`XetUploadCommit.wait_to_finish`.
+/// with :meth:`write`, then finalise with either:
+///
+/// - :meth:`finish` (explicit call), or
+/// - the context manager, which calls :meth:`finish` automatically on normal exit and :meth:`abort` on exception.
+///
+/// **:meth:`finish` (or the context manager) must complete before**
+/// :meth:`XetUploadCommit.wait_to_finish` is called.
+///
+/// ```python
+/// with commit.start_upload_stream(name="big.bin") as stream:
+///     for chunk in produce_chunks():
+///         stream.write(chunk)
+/// # finish() was called automatically
+/// ```
 #[pyclass(name = "XetStreamUpload")]
 #[derive(Clone)]
 pub struct PyXetStreamUpload {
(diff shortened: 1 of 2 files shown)


Wauplin (commented):
Haven't checked the implementation but love the API! 🫶

assafvayner (approved):

rajatarya (approved):
Clean implementation of the context-manager affordance from #792's review. Small, focused, and the test coverage is exactly the three scenarios that matter: normal exit, exception path, and explicit-`finish()`-inside-the-block.

A few things I particularly liked:

- **Idempotency via `try_finish().is_none()`** — using the existing `try_finish` probe to gate the second `finish()` is cleaner than tracking a separate `finished` flag on the Python wrapper.
- **`Ok(false)` from `__exit__`** — standard non-suppressing context-manager semantics; the docstring correctly calls this out.
- **GIL handling** — `__exit__` routes `finish()` through `self.finish(py)` so the existing `py.detach()` release still applies, while `abort()` is a direct (infallible) call on `self.inner`. Right split.
- **Docstring update** — the new bullet list + the inline `with` example reads well, and the `:meth:`/`:class:` cross-refs match the existing style in this file.

CI fully green across linux/macos/win + WASM. Ship it.

PR #835, opened 2026-05-07
Add napi smoke-test example for hf-xet

assafvayner:
Human context: in integrating new hf-hub usage in tokenizers, tokenizers also generated a napi binary. so we should validate that hf-hub/hf-xet are napi compat (hf-hub is pretty trivial if given that hf-xet is compatible).

## Summary

- Adds `examples/xet_pkg_napi/` — a minimal napi-rs binding that links `hf-xet` (the `xet` crate at `xet_pkg/`) into a Node.js native addon.
- Exposes `initLogging(version)` and `smokeTest()`. The smoke test builds a `XetSession` synchronously and constructs upload-commit + file-download-group builders to exercise lazy runtime startup.
- Crate is excluded from the xet-core workspace and carries its own `[workspace]` table so it stays standalone under git worktrees (where cargo would otherwise resolve through the canonical repo path).
- Build artifacts (`*.node`, `index.js`, `index.d.ts`, `node_modules/`) are gitignored; `Cargo.lock` and `package-lock.json` are committed for reproducibility.

The point of the smoke test is **not** a full JS API — it's to verify hf-xet compiles, links, and starts inside libuv (no pyo3, no host-owned tokio). If you can run `npm run smoke` and see `xet session built; runtime initialized`, the integration is ready for a fuller binding (async upload/download via `#[napi]` async fns, progress callbacks via `ThreadsafeFunction`).

## Test plan

- [x] `npm install` (in `examples/xet_pkg_napi/`)
- [x] `npm run build:debug` — compiles `hf-xet`, `xet-runtime`, `xet-client`, `xet-data`, `xet-core-structures` and the napi shim against napi 2.16
- [x] `npm run smoke` — outputs:
  ```
  loaded addon, exports: [ 'initLogging', 'smokeTest' ]
  smokeTest: xet session built; runtime initialized
  ```
- [x] Verify on Linux (only tested on darwin-arm64 locally)
- [x] Decide whether to wire into CI, or keep as a manual example

---

> [!NOTE]
> **Low Risk**
> Adds a standalone example project and build scripts without changing production crates; primary risk is repo bloat/noise from the committed lockfiles and an extra exclusion in the workspace.
> 
> **Overview**
> Adds a new standalone `examples/xet_pkg_napi` project to smoke-test that `hf-xet` can compile/link as a `napi-rs` Node native addon and perform a real file download via the blocking download APIs.
> 
> Updates the root `Cargo.toml` to **exclude** this example from the workspace, and includes the example’s build/run scaffolding (`package.json`, `smoke.mjs`, `build.rs`) plus committed lockfiles and a `.gitignore` for generated artifacts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cb628956f7d0c01bd9cbd5c351df90a1448121a1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


examples/xet_pkg_napi/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "scripts": {
+    "build": "napi build --platform --release",
+    "build:debug": "napi build --platform",
+    "smoke": "node smoke.mjs"
+  },
+  "devDependencies": {
+    "@napi-rs/cli": "^2.18.4"
+  },
+  "engines": {
+    "node": ">=16"
```
seanses: `"node": ">=16"` but `smoke.mjs` uses the native `fetch` API which requires Node 18+ (added experimentally in 18.0.0, not present in 16). Running on Node 16 will crash at the `fetch` call.

```diff
-    "node": ">=16"
+    "node": ">=18"
```

Also worth updating the requirement in README.md line 32.

examples/xet_pkg_napi/smoke.mjs
```diff
@@ -0,0 +1,109 @@
...
+const fileSize = Number(linkedSize);
+console.log(`  xet-hash:    ${xetHash}`);
+console.log(`  size:        ${fileSize.toLocaleString()} bytes`);
+console.log(`  sha256:      ${sha256 ?? "(unknown)"}`);
+
+mkdirSync(destDir, { recursive: true });
+const destPath = join(destDir, basename(filename));
+rmSync(destPath, { force: true });
+
+addon.initLogging("xet_pkg_napi/0.0.1 smoke");
```
seanses, line 83: `smokeTest()` is never called here. The whole point of that function is to exercise session/runtime startup without I/O, but the smoke script skips straight to `downloadFile`. The no-I/O path goes untested and the PR description output (`"smokeTest: xet session built; runtime initialized"`) does not match the actual script.

```js
addon.initLogging("xet_pkg_napi/0.0.1 smoke");
const testResult = addon.smokeTest();
console.log("smokeTest:", testResult);
```

seanses (commented):
Not an expert here. I noticed that `Verify on Linux (only tested on darwin-arm64 locally)` is not checked yet? Might as well also verify on Linux amd64

seanses (approved):

PR #841, opened 2026-05-12
feat(wasm): make hf-xet compile for wasm32-unknown-unknown

### LINKED ISSUE #840: hf-xet rust crate (xet_pkg) should support WASM target
hf-hub depends on the hf-xet rust crate. hf-hub is depended on by other projects that have a WebAssembly interface. In order for hf-hub to be able to provide be WASM compatible we need hf-xet to provide a WASM compatible interface.

The ideal case is that the XetSession is still available with the same interface when compiling for a wasm target.

Download functionality is more important than upload at the moment (note that both have been previously implemented, just not integrated into the public crate). The main end consuming libraries (tokenizers, candle, TEI) are only downloading files from the hub, not uploading.

assafvayner:
Closes #840

## Summary

Makes `hf-xet` (`xet_pkg`) and its streaming download + upload data-prep paths compile and **run** on `wasm32-unknown-unknown`, and reworks `wasm/hf_xet_wasm/` into a single example/smoke-test JS-binding crate (`cdylib`) that exposes both flows from one wasm module. (The earlier split into separate download/upload crates was reverted — one module is enough for the smokes and lets the example pages share it.) `wasm/hf_xet_thin_wasm/` is unchanged in scope but follows the new shared wasm-bindgen pinning (`=0.2.121`).

## Approach

1. **Compile compatibility** — cfg-gated `tokio` dep split; native-only gating of `_blocking`/path-based methods, `xet_pkg::legacy`, and native runtime pieces; `xet_client` cache gating.
2. **Streaming runtime compatibility** — make the streaming download/upload paths `!Send`-tolerant so they actually execute on wasm (where reqwest's futures are `!Send`), via conditional `?Send` async-traits and `use tokio_with_wasm::alias as tokio` at every spawn site.

A later simplification pass deduplicates the wasm/native split where the bodies were identical: `TaskRuntime::bridge_async{,_finalizing}` are written once via a `MaybeSend` shim, `XetSessionBuilder::build` / `TranslatorConfig::new` are single cfg-bound fns, the wasm shard dedup cache is a plain bounded FIFO, and the ci-smoke suite runs from one shared harness.

## Key changes

- **`tokio_with_wasm`** added to workspace deps and the wasm-target deps of `xet_pkg`/`xet_data`/`xet_runtime`/`xet_client`. Native uses real `tokio`; wasm aliases to `tokio_with_wasm` (spawns via `wasm_bindgen_futures::spawn_local`, so spawned futures needn't be `Send`).
- **Conditional `?Send` async-traits** for the streaming providers/writers (`URLProvider`, `DataWriter`, `XorbURLProvider`, `Sequential/UnorderedWriter`, `UploadSessionDataManager`). Trait bounds on `dyn T` stay `Send + Sync`; only the futures drop `Send` on wasm.
- **Native-only gating** of all `_blocking`/filesystem methods, `XetSession::sigint_abort`, `XetSessionBuilder::with_tokio_handle`, the `xorb_block.rs` chunk-cache spawn, and the entire `xet_pkg::legacy` module — surfaced as compile errors rather than runtime panics.
- **`xet_runtime`**: the wasm `XetRuntime` is a stub (no tokio runtime; bridges `.await` directly via `wasm_bindgen_futures`); `spawn_blocking` maps to `tokio_with_wasm`; native-only accessors and the `SystemMonitor` are gated off on wasm.
- **`xet_data`**: `processing` + `file_reconstruction` are now available on wasm; new `shard_interface/wasm.rs` provides an in-memory `SessionShardInterface` with a bounded FIFO dedup cache (no disk staging, no resume).
- **`wasm/hf_xet_wasm/`** (unified `cdylib`): `#[wasm_bindgen]` wrapper over `xet::xet_session::XetSession` exposing both upload and download from one module. Auth-free `XetSession`; auth lives on the per-commit / per-group builder. Ships `build_wasm.sh` + `examples/{download,upload}.html`. **Not a published browser SDK** — real consumers should depend on `hf-xet` directly with their own `#[wasm_bindgen]` glue, or use a downstream SDK such as `hf-hub`.
- **CI** (`.github/actions/build-wasm` + `ci.yml`): builds both wasm crates, runs the `xet_pkg/build_wasm.sh` compile gate (`cargo +nightly check --target wasm32-unknown-unknown -p hf-xet -Z build-std`), checks the wasm `Cargo.lock`s are fresh, then runs the `wasm/ci-smoke/` headless-Chromium Playwright suite via `node run.mjs all`.

## WASM API surface (downloads)

```
XetSessionBuilder::build() → XetSession
XetSession::new_download_stream_group() → XetDownloadStreamGroupBuilder
  .with_endpoint(...)
  .with_token_info(token, expiry)
  .with_token_refresh_url(...)
  .build().await → XetDownloadStreamGroup
XetDownloadStreamGroup::download_stream(file_info, range).await → XetDownloadStream
XetDownloadStream::next().await → Option<Bytes>
XetDownloadStream::cancel()
XetSession::abort() / id() / config()
```

## WASM API surface (uploads)

```
XetSessionBuilder::build() → XetSession
XetSession::new_upload_commit() → XetUploadCommitBuilder
  .with_endpoint(...)
  .with_token_info(token, expiry)
  .build().await → XetUploadCommit
XetUploadCommit::upload_bytes(bytes, sha256, name).await → XetFileUpload
XetUploadCommit::upload_stream(name, sha256).await → XetStreamUpload
XetStreamUpload::write(chunk).await / finish().await / abort()
XetUploadCommit::commit().await → XetCommitReport
XetUploadCommit::abort()
```

The unified `hf_xet_wasm` crate exposes the JS classes `XetSession`, `XetDownloadStreamGroup`, `XetDownloadStream`, `XetUploadCommit`, `XetStreamUpload`, and `XetFileUpload` from a single wasm module.

## Testing

- **Native**: `hf-xet` (95 unit + 6 doctest) and `xet-data` (283 unit) tests pass.
- **WASM compile + build**: `xet_pkg/build_wasm.sh` (the wasm compile gate) and both crates' `build_wasm.sh` scripts produce `pkg/{js,d.ts,wasm}` (CI).
- **WASM browser smokes** (CI, headless Chromium + Playwright): **18 scenarios** in `wasm/ci-smoke/`, run in a single `node run.mjs all` step against prod Hub + CAS (`HF_SMOKE_TEST_TOKEN`). Network scenarios get one retry; any failure fails the job. Coverage: local input validation; single / multi / range / cancel / error downloads (streaming + `downloadToBytes`); single / multi / mixed / tiny / streaming uploads; multi-commit; `sha256-policy`; upload lifecycle/misuse; session-level dedup; and HMAC-keyed global dedup (re-upload fully dedups — `xorb_bytes_uploaded === 0`, no new xorb).
- **Manual**: serve `wasm/hf_xet_wasm/examples/{download,upload}.html` with a static server (the smoke `server.mjs` sets the required COOP/COEP headers).

---

> [!NOTE]
> **High Risk**
> Large cross-cutting changes to upload/download, dedup, HTTP, and async runtime behavior on wasm, plus CI that hits production Hub/CAS with write-capable tokens.
> 
> **Overview**
> Adds **`wasm32-unknown-unknown`** support for `hf-xet` and its transitive crates so browser consumers can depend on the library directly, with a **wasm-only public API subset** (async-only, no filesystem/`legacy`, stub `XetRuntime`).
> 
> **Runtime / async:** Introduces **`tokio_with_wasm`** and workspace-pinned **`wasm-bindgen` 0.2.121**; wasm code uses `tokio_with_wasm::alias as tokio`, conditional **`async_trait(?Send)`**, and **`web_time`** instead of std/tokio instants. Native blocking APIs, path downloads, disk caches, and **`xet_pkg::legacy`** are **`#[cfg(not(target_family = "wasm"))]`**. **`xet_data`** gains an in-memory **`shard_interface/wasm`** path; **`RemoteClient::upload_xorb`** uses non-streaming bodies on wasm with bulk progress reporting.
> 
> **`wasm/hf_xet_wasm`:** Reworked into a thin **`cdylib`** over `hf-xet` (unified upload/download JS smoke wrapper, not a published SDK). **`wasm/ci-smoke/`** adds **18 Playwright scenarios** against prod Hub/CAS, run from CI via **`node run.mjs all`** after wasm builds and **`xet_pkg/build_wasm.sh`** compile gate.
> 
> **CI / tooling:** **`build-wasm`** caches wasm-bindgen/wasm-pack binaries, bumps tool versions, and runs the **`hf-xet`** wasm check; docs and **`api_changes/update_260515_wasm_target_support.md`** document wasm differences and maintenance patterns.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d549241f78d91d75dfb8eae4a54bfa86f203aca2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


xet_client/src/cas_client/remote_client.rs
```diff
@@ -674,22 +675,38 @@ impl Client for RemoteClient {
...
-                        let upload_stream = UploadProgressStream::wrap_bytes_as_stream(
-                            serialized_data.clone(),
-                            block_size,
-                            upload_reporter.clone(),
-                        );
                         let url = url.clone();
+                        let serialized_data = serialized_data.clone();
+
+                        #[cfg(not(target_family = "wasm"))]
+                        let request = {
```
hoytak, line 685: nit: I'd find: 
```
let request = {
   #[cfg(not(...))] { ...}

 #[cfg(...)]
 { ... } 
}; 
```
to be a more readable structure.

xet_core_structures/src/utils/exp_weighted_moving_avg.rs
```diff
@@ -1,4 +1,11 @@
-use tokio::time::{Duration, Instant};
+use std::time::Duration;
+
+// Native: `tokio::time::Instant` so tests can use `tokio::time::{advance, pause}`.
+// Wasm: `web_time::Instant` since `std::time::Instant` panics on `wasm32-unknown-unknown`.
```
hoytak, line 4: This is about tokio::time -- does that panic on wasm ?

assafvayner: we need to use web_time here, tokio::time::Instance is a wrapper over the std::time version. and the tokio_with_wasm overshadows but does not export Instant

assafvayner (commented):
claude generated — inline notes documenting the non-gating behavioral changes in this PR (the "genuine logic changes"). These are the only changes that alter runtime behavior beyond cfg-gating; flagging them for reviewer attention.

xet_data/src/processing/file_upload_session.rs
```diff
@@ -518,16 +539,22 @@ impl FileUploadSession {
...
+        // absorbs (and silently drops) those late writes — manifests on wasm
+        // (single-threaded; tasks rarely complete before the `await` above
+        // returns) as `xorb_bytes_uploaded == 0`.
         let mut upload_tasks = take(&mut *self.xorb_upload_tasks.lock().await);
 
         while let Some(result) = upload_tasks.join_next().await {
             result??;
         }
 
+        let mut metrics = take(&mut *self.deduplication_metrics.lock().await);
```
assafvayner, line 594: claude generated — Correctness fix: `take(deduplication_metrics)` was moved to run *after* the `xorb_upload_tasks` join loop above. Each xorb-upload task records its transmitted-byte count into `deduplication_metrics` only after its CAS request resolves, so taking the metric before the join snapshots an empty `DeduplicationMetrics` and silently drops those late writes. On native (multi-threaded) the tasks usually finish before this point, masking the race; on single-threaded wasm it reliably surfaced as `xorb_bytes_uploaded == 0`. This also closes a latent native race — and is safe because no new upload tasks are queued after the join.

xet_data/src/deduplication/file_deduplication.rs
```diff
@@ -96,7 +96,7 @@ impl<DataInterfaceType: DeduplicationDataInterface> FileDeduper<DataInterfaceTyp
             .config
             .xorb
             .simulation_max_bytes
-            .map(|bs| (bs.as_u64() as usize).min(*MAX_XORB_BYTES))
+            .map(|bs| bs.as_u64().min(*MAX_XORB_BYTES as u64) as usize)
```
assafvayner, line 99: claude generated — Correctness fix (32-bit truncation): the order was swapped from `(bs.as_u64() as usize).min(*MAX_XORB_BYTES)` to `bs.as_u64().min(*MAX_XORB_BYTES as u64) as usize`. The old form cast a potentially huge `u64` to `usize` *before* clamping, which truncates on wasm32's 32-bit `usize`. Clamping in `u64` space first guarantees the result fits (MAX_XORB_BYTES = 64 MiB). Same change applied at `file_upload_session.rs:442`.

xet_data/src/processing/file_upload_session.rs
```diff
@@ -418,7 +439,7 @@ impl FileUploadSession {
                 .config
                 .xorb
                 .simulation_max_bytes
-                .map(|bs| (bs.as_u64() as usize).min(*MAX_XORB_BYTES))
+                .map(|bs| bs.as_u64().min(*MAX_XORB_BYTES as u64) as usize)
```
assafvayner, line 442: claude generated — Correctness fix (32-bit truncation): same fix as `file_deduplication.rs:99` — clamp `simulation_max_bytes` against `MAX_XORB_BYTES` in `u64` space *before* the `as usize` cast, so wasm32's 32-bit `usize` cannot truncate a large configured value.

xet_client/src/cas_client/remote_client.rs
```diff
@@ -674,22 +675,42 @@ impl Client for RemoteClient {
...
+                                    .with_extension(Api(api_tag))
+                                    .header(CONTENT_LENGTH, HeaderValue::from(n_upload_bytes)) // must be set because of streaming
+                                    .body(Body::wrap_stream(upload_stream))
+                            }
+
+                            // reqwest's wasm backend does not support streaming request bodies;
+                            // pass the raw Bytes directly (CONTENT_LENGTH is set by reqwest from the body length).
+                            #[cfg(target_family = "wasm")]
+                            {
+                                client.post(url).with_extension(Api(api_tag)).body(serialized_data)
```
assafvayner, line 704: claude generated — Behavior change: native and wasm `upload_xorb` are now a single code path. The previous separate wasm impl posted the body directly and skipped both the connection permit and `RetryWrapper`/adaptive-concurrency. wasm now flows through the same retry + permit logic; it differs only in the request body (raw `Bytes`, since reqwest's wasm backend can't stream request bodies — `CONTENT_LENGTH` is set by reqwest from the body) and emits a single bulk `report_progress(n_transfer_bytes)` after success instead of per-chunk streaming updates. Net improvement, but note the wasm upload path now newly exercises the retry/adaptive-concurrency machinery.

xet_data/src/processing/shard_interface/wasm.rs
```diff
@@ -0,0 +1,167 @@
...
+        }
+        Ok(true)
+    }
+
+    /// Query the in-memory session shard first, then the global-dedup cache.
+    /// The third tuple element ("already uploaded") is `false` for session
+    /// hits and `true` for dedup-cache hits, since cache entries come from
+    /// shards the server already has. Cache hits also bump the matched
+    /// shard to the most-recently-used position.
+    pub async fn chunk_hash_dedup_query(
```
assafvayner, line 102: claude generated — Correctness note: the `(usize, FileDataSequenceEntry, bool)` "already uploaded" contract matches native (`shard_interface/native.rs::chunk_hash_dedup_query`): session-shard hit → `false`, global-dedup-cache hit → `true`. wasm correctly omits native's third "resumed session" branch (no resume support on wasm) and mirrors only the two branches that apply. A miss here costs only a re-upload (never corruption), so the in-memory + bounded-LRU divergence from native's on-disk shard managers is low-risk for correctness.

.github/actions/build-wasm/action.yml
```diff
@@ -14,11 +14,31 @@ runs:
...
+    - name: Cache wasm-bindgen-cli and wasm-pack
+      uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7  # v5
+      with:
+        path: |
+          ~/.cargo/bin/wasm-bindgen
+          ~/.cargo/bin/wasm-bindgen-test-runner
+          ~/.cargo/bin/wasm-pack
+          ~/.cargo/.crates.toml
+          ~/.cargo/.crates2.json
+        key: ${{ runner.os }}-${{ runner.arch }}-cargo-tools-wbg-0.2.121-wp-0.14.0-rustc-${{ steps.rustc-version.outputs.version }}
```
seanses: "wasm-bindgen-cli" and "wasm-pack" are stable native host binaries that don't depend on nightly-only features. So embedding in the rustc nightly version is useless: compilation produces the same artifact as long as the binary release version is identical (e.g. 0.2.121).

.github/actions/build-wasm/action.yml
```diff
@@ -29,6 +49,14 @@ runs:
       working-directory: wasm/hf_xet_wasm
       run: |
         ./build_wasm.sh
+    - name: Check hf-xet compiles for wasm32-unknown-unknown
+      shell: bash
+      run: |
+        CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUSTFLAGS="-C target-feature=+atomics,+bulk-memory,+mutable-globals --cfg getrandom_backend=\"wasm_js\"" \
```
seanses: Let's similarly put the build details under the "hf-xet" package (xet_pkg) "build_wasm.sh". And comparing this to "wasm/hf_xet_wasm/build_wasm.sh", a lot of RUSTFLAGS seem missing.

.github/actions/cache-rust-build/action.yml
```diff
@@ -11,6 +11,6 @@ runs:
           ~/.cargo/git
           target
           hf_xet/target
-          wasm/hf_xet_wasm/target
           wasm/hf_xet_thin_wasm/target
+          wasm/hf_xet_wasm/target
```
seanses, line 15: Seems meaningless reorder but alright..

.github/workflows/ci.yml
```diff
@@ -183,6 +183,16 @@ jobs:
...
+    # Non-blocking: this job runs the full wasm test suite (version-pin check,
+    # wasm builds, the `cargo check -p hf-xet` compile gate, and all browser
+    # smoke tests) on every PR and main push, but a failure here must never
+    # fail the CI run or gate merges/releases. wasm is an additive target, so
+    # regressions surface as a red "Build WASM" job (visible signal) without
+    # turning the overall run red. Nothing `needs:` this job and the release
+    # workflows are independent, so it blocks no other jobs or releases either.
+    # (The per-step `continue-on-error` on the network smokes below keeps the
+    # job green for hub/CAS blips, so a red job means a real wasm regression.)
+    continue-on-error: true
```
seanses: This is contradictory to our expectation: if we ship WASM compatible hf-xet we don't want to break it in any version.

.github/workflows/ci.yml
```diff
@@ -204,6 +214,108 @@ jobs:
         run: |
           # the build steps would update Cargo.lock if it is out of date
           test -z "$(git status --porcelain Cargo.lock)" || (echo "hf_xet_wasm Cargo.lock has uncommitted changes!" && exit 1)
+      - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444  # v5.0.0
+        with:
+          node-version: '20'
```
seanses: Use "24". Github actions are phasing out "20", so should we

.github/workflows/ci.yml
```diff
@@ -204,6 +214,108 @@ jobs:
...
+      - name: Install WASM CI smoke test deps
```
seanses, line 208: It's better if we extract this big block of integration test driver into a shell script file or whatever reasonable. This CI definition doesn't need to know all the details, only if they pass or not.

.github/workflows/ci.yml
```diff
@@ -204,6 +214,108 @@ jobs:
...
+        continue-on-error: true
+        env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
+        working-directory: wasm/ci-smoke
+        run: node run.mjs download-multi
+      - name: WASM upload smoke test (Hub xet-write-token + CAS push)
+        # uploadBytes + commit() against prod hub/CAS. Non-blocking.
+        continue-on-error: true
+        env:
+          HF_SMOKE_TEST_TOKEN: ${{ secrets.HF_SMOKE_TEST_TOKEN }}
```
seanses: Why requiring two tokens ("HF_SMOKE_TEST_TOKEN" and "HF_TOKEN")? And I don't see "HF_TOKEN" set in our repo secrets settings, seems useless

api_changes/update_260316_xet_config_get_set_python.md
```diff
@@ -85,7 +85,9 @@ This now drives:
 - `with_config` / `get` dispatch
 - key enumeration
 
-`system_monitor` remains cfg-gated and is appended separately for non-wasm targets.
+`system_monitor` is appended separately to the dispatch sites; the config
```
seanses: No need to touch this file that describes a previous commit.

api_changes/update_260515_wasm_target_support.md
```diff
@@ -0,0 +1,225 @@
...
+`hf_xet_wasm` is a separate workspace under `[workspace.exclude]` and pins
+`wasm-bindgen = "=0.2.121"`, `wasm-bindgen-futures = "0.4"`, `js-sys = "0.3"`
+independently. The main workspace pins the same versions under
+`[workspace.dependencies]` (consumed by `xet_runtime` via `workspace = true`).
+**Bump both Cargo.toml files together** with the `WASM_BINDGEN_VERSION` in
+`wasm/hf_xet_wasm/build_wasm.sh` and the cargo-tools cache key in
+`.github/actions/build-wasm/action.yml`.
+
+The previous separate `wasm/hf_xet_wasm_download` and `wasm/hf_xet_wasm_upload`
+crates have been combined back into `wasm/hf_xet_wasm`.
```
seanses: Looks like this talks about some intermediate work in this PR, ask AI to trim such statements.

api_changes/update_260515_wasm_target_support.md
```diff
@@ -0,0 +1,225 @@
...
+### `xet_data::processing::data_client` selective gating
+
+`clean_bytes` remains available on wasm — it has no filesystem
+dependencies and just feeds the byte slice into the `FileUploadSession`
+clean handle. `clean_file`, `hash_files_async`, and the private
+`hash_single_file` are `#[cfg(not(target_family = "wasm"))]`: they open
+files from disk, and `hash_files_async` additionally routes through
+`XetRuntime::spawn_blocking` (which on wasm runs `f` inline anyway —
+unsuitable for the parallel-hash use case). Wasm consumers that need
+hashing must drive it from JS (e.g. `crypto.subtle.digest`) or feed
```
seanses: What is `crypto.subtle.digest`? That's not the Xet file hash.

wasm/hf_xet_thin_wasm/Cargo.toml
```diff
@@ -4,16 +4,16 @@ version = "1.4.0"
 edition = "2024"
 
 [lib]
-crate-type = ["cdylib", "rlib"]
+crate-type = ["cdylib"]
```
seanses, line 7: What's the reasoning behind removing "rlib"?

xet_runtime/src/utils/mod.rs
```diff
@@ -11,20 +11,13 @@ pub use configuration_utils::is_high_performance;
 
 #[cfg(not(target_family = "wasm"))]
 mod file_paths;
+#[cfg(target_family = "wasm")]
```
seanses: Instead of this, I'm thinking to completely cfg-gate XetConfig values that is of type "TemplatedPathBuf", so we don't need to provide a dummy "TemplatedPathBuf" for wasm which is never used.

xet_runtime/Cargo.toml
```diff
@@ -59,6 +58,11 @@ tokio = { workspace = true, features = [
     "rt",
     "time",
 ] }
+tokio_with_wasm = { workspace = true, features = ["rt", "sync", "time"] }
+web-time = { workspace = true }
+wasm-bindgen = { workspace = true }
```
seanses: Does "xet_runtime" need "wasm-bindgen", "wasm-bindgen-futures" and "js-sys"? I thought they are used to generate JS bindings, i.e. in "wasm/*"

seanses (commented):
first batch of comments

xet_runtime/src/core/runtime.rs
```diff
@@ -179,90 +117,38 @@ pub struct XetRuntime {
 
 #[cfg(not(target_family = "wasm"))]
 fn system_monitor_for_config(config: &XetConfig) -> Option<SystemMonitor> {
-    config
-        .system_monitor
-        .enabled
-        .then(|| {
-            SystemMonitor::follow_process(config.system_monitor.sample_interval, config.system_monitor.log_path.clone())
-                .ok()
```
seanses, line 187: If we want to inspect the error, we can just add `.inspect_err(|e| debug!(...))` before `.ok()`, instead of rewritting the function

xet_runtime/src/core/runtime.rs
```diff
@@ -3,15 +3,18 @@ use std::collections::HashMap;
 use std::fmt::Display;
 use std::future::Future;
 use std::panic::AssertUnwindSafe;
```
seanses: This file contains so many cfg-gated code and it's becoming difficult to read, and it still contains many code path that won't be used by WASM at all (e.g. all member fields of `XetRuntime`). A `XetRuntime` for WASM target can be extremely simple, how about splitting `XetRuntime` for WASM into a separate file, and keeping this only for native target: https://github.com/huggingface/xet-core/pull/873

seanses (commented):
Regarding `XetRuntime`

xet_client/src/cas_client/adaptive_concurrency/controller.rs
```diff
@@ -674,7 +676,7 @@ impl ConnectionPermit {
                 0.0
             };
 
-            tokio::spawn(async move {
+            tokio::task::spawn(async move {
```
seanses: Just double check if this is expected?

xet_data/src/file_reconstruction/file_reconstructor.rs
```diff
@@ -149,6 +153,7 @@ impl FileReconstructor {
     ///
     /// The writer receives data starting from its current position (position 0
     /// for a fresh writer), regardless of the byte range being reconstructed.
+    #[cfg(not(target_family = "wasm"))]
```
seanses, line 157: I think this function can be WASM compatible?

xet_data/src/processing/file_download_session.rs
```diff
@@ -346,6 +256,104 @@ impl FileDownloadSession {
...
+    /// Downloads a byte range of a file and writes it to the provided writer.
+    ///
+    /// The provided `source_range` is interpreted against the original file; output
+    /// starts at the writer's current position. Accepts any `RangeBounds<u64>`:
+    /// `4..12`, `5..`, `..100`, or `..` (full file).
+    ///
+    /// This path does not acquire the session-level file download semaphore.
+    #[instrument(skip_all, name = "FileDownloadSession::download_to_writer",
+        fields(hash = file_info.hash(), range_start = tracing::field::Empty, range_end = tracing::field::Empty))]
+    pub async fn download_to_writer<W: Write + Send + 'static>(
```
seanses, line 325: It seems `download_to_writer` can be made WASM compatible

xet_data/src/error.rs
```diff
@@ -92,6 +96,13 @@ pub enum DataError {
 
 pub type Result<T> = std::result::Result<T, DataError>;
 
+#[cfg(target_family = "wasm")]
+impl From<tokio_with_wasm::task::JoinError> for DataError {
+    fn from(err: tokio_with_wasm::task::JoinError) -> Self {
+        DataError::WasmTaskJoinError(Arc::new(err))
```
seanses: This takes the ownership of `err`, is `Arc` needed?

xet_pkg/src/xet_session/mod.rs
```diff
@@ -1,5 +1,20 @@
 //! Session-based file upload and download API for XetHub / HuggingFace Hub.
 //!
+//! ## WASM availability
```
seanses: This section "## WASM availability" should be placed at the bottom

xet_pkg/src/xet_session/session.rs
```diff
@@ -168,6 +171,13 @@ impl XetSessionBuilder {
     ///
     /// Each build creates a fresh [`XetContext`] around the selected runtime, so sessions
     /// can share the same execution backend while keeping independent config and common state.
+    ///
+    /// On wasm no tokio runtime is instantiated at all: `tokio_handle` is statically
+    /// `None` and `XetContext::with_config` creates a stub `XetRuntime`. The wasm
+    /// bridge variants in `task_runtime.rs` `.await` futures directly via
+    /// `wasm_bindgen_futures::spawn_local` and never call back into
+    /// `XetRuntime::handle()`, so the `XetContext` is purely a bookkeeping object
+    /// (config + common state + cancellation tokens).
```
seanses: No need to mention implementation details, just mention on WASM no tokio runtime is instantiated, and all tasks are spawned onto JavaScript event loop

xet_pkg/src/xet_session/task_runtime.rs
```diff
@@ -154,6 +157,130 @@ impl TaskRuntime {
...
+impl<T: Send> MaybeSend for T {}
+#[cfg(target_family = "wasm")]
+pub(super) trait MaybeSend {}
+#[cfg(target_family = "wasm")]
+impl<T> MaybeSend for T {}
+
+// Async task bridging, shared across targets. The state-machine logic lives
+// here once; only `run_inner_async` (XetRuntime offload on native, inline
+// select on wasm) differs per target.
+impl TaskRuntime {
```
seanses: This block is not cfg-gated, we can merge these functions into the above `impl TaskRuntime` block

xet_pkg/src/lib.rs
```diff
@@ -42,6 +42,31 @@
...
+//!
+//! For a hand-runnable browser example and a `#[wasm_bindgen]` wrapper
+//! pattern, see the example crate `wasm/hf_xet_wasm/` in the repository.
+//! That crate is an example / smoke-test wrapper, not a published browser
+//! SDK — real browser consumers should depend on this crate directly with
+//! their own `#[wasm_bindgen]` glue.
+//!
+//! See `api_changes/update_260515_wasm_target_support.md` for the full set
+//! of wasm-only differences and the patterns wasm-reachable code must
+//! follow.
```
seanses: Drop this section. Comments here should be self-contained and are shown on the docs.rs page. There's no way for users to see the "wasm" or "api_changes" directories.

seanses (commented):
Another batch of mostly nit issues

xet_runtime/src/core/runtime.rs
assafvayner: Good catch — this is now a no-op. The function rewrite was reverted in `008e8ebe` (`revert(wasm): restore main's system_monitor config gating and native as-casts`), so `system_monitor_for_config` (now in `xet_runtime/src/core/runtime/native.rs` after the runtime split) is byte-identical to `main` — no rewrite, just the original `.then(...).ok().flatten()`. Nothing left to simplify here, so resolving.

wasm/hf_xet_thin_wasm/Cargo.toml
assafvayner: claude generated response

Yes — `hf_xet_thin_wasm` is a JS-binding leaf crate, built only via `wasm-pack build` (which emits the `cdylib` `.wasm`). The `rlib` output was unused here:

- Nothing in the workspace depends on it as a Rust library — the only reference is its workspace-member entry in the top-level `Cargo.toml`.
- It has no tests (no `tests/` dir, no `#[test]`/`#[wasm_bindgen_test]` in `src`), so there is no `cargo test` / `wasm-pack test` that would need `rlib`.
- The sibling `hf_xet_wasm` is already `crate-type = ["cdylib"]`, so dropping `rlib` keeps the two consistent.

So `rlib` was just an extra unused artifact (slower builds, no consumer). Leaving it removed.

.github/actions/cache-rust-build/action.yml
assafvayner: I think claude wanted to sort them alphabetically, I think that's ok; the more dirs are in this list the better it is we have sorting. I'll leave it be.

api_changes/update_260515_wasm_target_support.md
assafvayner: claude generated response

Agreed — that reference was wrong. `crypto.subtle.digest` is the Web Crypto digest primitive (SHA-256/etc.), not the Xet file hash, which is a content-defined-chunking merkle hash (`Chunker` → chunk hashes → `merklehash::file_hash`). Removed it; the doc now points wasm consumers at the `XetFileInfo` returned by `clean_bytes` — the only way to obtain the Xet hash on wasm. Fixed in a37faf65.

xet_runtime/Cargo.toml
assafvayner: claude generated response

Correct, `xet_runtime` did not use any of them (no `use`, no `#[wasm_bindgen]`, no path references — only a doc-comment mention). Removed `wasm-bindgen`, `wasm-bindgen-futures`, and `js-sys` from its wasm-target deps. They remain available transitively (via `reqwest` / `tokio_with_wasm` / `web-time`) and are still declared directly by the `wasm/*` binding crates that actually generate the JS bindings. wasm build still passes. Fixed in a37faf65.

xet_data/src/file_reconstruction/file_reconstructor.rs
assafvayner: claude generated response

Yes — made it wasm-compatible. It was gated because it goes through the `SequentialWriter` constructor that spawns a background blocking OS thread (`spawn_blocking` + `SyncWriterThread`); on wasm `spawn_blocking` runs inline and would deadlock waiting for data the event loop has not produced yet. The wasm variant instead drives the already-wasm-compatible `reconstruct_to_stream()` and copies each chunk into the writer inline (native keeps the background-thread path for vectored writes). Covered by a new `download-to-bytes` wasm smoke scenario. Done in 5aefebe0.

xet_data/src/processing/file_download_session.rs
assafvayner: claude generated response

Done. Moved `download_to_writer` out of the native-only `impl` block into a cross-target one — it writes to a `Write` sink and needs no filesystem (the truly fs-bound `download_file` / `download_file_with_id` stay native). It calls the now-wasm-compatible `reconstruct_to_writer` (sibling thread). Added a JS-reachable path to smoke-test it on wasm: `XetDownloadStreamGroup::download_to_bytes` (xet_pkg) → `downloadToBytes` (hf_xet_wasm) → a `download-to-bytes` ci-smoke scenario. Verified end-to-end: built the wasm and ran the scenario in headless Chromium against prod hub+CAS → PASS, sha256 matches the pinned file. Done in 5aefebe0.

.github/actions/build-wasm/action.yml
assafvayner: claude generated

Moved the inline `cargo check -p hf-xet` step out of the composite action into `xet_pkg/build_wasm.sh`, so the hf-xet package builds the same way the wasm crates do (each step just runs its own `build_wasm.sh`). The action step is now `working-directory: xet_pkg` + `run: ./build_wasm.sh`.

On the missing RUSTFLAGS: that step is a `cargo check`, which only emits metadata and never links — so the `-C link-arg=...` flags from `wasm/hf_xet_wasm/build_wasm.sh` (`--shared-memory`, `--import-memory`, the TLS/heap `--export`s) are no-ops for it and are intentionally omitted (documented in the script). The flags that actually affect a check are kept in sync: `-C target-feature=+atomics,+bulk-memory,+mutable-globals`, `--cfg getrandom_backend="wasm_js"`, and `-Z build-std=std,panic_abort`.

Done in cc7924cd.

.github/workflows/ci.yml
assafvayner: claude generated

Dropped `HF_TOKEN` and consolidated every smoke scenario (download + upload) on `HF_SMOKE_TEST_TOKEN` (a097860d). You were right that `HF_TOKEN` was useless: it isn't in the repo secrets, uploads already preferred `HF_SMOKE_TEST_TOKEN`, and the read scenarios run against a public pinned repo so they work anonymously.

Verified in CI run 27659876426: the Build WASM job passed with all 17 smoke scenarios green, including the 11 write-scoped upload scenarios that hard-require `HF_SMOKE_TEST_TOKEN` — confirming it is the configured token and `HF_TOKEN` was redundant.

.github/workflows/ci.yml
assafvayner: claude generated

Extracted the driver: `run.mjs` now has a `node run.mjs all` mode (per-scenario logic factored into `runScenario()`) that runs every scenario in the `SCENARIOS` table, prints a pass/fail summary plus a `::error` annotation per failure, retries transient network blips, and exits non-zero if any fail. `node run.mjs <name>` still runs a single scenario for local debugging.

The CI block now collapses the per-scenario steps into a single `node run.mjs all` step — the workflow just sees pass/fail; per-scenario detail lives in `run.mjs`. Also removed the noisy comments under the Build WASM job.

Verified locally end-to-end against prod hub + CAS with a write-scoped token: 18/18 scenarios pass. Done in bb51e14a.

api_changes/update_260316_xet_config_get_set_python.md
seanses: Done — the edit to this file was reverted; it now matches main.

xet_client/src/cas_client/adaptive_concurrency/controller.rs
seanses: Yes, this was reverted — current code uses `tokio::spawn` (not `tokio::task::spawn`). Both work since `tokio_with_wasm::alias` re-exports `spawn` at the top level, but keeping the original form is cleaner.

xet_pkg/src/xet_session/session.rs
seanses: Done — simplified to one line: "On wasm no tokio runtime is instantiated; all tasks are spawned onto the JavaScript event loop."

seanses (approved):
LGTM, thanks for working on this!

xet_pkg/src/xet_session/upload_commit.rs
```diff
@@ -512,7 +512,6 @@ impl XetUploadCommit {
     ///
     /// - `file_path`: path to the file. Resolved to an absolute path so the upload is unaffected by later
     ///   working-directory changes.
-    /// - `sha256`: SHA-256 handling policy for this file.
```
seanses: Don't delete this

PR #842, opened 2026-05-13
feat(xet_client): fall back to single-range GETs when a multi-range URL 403s

assafvayner:
## Context

V2 reconstruction (`QueryReconstructionResponseV2`) packs multiple byte ranges per xorb into a single signed URL (`XorbMultiRangeFetch.ranges` may have `len > 1`). One storage backend rejects multi-range Range headers at the edge with `403 Invalid Range Specified` before the request reaches the origin. Today's retry path on 403 (`xet_client/src/cas_client/remote_client.rs`) refreshes the URL via `URLProvider::refresh_url` and retries with the same multi-range shape, so retries are doomed against that backend. This PR adds client-side resilience that is independent of any server-side fix.

## What changed

- `xet_client/src/cas_client/interface.rs`: new `SingleRangeURLProvider` wrapping a multi-range `URLProvider` plus a range index, exposing exactly one of the inner ranges as a single-range fetch. `refresh_url` is delegated to the inner provider so signature-refresh semantics are unchanged. A `split_into_single_range_providers` helper fans a multi-range provider out into N single-range providers. The `URLProvider` trait itself is unchanged — this is purely additive.
- `xet_client/src/cas_client/remote_client.rs`: `RemoteClient::get_file_term_data` is split into a per-provider `fetch_xorb_via_provider` helper and the public entry point. When the multi-range fetch returns a `ClientError` with status `403 Forbidden` after the `with_retry_on_403` retry budget is exhausted and `url_ranges.len() > 1`, the entry point fans the request out into N independent single-range GETs (each with its own retry budget and download permit), concatenating the decompressed bytes + chunk indices via `append_chunk_segment` in the same order the multi-range fetch would have produced.

## Behavior

- Single-range URLs and successful multi-range URLs are untouched (no extra requests on healthy paths).
- The fallback only triggers after the multi-range retry budget is exhausted with a 403; it never triggers on transient errors, timeouts, or any non-403 failure.
- Scope of the fallback is per failing `XorbMultiRangeFetch`, not the whole file.
- Tradeoff: a failing xorb now costs N storage requests instead of 1. This is acceptable since (a) the alternative is failing the download, and (b) the trigger condition is narrow.

## Test plan

Added three unit tests in `xet_client/src/cas_client/remote_client.rs` against a wiremock storage backend:
- `test_get_file_term_data_multi_range_happy_path`: multi-range 206 multipart/byteranges succeeds first try; verifies no fallback is invoked and bytes round-trip.
- `test_get_file_term_data_multi_range_403_falls_back_to_single_range`: multi-range URL always 403s; each subrange GET succeeds; verifies the fallback produces the correctly-concatenated bytes.
- `test_get_file_term_data_single_range_fallback_errors_propagate`: every GET 403s; verifies the error propagates with 403 status, the request count is bounded (no infinite loop).

Local verification:
- `cargo +nightly fmt --all -- --check`
- `cargo clippy -r -p xet-client -- -D warnings` (release CI invocation)
- `cargo test -p xet-client --features simulation`
- `cargo test -p xet-data`
- `cargo +nightly check --target wasm32-unknown-unknown` for both `wasm/hf_xet_thin_wasm` and `wasm/hf_xet_wasm`

### DIFF
--- a/xet_client/src/cas_client/interface.rs
+++ b/xet_client/src/cas_client/interface.rs
@@ -69,9 +69,7 @@ impl URLProvider for SingleRangeURLProvider {
 ///
 /// Returns an empty vec if the inner provider has no ranges (caller should treat this as an
 /// error condition).
-pub async fn split_into_single_range_providers(
-    inner: Arc<Box<dyn URLProvider>>,
-) -> Result<Vec<Box<dyn URLProvider>>> {
+pub async fn split_into_single_range_providers(inner: Arc<Box<dyn URLProvider>>) -> Result<Vec<Box<dyn URLProvider>>> {
     let (_, ranges) = inner.retrieve_url().await?;
     Ok((0..ranges.len())
         .map(|i| Box::new(SingleRangeURLProvider::new(inner.clone(), i)) as Box<dyn URLProvider>)
--- a/xet_client/src/cas_client/interface.rs
+++ b/xet_client/src/cas_client/interface.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use bytes::Bytes;
 use xet_core_structures::merklehash::MerkleHash;
 use xet_core_structures::metadata_shard::file_structs::MDBFileInfo;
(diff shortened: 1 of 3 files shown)


assafvayner:
not needed

PR #855, opened 2026-05-26
feat: expose FileUploadSession low-level functions

jgodlew:
A few functions on the FileUploadSession are useful for clients building composite files that aren't natively supported yet in xet-core. For example, when concatenating multiple existing xet files into a single one, having the ability to pull chunk/reconstruction info to build the composite file's reconstruction info and then register that with the session are needed. 

---

> [!NOTE]
> **Low Risk**
> Small visibility and guard changes on upload-session helpers; no auth or data-path redesign.
> 
> **Overview**
> Makes **`FileUploadSession`** usable for advanced composite uploads by exposing a few session APIs that were previously crate-private.
> 
> **`register_composed_file`** is now **public** and rejects calls after **`finalize()`** via **`check_not_finalized()`**, matching other mutating entry points. A new **`client()`** accessor returns a clone of the session’s CAS **`Client`**, so callers can fetch chunk/reconstruction data while building an **`MDBFileInfo`** plan before registering it on the session.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d264d3c0888c32b65acae84fe991016d93d90632. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/xet_data/src/processing/file_upload_session.rs
+++ b/xet_data/src/processing/file_upload_session.rs
@@ -13,6 +13,8 @@ use tokio::sync::Mutex;
 use tokio::task::{JoinHandle, JoinSet};
 use tracing::{Instrument, Span, info_span, instrument};
 use xet_client::cas_client::{Client, ProgressCallback};
+use xet_client::cas_types::{FileChunkHashesResponse, FileRange};
+use xet_core_structures::merklehash::MerkleHash;
 use xet_core_structures::metadata_shard::file_structs::MDBFileInfo;
 use xet_core_structures::xorb_object::SerializedXorbObject;
 use xet_runtime::core::XetContext;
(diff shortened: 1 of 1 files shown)


XciD (approved):

seanses (commented):
I'm OK to get this in -- but the added APIs don't seem to fit into this struct. If creating a FileUploadSession is truly needed, can we just expose its `client` field?

jgodlew:
> I'm OK to get this in -- but the added APIs don't seem to fit into this struct. If creating a FileUploadSession is truly needed, can we just expose its `client` field?

Yeah, that makes sense, I exposed the inner client instead of having those methods. It is still desired to have this on the FileUploadSession since I am still uploading a file, but I need access to call APIs on the XetServer to download existing file info (i.e. MerkleSubtrees and reconstruction info) from files already uploaded earlier.

PR #856, opened 2026-05-28
hf-xet BackgroundProgress poller with condvar shutdown and final callback

seanses:
## Problems addressed

1. **Missing final progress update** — Progress was only emitted on a fixed poll interval, so the last snapshot could be skipped when upload/download finished between ticks (e.g. UI stuck below 100%).
2. **Panic on shutdown** — The background progress thread used `Python::attach` while the interpreter was tearing down, which can panic when the GIL is no longer available. Resolves https://github.com/huggingface/huggingface_hub/pull/4116#discussion_r3272527522

## Approach

- Introduce `BackgroundProgress` (`hf_xet/src/background_progress.rs`): a shared poller for upload commits and download groups.
- Poll loop uses `Condvar::wait_timeout_while` so shutdown returns immediately when the flag is already set.
- `stop_and_join(py)` sets the shutdown flag, notifies, then `join` the background thread.
- After `wait_to_finish` / `commit_blocking` completes, `stop_and_emit` runs one final callback on the caller thread with the latest snapshot.
- Use `Python::try_attach` in the poller loop so shutdown does not crash if the interpreter is gone.

## Test plan

- [x] `cargo test` in `hf_xet` — unit test spawns a poller with a 60s interval and calls `stop_and_join` immediately (no sleep); asserts join finishes in under 2s (covers pre-set shutdown + `wait_timeout_while`, not only condvar wake while blocked).
- [x] Manual: upload/download with a progress callback and confirm final 100% (or terminal state) is reported
- [x] Manual: run `hf upload/download` cli so Python interpreter shuts down immediately after `wait_to_finish` returns, confirm no panics.

---

> [!NOTE]
> **Medium Risk**
> Changes background-thread and GIL interaction for progress callbacks and makes abort(py) required on two pyclasses; mistakes could cause missed updates, deadlocks, or shutdown panics, but scope is limited to optional progress reporting.
> 
> **Overview**
> Replaces ad-hoc progress polling threads on **`XetUploadCommit`** and **`XetFileDownloadGroup`** with a shared **`BackgroundProgress`** helper that polls Rust progress atomics and calls the Python callback via **`Python::try_attach`** (avoids panics during interpreter shutdown).
> 
> The poller waits on a **condvar** with a timeout so **`stop_and_join`** returns immediately instead of sleeping the full interval; joining releases the **GIL** via **`py.detach`** to avoid deadlocks. On successful **`wait_to_finish`**, **`stop_and_emit`** runs **one final** callback on the caller thread so UIs can reach 100%; **`abort`** only stops the poller (no extra callback). **`abort`** now takes **`py`** so the background thread is joined cleanly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 79b00ae33b0c8027e2426b1f77db1e8a204a264f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/hf_xet/src/background_progress.rs
+++ b/hf_xet/src/background_progress.rs
@@ -1,0 +1,168 @@
+//! Background progress callbacks for upload commits and download groups.
+//!
+//! A dedicated thread polls Rust-side progress atomics and invokes a Python
+//! callable via [`Python::try_attach`].  The poller sleeps on a condvar so
+//! [`BackgroundProgress::stop_and_join`] can wake it immediately instead of
+//! waiting for the full poll interval.
+
+use std::collections::HashMap;
+use std::sync::{Arc, Condvar, Mutex};
+use std::thread::JoinHandle;
+use std::time::Duration;
+
+use pyo3::exceptions::PyRuntimeError;
+use pyo3::prelude::*;
+use xet_pkg::xet_session::{GroupProgressReport, ItemProgressReport, UniqueId};
+
+fn lock_poisoned_err() -> PyErr {
+    PyRuntimeError::new_err("progress poller lock poisoned")
+}
+
+fn poller_panicked_err() -> PyErr {
+    PyRuntimeError::new_err("progress poller thread panicked")
+}
+
+/// Background thread that polls progress and invokes a Python callback.
+pub(crate) struct BackgroundProgress {
+    /// ``false`` while the poller is running; set to ``true`` in ``stop_and_join``.
+    /// Paired with the condvar so setting shutdown wakes a thread blocked in
+    /// ``wait_timeout`` (the mutex is released for the duration of the wait).
+    wake: Arc<(Mutex<bool>, Condvar)>,
+    join: Mutex<Option<JoinHandle<PyResult<()>>>>,
+    callback: Py<PyAny>,
+}
+
+impl BackgroundProgress {
+    /// Spawn a thread that calls ``snapshot`` every ``interval_ms`` until shutdown,
+    /// the snapshot reports a terminal task state, or the callback fails.
+    pub(crate) fn spawn<F>(py: Python<'_>, callback: Py<PyAny>, interval_ms: u64, mut snapshot: F) -> Self
+    where
+        F: FnMut() -> (GroupProgressReport, HashMap<UniqueId, ItemProgressReport>, bool) + Send + 'static,
+    {
+        let wake = Arc::new((Mutex::new(false), Condvar::new()));
+        let wake_for_thread = Arc::clone(&wake);
+        let callback_for_thread = callback.clone_ref(py);
+        let interval = Duration::from_millis(interval_ms);
+
+        let join = std::thread::spawn(move || {
+            progress_thread_loop(callback_for_thread, interval, &wake_for_thread, &mut snapshot)
+        });
+
+        Self {
+            wake,
+            join: Mutex::new(Some(join)),
+            callback,
+        }
+    }
+
+    /// Request shutdown, wake the poller if it is sleeping, and block until the thread exits.
+    pub(crate) fn stop_and_join(&self) -> PyResult<()> {
+        {
+            let mut guard = self.wake.0.lock().map_err(|_| lock_poisoned_err())?;
+            *guard = true;
+        }
+        self.wake.1.notify_one();
+        let handle = self.join.lock().map_err(|_| lock_poisoned_err())?.take();
+        if let Some(handle) = handle {
+            match handle.join() {
+                Ok(result) => result,
+                Err(_) => Err(poller_panicked_err()),
+            }
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Stop the poller, then invoke the callback once on the calling thread (final snapshot).
+    pub(crate) fn stop_and_emit<F>(&self, py: Python<'_>, snapshot: F) -> PyResult<()>
+    where
+        F: FnOnce() -> (GroupProgressReport, HashMap<UniqueId, ItemProgressReport>),
+    {
+        self.stop_and_join()?;
+        let (group_report, item_reports) = snapshot();
+        self.callback.call1(py, (group_report, item_reports))?;
+        Ok(())
+    }
+}
+
+fn progress_thread_loop<F>(
+    callback: Py<PyAny>,
+    interval: Duration,
+    wake: &Arc<(Mutex<bool>, Condvar)>,
+    snapshot: &mut F,
+) -> PyResult<()>
+where
+    F: FnMut() -> (GroupProgressReport, HashMap<UniqueId, ItemProgressReport>, bool),
+{
+    loop {
+        // Sleep up to `interval`, but return promptly when stop_and_join sets shutdown and notifies.
+        let shutdown = {
+            let guard = wake.0.lock().map_err(|_| lock_poisoned_err())?;
+            let (guard, _) = wake.1.wait_timeout(guard, interval).map_err(|_| lock_poisoned_err())?;
+            *guard
+        };
+
+        if shutdown {
+            break;
+        }
+
+        // Mutex is not held across snapshot or the Python callback.
+        let (group_report, item_reports, is_terminal) = snapshot();
+
+        match Python::try_attach(|py| callback.call1(py, (group_report, item_reports))) {
+            None => break, // interpreter shutting down
+            Some(Ok(_)) => {},
+            Some(Err(e)) => {
+                let _ = Python::try_attach(|py| e.print(py));
+                break;
+            },
+        }
+
+        if is_terminal {
+            break;
+        }
+    }
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use std::collections::HashMap;
+    use std::time::{Duration, Instant};
+
+    use pyo3::Python;
+    use pyo3::ffi::c_str;
+    use xet_pkg::xet_session::GroupProgressReport;
+
+    use super::*;
+
+    #[test]
+    fn stop_and_join_returns_before_poll_interval_elapses() {
+        /// Poll interval long enough that a blocking sleep would fail the test.
+        const LONG_INTERVAL_MS: u64 = 60_000;
+        const STOP_DEADLINE: Duration = Duration::from_secs(2);
+
+        Python::attach(|py| -> PyResult<()> {
+            py.run(c_str!("def _bg_progress_test_cb(_group, _items): pass"), None, None)?;
+            let callback: Py<PyAny> = py.eval(c_str!("_bg_progress_test_cb"), None, None)?.unbind();
+
+            let progress = BackgroundProgress::spawn(py, callback, LONG_INTERVAL_MS, || {
+                (GroupProgressReport::default(), HashMap::new(), false)
+            });
+
+            // Let the poller enter its timed wait on the condvar.
+            std::thread::sleep(Duration::from_millis(50));
+
+            let start = Instant::now();
+            progress.stop_and_join()?;
+            let elapsed = start.elapsed();
+
+            assert!(
+                elapsed < STOP_DEADLINE,
+                "stop_and_join took {elapsed:?}; expected prompt return (poll interval is {LONG_INTERVAL_MS}ms)"
+            );
+            Ok(())
+        })
+        .unwrap();
+    }
+}
--- a/hf_xet/src/background_progress.rs
+++ b/hf_xet/src/background_progress.rs
@@ -56,21 +56,25 @@ impl BackgroundProgress {
     }
 
     /// Request shutdown, wake the poller if it is sleeping, and block until the thread exits.
-    pub(crate) fn stop_and_join(&self) -> PyResult<()> {
+    pub(crate) fn stop_and_join(&self, py: Python<'_>) -> PyResult<()> {
         {
             let mut guard = self.wake.0.lock().map_err(|_| lock_poisoned_err())?;
             *guard = true;
         }
         self.wake.1.notify_one();
         let handle = self.join.lock().map_err(|_| lock_poisoned_err())?.take();
-        if let Some(handle) = handle {
-            match handle.join() {
-                Ok(result) => result,
-                Err(_) => Err(poller_panicked_err()),
+        // Releases the GIL while joining so the poller can acquire it inside
+        // [`Python::try_attach`] without deadlocking the caller.
+        py.detach(|| {
+            if let Some(handle) = handle {
+                match handle.join() {
+                    Ok(result) => result,
+                    Err(_) => Err(poller_panicked_err()),
+                }
+            } else {
+                Ok(())
             }
-        } else {
-            Ok(())
-        }
+        })
     }
 
     /// Stop the poller, then invoke the callback once on the calling thread (final snapshot).
@@ -78,7 +82,7 @@ impl BackgroundProgress {
     where
         F: FnOnce() -> (GroupProgressReport, HashMap<UniqueId, ItemProgressReport>),
     {
-        self.stop_and_join()?;
+        self.stop_and_join(py)?;
         let (group_report, item_reports) = snapshot();
         self.callback.call1(py, (group_report, item_reports))?;
         Ok(())
@@ -95,10 +99,13 @@ where
     F: FnMut() -> (GroupProgressReport, HashMap<UniqueId, ItemProgressReport>, bool),
 {
     loop {
-        // Sleep up to `interval`, but return promptly when stop_and_join sets shutdown and notifies.
+        // Sleep up to `interval`, but return promptly when shutdown is already set or notified.
         let shutdown = {
             let guard = wake.0.lock().map_err(|_| lock_poisoned_err())?;
-            let (guard, _) = wake.1.wait_timeout(guard, interval).map_err(|_| lock_poisoned_err())?;
+            let (guard, _) = wake
+                .1
+                .wait_timeout_while(guard, interval, |shutdown| !*shutdown)
+                .map_err(|_| lock_poisoned_err())?;
             *guard
         };
 
@@ -150,11 +157,8 @@ mod tests {
                 (GroupProgressReport::default(), HashMap::new(), false)
             });
 
-            // Let the poller enter its timed wait on the condvar.
-            std::thread::sleep(Duration::from_millis(50));
-
             let start = Instant::now();
-            progress.stop_and_join()?;
+            progress.stop_and_join(py)?;
             let elapsed = start.elapsed();
 
             assert!(
--- a/hf_xet/src/background_progress.rs
+++ b/hf_xet/src/background_progress.rs
@@ -84,7 +84,10 @@ impl BackgroundProgress {
     {
         self.stop_and_join(py)?;
         let (group_report, item_reports) = snapshot();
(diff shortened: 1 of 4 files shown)


hf_xet/src/background_progress.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    /// Stop the poller, then invoke the callback once on the calling thread (final snapshot).
+    pub(crate) fn stop_and_emit<F>(&self, py: Python<'_>, snapshot: F) -> PyResult<()>
+    where
+        F: FnOnce() -> (GroupProgressReport, HashMap<UniqueId, ItemProgressReport>),
+    {
+        self.stop_and_join(py)?;
+        let (group_report, item_reports) = snapshot();
+        self.callback.call1(py, (group_report, item_reports))?;
+        Ok(())
```
rajatarya, line 91: Potential double-fire of the final callback. Consider the sequence:

1. `commit_blocking()` returns on the caller thread.
2. Before `stop_and_emit` runs, the poller wakes from `wait_timeout_while` (timeout, not notified yet), calls `snapshot()`, observes `is_terminal == true`, invokes the callback, breaks the loop.
3. `stop_and_emit` calls `stop_and_join` (no-op, thread already exited), takes a *second* snapshot, invokes the callback again.

Net result: two callbacks for the final state. The PR description and the doc above promise one. In practice this is benign for most progress UIs (idempotent rendering), but the contract as written is stricter than what the code guarantees.

A couple of options if you want to tighten it:
- Have the poller skip the `is_terminal` callback (just `break`) and let `stop_and_emit` always own the final emission.
- Track a `final_emitted: AtomicBool` shared with the poller, and have `stop_and_emit` only fire if the poller didn't already.

First option is simpler and reads cleaner — the poller is purely "running updates," the caller thread owns the terminal one.

hf_xet/src/py_upload_commit.rs
```diff
@@ -308,13 +307,28 @@ impl PyXetUploadCommit {
...
         let inner = self.inner.clone();
-        blocking_call_with_signal_check(py, move || inner.commit_blocking())
+        let result = blocking_call_with_signal_check(py, move || inner.commit_blocking());
+        if let (Some(handles), Some(progress)) = (&self.upload_handles, &self.background_progress) {
+            // ignore any error from progress update
+            let _ = progress.stop_and_emit(py, || {
+                let item_reports = item_reports_from_upload_handles(handles);
+                (self.inner.progress(), item_reports)
+            });
+        }
```
rajatarya, line 329: When `commit_blocking()` returns an error (including `KeyboardInterrupt` propagated through `blocking_call_with_signal_check`), we still call `stop_and_emit`, which fires the Python callback with whatever snapshot we can read at that moment.

Is that intentional? I'd expect a caller catching the exception to see the *exception* and reason about state from there, not also receive a progress callback as a side effect. The same applies to `py_file_download_group.rs` line 203.

If the intent is "always emit one final terminal snapshot regardless of outcome," that's defensible — but worth saying so in the docstring. If not, gating on `result.is_ok()` would match the more conservative interpretation.

rajatarya (commented):
Hi Di — thanks for the careful writeup, the problem statement is clear and the condvar/`try_attach` refactor is a nice cleanup from the previous sleep-loop. A few things I'd like to discuss before this lands; none are blockers but the first two feel material to the stated contract.

**What I like**
- Centralizing the poller into `BackgroundProgress` removes a lot of duplicated code between upload commit and download group — good factoring.
- `wait_timeout_while` + `py.detach` around `join` are both correct now (cursor[bot] flagged the earlier versions; the current commit looks right to me).
- `Python::try_attach` is the right call for the interpreter-teardown panic — and the matching `e.print` is also wrapped in `try_attach`, which is easy to forget.
- Field-level rustdoc on `download_handles`/`progress` is helpful.

**Items to discuss**

1. **"Final callback fires exactly once" isn't quite guaranteed today.** The PR title and docstring promise one final callback, but I think we can fire it twice in the common case. See the inline on `stop_and_emit`. (Most consequential.)
2. **Final callback also fires on the error path** of `wait_to_finish` — including `KeyboardInterrupt`. That may surprise callers who expect the callback only after a successful terminal state. Worth confirming the intent. Inline on `wait_to_finish`.
3. **Public Python API change**: `abort()` now requires `py` and the Python signature picks up an implicit argument. PyO3 will inject `py` so Python-side callers are unaffected, but Rust-side callers (and any external consumers if this crate is ever embedded elsewhere) need to update. Worth a note in the changelog/release notes.
4. **Naming inconsistency between the two pyclasses**: `PyXetUploadCommit::background_progress` vs `PyXetFileDownloadGroup::progress` for the same type. Minor but it tripped me up while reading. Inline.
5. **Test coverage gap on the new contract**: the one unit test only exercises pre-set-shutdown wakeup. There's no test for the "exactly once" final-callback property or for the snapshot-then-stop race. Inline on the test.
6. **Silent error swallowing** in `wait_to_finish` (`let _ = progress.stop_and_emit(...)`). I get why — we don't want to mask the real commit result — but a callback raising might be a real user bug. At minimum I'd `tracing::warn!` the error. Inline.

Let me know what you think on (1) and (2) in particular — happy to be wrong about the race if I'm misreading the ordering.

hf_xet/src/py_upload_commit.rs
```diff
@@ -158,6 +154,9 @@ pub struct PyXetUploadCommit {
     pub(crate) inner: XetUploadCommit,
     /// Per-file handles shared with the progress thread; None when no callback was registered.
     upload_handles: Option<Arc<RwLock<Vec<XetFileUpload>>>>,
+    /// Background thread that polls progress and invokes the Python callback; None when no
+    /// callback was registered. Stopped and joined from ``wait_to_finish`` and ``abort``.
+    background_progress: Option<BackgroundProgress>,
```
rajatarya: nit: this field is named `background_progress` here but `progress` in `PyXetFileDownloadGroup` (`py_file_download_group.rs:104`). Same type, same role — would be nice to pick one. I have a slight preference for `progress` since `background_` is implied by the type name.

hf_xet/src/py_upload_commit.rs
```diff
@@ -308,13 +307,28 @@ impl PyXetUploadCommit {
...
     /// 100 ms so that Ctrl-C is delivered promptly.
+    ///
+    /// When a progress callback was registered, it is invoked once more on the
+    /// calling thread with the final snapshot after all work completes.
     pub fn wait_to_finish(&self, py: Python<'_>) -> PyResult<XetCommitReport> {
         let inner = self.inner.clone();
-        blocking_call_with_signal_check(py, move || inner.commit_blocking())
+        let result = blocking_call_with_signal_check(py, move || inner.commit_blocking());
+        if let (Some(handles), Some(progress)) = (&self.upload_handles, &self.background_progress) {
+            // ignore any error from progress update
```
rajatarya: nit: `let _ = ...` silently drops any error from the callback (including a Python exception raised by user code). I'd at least `tracing::warn!` so it shows up in logs — otherwise a buggy callback fails invisibly only on the final tick. Same at `py_file_download_group.rs:197`.

hf_xet/src/background_progress.rs
```diff
@@ -0,0 +1,172 @@
...
+            let start = Instant::now();
+            progress.stop_and_join(py)?;
+            let elapsed = start.elapsed();
+
+            assert!(
+                elapsed < STOP_DEADLINE,
+                "stop_and_join took {elapsed:?}; expected prompt return (poll interval is {LONG_INTERVAL_MS}ms)"
+            );
+            Ok(())
+        })
```
rajatarya, line 182: The test verifies the prompt-wakeup path (good — that's the main thing `wait_timeout_while` buys us), but it doesn't cover the headline contract in the PR title: "final callback." Two cases worth adding if you have appetite:

1. **Final callback fires exactly once on normal completion** — spawn poller with a counter callback, drive the snapshot to return `is_terminal=true`, call `stop_and_emit`, assert callback was invoked exactly once with the final state (or document the actual count if you accept the double-fire I noted on `stop_and_emit`).
2. **Final callback fires once on abort** — spawn poller, call `stop_and_join` (the abort path), assert the callback was *not* invoked again afterward.

These would also serve as living documentation of the intended semantics.

hf_xet/src/py_upload_commit.rs
```diff
@@ -308,13 +307,28 @@ impl PyXetUploadCommit {
...
+                let item_reports = item_reports_from_upload_handles(handles);
+                (self.inner.progress(), item_reports)
+            });
+        }
+        result
     }
 
     /// Cancel all active uploads in this commit.
-    pub fn abort(&self) -> PyResult<()> {
+    pub fn abort(&self, py: Python<'_>) -> PyResult<()> {
```
rajatarya, line 334: The Python signature of `abort()` changes from `abort()` to `abort()` (PyO3 fills in `py` implicitly), so Python callers are fine — but anyone calling this from Rust (or relying on the old signature for FFI stubs / typeshed) will need to update. Worth a line in the changelog / release notes for the next `hf_xet` bump. Not blocking; just flagging.

hf_xet/src/background_progress.rs
seanses: Indeed it can double fire the final progress, I adopted the first option.

hf_xet/src/py_upload_commit.rs
seanses: The main intention here is to stop the progress background thread, whether the result is Ok or Err, so I updated this logic to call `stop_and_emit` when result is Ok and call `stop_and_join` when result is Err.

hf_xet/src/py_upload_commit.rs
seanses: The main purpose of this is to stop the progress thread, and ignore any error that may happen during the join. I'd rather not `warn!` here to confuse users. Regarding "a buggy callback", I do log to Python if the callback fails in the background progress loop:
```
match Python::try_attach(|py| callback.call1(py, (group_report, item_reports))) {
            None => break, // interpreter shutting down
            Some(Ok(_)) => {},
            Some(Err(e)) => {
                let _ = Python::try_attach(|py| e.print(py));
                break;
            },
        }
```

hf_xet/src/background_progress.rs
seanses: Added the suggested two unit tests.

hf_xet/src/py_upload_commit.rs
seanses: I doubt anyone will call this from Rust instead of using the hf-xet Rust crate directly, after all this is only a Python wrapper for the hf-xet Rust crate.

hf_xet/src/py_upload_commit.rs
seanses: Renamed to `progress`.

hf_xet/src/background_progress.rs
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        match Python::try_attach(|py| callback.call1(py, (group_report, item_reports))) {
+            None => break, // interpreter shutting down
+            Some(Ok(_)) => {},
+            Some(Err(e)) => {
+                let _ = Python::try_attach(|py| e.print(py));
+                break;
+            },
+        }
```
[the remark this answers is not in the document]
seanses: This is not double invocation - if `is_terminal` is false then that's not the final snapshot.

seanses:
The "Cargo Audit" CI failure requires bumping `russh` which is used under "git_xet" for integration test. Addressing that in a separate PR https://github.com/huggingface/xet-core/pull/860

hf_xet/src/py_upload_commit.rs
```diff
@@ -308,13 +307,32 @@ impl PyXetUploadCommit {
...
     /// 100 ms so that Ctrl-C is delivered promptly.
+    ///
+    /// When a progress callback was registered, it is invoked once more on the
+    /// calling thread with the final snapshot after all work completes.
     pub fn wait_to_finish(&self, py: Python<'_>) -> PyResult<XetCommitReport> {
         let inner = self.inner.clone();
-        blocking_call_with_signal_check(py, move || inner.commit_blocking())
+        let result = blocking_call_with_signal_check(py, move || inner.commit_blocking());
+        if let (Some(handles), Some(progress)) = (&self.upload_handles, &self.progress) {
+            // ignore any error from progress update
```
rajatarya: This `let _ = …` actually swallows two distinct error classes that I think are worth distinguishing:

- `stop_and_join` failures (lock poisoning, poller-thread panic) — these are pure diagnostics, never something the Python caller should see, but losing them silently makes future debugging harder. A `tracing::warn!` would preserve them.
- `stop_and_emit` failures, which include a `PyErr` raised by the **final** callback. The poller loop in `background_progress.rs` lines 126–129 catches mid-flight callback errors and prints them via `e.print(py)`. This path doesn't — so a Python callback that only raises on the terminal snapshot will produce no output anywhere.

What do you think about splitting the two branches and either calling `e.print(py)` on the emit branch's `Err` (matching the poller's behavior) or at least logging both via `tracing`? Same comment applies to `py_file_download_group.rs:197`.

hf_xet/src/py_file_download_group.rs
```diff
@@ -188,13 +187,32 @@ impl PyXetFileDownloadGroup {
...
     /// 100 ms so that Ctrl-C is delivered promptly.
+    ///
+    /// When a progress callback was registered, it is invoked once more on the
+    /// calling thread with the final snapshot after all work completes.
     pub fn wait_to_finish(&self, py: Python<'_>) -> PyResult<XetDownloadGroupReport> {
         let group = self.inner.clone();
-        blocking_call_with_signal_check(py, move || group.finish_blocking())
+        let result = blocking_call_with_signal_check(py, move || group.finish_blocking());
+        if let (Some(handles), Some(progress)) = (&self.download_handles, &self.progress) {
+            // ignore any error from progress update
```
rajatarya: Mirror of the comment on `py_upload_commit.rs:317` — same `let _ = …` swallows both shutdown diagnostics and final-callback exceptions. Whatever you settle on there, please apply here too.

hf_xet/src/background_progress.rs
```diff
@@ -0,0 +1,223 @@
...
+
+            let n = call_count(py, &ns)?;
+            assert_eq!(n, 1, "expected exactly one final callback, got {n}");
+            Ok(())
+        })
+        .unwrap();
+    }
+
+    #[test]
+    fn abort_path_does_not_invoke_callback() {
```
rajatarya: nit: this test actually exercises `stop_and_join` directly — `abort()` lives in the call-site structs and isn't touched here. `stop_and_join_does_not_invoke_callback` would more accurately describe what's being tested.

rajatarya (commented):
Thanks Di — clean refactor that lifts the inline polling loops into a focused helper. The condvar-based shutdown is the right primitive here, the `try_attach` fix for interpreter teardown is a real bug squashed, and the prompt-shutdown test (60s interval, 2s deadline) is a nice guard against regressions.

A few items worth a look — none blocking:

1. **Final-callback exceptions are silently dropped now** — see comment on `py_upload_commit.rs:317`. In the old loop, callback exceptions on the terminal tick were printed via `e.print(py)`. In the new flow, the final callback runs inside `stop_and_emit` and its `PyErr` is discarded at the `let _ = …` in `wait_to_finish`.

2. **`stop_and_join` errors are also dropped** at the same call sites — that path masks lock-poisoning / poller-panic diagnostics. A `tracing::warn!` would preserve those without surfacing them to Python.

3. **Behavioral FYI:** the final callback now fires on the *caller* thread, not the poller thread. Python callbacks are GIL-serialized so this should be transparent, but worth keeping in the back pocket if anything downstream had implicit thread-identity assumptions.

4. One small test-name nit inline.

hf_xet/src/py_upload_commit.rs
seanses: Now if the Python callback in `stop_and_emit` returns an Err, it is print as a standard Python traceback. Also added `tracing::warn!` to log the error if returned from `stop_and_emit`.

hf_xet/src/py_file_download_group.rs
seanses: Applied the same changes as for py_upload_commit.

hf_xet/src/background_progress.rs
seanses: Renamed

hf_xet/src/py_upload_commit.rs
```diff
@@ -308,13 +307,35 @@ impl PyXetUploadCommit {
...
```
rajatarya: nit: this comment is now slightly stale — the error is no longer ignored, it's logged via `tracing::warn!` below. Feel free to drop or reword whenever you next touch this file (same in `py_file_download_group.rs:197`). The matching comment in `abort` is still accurate.

rajatarya (approved):
Thanks Di — I went back through both rounds of feedback against the current head (075f5db) and everything is addressed. Approving.

**How each thread was resolved:**

1. **Double-fire of the final callback** — the poller now breaks on `is_terminal` *before* invoking the callback, so `stop_and_emit` solely owns the final emission. I also re-checked the in-flight race bugbot raised: since `stop_and_emit` joins the poller thread before emitting, any in-flight regular update strictly happens-before the final callback, and the final callback can't fire twice. Your rebuttal holds.

2. **Final callback on the error path** — `wait_to_finish` now gates on `result.is_ok()` (emit on Ok, join-only on Err) in both `PyXetUploadCommit` and `PyXetFileDownloadGroup`, so a `KeyboardInterrupt` no longer triggers a side-effect callback.

3. **Error visibility** — final-callback `PyErr`s are printed via `e.print(py)` inside `stop_and_emit`, and join failures are logged with `tracing::warn!` in both `wait_to_finish` impls. That's a good balance against `abort`'s intentional best-effort silence.

4. **Tests** — both suggested tests landed (`final_callback_fires_exactly_once_via_stop_and_emit`, `stop_and_join_does_not_invoke_callback` — rename included) and they assert exactly the contract. I also pulled the branch locally: clean build, all 36 lib tests pass, and the `background_progress` tests survive a 30x multithreaded stress run with no hangs — the `py.detach`-around-join means the poller can always drain through `try_attach`, so no GIL-vs-join deadlock.

5. **Naming** — field unified to `progress` in both structs.

One truly take-it-or-leave-it nit inline about a now-stale comment. Nice work on this one — the final shape of `BackgroundProgress` is a clean, well-tested primitive.

hf_xet/src/py_file_download_group.rs
```diff
@@ -188,13 +187,35 @@ impl PyXetFileDownloadGroup {
...
+        if let (Some(handles), Some(progress)) = (&self.download_handles, &self.progress) {
+            // Not propagate any error from progress update
+            let progress_join_ret = if result.is_ok() {
+                progress.stop_and_emit(py, || {
+                    let item_reports = item_reports_from_download_handles(handles);
+                    (self.inner.progress(), item_reports)
+                })
+            } else {
+                progress.stop_and_join(py)
+            };
```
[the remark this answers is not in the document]
seanses: This is expected

hf_xet/src/py_upload_commit.rs
seanses: rephrased

PR #860, opened 2026-06-02
Bump russh for cargo audit CI check

seanses:
This PR bumps the `russh` crate version to 0.60 to address the cargo audit CI failure in https://github.com/huggingface/xet-core/pull/856. This crate is used under `git_xet` for integration test. Also updated code due to API change.


rajatarya (approved):
Seems reasonable. Having the key in the codebase might trip up some static analysis workflows.

PR #861, opened 2026-06-03
fix(simulation): remove dirty-range extension from simulation CAS client

jgodlew:
The simulation client's `build_file_chunk_hashes_response` was calling `extend_dirty_ranges_to_stable_windows`, which expands the dirty byte ranges to the next stable CDC chunk boundary before returning them. The real Xet server expects the ranges provided to it to already be stable chunk boundaries and thus, uses the dirty_ranges from the request as-is. This discrepancy means that the simulation produces different window boundaries than production, which impacts the ability to test new features relying on this API (get_file_chunk_hashes).

Changes:
- Removes `extend_dirty_ranges_to_stable_windows` and `next_stable_end_for_range` entirely
- Updates tests to match the corrected no-extension behavior

---

> [!NOTE]
> **Low Risk**
> Behavior change is limited to the local simulation CAS client and its tests; production paths are unchanged and the fix reduces sim/prod divergence.
> 
> **Overview**
> The simulation path for **`get_file_chunk_hashes`** no longer **widens** caller-supplied dirty byte ranges to the next stable CDC chunk boundary before building chunk windows.
> 
> **`build_file_chunk_hashes_response`** in `chunk_window_builder.rs` drops `extend_dirty_ranges_to_stable_windows`, chunk-boundary derivation, and the `next_stable_chunk_boundary` dependency. Dirty ranges are only clamped to file size and validated; **`ChunkWindowBuilder`** then derives windows from those ranges as given—aligned with production CAS, which treats request ranges as already stable.
> 
> Tests that asserted extension/coalescing at stable boundaries are replaced with cases that expect windows to follow the **exact** dirty ranges (including two separate windows for far-apart ranges). **`local_client`** has a small refactor when assembling per-segment chunk lists (no behavior change).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c3c13f5252f28cc6ec79ea418924b6ef35fb6b78. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #863, opened 2026-06-03
fix(chunk-windows): fold stable-boundary extension into ChunkWindowBuilder

XciD:
## Context

Alternative to #861. Same goal (align sim and prod behavior of `get_file_chunk_hashes`) but moves the stable-CDC-boundary extension **into `ChunkWindowBuilder`** instead of removing it, so both consumers get it automatically.

## Problem

`ChunkWindowBuilder` is the shared state machine used by:
- the simulation (`build_file_chunk_hashes_response`)
- the production xetcas server (`cas_server/src/core.rs::build_chunk_windows`)

Pre-existing sim path: `build_file_chunk_hashes_response` pre-processed dirty ranges via `extend_dirty_ranges_to_stable_windows` before feeding the builder, so its windows ended at stable CDC boundaries.

Pre-existing prod path: xetcas never called that helper. It fed raw dirty ranges to the builder. Sim and prod diverged.

This drift broke two things:
1. `xet-data::upload_ranges` requires windows to end at stable CDC boundaries (so the cleaner's re-chunked window output can be glued back to the original chunks without hash drift). It worked in sim, would fail in prod.
2. The s3-gateway `part_merger`'s strict equality check between sim and prod failed for multi-XORB parts.

#861 fixed the drift by removing the extension everywhere, which kept the s3-gateway check happy but breaks `upload_ranges` in both sim and prod.

## Approach

Fold the extension into `ChunkWindowBuilder` itself, so it's impossible to forget:

- Builder tracks the previous clean-chunk size while a window is in its post-dirty extension phase.
- When that size and the current chunk size both satisfy the CDC stability predicate (both in `[2*min, max-min)`), the window closes at the end of the current chunk.
- Tail resets on any chunk that re-enters dirty content, on back-to-back dirty ranges, and on \`skip_dirty_entry\`.
- The external \`extend_dirty_ranges_to_stable_windows\` / \`next_stable_end_for_range\` helpers (sim-only and fragile to forget) are removed.
- \`is_stable_chunk_size\` is lifted into \`xet_core_structures::xorb_object::constants\` so \`ChunkWindowBuilder\` and \`next_stable_chunk_boundary\` share one source of truth.

Also drops stale diagnostic prints from \`local_client.rs::get_file_chunk_hashes\` (carried over from #861).

## Result

- \`processing::range_upload::tests::test_regression_hash_matches_clean_upload_seed1_round17\` passes.
- Full \`xet-data --lib\` (283 tests) green.
- \`chunk_window_builder\` builder tests green.
- After xetcas bumps the \`xet-core\` dependency past this commit, its prod \`get_file_chunk_hashes\` starts returning extended windows for free. Sim/prod aligned, s3-gateway strict equality satisfied, \`upload_ranges\` works in prod.

## Trade-offs

- Sliding 2-chunk tail is more state than the previous pre-processing approach, but it avoids the upfront pass over chunk_boundaries that the helper needed.
- Per-chunk overhead in extension phase: one is_stable_chunk_size check (3 lazy_static derefs + comparisons). Only runs for chunks in the post-dirty tail, so the practical cost is a handful of cycles per request.

---

> [!NOTE]
> **Medium Risk**
> Changes core chunk-window semantics for file-chunk-hashes used by upload_ranges and sim/prod parity; behavior is intentional but affects how dirty byte ranges are computed across the stack.
> 
> **Overview**
> **Stable CDC window extension** is moved from a sim-only pre-pass into **`ChunkWindowBuilder`**, so simulation and production `get_file_chunk_hashes` both extend dirty windows until two consecutive clean chunks satisfy the stability size predicate (or the file ends).
> 
> The sim driver **`build_file_chunk_hashes_response`** no longer builds chunk boundaries or calls **`extend_dirty_ranges_to_stable_windows`** / **`next_stable_end_for_range`**; those helpers are removed. The builder gains **`tail_prev_size`**, post-dirty extension logic, **`overlaps_dirty_at`** (including the next dirty range), and **`extend_open_window`**.
> 
> **`is_stable_chunk_size`** is added in **`xorb_object::constants`** and shared with **`next_stable_chunk_boundary`**, which now delegates to it instead of inlining the same bounds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7c5cf4d6cd5bd330cd7fcc2d7f4364b2751717cd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/xet_client/src/cas_client/chunk_window_builder.rs
+++ b/xet_client/src/cas_client/chunk_window_builder.rs
@@ -4,7 +4,7 @@
 
 use xet_core_structures::merklehash::{MerkleHash, MerkleHashSubtree};
 use xet_core_structures::metadata_shard::file_structs::MDBFileInfo;
-use xet_core_structures::xorb_object::constants::next_stable_chunk_boundary;
+use xet_core_structures::xorb_object::constants::is_stable_chunk_size;
 
 use crate::cas_types::{ChunkWindow, FileChunkHashesResponse, FileRange};
 use crate::error::{ClientError, Result};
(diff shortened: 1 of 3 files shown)


jgodlew (approved):
Looks good. This is a better fix for the inconsistency between the simulator and xet server.

PR #864, opened 2026-06-04
fix: debug assertions for deserialized MerkleHashSubtree

jgodlew:
Fixes a false-positive debug assertion in `MerkleHashSubtree` that panics when a
  subtree built via `from_chunks` is merged with one obtained by deserialization
  (e.g. a stable interior fetched from the Xet server over the wire).

  ## Background

  `MerkleHashSubtree` keeps a debug-only `debug_chunks` field and an end-to-end
  invariant check, `verify_invariants`, which asserts that `final_hash()` equals
  `aggregated_node_hash(debug_chunks)`. The field is populated by `from_chunks`
  but reset to empty by `deserialize`, thus the original chunk list is not recoverable
  from the compacted hump representation.

  `merge_into_impl` blindly concatenated `debug_chunks` from both operands. When
  one operand was deserialized (nodes present, `debug_chunks` empty), the merged
  `debug_chunks` held only the *tracked* half's chunks, while `final_hash()`
  correctly reflected the *full* merged tree. On the final fully-closed merge,
  `verify_invariants` then compared the correct full-tree hash against
  `aggregated_node_hash` of a partial chunk list and panicked:

      assertion `left == right` failed: MerkleHashSubtree invariant: final_hash mismatch.
      Num debug_chunks: 6, num_nodes: 1

  This surfaced in a downstream multipart-merge use case (S3 gateway) that stitches
  deserialized stable interiors from Xet server together with locally re-chunked dirty
  boundary runs. The produced hash was always correct, which was confirmed by running
  the affected integration tests in release mode (no `debug_assertions`) and validating with
  the file-hash of the full content. The panic was purely the debug-only invariant misfiring in debug builds.

  ## Fix

  `debug_chunks` reconstructs the full chunk list end-to-end only when **both**
  merge operands tracked theirs. When either side has nodes but no `debug_chunks`
  (i.e. was deserialized), the merged result drops `debug_chunks` entirely, so
  `verify_invariants` skips rather than comparing against a partial list. The
  invariant remains fully active for the all-`from_chunks` case.

---

> [!NOTE]
> **Low Risk**
> Debug-build-only invariant and test changes; production hashes and merge logic are unchanged.
> 
> **Overview**
> Fixes a **debug-only false panic** when merging a `from_chunks` `MerkleHashSubtree` with one that was **deserialized** (e.g. stable interior from CAS), where `debug_chunks` is empty but the hump `nodes` are intact.
> 
> `merge_into_impl` no longer always appends both operands’ `debug_chunks`. If either side has **nodes but no `debug_chunks`** (untracked / wire form), the merged result **clears** `debug_chunks` so `verify_invariants` skips instead of comparing `final_hash()` on the full tree to `aggregated_node_hash` on a **partial** chunk list. When **both** sides tracked chunks, behavior is unchanged and the invariant still runs.
> 
> Docs on `debug_chunks` describe deserialize vs merge behavior. A regression test serde-round-trips one merge operand and asserts merges still match the expected hash without tripping the invariant.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d908904a5aa1295a7779ca44199277cda50ded98. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/xet_core_structures/src/merklehash/merkle_hash_subtree.rs
+++ b/xet_core_structures/src/merklehash/merkle_hash_subtree.rs
@@ -424,7 +424,9 @@ pub fn find_stable_end(nodes: &[Node]) -> Option<usize> {
 /// - `at_start`: `true` if this range begins at position 0 of the full chunk sequence (left boundary is known).
 /// - `at_end`: `true` if this range ends at the last chunk of the full sequence (right boundary is known).
 /// - `debug_chunks`: (debug builds only) the original level-0 chunks, retained to verify that `final_hash()` matches
-///   `aggregated_node_hash`.
+///   `aggregated_node_hash`. Populated by [`from_chunks`] and empty after [`deserialize`] (the chunk list is not
+///   recoverable from the hump). A merge keeps `debug_chunks` only when both operands tracked theirs; otherwise it is
+///   cleared so the partial list never drives a false-positive invariant check.
 #[derive(Clone, Debug)]
 pub struct MerkleHashSubtree {
     nodes: Vec<Node>,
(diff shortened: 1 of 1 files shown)


hoytak (approved):
LGTM!  Makes sense.

PR #865, opened 2026-06-05
feat(xorb): per-upload uniqueness nonce APIs for S3 ETag differentiation

sirahd:
## What

Adds APIs to stamp a per-upload **uniqueness nonce** into a xorb's serialized footer, so that byte-identical re-uploads of the same xorb produce a different S3 ETag.

- `XorbObjectInfoV1::set_uniqueness_nonce(&mut self, [u8; 16])` / `uniqueness_nonce(&self) -> [u8; 16]` — accessors over the existing 16-byte `_buffer` extensibility field.
- `XorbObject::overwrite_uniqueness_nonce(serialized: &mut [u8], nonce: [u8; 16]) -> Result<(), CoreError>` — rewrites the nonce in already-serialized bytes without re-serializing. Writes into `[len-20, len-4)` (the `_buffer` slot, which precedes the trailing 4-byte `info_length`). Works for **both V0 and V1** objects, since both footers end with `_buffer(16) || info_length(u32)`. Validates the trailing `info_length` and returns `Err` on too-short/implausible input rather than corrupting bytes.

## Why

The nonce is **excluded from the xorb hash** (the hash is a merkle tree over chunk contents only), so it does not change the xorb's content address or S3 key — only the serialized bytes, and hence the S3 ETag.

Garbage collection snapshots `(xorb_hash, S3 ETag)` and later conditionally soft-deletes (`HeadObject If-Match` → `PutObjectTagging(gc-delete=true)`). Giving every (re-)upload a fresh ETag makes the `If-Match` fail for a resurrected object, preventing GC from deleting live data.

## Downstream

Consumed by xetcas / cas_server (stacked PR): stamp a nonce on first upload before `serialize_given_info`; on re-upload of an existing object, rewrite the nonce in the ground-truth bytes and force-PUT to refresh the ETag.

## Compatibility

Purely additive. No changes to serialization layout, `serialized_length`, the wire format, or existing call sites. Objects written before this change have an all-zero `_buffer` and remain valid. See `api_changes/update_260604_xorb_uniqueness_nonce.md`.

## Tests

- nonce roundtrip (hash unaffected by nonce)
- in-place overwrite for V1 and V0-footer'd xorbs (only the nonce slot changes; still deserializes to the same hash; malformed input rejected)
- `cargo test -p xet-core-structures` (162 passing), `cargo +nightly fmt`, `cargo clippy -r -- -D warnings` clean

---

> [!NOTE]
> **Medium Risk**
> Changes how serialized xorb footers are written and patched in core storage format code; wire layout is unchanged but incorrect nonce handling could affect upload/GC ETag behavior downstream.
> 
> **Overview**
> Adds **per-upload uniqueness nonce** support in the xorb footer so byte-identical content can produce different serialized bytes (e.g. distinct S3 ETags) without changing the content hash.
> 
> `XorbObjectInfoV1` renames the private footer extensibility field to `_nonce_buffer` and exposes `set_uniqueness_nonce` / `uniqueness_nonce` for the leading **4 bytes** of the existing 16-byte footer slot (12 bytes reserved). New constants document footer buffer layout. `XorbObject::overwrite_uniqueness_nonce` patches that nonce in already-serialized bytes for V0 and V1 footers after validating trailing `info_length`, leaving chunk data and hash semantics unchanged.
> 
> Includes `api_changes/update_260604_xorb_uniqueness_nonce.md` and tests for roundtrip, in-place overwrite, and malformed-input rejection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7e004ee441a91aeb9b748c11acd11724a69d3f56. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/xet_core_structures/src/xorb_object/xorb_object_format.rs
+++ b/xet_core_structures/src/xorb_object/xorb_object_format.rs
@@ -32,8 +32,19 @@ const _XORB_OBJECT_INFO_DEFAULT_LENGTH_V0: u32 = 60;
 const XORB_OBJECT_INFO_DEFAULT_LENGTH: u32 = 92;
 
 /// Length in bytes of the footer's trailing extensibility buffer (`_nonce_buffer` in V1, `_buffer`
-/// in the legacy V0), which may hold a per-upload uniqueness nonce. Excluded from the xorb hash.
+/// in the legacy V0). Excluded from the xorb hash.
 pub(crate) const XORB_OBJECT_FORMAT_FOOTER_BUFFER_LEN: usize = 16;
+
+/// Number of leading bytes of the footer buffer used for the per-upload uniqueness nonce. The
+/// remaining `XORB_OBJECT_FORMAT_FOOTER_BUFFER_LEN - XORB_OBJECT_FORMAT_NONCE_LEN` bytes stay zero,
+/// reserved for future use.
+///
+/// A 32-bit random nonce collides with any one specific previously-seen nonce with probability
+/// ~1/2^32. The nonce only matters in the narrow window where a dereferenced xorb is snapshotted and
+/// then identical content is re-uploaded before deletion; even at a generous 10^8 such events per
+/// year that is ~0.02 expected collisions/year, so 4 bytes is sufficient while leaving 12 of the 16
+/// buffer bytes free for future use.
+pub(crate) const XORB_OBJECT_FORMAT_NONCE_LEN: usize = 4;
 
 // Decide array preallocation size based on the declared size, to prevent an adversarial
 // giant size that leads to OOM on allocation.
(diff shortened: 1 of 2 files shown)


xet_core_structures/src/xorb_object/xorb_object_format.rs
```diff
@@ -75,8 +79,8 @@ pub struct XorbObjectInfoV0 {
     pub chunk_hashes: Vec<MerkleHash>,
 
     #[serde(skip)]
-    /// Unused 16-byte buffer to allow for future extensibility.
-    _buffer: [u8; 16],
+    /// Extensibility buffer for future use; see `XorbObjectInfoV1::_buffer`.
+    _buffer: [u8; XORB_OBJECT_FORMAT_FOOTER_BUFFER_LEN],
 }
 
 impl Default for XorbObjectInfoV0 {
```
hoytak, line 97: V0 needs to be the same for backwards compatibility; this should only go in V1.

xet_core_structures/src/xorb_object/xorb_object_format.rs
```diff
@@ -382,6 +390,17 @@ impl Default for XorbObjectInfoV1 {
 }
 
 impl XorbObjectInfoV1 {
+    /// Set the per-upload uniqueness nonce stored in the extensibility buffer. Does
+    /// not affect the xorb hash; see the `_buffer` field docs.
+    pub fn set_uniqueness_nonce(&mut self, nonce: [u8; XORB_OBJECT_FORMAT_FOOTER_BUFFER_LEN]) {
+        self._buffer = nonce;
```
hoytak: I'm fine to rename the _buffer part to _random_buffer or something to reflect this.

hoytak (approved):
LGTM!

xet_core_structures/src/xorb_object/xorb_object_format.rs
```diff
@@ -31,6 +31,10 @@ pub(crate) const XORB_OBJECT_FORMAT_BOUNDARIES_VERSION: u8 = 1;
 const _XORB_OBJECT_INFO_DEFAULT_LENGTH_V0: u32 = 60;
 const XORB_OBJECT_INFO_DEFAULT_LENGTH: u32 = 92;
 
+/// Length in bytes of the footer's trailing extensibility buffer (`_nonce_buffer` in V1, `_buffer`
+/// in the legacy V0), which may hold a per-upload uniqueness nonce. Excluded from the xorb hash.
+pub(crate) const XORB_OBJECT_FORMAT_FOOTER_BUFFER_LEN: usize = 16;
```
assafvayner, line 36: can we set a shorter nonce to maintain more buffer space for the future

sirahd: Good call. The nonce now occupies just the leading 4 bytes of the 16-byte footer buffer, leaving 12 bytes reserved for future use (XORB_OBJECT_FORMAT_NONCE_LEN = 4). With 4 bytes, the probability of two xorb reuploads producing the same nonce is 1/2^32 which is statistically negligible given that xorb reupload is already quite rare.

assafvayner (approved):
approved conditional on the serverside change. (maybe wait till the serverside change is ready to merge before merging this one; in case you need more changes)

PR #872, opened 2026-06-12
feat: gate internal-only dependencies behind an `internal-tools` feature

kszucs:
Daft has restricted the buckets support to only downloading because opendal-hf would pull in ~80 additional crates to their build https://github.com/Eventual-Inc/Daft/pull/6731, hence I'm trying to slightly slim down hf-xet to make opendal-hf builds lighter.

`clap`, `csv`, and `walkdir` were unconditional dependencies used only by the internal
binaries (`x`, `xtool`, `shard_benchmark`, `collect_compression_stats`) and the
simulation/analysis tooling, yet they compiled for every library consumer of
`hf-xet`. This gates them behind features so they no longer reach library users.

### Changes

- `xet_data` / `xet_core_structures`: `clap`/`csv`/`walkdir`/`tempfile` optional behind a new `internal-tools`
  feature; internal binaries gated with `required-features = ["internal-tools"]`.
- `xet_data`: examples pull `clap` from `[dev-dependencies]` so they build without the flag.
- `xet_client`: `clap`/`tracing-subscriber` optional, wired into the existing
  `analysis`/`simulation` features that use them.
- ci: `internal-tools` added to the test feature set so the binaries stay compile-checked.

### Effect

No library API or runtime behavior change. Removes 17 crates from the default
`hf-xet` dependency tree (218 → 201):

```
clap            clap_builder   clap_derive    clap_lex
anstream        anstyle        anstyle-parse  anstyle-query
colorchoice     utf8parse      is_terminal_polyfill  strsim
csv             csv-core       heck           ryu
walkdir
```

The build-time win is larger than the count suggests, since `clap_derive` is a proc-macro.

Verified: default `hf-xet` tree has zero `clap`/`csv`/`walkdir`; libraries and examples build with
default features; internal bins build with `--features internal-tools`; simulation/analysis bins build with
their features; `cargo machete` passes.


kszucs:
The PyO3 dependency bump is not related.

kszucs:
@seanses @hoytak opinions?

everettVT:
Thanks @kszucs, What will the size be following this PR?

seanses (approved):
Thank you @kszucs! I noticed the concerned build targets are mainly benchmark and test utilities, not any user-facing CLI, so I'd prefer to call this feature "internal-tools" instead of "cli". The rest looks good to me!

kszucs:
I'm OOO this week so feel free to update. Also what do you think, are there further options to slim down the cet crate? Especially from opendal's perspective.

seanses:
> I'm OOO this week so feel free to update. Also what do you think, are there further options to slim down the cet crate? Especially from opendal's perspective.

There's probably more we can mark as optional, let me see if there's any tool to help us diagnose

PR #877, opened 2026-06-17
Drop hf-xet Python release for Python 3.13t

seanses:
`pyo3` (so `maturin`) has [dropped](https://github.com/PyO3/pyo3/releases/tag/v0.29.0) support for Python `3.13t`, so our hf-xet Python release CI is failing. This PR drops releases for Python `3.13t`.

---

> [!NOTE]
> **Low Risk**
> CI matrix-only change; narrows published wheel variants with no runtime or application code changes.
> 
> **Overview**
> Stops building and publishing **hf-xet** wheels for **Python 3.13t** because PyO3/maturin no longer supports that interpreter.
> 
> The change removes `3.13t` from the `python-version` matrix in the **linux**, **musllinux**, **windows**, and **macos** jobs in `.github/workflows/release.yml`. Releases continue for **3.14** and **3.14t** on the same platform targets.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5e6f0add69f8a24f7365b1b7b6b8ba50e98192c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,7 +16,6 @@ on:
         description: 'Semantic version for PyPI release (tag will share the same name)'
         required: true
         default: 'v0.1.0'
-  pull_request:
 
 permissions:
   contents: read
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,6 +16,7 @@ on:
         description: 'Semantic version for PyPI release (tag will share the same name)'
         required: true
         default: 'v0.1.0'
+  pull_request:
 
 permissions:
   contents: read
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,7 +34,6 @@ jobs:
             manylinux: manylinux_2_28
         python-version:
           - 3.14
-          - 3.13t
           - 3.14t
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
@@ -130,7 +129,6 @@ jobs:
             target: aarch64
         python-version:
           - 3.14
-          - 3.13t
           - 3.14t
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
@@ -220,7 +218,6 @@ jobs:
             rust_target: aarch64-pc-windows-msvc
         python-version:
           - 3.14
-          - 3.13t
           - 3.14t
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -34,7 +34,6 @@ jobs:
             manylinux: manylinux_2_28
         python-version:
           - 3.14
-          - 3.13t
           - 3.14t
```
rajatarya, line 37: nit (non-blocking): with `3.13t` removed, `3.14t` here is now the only free-threaded target we publish a prebuilt wheel for. That lines up with the `Programming Language :: Python :: Free Threading` classifiers still present in `hf_xet/pyproject.toml`, so those classifiers are correct to keep. Just confirming the two are consistent — nothing to change.

rajatarya (approved):
Thanks Di. This is a clean, well-scoped change and the rationale checks out: PyO3 0.29 dropped `3.13t` support (https://github.com/PyO3/pyo3/releases/tag/v0.29.0), so the release CI for that interpreter can't succeed and dropping it is the right call.

I verified the removal is complete and consistent:
- `3.13t` is gone from all four wheel-build matrices (`linux`, `musllinux`, `windows`, `macos`) and `3.14t` (the other free-threaded target) is intact in all four — no half-removal or dangling matrix entry.
- A repo-wide search turns up no other references to `3.13t` / `cp313t` / `313t` outside this file. The test/CI workflows (`ci.yml`, `hf-xet-tests.yml`, `pre-release-testing.yml`) pin a single interpreter (3.10) and never enumerated `3.13t`, so there's nothing dangling there either.
- The `sdist` job has no version matrix, so source builds for 3.13t remain available — free-threaded 3.13 users can still build from source, which is the reasonable fallback here.
- Nothing unrelated is swept in (0 additions, 4 deletions, single file).

One user-impact note worth confirming for my own understanding: `hf_xet/Cargo.toml` pins `pyo3` with `abi3-py37`, so the GIL-enabled builds ship a single abi3 wheel that already covers 3.8–3.13+. The free-threaded interpreters can't use the stable ABI, so `3.13t` was the one variant that gave 3.13 free-threading users a prebuilt wheel. After this change those users fall back to building from source (or staying on a prior hf-xet). Given upstream forced our hand, that seems acceptable — just flagging it as the concrete user-facing effect.

Two small, non-blocking thoughts:
- What do you think about a one-line note in the PR description (or a short `api_changes/` entry, per the repo convention) calling out that 3.13t users now build from source? It'd save a future "where did my wheel go" question. Not blocking.
- nit: `hf_xet/pyproject.toml` still carries the `Programming Language :: Python :: Free Threading` / `Free Threading :: 2 - Beta` trove classifiers. Those stay accurate because `3.14t` is still released, so no action needed — just noting they're intentionally retained and not a leftover from this drop.

Approving — the change does exactly what it says and does it consistently.

.github/workflows/release.yml
seanses: Yes keeping "Programming Language :: Python :: Free Threading" as we still build for "3.14t"

PR #879, opened 2026-06-26
fix(client): call versioned /v1/shards endpoint for shard uploads

assafvayner:
The shard-upload client was still calling the old unversioned `/shards` path while every other CAS endpoint (chunks, reconstructions, xorbs, file-chunk-hashes) already uses a version prefix. This switches `upload_shard` to the canonical `/v1/shards` and registers that route on the simulation mock server.

---

> [!NOTE]
> **Low Risk**
> Small URL change only; risk is limited to environments that still expose only unversioned `/shards` without `/v1/shards`.
> 
> **Overview**
> Shard uploads from `RemoteClient::upload_shard` now POST to **`/v1/shards`** instead of the legacy unversioned **`/shards`**, matching other versioned CAS routes (`/v1/chunks`, `/v1/xorbs`, etc.).
> 
> The simulation **local server** registers **`POST /v1/shards`** under the `/v1` router and drops the root-level **`/shards`** alias so local/integration tests hit the same path the client uses.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a4ba3aa75e0236611315169364e5d1087cb9f2f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/xet_client/src/cas_client/remote_client.rs
+++ b/xet_client/src/cas_client/remote_client.rs
@@ -581,7 +581,7 @@ impl Client for RemoteClient {
         event!(INFORMATION_LOG_LEVEL, call_id, size = n_upload_bytes, "Starting upload_shard API call",);
 
         let api_tag = "cas::upload_shard";
-        let url = Url::parse(&format!("{}/shards", self.endpoint))?;
+        let url = Url::parse(&format!("{}/v1/shards", self.endpoint))?;
 
         // Use the no-read-timeout client for shard uploads. reqwest's per-request timeout()
         // does NOT override the client-level read_timeout(), so we use a separate client
--- a/xet_client/src/cas_client/simulation/local_server/server.rs
+++ b/xet_client/src/cas_client/simulation/local_server/server.rs
@@ -179,6 +179,7 @@ impl LocalServer {
                     .route("/reconstructions/{file_id}", get(handlers::get_reconstruction))
                     .route("/chunks/{prefix}/{hash}", get(handlers::get_dedup_info_by_chunk))
                     .route("/xorbs/{prefix}/{hash}", head(handlers::head_xorb).post(handlers::post_xorb))
+                    .route("/shards", post(handlers::post_shard))
                     .route("/files/{file_id}", head(handlers::head_file))
                     .route("/get_xorb/{prefix}/{hash}/", get(handlers::get_file_term_data))
                     .route("/fetch_term", get(handlers::fetch_term)),
@@ -193,7 +194,6 @@ impl LocalServer {
             )
             // Routes used by RemoteClient without /v1/ prefix
             .route("/reconstructions", get(handlers::batch_get_reconstruction))
-            .route("/shards", post(handlers::post_shard))
             .layer(CorsLayer::very_permissive())
             .with_state(handlers::ServerState {
                 client: self.client.clone(),


XciD (approved):

xet_client/src/cas_client/simulation/local_server/server.rs
```diff
@@ -193,7 +194,6 @@ impl LocalServer {
             )
             // Routes used by RemoteClient without /v1/ prefix
             .route("/reconstructions", get(handlers::batch_get_reconstruction))
```
seanses, line 196: Might as well remove this route

xet_client/src/cas_client/simulation/local_server/server.rs
```diff
@@ -193,7 +194,6 @@ impl LocalServer {
             )
             // Routes used by RemoteClient without /v1/ prefix
```
seanses, line 195: and this comment

seanses (approved):

xet_client/src/cas_client/remote_client.rs
```diff
@@ -581,7 +581,7 @@ impl Client for RemoteClient {
         event!(INFORMATION_LOG_LEVEL, call_id, size = n_upload_bytes, "Starting upload_shard API call",);
 
         let api_tag = "cas::upload_shard";
-        let url = Url::parse(&format!("{}/shards", self.endpoint))?;
+        let url = Url::parse(&format!("{}/v1/shards", self.endpoint))?;
```
rajatarya, line 584: Confirmed this is safe against the current server: `xetcas` `serve.rs` merges `upload_public_cas_routes` (which defines `/shards`) both under the `/v1` nest and at the root, so `/v1/shards` has been served all along and the old root `/shards` stays live for older clients.

One thing worth confirming before this ships: are there any long-lived or pinned CAS deployments (older self-hosted instances, pinned images) that predate the `/v1` nest and would still only expose root `/shards`? If not, no concern here.

xet_client/src/cas_client/simulation/local_server/server.rs
```diff
@@ -179,6 +179,7 @@ impl LocalServer {
                     .route("/reconstructions/{file_id}", get(handlers::get_reconstruction))
                     .route("/chunks/{prefix}/{hash}", get(handlers::get_dedup_info_by_chunk))
                     .route("/xorbs/{prefix}/{hash}", head(handlers::head_xorb).post(handlers::post_xorb))
+                    .route("/shards", post(handlers::post_shard))
```
rajatarya, line 182: Nice — moving this under `/v1` matches the client, and keeping root `/reconstructions` at line 196 mirrors the still-unversioned batch call in `remote_client.rs`, so the mock tracks exactly what the client calls.

One difference from the real server: `xetcas` serves `/shards` at *both* `/v1` and root (for backward compat), whereas the mock now serves only `/v1/shards`. Since the client only calls `/v1/shards`, dropping the alias is reasonable — just checking whether you'd want the mock to keep root `/shards` too so the backward-compat path stays exercised, or whether versioned-only coverage is intended here.

rajatarya (commented):
Thanks, Assaf — clean consistency fix. This brings `upload_shard` in line with every other CAS route (`/v1/chunks`, `/v1/xorbs`, `/v1/reconstructions`, `/v2/file-chunk-hashes`), and the mock's `post_shard` handler doc comment already reads `POST /v1/shards`, so the route now matches its own docstring.

**Backward compatibility (the main thing I checked)**
The deployed server is fine: `xetcas` `serve.rs` merges `upload_public_cas_routes` (which registers `/shards`) both under the `/v1` nest *and* at the root, so `/v1/shards` has been served all along, and the old root `/shards` stays live for older clients. I also confirmed there are no other in-tree callers of the old path. The only residual risk (same as the Cursor note) is a long-lived / pinned CAS deployment predating the `/v1` nest — see the inline question.

**What's good**
- Route matches the rest of the client and the handler's documented path.
- The mock change is principled: it moves `/shards` under `/v1` (matching the client) while keeping root `/reconstructions`, which the batch call at `remote_client.rs:324` still hits unversioned — so the mock keeps mirroring exactly what the client calls.

**Items to consider**
1. Stale doc comments still describing the old routing (just outside this diff, but part of the same change): the `create_router` doc comment (`server.rs:168-171`, "Root-level ... `/shards` for uploads") and `local_server/main.rs:36` (`POST /shards`). Worth a one-line update here while you're in the file.
2. (question, non-blocking) The batch reconstruction call at `remote_client.rs:324` is still on the unversioned root `/reconstructions` while everything else is versioned. Given the stated goal is CAS-endpoint consistency, is leaving it unversioned intentional / a follow-up, or in scope here?

Not approving (leaving this as a comment) — happy to approve once the doc cleanup and the backward-compat question are settled.

rajatarya (approved):
Minor comments, good consistency fix.

PR #880, opened 2026-06-29
feat(file-reconstruction): add read_timeout to bound stalled download streams

XciD:
## Problem

`DownloadStream::next()` awaits the next reconstructed chunk with **no upper bound**. If the underlying CAS/CDN connection stalls (delivers no data and no error), `next()` never returns.

Consumers that block a thread on `next()` are then parked indefinitely. A concrete case: a FUSE filesystem (e.g. mountpoint-style mounts, or `hf-mount`) does `block_on(stream.next())` on a bounded pool of worker threads — a single stalled stream pins a worker forever, and once enough stalled reads accumulate, every worker is stuck and the whole mount silently wedges (cold reads hang while metadata ops still work). The only recourse today is to cancel the entire stream.

## Change

Add an optional **inactivity timeout** to the streaming reconstruction path:

- `FileReconstructor::with_read_timeout(Duration)` — opt-in, stored as `Option<Duration>` (default `None`).
- `DownloadStream::next()` — when a timeout is set, a call that waits longer than `timeout` for the next chunk returns the new `FileReconstructionError::ReadTimeout(Duration)` and **cancels the reconstruction** (via the existing `RunState`), so the in-flight fetch is aborted rather than left running detached.

`None` preserves the current unbounded behavior, so this is fully backwards compatible.

The point is to put the stall bound **in the client library**, where every consumer benefits, instead of each one wrapping `next()` in its own `tokio::time::timeout` (which leaves the underlying fetch running).

## Scope

Applies to the ordered streaming path (`reconstruct_to_stream` → `DownloadStream::next`). Left unchanged, as candidates for the same treatment later:
- `DownloadStream::blocking_next()` (sync path; `blocking_recv` has no timeout variant),
- `UnorderedDownloadStream`,
- the file/writer reconstruction paths (`reconstruct_to_file` / `reconstruct_to_writer`), which have no consumer-side `next()` to bound — they'd need the timeout inside the reconstruction loop itself.

## Tests

- `read_timeout_fires_when_reconstruction_stalls`: a 0-permit custom buffer semaphore (never seeded, since dynamic scaling only runs for the global limiter) stalls the reconstruction at buffer acquisition before any data is produced — standing in for a stalled CAS/CDN connection. `next()` returns `ReadTimeout`, and the next call returns `Ok(None)` (stream finished/cancelled).
- `no_read_timeout_streams_normally`: with no timeout set, a healthy reconstruction streams to completion unchanged.

---

> [!NOTE]
> **Medium Risk**
> Touches core streaming download behavior and cancellation on timeout; default is opt-in `None`, but misconfigured timeouts could abort slow-but-valid reads.
> 
> **Overview**
> Adds an **opt-in inactivity timeout** on the ordered streaming path so consumers are not blocked forever when CAS/CDN stops delivering chunks.
> 
> **`FileReconstructor::with_read_timeout(Duration)`** stores an optional timeout (default **`None`** keeps unbounded waits). **`reconstruct_to_stream`** passes it into **`DownloadStream`**, which wraps async **`next()`** with **`tokio::time::timeout`**. If a wait exceeds the limit, the stream returns **`FileReconstructionError::ReadTimeout`**, marks the stream finished, and **cancels reconstruction** via existing **`RunState`** so the background fetch is torn down—not left running after an outer timeout.
> 
> A new **`next_inner`** holds the prior receive logic; **`blocking_next`**, unordered streams, and file/writer paths are unchanged. Tests cover a stalled reconstruction (0-permit buffer semaphore) surfacing **`ReadTimeout`** and the default path still streaming to completion.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ba891f33ee0f3eb8b710b28b4799bc3dbe1c79f4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/xet_data/src/file_reconstruction/data_writer/download_stream.rs
+++ b/xet_data/src/file_reconstruction/data_writer/download_stream.rs
@@ -1,4 +1,5 @@
 use std::sync::Arc;
+use std::time::Duration;
 
 use bytes::Bytes;
 use tokio::sync::Notify;
@@ -39,6 +40,10 @@ pub struct DownloadStream {
     /// Signal to unblock the spawned reconstruction task. `Some` means
     /// `start()` has not yet been called; the spawned task is waiting.
     start_signal: Option<Arc<Notify>>,
+    /// Optional inactivity timeout: when set, a [`next`](Self::next) that waits
+    /// longer than this for the next chunk fails with
+    /// [`FileReconstructionError::ReadTimeout`] and cancels the reconstruction.
+    read_timeout: Option<Duration>,
 }
 
 impl DownloadStream {
@@ -49,7 +54,11 @@ impl DownloadStream {
     /// # Panics
     ///
     /// Panics if called outside a tokio runtime context.
-    pub(crate) fn new(reconstructor: FileReconstructor, run_state: Arc<RunState>) -> Self {
+    pub(crate) fn new(
+        reconstructor: FileReconstructor,
+        run_state: Arc<RunState>,
+        read_timeout: Option<Duration>,
+    ) -> Self {
         let (data_writer, receiver) = SequentialWriter::new_streaming(run_state.clone());
         let start_signal = Arc::new(Notify::new());
 
@@ -66,6 +75,7 @@ impl DownloadStream {
             finished: false,
             run_state,
             start_signal: Some(start_signal),
+            read_timeout,
         }
     }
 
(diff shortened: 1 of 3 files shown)


xet_data/src/file_reconstruction/data_writer/download_stream.rs
```diff
@@ -149,12 +159,33 @@ impl DownloadStream {
...
+    /// [`FileReconstructionError::ReadTimeout`] and cancels the reconstruction
+    /// (aborting the in-flight fetch).
     pub async fn next(&mut self) -> Result<Option<Bytes>> {
         if self.finished {
             return Ok(None);
         }
         self.ensure_started();
 
+        match self.read_timeout {
+            Some(timeout) => match tokio::time::timeout(timeout, self.next_inner()).await {
```
seanses, line 174: Note that `self.next_inner()` returns an entire file term, which can be as small as one chunk (~64 KiB), and as large as one full Xorb (64 MiB). So it can be difficult to set a "good" timeout for this. Is this really the unit we want to apply read timeout to? FYI, the underlying http client for retrieving file term has a default read timeout set to 300s (https://github.com/huggingface/xet-core/blob/bebdf4b5aaad56318cfb9999f80e75760da59a5e/xet_client/src/common/http_client.rs#L89, https://github.com/huggingface/xet-core/blob/bebdf4b5aaad56318cfb9999f80e75760da59a5e/xet_runtime/src/config/groups/client.rs#L60), maybe set or re-configure this value?

XciD: oh you are right. Missed this one. Thanks for the pointer

PR #884, opened 2026-07-02
Stream shard finalization progress via POST /v2/shards

seanses:
## Summary

Adds client support for `POST /v2/shards`, a streaming NDJSON variant of shard upload that reports server-side finalization progress (validation against xorb metadata, then durable commit to S3/DynamoDB) instead of blocking silently until the whole thing completes. This closes a visibility gap for large shards, where server-side finalization can take a while after the bytes have already been transferred.

- `RemoteClient::upload_shard` auto-detects V2 vs V1: it tries `/v2/shards` first and falls back to `/v1/shards` on `404`/`501`, caching the detected version per client (same pattern already used for the reconstruction API). A new `client.shard_api_version` config option (`HF_XET_CLIENT_SHARD_API_VERSION`) can force a specific version with no fallback.
- New `ShardUploadEvent` wire type (`Validating`, `Committing{stage}`, `Result`, `Error`) models the NDJSON frames sent by the server. It has a custom, deliberately partial `PartialOrd`/`precede` so the client can safely ignore stale, duplicate, or out-of-order frames instead of naively overwriting already-observed state.
- New `shard_upload_v2` module parses the response body frame-by-frame (correctly handling frames split across TCP chunks), invokes an optional progress callback per frame, and resolves once a terminal `Result`/`Error` frame arrives.

## Progress tracking refactor

This also reworks the internal progress-tracking types to plug in shard finalization progress alongside existing file/xorb progress:

- Added `ShardUploadProgress`: tracks shard byte-transfer progress plus the new server-reported validation/commit counters (`total_shard_validation_entries[_completed]`, `total_shards_uploaded_to_store`, `total_shards_synced`, `total_shards_completed`), keyed per-shard by `UniqueId` so out-of-order/duplicate NDJSON frames can't regress already-recorded state.
- Introduced `UploadGroupProgress`, combining file/xorb progress (`GroupProgress`) and the new `ShardUploadProgress` into a single handle now owned by `CompletionTracker` (previously `FileUploadSession` held a bare `GroupProgress` directly).
- `CompletionTracker::register_new_file` now creates the per-file `ItemProgressUpdater` internally instead of requiring callers to construct one and pass it in, simplifying call sites in `FileUploadSession`.
- Removed two redundant aliases on `ItemProgressUpdater` (`report_bytes_written`, `report_transfer_progress`) in favor of their canonical names (`report_bytes_completed`, `report_transfer_bytes_completed`).

## Other changes

- `upload_xorb` now moves the serialized xorb bytes (`mem::take` + `Bytes::from`) instead of cloning them before streaming.
- Simulation/local test server and all `Client` impls (`LocalClient`, `MemoryClient`, `RemoteSimulationClient`, `SimulationControlClient`, `LocalTestServer`) updated to thread the new `progress_callback` parameter through `upload_shard`, and the simulation router's `/shards` route was consolidated under the `/v1` nest.
- Substantial new unit test coverage: `ShardUploadEvent` serde round-trips and a full `PartialOrd`/`precede` ordering matrix, `ShardUploadProgress::update` behavior under out-of-order/skipped/duplicate frames (including a multi-shard aggregation test), and `shard_upload_v2` NDJSON parsing edge cases (mid-line frame splits, blank lines, missing terminal frame, callback ordering).

## Related PRs
CAS server side: https://github.com/huggingface-internal/xetcas/pull/1140
huggingface_hub side: https://github.com/huggingface/huggingface_hub/pull/4478

Resolves client side of https://github.com/huggingface-internal/xetcas/pull/996

---

> [!NOTE]
> **Medium Risk**
> Touches core CAS upload paths, API version fallback, and progress accounting across retries; large surface area but well-tested and mostly additive with V1 fallback preserved.
> 
> **Overview**
> **Shard uploads** now prefer `POST /v2/shards` with NDJSON progress (validation, commit stages, terminal result/error), auto-detecting V2 vs `/v1/shards` on 404/501 like reconstruction API versioning (`HF_XET_CLIENT_SHARD_API_VERSION` to force). `upload_shard` takes an optional progress callback (transfer bytes, decrement on failed V2 before V1 fallback, response events) and returns `Result<()>` instead of `bool`.
> 
> **Progress plumbing** adds `ShardUploadEvent` wire types, `ShardUploadProgress` / `UploadGroupProgress`, and wires shard callbacks through native `SessionShardInterface` via `CompletionTracker`. Commit reports gain optional `shard` stats; Python exposes `ShardUploadProgressReport`.
> 
> **Supporting fixes:** `UploadProgressStream` flushes the last chunk on drop; V1/fallback synthesizes transfer + `Result` so UI counters complete; batch reconstructions use `/v1/reconstructions`; removed `RetryWrapper::run_and_extract_custom` alias.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9f61c376f903a2d32adc845835647fc7e311d424. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


seanses:
Demo (with huggingface_hub [PR](https://github.com/huggingface/huggingface_hub/pull/4478)):

https://github.com/user-attachments/assets/54fc504b-26cc-4519-96e1-0e8771438adb

xet_client/src/cas_client/remote_client.rs
```diff
@@ -308,6 +313,165 @@ impl RemoteClient {
...
+            INFORMATION_LOG_LEVEL,
+            call_id,
+            size = n_upload_bytes,
+            api_version = "v2",
+            "Starting upload_shard API call",
+        );
+
+        // One target of "/v2/shards" API compared to "v1/shards" is to eliminate
+        // the http client without read timeout.
+        let client = self.authenticated_http_client.clone();
```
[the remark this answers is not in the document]
seanses: This is by design

xet_client/src/cas_client/remote_client.rs
```diff
@@ -308,6 +313,165 @@ impl RemoteClient {
...
+        let response: UploadShardResponse = RetryWrapper::new(self.ctx.clone(), api_tag)
+            .with_connection_permit(upload_permit, Some(shard_data.len() as u64))
+            .run_and_extract_json(move || {
+                client
+                    .post(url.clone())
+                    .with_extension(Api(api_tag))
+                    .body(shard_data.clone())
+                    .send()
+            })
+            .await?;
```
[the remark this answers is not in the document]
seanses: v1 path has no progress reporting, this is by design.

XciD:
> Demo (with huggingface_hub [PR](https://github.com/huggingface/huggingface_hub/pull/4478)):
> 
>  Screen.Recording.2026-07-03.at.7.58.55.AM.mov

could we ad a it/s for the shard validation ? with an approx ETA ? something dumb is okay I think

XciD:
I've added a `retryable: true` on server side implementation to handle transient connection reset during server side validation.

LGTM

xet_client/src/cas_client/shard_upload_v2.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    while let Some(chunk) = stream.try_next().await? {
+        buffer.extend_from_slice(&chunk);
+        while let Some(newline_idx) = buffer.iter().position(|&b| b == b'\n') {
+            let line = buffer.drain(..=newline_idx).collect::<Vec<u8>>();
+            let line = line.strip_suffix(b"\n").unwrap_or(&line);
+            if line.is_empty() {
+                continue;
+            }
+            let event: ShardUploadEvent = serde_json::from_slice(line).map_err(|err| {
```
XciD, line 36: I think `ShardUploadEvent` should add `#[serde(other)]` to avoid fatal error.
In case we add new type in the backend

xet_client/src/cas_client/remote_client.rs
```diff
@@ -308,6 +313,165 @@ impl RemoteClient {
...
+        let mut upload_reporter = StreamProgressReporter::new(n_upload_bytes as u64)
+            .with_adaptive_concurrency_reporter(upload_permit.get_partial_completion_reporting_function());
+        if let Some(cb) = &progress_callback {
+            let cb = cb.clone();
+            upload_reporter = upload_reporter
+                .with_progress_callback(Arc::new(move |delta, _, _| cb(ShardUploadProgressType::Transfer(delta))));
+        }
+
+        let response = RetryWrapper::new(self.ctx.clone(), api_tag)
+            .with_connection_permit(upload_permit, Some(shard_data.len() as u64))
```
XciD, line 408: nit: The permit will be release once we received the header, not during the streaming consumption. maybe we should release after L415

XciD (approved):

xet_data/src/processing/shard_interface/native.rs
```diff
@@ -281,8 +286,22 @@ impl SessionShardInterface {
...
+                                completion_tracker.increment_shard_transfer_progress(shard_progress_id, nbytes)
+                            },
+                            ShardUploadProgressType::Response(event) => {
+                                completion_tracker.register_shard_upload_progress(shard_progress_id, event)
+                            },
+                        }));
+
                     // Upload the shard.
-                    shard_client.upload_shard(data, upload_permit).await?;
+                    shard_client.upload_shard(data, upload_permit, progress_callback).await?;
```
rajatarya, line 306: `register_shard_transfer` (line 291) and this callback always run, but only the V2 `RemoteClient` path ever invokes the callback. When the client ends up on V1 -- forced via `shard_api_version`, after a one-time 404/501 fallback, or for `LocalClient`/`MemoryClient`/wasm -- `total_shards` and `total_shard_bytes` get bumped but no transfer/validation/commit completion is ever reported, so the shard portion of `GroupProgressReport` stays stuck at 0 completed. Is a permanently-incomplete shard section OK for the hub-side consumer, or should registration be gated on V2 (or V1 report a synthetic completion)? Curious what you're expecting downstream.

xet_data/src/processing/shard_interface/native.rs
```diff
@@ -281,8 +286,22 @@ impl SessionShardInterface {
                         return Ok(());
                     }
 
+                    // Shard upload progress
+                    let shard_progress_id =
+                        completion_tracker.register_shard_transfer(data.len().try_into().unwrap_or(0));
```
rajatarya: nit: `data.len().try_into().unwrap_or(0)` will silently record a 0-byte shard if the conversion ever fails. `data.len() as u64` is lossless on all supported targets and reads more clearly here.

xet_data/src/progress_tracking/progress_types.rs
```diff
@@ -211,6 +207,243 @@ impl std::fmt::Debug for GroupProgress {
...
+            ShardUploadEvent::Error { .. } => {},
+            // `Committing(*)` and `Result`: stage milestones. `Error` and `Validating` are
+            // handled above, so this only ever sees forward-moving stages.
+            _ => {
+                self.credit_skipped_validation(&maybe_old);
+                self.credit_skipped_store_stage(&maybe_old, event);
+                if maybe_old.is_none_or(|old| old.precede(event)) {
+                    guard.insert(id, event.clone());
+                    if matches!(event, ShardUploadEvent::Result { .. }) {
+                        self.total_shards_synced.fetch_add(1, Ordering::Relaxed);
```
rajatarya, line 298: `total_shards_synced` and `total_shards_completed` are only ever incremented here, together, on `Result` -- so they're always equal in every report. Given `total_shards_uploaded_to_store` already marks reaching the store/sync stage, I'd have expected `synced` to track the sync step finishing distinctly from overall `completed`. Are the two intentionally identical for now, or is one meant to advance earlier (e.g. `synced` on `Committing(Syncing)`)? Since `ShardUploadProgressReport` is exposed as a `pyclass`, worth locking the semantics down before it ships.

xet_client/src/cas_client/interface.rs
```diff
@@ -7,7 +9,17 @@ use super::adaptive_concurrency::ConnectionPermit;
 use super::progress_tracked_streams::ProgressCallback;
 use crate::cas_types::{
     BatchQueryReconstructionResponse, FileChunkHashesResponse, FileRange, HttpRange, QueryReconstructionResponseV2,
+    ShardUploadEvent,
 };
+
+/// Progress callback for shard upload finalization (v2 NDJSON stream).
```
rajatarya: nit: the doc reads "Progress callback for shard upload finalization", but this type is the payload delivered *to* the callback (`ShardUploadProgressCallback` is the callback itself). Something like "Progress update passed to the shard-upload callback" might read truer. Also the variant comments on the next lines use `//` rather than `///` so they won't render as docs, and "transfered" -> "transferred".

xet_client/src/cas_client/interface.rs
```diff
@@ -7,7 +9,17 @@ use super::adaptive_concurrency::ConnectionPermit;
...
+pub enum ShardUploadProgressType<'a> {
+    // Number of incremental bytes transfered
+    Transfer(u64),
+    // Progress event from server
+    Response(&'a ShardUploadEvent),
+}
+pub type ShardUploadProgressCallback = Arc<dyn Fn(ShardUploadProgressType) + Send + Sync>;
 use crate::error::Result;
```
rajatarya: nit: this `use crate::error::Result;` ended up below the new `pub enum` / `pub type` definitions. Consider moving it back up with the other `use`s so the imports stay grouped at the top of the file.

xet_client/src/cas_client/shard_upload_v2.rs
```diff
@@ -0,0 +1,239 @@
...
+/// Parse the `/v2/shards` NDJSON response stream and return whether the shard was newly inserted.
+pub(crate) async fn read_shard_upload_ndjson(
+    response: Response,
+    progress_callback: Option<ShardUploadProgressCallback>,
+) -> Result<bool> {
+    let mut buffer = Vec::new();
+    let mut stream = response.bytes_stream();
+
+    while let Some(chunk) = stream.try_next().await? {
+        buffer.extend_from_slice(&chunk);
```
rajatarya, line 26: nit / optional: `buffer` has no upper bound, so a server or proxy that streams bytes without ever emitting a newline would grow it without limit. The CAS server is trusted so this is low risk, but a sanity cap on line length (erroring past some size) would make this defensive against a stuck stream. Fine to leave as-is if you'd rather not.

rajatarya (commented):
Nice work, Di -- this is a big change but it's well-organized, and the test coverage is genuinely impressive. The `PartialOrd`/`precede` matrix, the out-of-order / skipped / duplicate-frame cases in `ShardUploadProgress::update`, and the NDJSON split-frame parsing tests all make the tricky ordering logic easy to trust. Auto-detecting V2 with a cached per-client version and a 404/501 fallback (mirroring the reconstruction API) is a sensible rollout strategy, and the `upload_xorb` `mem::take` (dropping the clone of the serialized bytes before streaming) is a clean win.

A few behavioral changes worth surfacing for reviewers:
- **V2-first by default.** `upload_shard` now tries `/v2/shards` first and falls back to `/v1/shards` on 404/501, caching the result per client. Risk is bounded, but worth confirming the CAS server PR is rolled out everywhere before this lands.
- **Routes moved under `/v1`.** `batch_get_reconstruction` now targets `/v1/reconstructions`, and the sim server dropped the un-prefixed `/reconstructions` + `/shards` routes. Coordinated with the server side -- flagging so it isn't missed.
- **Public surface shifts (internal crates).** `GroupProgress::new`/`with_speed_config` now return `Self` instead of `Arc<Self>`; `FileUploadSession::progress()` now returns `&Arc<UploadGroupProgress>`; and the `report_bytes_written` / `report_transfer_progress` aliases plus `ItemProgressUpdater::group()` were removed. All call sites are updated here; just noting for anyone tracking the surface.
- **`UploadProgressStream` now flushes the final chunk on `Drop`.** Good catch -- the last chunk was previously never reported. The comment explaining it is excellent, and callers were updated.

Items to consider (linked inline):
1. Shard progress counters are incremented unconditionally in `native.rs`, but only the V2 `RemoteClient` path ever reports transfer/validation completion -- against a V1 server the shard section of the report looks permanently incomplete. Is that acceptable for the hub consumer? (inline)
2. `total_shards_synced` and `total_shards_completed` are always incremented together (both only on `Result`), so they're always equal. Since `total_shards_uploaded_to_store` already marks the syncing stage, did you intend `synced` to advance distinctly? These are public `pyclass` fields, so worth pinning down. (inline)
3. A few nits: doc/`use` placement in `interface.rs`, `try_into().unwrap_or(0)` in `native.rs`, and an optional buffer-bound thought in `shard_upload_v2.rs`.

Thanks, Di -- happy to talk through any of these.

rajatarya:
> Demo (with huggingface_hub [PR](https://github.com/huggingface/huggingface_hub/pull/4478)):
> 
>  Screen.Recording.2026-07-03.at.7.58.55.AM.mov

image

I'm not sure users understand shards - that is a pretty low-level Xet specific concept. I think for the UX we might want to simplify the reporting to just a 'Validating' progress bar. I think it would be confusing for users to see two Committing lines. 

Their entire UX is based on files - shards are an internal representation of their files to help with dedupe.

Instead I suggest the Uploading shard, validating shard, and Committing shard progress bars are collapsed into one - 'Validating' progress bar, which is before the users 'Committing' bar.

This way shard operations are just part of the file upload validation process - which seems right given shards are part of how we represent the user data.

xet_client/src/cas_client/shard_upload_v2.rs
```diff
@@ -0,0 +1,239 @@
+use futures::TryStreamExt;
+use reqwest::Response;
+
+use super::interface::{ShardUploadProgressCallback, ShardUploadProgressType};
+use crate::cas_types::{ShardUploadEvent, UploadShardResponseType};
+use crate::error::{ClientError, Result};
+
+/// Parse the `/v2/shards` NDJSON response stream and return whether the shard was newly inserted.
+pub(crate) async fn read_shard_upload_ndjson(
```
assafvayner, line 18: I would like a small refactor to have 2 functions, 1 which reads the response stream until next newline and maintains a buffer and another to handle the ShardUploadEvent specifically.

Might also suggest, although it looks like more overkill to use common utilities: wrap Stream in StreamReader and BufReader then call read_line.

assafvayner (commented):
Are we resilient to a double send of the same shard upload event?

xet_data/src/progress_tracking/progress_types.rs
```diff
@@ -211,6 +207,243 @@ impl std::fmt::Debug for GroupProgress {
...
+
+impl std::fmt::Debug for ShardUploadProgress {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_fmt(format_args!("{:?}", self.report()))
+    }
+}
+
+pub struct UploadGroupProgress {
+    // File read and Xorb upload progress
+    pub file_data: Arc<GroupProgress>,
```
assafvayner, line 381: non-blocking, GroupProgress is no longer a good name for the previous progress tracking struct.

Maybe FileIngestionXorbUploadProgress and/or split the 2 pieces into 2 parts and have UploadGroupProgress have 3 sub-component fields. I think this split is the most "clean" but is a lot of heavy lifting (maybe another PR for this refactor?)

xet_data/src/progress_tracking/progress_types.rs
```diff
@@ -372,6 +589,20 @@ pub struct GroupProgressReport {
     pub total_transfer_bytes: u64,
     pub total_transfer_bytes_completed: u64,
     pub total_transfer_bytes_completion_rate: Option<f64>,
+    pub shard: Option<ShardUploadProgressReport>,
+}
```
assafvayner, line 608: I assume we're keeping this shape and adding sub-struct for backwards compatibility with mismatch versions of huggingface_hub. maybe we should have a comment about this/instructions for future changes.

if that's not the reason and we can change the shape, then I'd suggest 2 fields/sub-structs 1 for shard, 1 for the file data upload progress.

assafvayner (approved):
some nits, but lgtm otherwise

xet_client/src/cas_client/shard_upload_v2.rs
seanses: Added

xet_client/src/cas_client/remote_client.rs
seanses: Added retry logic for error event with "retry = true", now the permit is not released until a correct `Result` event is received, i.e. it survives through retries.

xet_data/src/processing/shard_interface/native.rs
seanses: Now the v1 upload path also bumps progress counters after the entire shard is uploaded.

xet_data/src/processing/shard_interface/native.rs
seanses: updated

xet_client/src/cas_client/interface.rs
seanses: Updated.

xet_client/src/cas_client/interface.rs
seanses: fixed

xet_data/src/progress_tracking/progress_types.rs
seanses: Yeah this struct is wired into many places including for download, so I'd rather not change it drastically in this PR.

xet_data/src/progress_tracking/progress_types.rs
seanses: Yes it's for backward compat (added comments). This is directly read in Python.

xet_data/src/progress_tracking/progress_types.rs
seanses: The `Syncing` event is sent before the shard syncing operation starts on the server side so we can't advance the counter `total_shards_synced` yet on receiving this event. Now `total_shards_synced` and `total_shards_completed` but there's room for more detailed reporting of the "Syncing" operation (because of may-be-slow DDB batch insertions). Let's wait for further evidence if that's needed (maybe with the help of client telemetry?). We can always bump the required `hf-xet` version in `huggingface_hub` when the semantics change.

xet_data/src/progress_tracking/progress_types.rs
```diff
@@ -211,6 +207,243 @@ impl std::fmt::Debug for GroupProgress {
...
+            // are handled above, so this only ever sees forward-moving stages.
+            _ => {
+                self.credit_skipped_validation(&maybe_old);
+                self.credit_skipped_store_stage(&maybe_old, event);
+                if maybe_old.is_none_or(|old| old.precede(event)) {
+                    guard.insert(id, event.clone());
+                    if matches!(event, ShardUploadEvent::Result) {
+                        self.total_shards_synced.fetch_add(1, Ordering::Relaxed);
+                        self.total_shards_completed.fetch_add(1, Ordering::Relaxed);
+                    }
```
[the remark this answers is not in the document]
seanses: This is by design.

xet_client/src/cas_client/shard_upload_v2.rs
seanses: Now caps each line at 1 MiB.

xet_data/src/progress_tracking/progress_types.rs
```diff
@@ -211,6 +207,254 @@ impl std::fmt::Debug for GroupProgress {
...
+    /// recorded `(verified, total)` via `saturating_sub`, so a stale/duplicate frame with lower
+    /// values contributes nothing.
+    fn credit_validating_progress(&self, maybe_old: Option<&mut ShardUploadEvent>, verified: u64, total: u64) {
+        let Some(ShardUploadEvent::Validating {
+            verified: old_verified,
+            total: old_total,
+        }) = maybe_old
+        else {
+            return;
+        };
```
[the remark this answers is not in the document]
seanses: That's Ok.

xet_client/src/cas_client/shard_upload_v2.rs
```diff
@@ -0,0 +1,344 @@
...
+                )))
+            })?;
+            if handle_shard_upload_event(event, progress_callback.as_ref())? {
+                return Ok(());
+            }
+        }
+        // Incomplete frame (no newline yet) must not grow without bound.
+        if buffer.len() > MAX_NDJSON_LINE_BYTES {
+            return Err(oversized_ndjson_frame_error(buffer.len()));
+        }
```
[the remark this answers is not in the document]
seanses: The underlying network stack (TCP) won't deliver large packets without dissecting.

PR #886, opened 2026-07-03
fix: do not count defrag-prevented ranges as deduped in metrics

XciD:
## Problem

In `FileDeduper::process_chunks`, a matched dedup range is counted into `deduped_chunks`, `deduped_bytes`, `total_chunks` and `total_bytes` *before* the defrag-prevention gate (`allow_dedup_on_next_range`). When the gate rejects the range, its chunks fall through to the new-data path below, which does its own `total_*`/`new_*` accounting. As a result:

- every defrag-prevented byte is double-counted in `total_bytes` (measured: `total_bytes = file_size + defrag_prevented_dedup_bytes`, exactly),
- those bytes are also misreported as `deduped_bytes` even though they were re-stored,
- any **debug build** cleaning a file that trips defrag prevention panics on the `file_size == total_bytes` `debug_assert` in `SingleFileCleaner::finish`.

File reconstruction info is unaffected (the registered file size and segments are correct); this is a metrics/telemetry bug plus a debug-build crash.

## Fix

Move the four increments after the defrag gate so each chunk is counted exactly once, either as deduped (gate passed) or as new (gate rejected, new-data path). `defrag_prevented_*` counters keep their existing semantics.

## Regression test

`test_dedup_metrics_when_defrag_prevention_triggers` reproduces the scenario organically rather than by mocking the tracker: upload content as small shuffled pieces (fragmented CAS layout, as produced by out-of-order writers), then cleanly re-upload the assembled file from a client with an empty shard cache, so dedup discovery goes through the sampled global dedup queries and only matches in piece-sized runs. The CPR tracker then rejects ranges (`defrag_prevented_dedup_bytes > 0` is asserted so the test cannot pass vacuously). Before the fix the test fails on the `SingleFileCleaner` debug_assert with `total_bytes` inflated by exactly `defrag_prevented_dedup_bytes`; after the fix `deduped + new == total == file_size` holds.

Side note from the same investigation, out of scope here: on such fragmented references, defrag prevention re-stores a significant share of already-stored content (~17% in the test above, ~50% observed on production bucket files, i.e. ~2x stored bytes with heavily fragmented layouts). That looks like a design discussion worth having separately.

### DIFF
--- a/xet_data/src/deduplication/file_deduplication.rs
+++ b/xet_data/src/deduplication/file_deduplication.rs
@@ -197,16 +197,20 @@ impl<DataInterfaceType: DeduplicationDataInterface> FileDeduper<DataInterfaceTyp
             }
 
             if let Some((n_deduped, fse, is_external)) = dedupe_query {
-                dedup_metrics.deduped_chunks += n_deduped as u64;
-                dedup_metrics.deduped_bytes += fse.unpacked_segment_bytes as u64;
-                dedup_metrics.total_chunks += n_deduped as u64;
-                dedup_metrics.total_bytes += fse.unpacked_segment_bytes as u64;
-
                 // check the fragmentation state and if it is pretty fragmented,
                 // we skip dedupe.  However, continuing the previous is always fine.
                 if self.file_data_sequence_continues_current(&fse)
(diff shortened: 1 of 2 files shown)


seanses:
Close in favor of https://github.com/huggingface/xet-core/pull/931

Issue #883, opened 2026-07-01
XET is hanging when downloading large models (>=80% of the time)


ri938:
### Describe the bug

I am getting recurring hangs when downloading files using XET 

Using snapshot download from huggingface hub

```
    snapshot_download(
        repo_id,
        revision=revision,
        token=token,
        local_dir=download_path,
        ignore_patterns=ignore_patterns
    )
```

somtimes it completes very quickly but maybe 2/3 of the time its getting stuck when downloading. (it will time out after 1 hour+)

e.g I just launched 2 downloads earlier

<server 1>
Wed Jul  1 19:34:11 UTC 2026
2.2G    ./.cache
528G    .

Wed Jul  1 19:58:55 UTC 2026
2.2G    ./.cache
528G    .

<server 2>
Wed Jul  1 19:34:16 UTC 2026
3.5G    ./.cache
524G    .

Wed Jul  1 19:59:07 UTC 2026
3.5G    ./.cache
524G    .

confirmed it works (but very slow) if I disable XET

### Logs

```shell

```

### System info

```shell
hf-xet==1.5.1
```

rajatarya:
@ri938 Can you share client logs from a failed upload attempt? The logs should be located in ~/.cache/huggingface/xet/logs.

What OS are you attempting to download from? Can you share more info from your environment (run `hf env` and share that as well)?

ri938:
root@<hostname>:/code# hf env
Hint: A new version of huggingface_hub (1.22.0) is available! You are using version 1.20.1.
To update, run: hf update

Copy-and-paste the text below in your GitHub issue.

- huggingface_hub version: 1.20.1
- Platform: Linux-6.5.13-65-650-4141-22041-coreweave-amd64-85c45edc-x86_64-with-glibc2.35
- Python version: 3.12.13
- Running in iPython ?: No
- Running in notebook ?: No
- Running in Google Colab ?: No
- Running in Google Colab Enterprise ?: No
- Token path ?: /root/.cache/huggingface/token
- Has saved token ?: False
- Configured git credential helpers:
- Run by AI agent ?: No
- Installation method: pip
- httpx: 0.28.1
- hf_xet: 1.5.1
- gradio: N/A
- tensorboard: N/A
- ENDPOINT: https://huggingface.co
- HF_HUB_CACHE: /root/.cache/huggingface/hub
- HF_ASSETS_CACHE: /root/.cache/huggingface/assets
[...]

Issue #869, opened 2026-06-10
hf-xet download stalls indefinitely for facebook/sam3.1 large checkpoint, while regular HTTP download succeeds


Aryo-Zare:
## Summary

I encountered a reproducible issue where downloads of the SAM3.1 checkpoint stall indefinitely when `hf-xet` is installed. 
Uninstalling `hf-xet` causes the exact same download to complete successfully.
note : the following has no effect on this problem :

```python
os.environ["HF_HUB_DISABLE_XET"] = "1"
```
The issue occurs on both Windows and WSL Ubuntu, and on multiple network connections ( including a mobile hotspot ).

### Windows

* Windows 11
* Conda environment
* Python 3.12
* huggingface_hub 1.18.0
* hf-xet 1.5.0

### Linux

* WSL2 Ubuntu
* Conda environment
* Python 3.12
* huggingface_hub 1.18.0
* hf-xet 1.5.0

## Repository

facebook/sam3.1

Large checkpoint:

sam3.1_multiplex.pt

Approximate size:

3.5 GB

## Reproduction

```python
from huggingface_hub import hf_hub_download

hf_hub_download(
    repo_id="facebook/sam3.1",
    filename="sam3.1_multiplex.pt",
    force_download=True
)
```

## Observed behavior with hf-xet installed

The download starts but appears to stall indefinitely.

Typical output:

```
Downloading (incomplete total...): 0%| | 5.07M/3.50G [...]
```

The transfer never progresses further.

The local Hugging Face cache remains nearly unchanged in size.

The process can remain in this state for hours.

No useful exception is raised.

Interrupting with Ctrl+C eventually produces a KeyboardInterrupt inside the huggingface_hub/tqdm/concurrent code path.

The same behavior occurred:

* On Windows 11
* On WSL2 Ubuntu
* On a university network
* On a mobile hotspot

## Additional observations

I also tried:

* snapshot_download()
* hf_hub_download()
* single-threaded downloads
* different destination directories
* different drives
* disabling Windows Defender

None of these changed the behavior.

## Workaround

After uninstalling hf-xet:

```bash
pip uninstall hf-xet
```

the exact same download succeeds immediately.

The library reports:

```
Xet Storage is enabled for this repo, but the 'hf_xet' package is not installed.
Falling back to regular HTTP download.
```

and then downloads successfully:

```
sam3.1_multiplex.pt: 100%|██████████| 3.50G/3.50G
```

Download completed successfully in approximately 24 minutes.

[...]

Issue #850, opened 2026-05-23
Downloads Stall when hf-xet is Enabled


ConnoisseurProtege:
### Describe the bug

My downloads using `hf download` seem to stall when I enable hf-xet regardless of what I'm downloading. When I use `HF_HUB_DISABLE_XET=1` the download works fine. Log is too long so I had to cut off the beginning. 

### Reproduction

Just run `hf download` to try to download anything. More specifically for the logs I was downloading https://huggingface.co/bartowski/Qwen_Qwen3.5-122B-A10B-GGUF

### Logs

```shell
{"timestamp":"2026-05-23T01:23:27.533308Z","level":"INFO","fields":{"message":"Concurrency control for download: Increased concurrency from 64 to 64; reason: success ratio 1.000 is above threshold 0.800 and predicted RTT for 0MB at new concurrency is 34.28s < target 60.0s"},"filename":"D:\\a\\xet-core\\xet-core\\xet_client\\src\\cas_client\\adaptive_concurrency\\controller.rs","line_number":543}
{"timestamp":"2026-05-23T01:23:27.533319Z","level":"INFO","fields":{"message":"Concurrency control for download: Current concurrency = 64; predicted bandwidth = 1769489; success_ratio = 1.000; reference_size = 1.0MB; observed bytes sent so far = 2269162817; completed transmissions = 298"},"filename":"D:\\a\\xet-core\\xet-core\\xet_client\\src\\cas_client\\adaptive_concurrency\\controller.rs","line_number":597}
{"timestamp":"2026-05-23T01:23:27.882048Z","level":"INFO","fields":{"message":"Concurrency control for download: Increased concurrency from 62 to 63; reason: success ratio 1.000 is above threshold 0.800 and predicted RTT for 12MB at new concurrency is 59.37s < target 60.0s"},"filename":"D:\\a\\xet-core\\xet-core\\xet_client\\src\\cas_client\\adaptive_concurrency\\controller.rs","line_number":543}
{"timestamp":"2026-05-23T01:23:28.150946Z","level":"INFO","fields":{"message":"Concurrency control for download: Increased concurrency from 64 to 64; reason: success ratio 1.000 is above threshold 0.800 and predicted RTT for 0MB at new concurrency is 34.28s < target 60.0s"},"filename":"D:\\a\\xet-core\\xet-core\\xet_client\\src\\cas_client\\adaptive_concurrency\\controller.rs","line_number":543}
[...]

PR #893, opened 2026-07-09
Fix retry strategy configuration

seanses:
## Summary

Fixes incorrect `ExponentialBackoff` construction in `RetryWrapper`. The old code passed `retry_base_delay` directly to `ExponentialBackoff::from_millis()`, which uses that value as **both** the first retry sleep and the per-step multiplier.

With the default config (`retry_base_delay = 3s`, `retry_max_attempts = 5`), that produced:

| Retry | Old delay |
|-------|-----------|
| 1 | 3s |
| 2 | **9000s (~2.5 hours)** |
| 3 | 7500h |
| 4 | 22_500_000h |
| 5 | 67_500_000_000h |

The intended behavior is 3s, 9s, 27s, 81s, 243s with defaults.

This PR corrects this configuration to match the intent.

See `test_exponential_retry_strategy` in `xet_client/src/cas_client/retry_wrapper.rs` for a side-by-side comparison of the old vs new delay sequences.

## Test plan

- [x] `cargo test -p xet-client --lib cas_client::retry_wrapper`
- [x] `test_exponential_retry_strategy` — documents old delays (including the 2.5h second retry) and asserts the corrected 3s/9s/27s/81s/243s sequence
- [x] `test_exponential_retry_strategy_caps_at_max_duration` — verifies `retry_max_duration` caps individual sleeps

---

> [!NOTE]
> **Medium Risk**
> Changes retry timing for all CAS HTTP paths that use `RetryWrapper`; the fix is clearly necessary but alters real-world wait behavior and how `HF_XET_CLIENT_RETRY_MAX_DURATION` is interpreted.
> 
> **Overview**
> **Fixes broken exponential backoff** in `RetryWrapper` where `retry_base_delay` was passed straight into `ExponentialBackoff::from_millis`, so the same millisecond value acted as both the first sleep and the per-step multiplier. With defaults (3s base, 5 attempts), the second retry waited ~2.5 hours instead of 9s.
> 
> Backoff is now built via **`exponential_retry_delays`**: multiplier from `base_delay` in whole seconds, `factor` so the first delay matches `base_delay`, then jitter as before (e.g. 3s → 9s → 27s → 81s → 243s).
> 
> **`retry_max_duration`** is wired from config into `RetryWrapper` (plus `with_max_duration`) and applied as **`.max_delay()`** on each retry sleep. Config docs now describe it as a per-attempt cap, not a total retry budget.
> 
> Adds **`Debug` for `RetryWrapper`**, logs `retry=?self`, and unit tests that document the old delay sequence and assert capping at `max_duration`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2c442ea8bd946d61119e0d4c1e8ce6b0cce78dab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/xet_client/src/cas_client/retry_wrapper.rs
+++ b/xet_client/src/cas_client/retry_wrapper.rs
@@ -28,6 +28,7 @@ struct ConnectionPermitInfo {
 pub struct RetryWrapper {
     max_attempts: usize,
     base_delay: Duration,
+    max_duration: Duration,
     no_retry_on_429: bool,
     retry_on_403: bool,
     expected_416: bool,
@@ -41,9 +42,11 @@ impl RetryWrapper {
     pub fn new(ctx: XetContext, api_tag: &'static str) -> Self {
         let max_attempts = ctx.config.client.retry_max_attempts;
         let base_delay = ctx.config.client.retry_base_delay;
+        let max_duration = ctx.config.client.retry_max_duration;
         Self {
             max_attempts,
             base_delay,
+            max_duration,
             no_retry_on_429: false,
             retry_on_403: false,
             expected_416: false,
@@ -61,6 +64,11 @@ impl RetryWrapper {
 
     pub fn with_base_delay(mut self, delay: Duration) -> Self {
         self.base_delay = delay;
+        self
+    }
+
+    pub fn with_max_duration(mut self, duration: Duration) -> Self {
+        self.max_duration = duration;
         self
     }
 
@@ -100,7 +108,26 @@ impl RetryWrapper {
         }));
         self
     }
+}
 
+impl std::fmt::Debug for RetryWrapper {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("RetryWrapper")
+            .field("max_attempts", &self.max_attempts)
+            .field("base_delay", &self.base_delay)
+            .field("max_duration", &self.max_duration)
+            .field("no_retry_on_429", &self.no_retry_on_429)
+            .field("retry_on_403", &self.retry_on_403)
+            .field("expected_416", &self.expected_416)
+            .field("expected_404", &self.expected_404)
+            .field("log_errors_as_info", &self.log_errors_as_info)
+            .field("api_tag", &self.api_tag)
+            .field("has_connection_permit", &self.connection_permit.is_some())
+            .finish()
+    }
+}
+
+impl RetryWrapper {
     fn process_error_response(&self, try_idx: usize, err: reqwest_middleware::Error) -> RetryableReqwestError {
         let api = &self.api_tag;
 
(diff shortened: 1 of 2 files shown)


rajatarya:
Fixes #883

rajatarya:
Fixes #869 
Fixes #850

rajatarya (approved):
This is a great find. Love seeing the tests being added confirming the behavior and documenting it so we cannot regress it again.

rxmox:
Confirming this fix resolves a serious production failure mode, posting the field evidence since "apparent hang" is exactly what users of hf_xet ≤ 1.5.1 will be searching for.

**Symptom (hf_xet 1.5.1, production AWS Batch GPU workers, us-east-1):** cold-cache downloads of a 32 MB `tokenizer.json` (Xet-backed repo, `huggingface_hub` 1.23.0 → `xet_get()`) would occasionally stop dead: no exception, no log output, process alive, GPU held. Observed durations ranged 27 min to 21.5 h; one worker "recovered" spontaneously after ~44 min.

**Captured state during a live hang** (reproduced during a degraded-CDN window):
- Python main thread parked at `huggingface_hub/file_download.py:563 in xet_get` (faulthandler dump)
- all 23 threads sleeping on futexes, tokio epoll idle, **zero** TCP retransmits
- exactly one TCP connection to `cas-bridge.xethub.hf.co` (CloudFront) stuck in **`CLOSE_WAIT` with 25 unread bytes** — same socket, unchanged for 4.6 h, no new connection ever attempted

So the client was not stuck on I/O — it was asleep in the backoff ladder this PR fixes (base delay used as both initial delay and multiplier: 3 s → 9,000 s → ~312 days). Every hang duration we observed over several days is explained by which rung of that ladder the worker was sleeping on.

**A/B on the same host in the same degraded-CDN window** (fully cold pulls, each in a subprocess hard-killed at 900 s):

| | completed | raised (bounded, ~1–6 min each) | unbounded hang (>900 s) |
|---|---|---|---|
| **hf_xet 1.5.1** | 0/6 | 0 | **6/6** |
| **hf_xet 1.5.2** | 1/60 | 59/60 | **0/60** |

The high raise-rate is just how bad that CDN window was — the point is the hang column: on 1.5.2 every pull terminates, and failures surface as `RuntimeError: … CAS Client Error: … error decoding response body`, so orchestrator-level retries (AWS Batch `retryStrategy` in our case) can finally do their job. After deploying 1.5.2 we also observed one live CDN failure recover through the corrected ladder in ~264 s in production — an event that on 1.5.1 meant a ≥2.5 h silent hang holding a GPU.

Minor suggestion: the v1.5.2 release notes mention the retry fix but don't link this PR — cross-referencing #893 there would help people diagnosing hangs find it. Thanks for the fix!

PR #899, opened 2026-07-14
Bump sha2 to 0.11

assafvayner:
## Summary
- bump workspace sha2 dependency to 0.11
- remove the obsolete sha2 asm feature from xet-data
- add a canonical byte-order constructor on DataHash/Sha256 and use it for sha2 0.11 digest bytes

## Tests
- cargo check -p xet-data
- cargo test -p xet-core-structures test_hash_from_be_bytes
- cargo test -p xet-data --features simulation sha_tests
- cargo fmt --check -p xet-data -p xet-core-structures
- git diff --check

---

> [!NOTE]
> **Medium Risk**
> Changes how SHA-256 digests are converted into merkle hashes (byte order vs hex round-trip); incorrect mapping would break content verification, though tests cover the path.
> 
> **Overview**
> Upgrades the workspace **`sha2`** dependency from **0.10** to **0.11** and refreshes related lockfiles (including **`hf_xet`** and WASM crates). **`sha2` 0.11** no longer ships the **`asm`** / **`sha2-asm`** path, so **`xet-data`** stops enabling **`asm`** on non-Windows and uses plain workspace **`sha2`** everywhere.
> 
> **`Sha256Generator`** now maps digests with **`Sha256::from_be_bytes`** instead of formatting hex and parsing back. **`DataHash`** gains **`from_be_bytes`** (canonical hex byte order → internal **`[u64; 4]`** layout), with a unit test.
> 
> A standalone **`simulation/sha256_bench`** crate plus **`benches/results.md`** document throughput comparing **0.10.9+asm** vs **0.11** to justify dropping **`asm`**. The root workspace **removes `simulation` from `members`** and **excludes** the whole **`simulation`** tree (replacing the narrower **`chunk_cache_bench`** exclude).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b79ece96cad6ead51f390f90136bc611603f0d66. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


xet_core_structures/src/merklehash/data_hash.rs
```diff
@@ -219,6 +219,17 @@ impl DataHash {
         Ok(hash)
     }
 
+    /// Constructs a DataHash from bytes ordered as the canonical hexadecimal representation.
+    pub fn from_be_bytes(value: [u8; 32]) -> Self {
```
seanses, line 223: Why do we need to introduce this function? It seems from the below usage the `from_slice` function is enough. Also we don't want to deal with endianness when everything is in-memory. See comments for `fn hex(&self)` above.

xet_data/src/processing/sha256.rs
```diff
@@ -45,9 +45,8 @@ impl Sha256Generator {
             None => return Ok(Sha256::default()),
         };
 
-        let sha256 = hasher.finalize();
-        let hex_str = format!("{sha256:x}");
-        Ok(Sha256::from_hex(&hex_str).expect("Converting sha256 to merklehash."))
+        let sha256_bytes: [u8; 32] = hasher.finalize().into();
```
seanses, line 48:
```
let sha256 = hasher.finalize();
Ok(Sha256::from_slice(&sha256).expect("Converting sha256 to merklehash."))
```

Cargo.toml
```diff
@@ -14,6 +14,7 @@ members = [
 ]
 exclude = [
     "simulation/chunk_cache_bench",
+    "simulation/sha256_bench",
```
seanses: Curious if we can just exclude the "simulation" parent directory?

xet_data/src/processing/sha256.rs
assafvayner: claude generated

`from_slice` does a raw byte copy into the internal `[u64; 4]` with no reordering, and `hex()` prints each `u64` via `{:016x}` (most-significant-byte first). On a little-endian host, a raw copy followed by that print reverses the byte order within each 8-byte chunk relative to the input bytes.

`from_be_bytes` cancels that reversal so `hex()` reproduces the actual canonical sha256 hex digest (e.g. matching `sha256sum` output), independent of host endianness. This is exercised by `test_sha_generation_builder`/`test_sha_multiple_updates` in `sha256.rs` (compared against a real `sha256sum` reference) and by `test_hash_hex_string_endianness` vs `test_hash_from_be_bytes` in `data_hash.rs`, which feed byte-reversed-per-chunk input to `from_slice` to get the same hex output as `from_be_bytes` gets from canonically-ordered bytes.

Swapping to `from_slice` here would silently break the sha256 hex output on little-endian hosts — it would no longer match external sha256 tools — so I'd like to keep `from_be_bytes`.

xet_core_structures/src/merklehash/data_hash.rs
assafvayner: claude generated

`from_slice` does a raw byte copy into the internal `[u64; 4]` with no reordering, and `hex()` prints each `u64` via `{:016x}` (most-significant-byte first). On a little-endian host, a raw copy followed by that print reverses the byte order within each 8-byte chunk relative to the input bytes.

`from_be_bytes` cancels that reversal so `hex()` reproduces the actual canonical sha256 hex digest (e.g. matching `sha256sum` output), independent of host endianness. This is exercised by `test_sha_generation_builder`/`test_sha_multiple_updates` in `sha256.rs` (compared against a real `sha256sum` reference) and by `test_hash_hex_string_endianness` vs `test_hash_from_be_bytes` in `data_hash.rs`, which feed byte-reversed-per-chunk input to `from_slice` to get the same hex output as `from_be_bytes` gets from canonically-ordered bytes.

Swapping to `from_slice` here would silently break the sha256 hex output on little-endian hosts — it would no longer match external sha256 tools — so I'd like to keep `from_be_bytes`.

seanses (approved):

PR #900, opened 2026-07-16
Use `LazyLock` instead of `lazy_static`

Expyron:
> [!NOTE]
> **Low Risk**
> Mechanical std-library migration with preserved lazy-init semantics; the postcard default-features change only affects dependency features, not obvious runtime logic.
> 
> **Overview**
> Replaces the **`lazy_static`** crate with **`std::sync::LazyLock`** across the workspace and drops **`lazy_static`** from workspace and crate manifests.
> 
> **`test_configurable_constants!`** in `xet_runtime` now defines env-overridable constants as **`LazyLock`** instead of **`lazy_static!`**, with the same debug-only **`HF_XET_*`** behavior; **`HIGH_PERFORMANCE`** and privilege detection follow the same pattern. Call sites (SIGINT/runtime globals in **`hf_xet`**, profiling session, shard filename regex, **`MAX_CHUNK_SIZE`**, simulation URL reference time, etc.) are updated accordingly, and the duplicate **`MAX_CHUNK_SIZE`** lazy static in **`xet_data`** chunking is removed in favor of the core-structures definition.
> 
> Workspace **`postcard`** is pinned with **`default-features = false`** (still **`alloc`**), which prunes **`heapless`** and related transitive deps from **`Cargo.lock`**. **`FN_CALL_ID`** in the remote CAS client is a plain **`AtomicU64`** static rather than a lazy wrapper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0e667fd2a4cffa6244dbb57ce29d96daf41a567f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


seanses (approved):
LGTM!

PR #902, opened 2026-07-16
Bump hf-xet to 1.5.2

seanses:
Update `version` in `hf_xet/Cargo.toml` then `maturin develop`.


rajatarya (approved):

seanses:
Re the failed "Cargo Audit": the yanked `spin` crate is a dependency of `postcard` which is only a dev dependency. Ignore it for now.

PR #908, opened 2026-07-21
fix: stack overflow in AuthGroupBuilder::build() from unboxed futures

kszucs:
## Summary

`AuthGroupBuilder<T>::build()`'s compiler-generated future was large enough to overflow a thread's default stack on platforms with a small default (e.g. ~524 KiB for non-main threads on macOS).

Root cause: `bridge_async`/`bridge_async_finalizing` and their downstream helpers were `async fn`s generic over `F`, so a large future crossing one of these boundaries by value got duplicated across nested generator frames instead of being shared — the same failure mode #901 partially addressed but didn't fully close.

**Fix:** box the future once, as early as possible (in `xet_session::TaskRuntime`), and pass the boxed pointer through downstream layers.

| | Before | After |
|---|---|---|
| `AuthGroupBuilder::build()` future size | 23,664 bytes | ~536–800 bytes (see below) |
| Min. viable stack (release build) | ~128–256 KiB | ~65.5–66 KiB |

One crate boundary further downstream, `xet_runtime::XetRuntime::bridge_async`/`bridge_to_owned` can either re-box or stay generic over `F` — we measured both (`-Zprint-type-sizes`, nightly, debug build):

| `AuthGroupBuilder<T>` instantiation | Re-box at `xet_runtime` boundary | Generic (current) |
|---|---:|---:|
| `download_stream_group` | 536 bytes | 808 bytes |
| `file_download_group` | 528 bytes | 800 bytes |
| `upload_commit` | 528 bytes | 800 bytes |

Both are trivially safe vs. the 23,664-byte pre-fix size. Re-boxing is ~34% smaller and makes that boundary self-defending for any caller, not just `xet_session` — we opted for the simpler single-boxing-point approach instead, accepting the larger (still tiny) future and leaving "box early" as the caller's responsibility.

## Test plan

- [x] Regression test per `AuthGroupBuilder<T>` instantiation, each running the real `build()` call on a small stack in a re-exec'd child process (verified: fails without the fix, passes with it)
- [x] Full test suites, workspace + wasm builds, `fmt`/`clippy` all clean

---

> [!NOTE]
> **Medium Risk**
> Touches core async bridging used by all session build/upload/download paths; behavior should be equivalent but incorrect boxing could affect cancellation or wasm Send bounds.
> 
> **Overview**
> Fixes **stack overflows** when calling `AuthGroupBuilder::build()` on threads with small default stacks (e.g. non-main threads on macOS). Generic `async fn` bridges were embedding huge caller futures (~24KB) into every nested generator frame.
> 
> **`TaskRuntime::bridge_async` / `bridge_async_finalizing`** are now plain functions that **`Box::pin` the incoming future before** building the outer `async` block, with a shared `BoxedBridgeFuture` type so only a pointer is carried through downstream awaits. Redundant boxing was removed from **`run_inner_async`** (wasm) and **`XetRuntime::bridge_async`** in External mode, since erasure happens once at the session layer.
> 
> Adds **`test_utils::stack_regression`**, which re-runs targeted tests in a **child process on a 128KiB stack** (512KiB on Windows), plus regression tests for upload commit, file download group, and download stream group `build().await`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8d06e277356a5bec3238a81e333cbc6325d7646f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/xet_pkg/src/xet_session/test_utils.rs
+++ b/xet_pkg/src/xet_session/test_utils.rs
@@ -1,0 +1,64 @@
+//! Shared test-only helpers for `xet_session`'s unit tests.
+
+#[cfg(all(test, not(target_family = "wasm")))]
+pub(crate) mod stack_regression {
+    //! Regression-test helper guarding against a large, generic future crossing an
+    //! `.await`-spanning function boundary by value without indirection — the compiler
+    //! then bakes the full, un-erased type into every generator frame it passes through,
+    //! which can silently balloon a state machine's size until it overflows a thread's
+    //! default stack (small on some platforms, e.g. ~524 KiB for non-main threads on
+    //! macOS).
+    //!
+    //! A stack overflow aborts the whole process (it can't be caught with
+    //! `catch_unwind`), so [`run`] executes `build_in_small_stack` in a *child process*
+    //! with an explicitly small stack: if it regresses, only the child aborts and the
+    //! caller's `#[test]` fails cleanly, instead of taking down the entire test binary.
+
+    const MARKER_ENV: &str = "XET_STACK_REGRESSION_CHILD";
+    // 128 KiB is tight enough to actually catch the regression on macOS/Linux (verified:
+    // fails at this size without the fix, passes with it). Windows needs more baseline
+    // stack for the same tokio runtime + thread setup regardless of this bug, so it gets
+    // a separate, larger bound.
+    #[cfg(not(windows))]
+    const STACK_SIZE: usize = 128 * 1024;
+    #[cfg(windows)]
+    const STACK_SIZE: usize = 512 * 1024;
+
+    /// Call from a `#[test]` fn. `test_name` must be that test's own (unique) name —
+    /// used to re-invoke just this test in a child process. `build_in_small_stack`
+    /// performs the actual risky call and only ever runs inside the child, on a thread
+    /// with an explicit small stack.
+    pub(crate) fn run(test_name: &'static str, build_in_small_stack: impl FnOnce() + Send + 'static) {
+        if std::env::var_os(MARKER_ENV).is_some() {
+            let t = std::thread::Builder::new()
+                .stack_size(STACK_SIZE)
+                .spawn(build_in_small_stack)
+                .unwrap();
+            t.join().unwrap();
+            return;
+        }
+
+        let exe = std::env::current_exe().unwrap();
+        let output = std::process::Command::new(exe)
+            .arg(test_name)
+            .arg("--test-threads=1")
+            .env(MARKER_ENV, "1")
+            .output()
+            .unwrap();
+        let stdout = String::from_utf8_lossy(&output.stdout);
+        assert!(
+            output.status.success(),
+            "child process exited abnormally ({:?}) - likely a stack overflow.\n{stdout}",
+            output.status
+        );
+        // libtest exits 0 even when a filter matches no tests, so a typo'd or renamed
+        // `test_name` would otherwise make this helper silently pass without the risky
+        // call ever running. Require exactly the one expected test to have run.
+        assert!(
+            stdout.contains("1 passed"),
+            "expected the child process to run exactly one test matching {test_name:?}, but it \
+             didn't (the test name filter may not have matched anything, or matched more than \
+             one test).\n{stdout}"
+        );
+    }
+}
--- a/xet_pkg/src/xet_session/test_utils.rs
+++ b/xet_pkg/src/xet_session/test_utils.rs
@@ -14,16 +14,25 @@ pub(crate) mod stack_regression {
     //! with an explicitly small stack: if it regresses, only the child aborts and the
     //! caller's `#[test]` fails cleanly, instead of taking down the entire test binary.
 
+    /// `module_path!()` is prefixed with the crate name, but libtest's own test paths
+    /// (and thus its `--exact` filter) are not. Call as
+    /// `test_path(module_path!(), "my_test_fn")` from within the `#[test]` fn itself.
+    pub(crate) fn test_path(module_path: &str, fn_name: &str) -> String {
+        let module_path = module_path.split_once("::").map_or(module_path, |(_, rest)| rest);
+        format!("{module_path}::{fn_name}")
+    }
+
     const MARKER_ENV: &str = "XET_STACK_REGRESSION_CHILD";
     // 128 KiB is tight enough to actually catch the regression (verified: fails at this
     // size without the fix, passes with it).
     const STACK_SIZE: usize = 128 * 1024;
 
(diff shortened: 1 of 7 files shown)


xet_pkg/src/xet_session/test_utils.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    const MARKER_ENV: &str = "XET_STACK_REGRESSION_CHILD";
+    // 128 KiB is tight enough to actually catch the regression on macOS/Linux (verified:
+    // fails at this size without the fix, passes with it). Windows needs more baseline
+    // stack for the same tokio runtime + thread setup regardless of this bug, so it gets
+    // a separate, larger bound.
+    #[cfg(not(windows))]
+    const STACK_SIZE: usize = 128 * 1024;
+    #[cfg(windows)]
+    const STACK_SIZE: usize = 512 * 1024;
```
seanses, line 34: The PR description says the minimum viable stack (release build) before this fix was ~128–256 KiB. But for Windows after the fix it needs 512 KiB stack size. Is that a regression on Windows? Or the description isn't acurate?

xet_runtime/src/core/runtime/native.rs
```diff
@@ -404,24 +404,33 @@ impl XetRuntime {
...
-                self.creation_pid,
-                std::process::id(),
-            )));
-        }
-        match &self.backend {
-            // Box to heap-erase the future's deep type; inlining it here overflows rustc's
-            // type-layout query-depth limit in callers. OwnedThreadPool erases via `spawn`.
-            RuntimeBackend::External { .. } => Ok(Box::pin(fut).await),
-            RuntimeBackend::OwnedThreadPool { .. } => self.bridge_to_owned(task_name, fut).await,
+        let fut: Pin<Box<dyn Future<Output = T> + Send>> = Box::pin(fut);
```
seanses: Is boxing at this level still necessary (as the PR description says we box as early as possible (in xet_session TaskRuntime) instead of re-boxing at each hop)?

xet_pkg/src/xet_session/test_utils.rs
```diff
@@ -0,0 +1,64 @@
...
+                .spawn(build_in_small_stack)
+                .unwrap();
+            t.join().unwrap();
+            return;
+        }
+
+        let exe = std::env::current_exe().unwrap();
+        let output = std::process::Command::new(exe)
+            .arg(test_name)
+            .arg("--test-threads=1")
```
seanses, line 55: Let's add `.arg("--exact")` so it won't prefix match on test names (and so the assert below `stdout.contains("1 passed")` is more reliable).

xet_pkg/src/xet_session/test_utils.rs
kszucs: AFAIK the default stack size is 1MB on windows so it aimed to reduce that limit to manifest the issue, though I don't have a windows to verify it, setting it to 128K uniformly instead.

xet_runtime/src/core/runtime/native.rs
kszucs: Two boxes exist because the future gets re-wrapped in a new type at a second crate boundary, and each boundary needs its own guard against `F` leaking through generically:

1. `task_runtime.rs:241` — `TaskRuntime::bridge_async` boxes the original, huge session-level future the moment it enters `xet_session`'s bridging code, so it stays a small `Pin<Box<dyn Future>>` while passing through `xet_session`'s own internal helpers (`run_inner_async`, etc.).
2. `native.rs:420` — but `run_inner_async` then builds a *brand-new* future around that box: `async move { tokio::select! { cancel => ..., result = fut => result } }`. That `select!` block is itself a fresh, freshly-sized type, and it's what actually gets passed as `F` into `XetRuntime::bridge_async` — a public, generic entry point belonging to a *different* crate (`xet_runtime`). Its own downstream helper, `bridge_to_owned`, only gets to take a concrete `Pin<Box<dyn Future>>` (instead of being generic over `F`) because `bridge_async` boxes first, right here.

Remove either box, and that crate's internal chain goes back to being generic over `F` — reintroducing the exact bug this PR fixed, just at a different hop.

```
xet_session crate                          xet_runtime crate
──────────────────                         ──────────────────
huge generic future F
        │
        ▼
  Box::pin(fut)   ◄── box #1 (task_runtime.rs:241)
        │
        │  small Pin<Box<dyn Future>>
        ▼
  run_inner_async()
   wraps it in select! { cancel, fut } ──► NEW generic future F'
                                                   │
                                                   ▼
                                          XetRuntime::bridge_async(fut: F')
                                                   │
                                             Box::pin(fut)  ◄── box #2 (native.rs:420)
                                                   │
                                                   │ small Pin<Box<dyn Future>>
                                                   ▼
                                          bridge_to_owned(fut: Pin<Box<dyn Future>>)
                                          (stays non-generic — safe from bloat)
```

Each box is the "earliest boundary" for its own crate's generic entry point — they don't overlap, and both are load-bearing.

kszucs: It might be better to change `bridge_async`/`bridge_async_finalizing` (both in `xet_session` and `xet_runtime`) to take `Pin<Box<dyn Future<Output = T> + Send>>` directly rather than generic `F: Future + Send + 'static`, mirroring what `bridge_to_owned` already does.

Though this would be an API changes, not sure what our preference is.

xet_pkg/src/xet_session/test_utils.rs
kszucs: All three build_survives_small_stack_* tests aborted on windows-latest with exit status 3221225725 (0xC00000FD = STATUS_STACK_OVERFLOW) at the unified 128 KiB stack size, even though the underlying bug is fixed. This proves Windows genuinely needs more baseline stack for the tokio runtime/thread setup itself, unrelated to the future-boxing issue.

Reverted to the platform-specific split (128 KiB non-Windows / 512 KiB Windows).

xet_runtime/src/core/runtime/native.rs
seanses: My intuition is that after boxing in the xet_session crate the future is already small, and `run_inner_async()` just selects over that boxed fut. So that future passed into `XetRuntime::bridge_async` won't "bloat" badly. I don't think we need to accommodate for direct XetRuntime::bridge_async callers (that's what I think the inner box is for), because the `xet_runtime` is not a "runtime" crate for general usage.

See the last commit in https://github.com/huggingface/xet-core/pull/912 where I dropped the inner box and CI tests pass. WDYT?

seanses (approved):
LGTM! Thanks!

kszucs:
cc @Xuanwo this will resolve some stack related issues on the opendal side

PR #910, opened 2026-07-22
Local simulation test for v2 shards upload API

seanses:
## Summary

Adds local simulation coverage for the `POST /v2/shards` NDJSON streaming endpoint, following our established pattern of exercising new CAS-facing APIs against the in-process `LocalServer`/`LocalTestServer` harness rather than only unit-testing client-side parsing logic.

- `local_server/handlers.rs`: new `post_shard_v2` handler — validates the shard via the same `DirectAccessClient::upload_shard`, then emits a production-shaped NDJSON event stream (`Validating` → `Committing` → `Result`). V2→V1 fallback is forced via `/simulation/set_config?config=disable_v2_endpoints` (renamed from `disable_v2_reconstruction`; one status code now disables both V2 reconstruction and V2 shard upload).
- `local_server/server.rs` / `main.rs`: routes `/v2/shards` and updates route docs. Registering this route flips the default shard-upload path for existing simulation tests to V2-first (auto-detect); V1's JSON path is still covered by the explicit fallback checks.
- Renames `disable_v2_reconstruction` → `disable_v2_endpoints` across the `DirectAccessClient` trait, local/memory clients, simulation config, and downstream call sites.
- `simulation_server.rs`: two new checks —
  - `check_v2_shard_upload`: verifies the exact NDJSON stream (Content-Type, frame order, counters, stages) both over raw HTTP and through `RemoteClient`'s progress callback.
  - `check_v2_shard_upload_disabled_fallback`: verifies V2→V1 fallback at both 501 and 404, including a fresh `RemoteClient` for the 404 case so the auto-detect path is actually exercised rather than replaying a cached V1 preference from the earlier 501 check.

## Test plan

- [x] `cargo test --package xet-client --lib --features simulation test_local_server` (both in-memory and disk-backed storage variants)
- [x] `cargo clippy --package xet-client --features simulation --tests -- -D warnings` (no new warnings vs. base)
- [x] Verified the 404-fallback assertion actually catches a regression: temporarily broke the `NOT_FOUND` arm in `RemoteClient::upload_shard_with_version_override` and confirmed the test fails with the real 404 surfacing, then reverted.

---

> [!NOTE]
> **Low Risk**
> Changes are confined to the simulation/local-server test harness and a transitive lockfile bump; production client paths are only touched via renamed test helpers.
> 
> **Overview**
> Adds **local CAS simulation** for production-style **`POST /v2/shards`** uploads: NDJSON progress (`Validating` → `Committing` → `Result`) and **HTTP 200 + terminal `Error` frames** (fatal vs retryable), configurable via **`shard_upload_error_frame`** on `/simulation/set_config`.
> 
> Renames **`disable_v2_reconstruction` → `disable_v2_endpoints`** so one simulated status disables **both** V2 reconstruction and V2 shard upload for V2→V1 fallback tests.
> 
> **`simulation_server`** gains integration checks for the NDJSON stream (raw HTTP + `RemoteClient` callbacks), disabled-endpoint fallback (501/404, including a fresh client for 404 auto-detect), and in-stream error/retry behavior. **`Cargo.lock`** bumps `event-listener` 5.4.1 → 5.4.2.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f9ddab797bfdd7a444d472ad17bf61b70d4d265. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/xet_client/src/cas_client/simulation/direct_access_client.rs
+++ b/xet_client/src/cas_client/simulation/direct_access_client.rs
@@ -58,12 +58,12 @@ pub trait DirectAccessClient: Client + Send + Sync {
     /// Pass `None` to disable (default: returns full shards with no expiration).
     fn set_global_dedup_shard_expiration(&self, expiration: Option<Duration>);
 
-    /// Disables V2 reconstruction responses with the given HTTP status code.
-    /// When disabled, the V2 endpoint returns this status, forcing clients to
-    /// fall back to V1. Pass 0 to re-enable.
+    /// Disables V2 endpoints (reconstruction and shard upload) with the given HTTP status code.
+    /// When disabled, `/v2/*` handlers return this status, forcing clients to fall back to V1.
+    /// Pass 0 to re-enable.
     fn disable_v2_reconstruction(&self, status_code: u16);
 
-    /// Returns the HTTP status code the V2 endpoint should return when disabled,
+    /// Returns the HTTP status code V2 endpoints should return when disabled,
     /// or 0 if V2 is enabled.
     fn v2_disabled_status_code(&self) -> u16 {
         0
--- a/xet_client/src/cas_client/simulation/direct_access_client.rs
+++ b/xet_client/src/cas_client/simulation/direct_access_client.rs
@@ -61,7 +61,7 @@ pub trait DirectAccessClient: Client + Send + Sync {
     /// Disables V2 endpoints (reconstruction and shard upload) with the given HTTP status code.
     /// When disabled, `/v2/*` handlers return this status, forcing clients to fall back to V1.
     /// Pass 0 to re-enable.
-    fn disable_v2_reconstruction(&self, status_code: u16);
+    fn disable_v2_endpoints(&self, status_code: u16);
 
     /// Returns the HTTP status code V2 endpoints should return when disabled,
     /// or 0 if V2 is enabled.
--- a/xet_client/src/cas_client/simulation/local_server/handlers.rs
+++ b/xet_client/src/cas_client/simulation/local_server/handlers.rs
@@ -31,8 +31,9 @@ use xet_core_structures::merklehash::MerkleHash;
 use super::super::super::{DeletionControlableClient, DirectAccessClient};
 use super::latency_simulation::{LatencySimulation, ServerLatencyProfile};
 use crate::cas_types::{
-    FileRange, HexKey, HexMerkleHash, QueryReconstructionResponseV2, UploadShardResponse, UploadShardResponseType,
-    UploadXorbResponse, XorbRangeDescriptor, XorbReconstructionFetchInfo,
+    CommitStage, FileRange, HexKey, HexMerkleHash, QueryReconstructionResponseV2, ShardUploadEvent,
+    UploadShardResponse, UploadShardResponseType, UploadXorbResponse, XorbRangeDescriptor,
+    XorbReconstructionFetchInfo,
 };
 use crate::error::ClientError;
 
@@ -587,6 +588,71 @@ pub async fn post_shard(State(state): State<ServerState>, body: Body) -> Respons
     }
 }
 
+/// POST /v2/shards
+///
+/// Upload a shard with an NDJSON progress stream (same event types production CAS emits).
+/// Respects [`DirectAccessClient::v2_disabled_status_code`] so clients can exercise V2→V1 fallback.
+///
+/// On success the response is `200` with `application/x-ndjson` frames ending in `result`.
+/// Backend upload failures map to HTTP errors (the stream never starts).
+pub async fn post_shard_v2(State(state): State<ServerState>, body: Body) -> Response {
+    let connection_guard = state.latency_simulation.register_connection().await;
+    if let Some(simulated_error) = connection_guard.simulate_error() {
+        return simulated_error;
+    }
+
+    // Allow testing V1 fallback by simulating V2 endpoint unavailability.
+    let disabled_status = state.client.v2_disabled_status_code();
+    if disabled_status != 0 {
+        let code = StatusCode::from_u16(disabled_status).unwrap_or(StatusCode::NOT_FOUND);
+        return (code, "V2 shard upload endpoint disabled").into_response();
+    }
+
+    let data = match collect_body(body).await {
+        Ok(d) => d,
+        Err(e) => return (StatusCode::BAD_REQUEST, e).into_response(),
+    };
+
+    let permit = match state.client.acquire_upload_permit().await {
+        Ok(p) => p,
+        Err(e) => return error_to_response(e),
+    };
+
+    match state.client.upload_shard(data, permit, None).await {
+        Ok(()) => ndjson_shard_upload_success_response(),
+        Err(e) => error_to_response(e),
+    }
+}
+
+/// Builds a minimal production-shaped NDJSON success stream for `/v2/shards`.
+fn ndjson_shard_upload_success_response() -> Response {
+    let events = [
+        ShardUploadEvent::Validating { verified: 0, total: 1 },
+        ShardUploadEvent::Validating { verified: 1, total: 1 },
+        ShardUploadEvent::Committing {
+            stage: CommitStage::Uploading,
+        },
+        ShardUploadEvent::Committing {
+            stage: CommitStage::Syncing,
+        },
+        ShardUploadEvent::Result,
+    ];
+
+    let mut body = Vec::new();
+    for event in &events {
+        // Simulation frames are always valid `ShardUploadEvent` values.
+        serde_json::to_writer(&mut body, event).expect("ShardUploadEvent serializes");
+        body.push(b'\n');
+    }
+
+    let mut headers = HeaderMap::new();
+    headers.insert(
+        http::header::CONTENT_TYPE,
+        HeaderValue::from_static("application/x-ndjson"),
+    );
+    (StatusCode::OK, headers, body).into_response()
+}
+
 /// HEAD /v1/files/{file_id}
 ///
 /// Get the size of a file.
@@ -786,7 +852,7 @@ pub async fn set_config(State(state): State<ServerState>, uri: Uri, body: Body) 
         "disable_v2_reconstruction" => match value.parse::<u16>() {
             Ok(code) => {
                 state.client.disable_v2_reconstruction(code);
-                (StatusCode::OK, "V2 reconstruction config set").into_response()
+                (StatusCode::OK, "V2 endpoints (reconstruction + shard upload) config set").into_response()
             },
             Err(e) => (StatusCode::BAD_REQUEST, format!("Invalid status code: {e}")).into_response(),
         },
--- a/xet_client/src/cas_client/simulation/local_server/handlers.rs
+++ b/xet_client/src/cas_client/simulation/local_server/handlers.rs
@@ -14,6 +14,7 @@
 //! Errors are mapped to appropriate HTTP status codes via `error_to_response`.
 
 use std::sync::Arc;
+use std::sync::atomic::{AtomicU8, Ordering};
 use std::time::Duration;
 
 use axum::Json;
@@ -36,12 +37,25 @@ use crate::cas_types::{
 };
 use crate::error::ClientError;
 
+/// `/v2/shards` in-stream `Error` frame mode (`set_config` + handler encoding).
+#[repr(u8)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub(crate) enum ShardUploadErrorFrame {
+    #[default]
+    Off = 0,
+    Fatal = 1,
+    Retryable = 2,
+}
+
 /// Server state passed to all handlers.
 #[derive(Clone)]
 pub(crate) struct ServerState {
     pub(crate) client: Arc<dyn DirectAccessClient>,
     pub(super) latency_simulation: Arc<LatencySimulation>,
     pub(crate) deletion_client: Option<Arc<dyn DeletionControlableClient>>,
+    /// While set, `/v2/shards` emits an in-stream `Error` frame instead of committing.
+    /// Encodes [`ShardUploadErrorFrame`] as `u8`.
+    pub(crate) shard_upload_error_frame: Arc<AtomicU8>,
 }
 
 /// Represents the different forms a Range header can take.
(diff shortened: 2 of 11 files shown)


seanses:
Ignoring the "Cargo Audit" failure now, will bump the corresponding dep in another PR

xet_client/src/cas_client/simulation/simulation_server.rs
```diff
@@ -1202,6 +1202,181 @@ mod tests {
...
+            .await
+            .unwrap();
+        assert_eq!(response.status(), reqwest::StatusCode::OK);
+        let content_type = response
+            .headers()
+            .get(reqwest::header::CONTENT_TYPE)
+            .and_then(|v| v.to_str().ok())
+            .unwrap_or("");
+        assert!(
+            content_type.contains("ndjson") || content_type.contains("json"),
```
rajatarya: `content_type.contains("ndjson") || content_type.contains("json")` — the second disjunct subsumes the first (`"ndjson"` contains `"json"`), so this accepts `application/json`, `text/json`, anything JSON-ish. Given the handler sets the header via `HeaderValue::from_static("application/x-ndjson")`, the exact check is free:

```rust
assert_eq!(content_type, "application/x-ndjson");
```

I verified this isn't theoretical — see the mutation test in my next comment; changing the handler to `application/json` did not fail either new check.

xet_client/src/cas_client/simulation/simulation_server.rs
```diff
@@ -1202,6 +1202,181 @@ mod tests {
...
+            "unexpected content-type: {content_type}"
+        );
+
+        let body = response.text().await.unwrap();
+        let events: Vec<ShardUploadEvent> = body
+            .lines()
+            .filter(|line| !line.is_empty())
+            .map(|line| serde_json::from_str(line).expect("valid ShardUploadEvent NDJSON line"))
+            .collect();
+        assert!(
```
rajatarya: This is the one I'd most like addressed, because it undercuts what the PR is for.

I applied a single combined mutation to `ndjson_shard_upload_success_response`: reversed the pipeline to `Committing{Syncing}` → `Validating{verified:99,total:1}` → `Result`, dropped two frames, corrupted the counters, **and** changed the Content-Type to `application/json`. **Both new checks stayed green.**

The reason is that `events.iter().any(Validating)` + `any(Committing)` + `last() == Result` pins neither order, nor frame count, nor field values — and the Content-Type check above accepts `application/json`. So a four-way regression to the exact contract this test exists to pin sails straight through.

Worth noting the pre-existing unit test `test_read_ndjson_invokes_callback_for_every_frame_in_order` in `shard_upload_v2.rs` *does* assert exact order, so the new end-to-end check is currently weaker than existing coverage on the very property it claims to add. Since the handler is deterministic, exact equality should just work:

```rust
assert_eq!(events, vec![/* the 5 exact frames */]);
```

Same applies to the `seen` assertions further down — and the `Committing { stage: Uploading } | Committing { stage: Syncing }` or-pattern there is equivalent to `Committing { .. }` since `CommitStage` has only two variants, so the apparent precision can't actually detect a dropped stage. Asserting both stages were observed would.

xet_client/src/cas_client/simulation/direct_access_client.rs
```diff
@@ -58,12 +58,12 @@ pub trait DirectAccessClient: Client + Send + Sync {
...
     fn set_global_dedup_shard_expiration(&self, expiration: Option<Duration>);
 
-    /// Disables V2 reconstruction responses with the given HTTP status code.
-    /// When disabled, the V2 endpoint returns this status, forcing clients to
-    /// fall back to V1. Pass 0 to re-enable.
-    fn disable_v2_reconstruction(&self, status_code: u16);
+    /// Disables V2 endpoints (reconstruction and shard upload) with the given HTTP status code.
+    /// When disabled, `/v2/*` handlers return this status, forcing clients to fall back to V1.
+    /// Pass 0 to re-enable.
+    fn disable_v2_endpoints(&self, status_code: u16);
```
rajatarya, line 64: This rename (`disable_v2_reconstruction` → `disable_v2_endpoints`) is applied consistently — trait def, `LocalClient`, `MemoryClient`, both `LocalTestServer` impls, `SimulationControlClient`, the query-param match arm and its "Supported:" help text, and all 7 `xet_data` call sites. Nice and thorough.

Two things it needs, though, and per our convention a trait-method + wire-key rename isn't a nit:

1. **No `api_changes/` entry.** `api_changes/README.md` asks for one when a PR involves an API change potentially requiring downstream updates. This qualifies on both counts, and it's not simulation-internal: `simulation/mod.rs` declares and re-exports `direct_access_client` with no `#[cfg(feature = "simulation")]`, and `cas_client/mod.rs` re-exports it unconditionally — so `xet_client::cas_client::DirectAccessClient` is public API in the *default* feature set. An out-of-repo implementor breaks on a routine dep bump with nothing explaining why.

2. **A stale entry now advertises the dead key.** `api_changes/update_260328_simulation_global_dedup_expiration_and_config_controls.md:53` still lists `disable_v2_reconstruction` under supported `set_config` keys. That key now returns 400 "Unknown config". Since those files are dated historical records, a new `update_2607xx_disable_v2_endpoints_rename.md` is probably the right fix rather than editing line 53 — and item 1 needs that file anyway.

The silent-failure shape is what makes this worth doing: an external caller POSTing the old key gets a 400 it likely doesn't check, and simply observes "V2 is still enabled" with no error surfaced.

xet_client/src/cas_client/simulation/local_server/handlers.rs
```diff
@@ -587,6 +587,68 @@ pub async fn post_shard(State(state): State<ServerState>, body: Body) -> Respons
     }
 }
 
+/// POST /v2/shards
+///
+/// Upload a shard with an NDJSON progress stream (same event types production CAS emits).
+/// Respects [`DirectAccessClient::v2_disabled_status_code`] so clients can exercise V2→V1 fallback.
+///
+/// On success the response is `200` with `application/x-ndjson` frames ending in `result`.
+/// Backend upload failures map to HTTP errors (the stream never starts).
```
rajatarya: "Backend upload failures map to HTTP errors (the stream never starts)" is the opposite of the production contract: `cas_types/mod.rs` documents an `Error` frame delivered *inside* a 200 stream, and says clients MUST treat that frame as the error signal precisely because the status line is already committed. There's no knob here to force one, so the simulation can't reach that path at all.

That leaves the retry wiring untested end-to-end — the parser unit tests cover classification but call `read_shard_upload_ndjson` directly, never through `RetryWrapper::run_and_process`. So if the `retryable` flag's polarity gets inverted, or `run_and_process` stops folding `RetryableError` into a retry, a transient CAS fault becomes a hard upload failure in production and nothing in the repo catches it.

Happy for this to be a follow-up rather than in-scope here, but it's the single highest-value thing this harness could add next — a `set_config` knob like `shard_upload_error_frame=retryable|fatal` would make the 200-with-error-frame path reachable. At minimum, could the doc comment stop asserting the opposite of the real contract?

xet_client/src/cas_client/simulation/simulation_server.rs
```diff
@@ -1202,6 +1202,181 @@ mod tests {
...
+                .upload_shard_with_version_override(shard_data.clone(), permit, None, None)
+                .await
+                .unwrap();
+        }
+
+        // Re-enable, confirm V2 works, then exercise 404 fallback.
+        server.client().disable_v2_endpoints(0);
+        {
+            let permit = server.remote_client().acquire_upload_permit().await.unwrap();
+            // Force V2 so we don't stay on a cached V1 preference from the previous fallback.
```
rajatarya: nit: this comment is inaccurate and contradicts the (correct) one ~20 lines below. `forced_version = Some(2)` does *not* clear the cached preference — `remote_client.rs` guards the `store(2)` on `forced_version.is_none()`, so `detected_shard_api_version` stays at 1.

Worth fixing because of what it invites: a maintainer trusts this line, concludes the fresh-client block below is redundant, deletes it, and the 404 auto-detect arm quietly stops being tested while the suite stays green.

rajatarya (commented):
Thanks, Di — the design here is right and the harness is a genuinely useful addition. A few things I want to call out as solid before the asks, because I went looking for problems in them and didn't find any:

- **The NDJSON contract can't drift.** The handler constructs the *same* `ShardUploadEvent`/`CommitStage` types the client parser consumes, so tag/field/serde skew is structurally impossible rather than merely currently-correct. Emitted order matches the `PartialOrd`/`precede` pipeline contract.
- **Content-Type is right on the wire.** I chased a suspected duplicate-header bug here (axum's `Vec<u8>` body sets `application/octet-stream` first, then the tuple's `HeaderMap` merges) — it's fine, because `HeaderMap::into_iter` yields `(Option<HeaderName>, T)` and so selects http's *overwriting* `Extend` impl. Newline after every frame including the last.
- **The fresh-client trick is sound.** `detected_shard_api_version` is a per-client `AtomicU32`, and direct `upload_shard_v2` never writes the cache, so `check_v2_shard_upload` genuinely cannot pollute the fallback check.
- **Routes are clean** — `/v2/shards` sits in the `/v2` nest, `/v1/shards` untouched, no catch-all or `fallback` to shadow it, and the `main.rs` route docs got some good drive-by corrections.
- I ran the suite on the branch: **both `test_local_server` variants pass**, and production `remote_client.rs` / `shard_upload_v2.rs` / `cas_types/mod.rs` are untouched, so the "validated, not altered" claim holds for behavior.

Three things I'd like to land before this merges, all mechanical:

1. **The new assertions are weaker than they look — I mutation-tested them and they don't catch a four-way regression** to the exact contract this PR exists to pin. Reversed frame order, dropped frames, corrupted counters, *and* a wrong Content-Type all pass. Details and the one-line fixes inline. This is the main ask, since a simulation that silently stops matching production is worse than no simulation — the next reader will trust it as the contract reference.
2. **An `api_changes/` entry for the `disable_v2_reconstruction` → `disable_v2_endpoints` rename.** `DirectAccessClient` is re-exported without a `simulation` cfg gate, so this is public API in the default feature set; there's also a stale api_changes file still advertising the now-400 key. Inline.
3. **The PR description names the old knob** (`config=disable_v2_reconstruction`, "the existing knob") and never mentions the rename — which is the most cross-cutting change in the diff. Worth correcting; it's what reviewers and future archaeologists read first, and I suspect it's why the rename hasn't drawn scrutiny yet.

Two more for discussion rather than gating:

4. **Registering `/v2/shards` silently flips the default shard-upload path for every pre-existing simulation test.** `shard_api_version` defaults to `None` → auto-detect → V2 first, so before this PR every harness shard upload 404'd to V1, and now they all route through `post_shard_v2`. I confirmed by deleting the `state.client.upload_shard(...)` call — `check_basic_correctness` fails with `FileNotFound`, a check with nothing to do with V2. Probably desirable (more production-shaped!), but it deserves a sentence in the description, and it means V1's JSON/`Exists`-vs-`SyncPerformed` path is now only covered by the two explicit fallback calls.
5. **The production error contract isn't simulatable** — an `Error` frame inside a 200, which `cas_types` says clients MUST treat as the error signal. Inline; good follow-up if out of scope.

Holding off on approval for (1)–(3), but none of it is deep — happy to re-review quickly once the assertions tighten up.

xet_client/src/cas_client/simulation/simulation_server.rs
```diff
@@ -1202,6 +1202,181 @@ mod tests {
         check_chunk_hashes_correctness(server).await;
         check_simulation_set_config(server).await;
         check_simulation_dummy_upload(server).await;
+        check_v2_shard_upload(server).await;
+        check_v2_shard_upload_disabled_fallback(server).await;
+    }
+
+    /// Builds raw shard bytes suitable for `/v1/shards` and `/v2/shards` upload tests.
+    fn random_shard_bytes() -> bytes::Bytes {
```
rajatarya: nit: `gen_random_shard_with_xorb_references(0, ...)` — that first arg is a `seed`, so a fixed `0` means every call returns byte-identical output. The re-generations at the fallback call sites are therefore no-ops, and each "fresh" upload is really a duplicate-shard re-registration rather than a new sync. `fixed_test_shard_bytes()` would describe it honestly, or thread a varying seed if you want genuinely distinct shards.

Related, and benign today: this shard references 8 xorbs that are never uploaded, so these checks leave dangling refs in shared server state. Fine while `run_all_server_checks` never calls `verify_integrity()` — it'd bite whoever adds one.

xet_client/src/cas_client/simulation/local_server/handlers.rs
```diff
@@ -587,6 +587,68 @@ pub async fn post_shard(State(state): State<ServerState>, body: Body) -> Respons
...
+pub async fn post_shard_v2(State(state): State<ServerState>, body: Body) -> Response {
+    let connection_guard = state.latency_simulation.register_connection().await;
+    if let Some(simulated_error) = connection_guard.simulate_error() {
+        return simulated_error;
+    }
+
+    // Allow testing V1 fallback by simulating V2 endpoint unavailability.
+    let disabled_status = state.client.v2_disabled_status_code();
+    if disabled_status != 0 {
```
rajatarya, line 622: nit: this returns before reading `body`, which works today only because test shards are a few KB and fit in socket buffers. With a large shard the client is mid-`Body::wrap_stream` write when the server responds and closes, and reqwest can surface a write/connection error whose `.status()` is `None` — in which case the `matches!(e.status(), Some(NOT_FOUND) | Some(NOT_IMPLEMENTED))` guard in `remote_client.rs` never fires.

So the fallback assertions could flake, and more importantly real V2→V1 fallback would silently stop working for large shards against a CAS deployment that rejects `/v2/shards` before reading the body. This mirrors the existing `get_reconstruction_v2` pattern, but that's a GET with no body, so the precedent doesn't quite carry.

rajatarya (approved):
Overall looks good, think the one test caveat should get fixed before merge.

xet_client/src/cas_client/simulation/direct_access_client.rs
seanses: These are building blocks for internal testing only - there's no need to document in an api_changes entry.

xet_client/src/cas_client/simulation/simulation_server.rs
seanses: Updated

xet_client/src/cas_client/simulation/simulation_server.rs
seanses: Updated to the suggested assertions.

xet_client/src/cas_client/simulation/simulation_server.rs
seanses: Updated the comment.

xet_client/src/cas_client/simulation/local_server/handlers.rs
seanses: If that route is removed, Axum never enters post_shard_v2 (which is where request.into_body() is consumed) and returns 404 immediately with no application-level body drain. Auth/route_layer also only apply to defined routes, so they don’t run either. (That’s a 404, not 501 — 501 only if something explicitly returns NOT_IMPLEMENTED.)

So the early return here before reading body matches the “route unavailable” production path we’re simulating. Draining first would make the harness less production-shaped for that case.

The large-body .status() == None risk is real, but it’s a client↔CAS interaction (worth handling in RemoteClient if we see it), not something this sim should paper over by draining. Leaving as-is.

xet_client/src/cas_client/simulation/local_server/handlers.rs
seanses: Now simulates the error frame

xet_client/src/cas_client/simulation/simulation_server.rs
seanses: Renamed and added comments

Issue #895, opened 2026-07-11
Download fails with "Task error: Unable to parse string as hex hash value" (hf-xet 1.5.1)


btseytlin:
## Environment
- hf-xet 1.5.1 (latest), huggingface_hub 1.23.0, Python 3.13, macOS 15.6 (arm64)

## What happened
`snapshot_download` on a gated private dataset repo (14 files; one 11.9 GB xet-backed blob) fails shortly after starting:

```
File ".../huggingface_hub/file_download.py", line 1920, in _download_to_tmp_and_move
    xet_get(
File ".../huggingface_hub/file_download.py", line 563, in xet_get
    with session.new_file_download_group(
        token_refresh_url=xet_file_data.refresh_route,
        ...
    ) as group:
RuntimeError: Task error: Unable to parse string as hex hash value
```

Small files in the same snapshot download fine; the failure happens on the large blob. Auth token is valid (the same call succeeds with xet disabled).

## Workaround
`HF_HUB_DISABLE_XET=1` downloads the same snapshot successfully over plain HTTP.

Happy to provide more details or run a patched build — the repo is private so I can't share a public repro, but the failure is reproducible on this snapshot.

seanses:
Sounds like `huggingface_hub` is passing incorrect file hash to `hf-xet` leading to this "Unable to parse string as hex hash value" error. Can you share the repo id (if you need a private channel feel free to send it to my email: di@huggingface.co). cc @Wauplin

wzshiming:
I'm having the same problem.

- huggingface_hub version: 1.24.0
- hf_xet: 1.5.2

```
hf download google/translategemma-4b-it
Fetching 15 files:  40%|██████████████████████████████████████████████████████████▍                                                                                       | 6/15 [00:00<00:00, 20.20it/s]
Reconstructing (incomplete total...):   0%|                                                                                                                                 |  0.00B / 4.96GB            Traceback (most recent call last):..):   0%|                                                                                                                                 |  0.00B / 8.64GB
  File "/opt/homebrew/bin/hf", line 6, in <module>
    sys.exit(main())
             ~~~~^^
  File "/opt/homebrew/Cellar/hf/1.24.0/libexec/lib/python3.14/site-packages/huggingface_hub/cli/hf.py", line 128, in main
    app()
    ~~~^^
  File "/opt/homebrew/Cellar/hf/1.24.0/libexec/lib/python3.14/site-packages/click/core.py", line 1569, in __call__
[...]

PR #914, opened 2026-07-24
Embed input in the DataHash parse Error

seanses:
## Summary
- Include the failing input in DataHash parse errors so download failures like [#895](https://github.com/huggingface/xet-core/issues/895) surface `Unable to parse string as hex hash value (got '...')` instead of an opaque message, helps debugging.
- Extract hex/bytes parse errors into `merklehash/error.rs` as a `thiserror` `DataHashError` enum (`InvalidHex` / `InvalidBytes`), both carrying the bad input for display. This has been our code style.
- Wire `CoreError::HashParsing` to `DataHashError` (replacing the unused `Infallible` stub) and map it to `XetError::DataIntegrity`.

## Test plan
- [x] Unit test that `from_hex` errors retain the bad input
- [x] Reproduce / confirm improved error text on a bad `XetFileInfo.hash` download path

---

> [!NOTE]
> **Low Risk**
> Localized error-type refactor with clearer diagnostics; parsing behavior unchanged aside from richer messages and explicit `HashParsing` → data-integrity classification.
> 
> **Overview**
> **DataHash parse failures now include the offending input** in error text (truncated at 72 characters), so paths like download/`XetFileInfo` hash parsing surface messages such as `Unable to parse string as hex hash value (got '...')` instead of generic failures.
> 
> The old `DataHashHexParseError` / `DataHashBytesParseError` types are replaced by a **`thiserror` `DataHashError`** enum (`InvalidHex` / `InvalidBytes`) in `merklehash/error.rs`, with `from_hex`, `from_slice`, and `from_base64` building errors via helpers that capture and truncate input. **`CoreError::HashParsing`** now wraps `DataHashError` (replacing the unused `Infallible` stub and manual `From` impls), and **`XetError`** treats hash parsing as **data integrity**. Call sites in WASM, chunk cache, and `xet_data` are updated to the unified type; unit tests assert input retention and truncation behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ca009c5f1914823e0737f007efb03a0a7767d818. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/xet_core_structures/src/merklehash/data_hash.rs
+++ b/xet_core_structures/src/merklehash/data_hash.rs
@@ -1,9 +1,7 @@
 use std::cmp::{Eq, Ord, Ordering, PartialEq, PartialOrd};
-use std::error::Error;
 use std::hash::{Hash, Hasher};
 use std::io::Write;
 use std::mem::transmute_copy;
-use std::num::ParseIntError;
 use std::ops::{Deref, DerefMut};
 use std::{fmt, str};
 
@@ -16,6 +14,8 @@ use rand::rngs::SmallRng;
 use rand::{Rng, SeedableRng};
 use safe_transmute::{transmute_to_bytes, transmute_to_bytes_mut};
 use serde::{Deserialize, Serialize};
+
+use super::error::DataHashError;
 
 /************************************************************************* */
 /*  */
@@ -135,24 +135,6 @@ unsafe impl Zeroable for DataHash {
 #[cfg(not(target_family = "wasm"))]
 unsafe impl Pod for DataHash {}
 
-/// The error type that is returned if [DataHash::from_hex] fails.
-#[derive(Debug, Clone)]
-pub struct DataHashHexParseError;
-
-impl Error for DataHashHexParseError {}
-
-impl fmt::Display for DataHashHexParseError {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "Invalid hex input for DataHash")
-    }
-}
-
-impl From<ParseIntError> for DataHashHexParseError {
-    fn from(_err: ParseIntError) -> Self {
-        DataHashHexParseError {}
-    }
-}
-
 impl DataHash {
     /// Returns the hexadecimal printout of the hash.
     /// Different programming languages and platforms may have different byte order layout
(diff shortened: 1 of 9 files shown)


seanses:
Cherry-picked this into branch `di/v1.5.3-dev0` on top of `v1.5.2`, and released `hf-xet` Python [1.5.3.dev0](https://pypi.org/project/hf-xet/1.5.3.dev0/). With this release user will see bad hash in the error message.

Example:

```
File ~/hf/huggingface_hub/src/huggingface_hub/file_download.py:563, in xet_get(incomplete_path, xet_file_data, headers, expected_size, displayed_filename, tqdm_class, _tqdm_bar)
    553 with XetDownloadProgressReporter(
    554     reconstruction_desc=f"{displayed_filename}: reconstructing file",
    555     transfer_desc=f"{displayed_filename}: downloading bytes",
   (...)    560     external_reconstruction_bar=_tqdm_bar,
    561 ) as progress:
    562     try:
--> 563         with session.new_file_download_group(
    564             token_refresh_url=xet_file_data.refresh_route,
    565             token_refresh_headers=headers,
    566             custom_headers=xet_headers,
    567             progress_callback=progress.update_progress,
    568         ) as group:
    569             group.start_download_file(
    570                 XetFileInfo("abc", expected_size), str(incomplete_path.absolute())
    571             )
    572     except KeyboardInterrupt:

RuntimeError: Task error: Unable to parse string as hex hash value (got 'abc')
```

assafvayner (approved):

xet_core_structures/src/merklehash/error.rs
```diff
@@ -0,0 +1,35 @@
...
+    #[error("Invalid bytes input for DataHash (got '{input}')")]
+    InvalidBytes {
+        /// Human-readable form of the bad input (hex for raw bytes, or the original string).
+        input: String,
+    },
+}
+
+impl DataHashError {
+    /// Returns the bad input string for either variant.
+    pub fn hex_input(&self) -> &str {
```
rajatarya: This is a public accessor, so the name matters: for `InvalidBytes` the stored `input` is hex only on the `from_slice` path — on the `from_base64` decode-failure path it's the raw base64 string (`input: b64.to_owned()`), which your doc comment acknowledges ("hex for raw bytes, or the original string"). So `hex_input()` is misleading there. What do you think about a neutral name like `input()` or `bad_input()`?

xet_core_structures/src/merklehash/data_hash.rs
```diff
@@ -177,27 +159,30 @@ impl DataHash {
     }
 
     /// Parses a hexadecimal string as a DataHash, returning
-    /// Err(DataHashHexParseError) on failure.
-    pub fn from_hex(h: &str) -> Result<DataHash, DataHashHexParseError> {
+    /// Err(DataHashError) on failure.
+    pub fn from_hex(h: &str) -> Result<DataHash, DataHashError> {
+        let parse_err = || DataHashError::InvalidHex { input: h.to_owned() };
```
rajatarya: This embeds the full input regardless of length (the `len() != 64` branch still returns `parse_err()` with the whole string), and `from_slice`/`from_base64` do likewise. Callers here are hash-sized so it's low risk, but since the message ends up in logs and Python exceptions, would you consider capping the embedded value (e.g. first ~72 chars + ellipsis) as a guard against a malformed/oversized input?

xet_core_structures/src/merklehash/data_hash.rs
```diff
@@ -586,4 +561,16 @@ mod tests {
         assert_eq!(expected_hex_string, data_hash.hex().as_str());
         assert_eq!(DataHash::from_hex(expected_hex_string).unwrap(), data_hash);
     }
+
+    #[test]
+    fn test_from_hex_parse_error_includes_input() {
+        let empty = DataHash::from_hex("").unwrap_err();
```
rajatarya, line 565: nit: this asserts on `hex_input()`, but the actual feature is the `Display` message (`...(got '...')`). Consider also asserting `err.to_string().contains("abcdef")` and adding an `InvalidBytes` case (`from_slice`/`from_base64`) so the user-facing format is covered too.

rajatarya (commented):
Thanks Di — this is a clean, well-scoped change and a nice debugging win. I traced all 9 files plus the downstream consumers, and the refactor propagates completely: no stray references to the old `DataHashHexParseError`/`DataHashBytesParseError` types remain, every `#[from]` site (ChunkCacheError, CoreError, DataError) is updated, and there are no exhaustive matches elsewhere that break on the new enum. A few things I liked: consolidating into a `thiserror` enum in `merklehash/error.rs` matches the house style, `#[non_exhaustive]` leaves room to grow, and `format_bytes` is non-panicking. I also confirmed the `CoreError::HashParsing` rewire is sound — it was previously uninhabited (`Infallible`) and hash-parse failures fell through `CoreError::Other` → `XetError::Internal`; now they land on `HashParsing` → `XetError::DataIntegrity`, and since both map to `PyRuntimeError` the Python-facing exception class is unchanged. Good call.

A few things worth discussing:

1. **`hex_input()` naming (public API)** — the accessor returns the stored `input` for both variants, but for `InvalidBytes` that value is hex only on the `from_slice` path; on the `from_base64` decode-failure path it's the original base64 string (which your doc comment acknowledges: "hex for raw bytes, or the original string"). So `hex_input()` will sometimes return non-hex. What do you think about a neutral name like `input()` / `bad_input()`? (inline)

2. **Unbounded input in the error/log** — `from_hex` embeds the full input even on the `len() != 64` path, and `from_slice`/`from_base64` embed the whole (hex-expanded) value. Callers here are hash-sized so it's low risk, but since these strings flow into logs and Python exceptions, would you consider capping very long inputs (first N chars + ellipsis) as a guard against a malformed pointer file? (inline)

3. **Test covers the accessor, not the message** (nit) — `test_from_hex_parse_error_includes_input` asserts `hex_input()`, but the feature is really the `Display` output. Consider also asserting `err.to_string().contains(...)` and adding an `InvalidBytes` / `from_slice` case so the user-facing format is locked in. (inline)

One last optional thought (not blocking): `api_changes/update_260309_package_restructure.md` still documents the removed `DataHashHexParseError`/`DataHashBytesParseError` types and the old `→ Other(...)` mappings that this PR changes. If that file is meant as a dated historical snapshot it's fine as-is — just flagging in case it should get a follow-up note.

xet_core_structures/src/merklehash/error.rs
seanses: Renamed to `input()`.

xet_core_structures/src/merklehash/data_hash.rs
seanses: Now store a capped string.

xet_core_structures/src/merklehash/data_hash.rs
seanses: Added test cases.

PR #922, opened 2026-07-30
feat(telemetry): generated metrics schema and CI compatibility gate

sirahd:
> **Stacked on #919** — base is `sira/client-transfer-telemetry`, so the diff here is just this change. Merge #919 first; the base will retarget to `main` automatically.

Publishes the telemetry metric vocabulary as a generated, committed JSON Schema, and enforces the compatibility rules consumers depend on. This is **PR 2 of 2**.

## `telemetry/metrics.schema.json` is the source of truth for what we collect

All **65 properties** carry a description of what they measure and their unit, so a consumer needs nothing from this repo's source to interpret a document:

```json
"defrag_prevented_dedup_bytes": {
  "description": "Bytes that could have been deduplicated but were re-uploaded anyway,
                  because doing otherwise would have fragmented the xorb past the defrag
                  threshold. The cost of the fragmentation-vs-dedup tradeoff.",
  "type": "integer", "format": "uint64", "minimum": 0
}
```

A test fails the build if any property reaches the schema without a description. A metric documented only by its name is a metric nobody outside this repo can read.

`xet_data/src/telemetry/payload.rs` remains the source of truth for the *code*; the schema is generated from it and must never be hand-edited:

```bash
UPDATE_TELEMETRY_SCHEMA=1 cargo test -p xet-data --lib telemetry::schema
```

Without that variable, the same test asserts the committed file is current — so a payload change that forgets to regenerate fails `cargo test`.

`schemars` is a **dev-dependency** and the derives are `cfg(test)`; `cargo tree -p xet-data -e normal` shows zero occurrences of it or its transitive deps, so nothing ships.

## Scope: what this repo publishes

**What the client emits** — property names, JSON types, meanings.

It deliberately does **not** describe how any consumer stores, indexes, or aggregates these documents. That's the consumer's concern, this repo is public, and it has no way to keep a description of someone else's storage layer correct. A consumer types its own storage from each property's `type`:

| JSON Schema | Meaning |
|---|---|
| `"integer"` | fits an unsigned 64-bit integer |
| `"number"` | finite double; never NaN or infinity |
| `"boolean"` | — |
| `"string"` | short, low-cardinality except `transfer_id` |

Every value is a scalar: never null, never nested, never an array.

Two things the `api_changes` note flags for whoever builds the ingestion side: the numeric properties must be stored **as numbers**, or throughput and duration can't be averaged or range-queried — which is the entire point of collecting them. And the mapping should **tolerate an unknown property**, since consumers pin a tag and a client can ship ahead of a consumer rebuild.

## CI gate

`scripts/check_telemetry_schema_compat.py` diffs the branch schema against the merge target in a new `telemetry-schema-compat` job.

| Change | Verdict |
|---|---|
| Property added | pass |
| Property removed | **fail** |
| Property type changed | **fail** |

A missing baseline is treated as the introducing commit. Checkout needs `fetch-depth: 0` — the shallow default can't read the base ref.

## Testing

6 schema tests: drift, every property documented, scalars only, both directions present, cross-direction type agreement, alerting metrics are numeric. I verified the drift test actually fires by tampering with the committed artifact, and exercised the CI script against all six cases (identical / missing baseline / added / removed / retyped / malformed).

- `cargo test --features "strict simulation internal-tools"` — 38 test binaries, all green
- `cargo clippy -r --features "strict simulation internal-tools"` — clean
- `cargo +nightly fmt` — clean

## Changed since first review

Dropped `telemetry/es-index-template.json`. It described the receiving service's document shape and storage layout, which doesn't belong in a public repo. The generated JSON Schema now carries the type information a consumer needs to derive its own mapping, and the storage-specific guidance moved to prose in the `api_changes` note. Related wording elsewhere in these files was made storage-agnostic for the same reason.

> **Note:** `api_changes/update_260728_client_transfer_telemetry.md` from #919 still names the storage technology in several places. I left it alone rather than editing a file from an open PR — say the word and I'll clean it up in either PR.

---

> [!NOTE]
> **Low Risk**
> Mostly contract, tests, and CI; runtime telemetry behavior is unchanged and `schemars` stays dev-only.
> 
> **Overview**
> Introduces a **published, generated contract** for client transfer telemetry: committed `telemetry/metrics.schema.json` is built from `xet_data` payload types (via test-only `schemars` + `schema.rs`), not hand-edited. Payload fields gain `///` doc comments that become schema `description`s; tests require every property to be documented and keep the committed file in sync (`UPDATE_TELEMETRY_SCHEMA=1` to regenerate).
> 
> Adds **`scripts/check_telemetry_schema_compat.py`** and a **`telemetry-schema-compat`** CI job that diffs the branch schema against the merge target: new metrics allowed, removals or JSON `type` changes fail. `.gitattributes` pins the schema to LF for stable byte comparisons on Windows.
> 
> Documents the contract in **`api_changes/update_260729_telemetry_schema_contract.md`** (pin-by-tag URL, regeneration, consumer typing rules). `Cargo.lock` updates reflect `schemars` as a dev-only dependency path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 78197420270b9a9160c1971b25a598320a8ee666. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+# The telemetry schema is compared byte-for-byte against a generated string in
+# `test_committed_schema_is_current`, and is read by `scripts/check_telemetry_schema_compat.py`.
+# Pin it to LF so a Windows checkout does not rewrite it to CRLF.
+telemetry/metrics.schema.json text eol=lf
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,21 @@ jobs:
         run: |
           cargo fmt --manifest-path ./Cargo.toml --all -- --check
           cargo fmt --manifest-path ./hf_xet/Cargo.toml --all -- --check
+  telemetry-schema-compat:
+    name: Telemetry schema compatibility
+    runs-on:
+      group: aws-general-8-plus
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          # The baseline is read from the merge target, so the shallow default clone is not enough.
+          fetch-depth: 0
+      - name: Check for breaking metric changes
+        run: |
+          # Consumers of this schema type each property on first sight and cannot change it
+          # in place. Adding a metric is fine; removing one or changing its type is not.
+          python3 scripts/check_telemetry_schema_compat.py \
+            --baseline-ref "origin/${{ github.base_ref || github.event.repository.default_branch }}"
   detect-unused-dependencies:
     runs-on:
       group: aws-general-8-plus
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -357,7 +357,7 @@ checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -757,7 +757,7 @@ dependencies = [
  "heck",
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.117",
 ]
 
 [[package]]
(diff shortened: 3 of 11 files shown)


seanses (commented):
Trying to understand this PR, what's the purpose of generating a JSON schema? We don't expect users to consume or use the telemetry details right?

PR #931, opened 2026-08-07
fix: remove double counting when reducing fragmentation

jgodlew:
## Problem

`SingleFileCleaner::finish*` can report a `XetFileInfo::file_size` **larger than the bytes actually written**. The same input produces different sizes across runs while the file hash stays identical, so the file content is correct — only the reported size is wrong.

When defrag prevention rejects a match, the range has already been added to `total_bytes`, the
`else` branch never rolls it back, and the chunks are then counted a second time as new data.
`FileCleaner::finish_inner` reports `total_bytes` as the file size, so the over-count becomes the
file's size.

The over-count is therefore exactly `defrag_prevented_dedup_bytes`. 

Measured on the production file: reported `2150261441`, true size `2147483648`, excess `2777793`, which  matches `defrag_prevented_dedup_bytes`.

This is intermittent, only affecting dedup-heavy content.

## Changes

**1. `xet_data/src/deduplication/file_deduplication.rs`** - move the `deduped_chunks` /
`deduped_bytes` / `total_chunks` / `total_bytes` increments inside the branch where the dedup is
actually taken. `fse.unpacked_segment_bytes` is captured once as `deduped_bytes` (it was read four
times, and `fse` is moved by `add_file_data_sequence_entry`).

This also corrects `deduped_bytes` and `deduped_chunks`, which were over-counted the same way.

**2. `xet_data/src/processing/file_cleaner.rs`** - promote the existing `#[cfg(debug_assertions)]`
block to unconditional checks returning `DataError::InternalError`. That assertion already encoded
this exact invariant, but being debug-only meant release builds shipped a wrong size silently.

**3. New `xet_data/tests/test_defrag_prevented_dedup_metrics.rs`** - forces the rejected-dedup path with
tuned config for a smaller repro.

## Notes for reviewers

- **Behaviour change:** callers that previously received a silently-wrong size now get an `Err` from
  `finish()`. That is the intent, but if the double-count was occurring more widely than we know,
  this will surface it as a hard failure rather than bad metadata.

---

> [!NOTE]
> **Medium Risk**
> Changes upload dedup metrics and reported file size (persisted metadata); mismatches now error at finish instead of silently wrong sizes in release builds.
> 
> **Overview**
> Fixes **inflated `XetFileInfo::file_size`** when defrag prevention rejects a dedup match: the deduped range was counted in `total_bytes` before the defrag gate, then counted again as new data when the match was skipped.
> 
> **`file_deduplication.rs`** — Dedup/total chunk and byte metrics (and xorb dependency bytes) are updated only inside the branch where dedup is actually applied; rejected ranges only bump `defrag_prevented_*` and fall through to per-chunk “new data” accounting.
> 
> **`file_cleaner.rs`** — The debug-only size invariant is enforced in all builds: `finish_inner` returns `DataError::InternalError` if pending file info count ≠ 1 or described size ≠ `deduplication_metrics.total_bytes`.
> 
> **New regression test** — Forces the defrag-reject path and asserts `total_bytes` and reported file size match input length.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5e8f24f0a38a60756c0018fba1fa80153639585a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/xet_data/src/deduplication/file_deduplication.rs
+++ b/xet_data/src/deduplication/file_deduplication.rs
@@ -197,16 +197,23 @@ impl<DataInterfaceType: DeduplicationDataInterface> FileDeduper<DataInterfaceTyp
             }
 
             if let Some((n_deduped, fse, is_external)) = dedupe_query {
-                dedup_metrics.deduped_chunks += n_deduped as u64;
-                dedup_metrics.deduped_bytes += fse.unpacked_segment_bytes as u64;
-                dedup_metrics.total_chunks += n_deduped as u64;
-                dedup_metrics.total_bytes += fse.unpacked_segment_bytes as u64;
+                let deduped_bytes = fse.unpacked_segment_bytes as u64;
 
                 // check the fragmentation state and if it is pretty fragmented,
                 // we skip dedupe.  However, continuing the previous is always fine.
                 if self.file_data_sequence_continues_current(&fse)
                     || self.defrag_tracker.allow_dedup_on_next_range(n_deduped)
                 {
+                    // Only account for the range once the dedup is actually taken. Counting it
+                    // before this check meant a defrag-prevented range was added here and then
+                    // again below as new data, inflating `total_bytes` by exactly
+                    // `defrag_prevented_dedup_bytes` — and `total_bytes` is what
+                    // `FileCleaner::finish_inner` reports as the file's size.
+                    dedup_metrics.deduped_chunks += n_deduped as u64;
+                    dedup_metrics.deduped_bytes += deduped_bytes;
+                    dedup_metrics.total_chunks += n_deduped as u64;
+                    dedup_metrics.total_bytes += deduped_bytes;
+
                     // Report this as a dependency
                     // The case where it's dededuped against the present xorb is handled
                     // when the xorb gets cut and we know the hash.
@@ -214,7 +221,7 @@ impl<DataInterfaceType: DeduplicationDataInterface> FileDeduper<DataInterfaceTyp
                         xorb_dependencies.push(FileXorbDependency {
                             file_id: self.file_id,
                             xorb_hash: fse.xorb_hash,
-                            n_bytes: fse.unpacked_segment_bytes as u64,
+                            n_bytes: deduped_bytes,
                             is_external,
                         });
                     }
@@ -225,8 +232,10 @@ impl<DataInterfaceType: DeduplicationDataInterface> FileDeduper<DataInterfaceTyp
                     cur_idx += n_deduped;
                     continue;
                 } else {
+                    // The range is not deduped: it falls through and is counted as new data
+                    // below, one chunk at a time.
                     dedup_metrics.defrag_prevented_dedup_chunks += n_deduped as u64;
-                    dedup_metrics.defrag_prevented_dedup_bytes += fse.unpacked_segment_bytes as u64;
+                    dedup_metrics.defrag_prevented_dedup_bytes += deduped_bytes;
                 }
             }
 
--- a/xet_data/src/processing/file_cleaner.rs
+++ b/xet_data/src/processing/file_cleaner.rs
@@ -17,7 +17,7 @@ use super::deduplication_interface::UploadSessionDataManager;
 use super::file_upload_session::FileUploadSession;
 use super::sha256::Sha256Generator;
 use crate::deduplication::{Chunk, Chunker, DeduplicationMetrics, FileDeduper};
-use crate::error::Result;
+use crate::error::{DataError, Result};
 use crate::progress_tracking::upload_tracking::CompletionTrackerFileId;
 
 /// Controls how SHA-256 is handled during file cleaning.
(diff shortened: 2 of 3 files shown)


xet_data/src/deduplication/file_deduplication.rs
```diff
@@ -197,24 +197,31 @@ impl<DataInterfaceType: DeduplicationDataInterface> FileDeduper<DataInterfaceTyp
...
                 // we skip dedupe.  However, continuing the previous is always fine.
                 if self.file_data_sequence_continues_current(&fse)
                     || self.defrag_tracker.allow_dedup_on_next_range(n_deduped)
                 {
+                    // Only account for the range once the dedup is actually taken. Counting it
+                    // before this check meant a defrag-prevented range was added here and then
+                    // again below as new data, inflating `total_bytes` by exactly
+                    // `defrag_prevented_dedup_bytes` — and `total_bytes` is what
+                    // `FileCleaner::finish_inner` reports as the file's size.
+                    dedup_metrics.deduped_chunks += n_deduped as u64;
```
rajatarya, line 212: Verified this is complete rather than just moved — the accept path takes `cur_idx += n_deduped; continue;`, so it's the only route where the range is consumed whole and the only place these four counters should fire. The reject path falls through to the per-chunk accounting below, which increments `total_chunks`/`total_bytes` once per chunk as it walks the same range, so the totals still land correctly. No third path takes the range.

One thing I noticed while tracing that's worth knowing but is pre-existing and not yours to fix here: after a rejection the loop re-queries at `cur_idx + 1`, and `dedup_query_against_local_data` can match a *new*, shorter range overlapping the one just rejected. If that one is also rejected, `defrag_prevented_dedup_bytes` accumulates for those bytes again. So the counter is "bytes of rejected dedup opportunities", which can exceed the file's size, rather than "bytes we failed to dedup".

That doesn't affect your fix — the old over-count was the same sum, so the two matched by construction, which is exactly why your production numbers agreed to the byte. But if anyone ever tries to use `defrag_prevented_dedup_bytes` to estimate recoverable storage, it'll overstate. Might be worth a note on the field.

xet_data/src/processing/file_cleaner.rs
```diff
@@ -254,10 +254,27 @@ impl SingleFileCleaner {
...
-            debug_assert_eq!(remaining_file_data.pending_file_info[0].0.file_size(), deduplication_metrics.total_bytes)
+        // `file_size` above is reported from `total_bytes`, an accumulated metrics counter, while
+        // the file's actual content is described by the entries in `pending_file_info`. If the two
+        // disagree, the size handed back to the caller does not describe the data that was just
+        // written. Callers persist that size alongside the file hash, so a mismatch is only
+        // detected much later — and by then the two are permanently out of step. Fail here instead.
+        //
+        // This was a debug-only assertion, which meant release builds shipped a wrong size in
+        // silence; see the defrag-prevented dedup double-count fixed in `FileDeduper`.
+        if remaining_file_data.pending_file_info.len() != 1 {
```
rajatarya, line 265: I think this branch can't be reached. `finalize()` builds its `DataAggregator` through `DataAggregator::new`, which sets `pending_file_info: vec![(pending_file_info, ...)]` — a single element, unconditionally — and `finalize()` is the only thing producing the value on this path. The aggregator's `merge`/`append` paths grow the vec, but nothing between `finalize()` and here calls them.

So as a `debug_assert` it was a free tripwire, but as a returned `InternalError` it's an error string that can never be constructed.

No harm in keeping it, and I can see the argument for a guard against someone later making `finalize` compose multiple files. If that's the intent, could the comment say so? Otherwise a reader hits an error message describing an impossible state and spends a while working out what would cause it. The size check below is the one that does real work.

xet_data/src/processing/file_cleaner.rs
```diff
@@ -254,10 +254,27 @@ impl SingleFileCleaner {
...
+            return Err(DataError::InternalError(format!(
+                "expected exactly one pending file info after clean, found {}",
+                remaining_file_data.pending_file_info.len()
+            )));
+        }
+        let described_size = remaining_file_data.pending_file_info[0].0.file_size();
```
rajatarya, line 271: Agreed this should fail loudly rather than ship a wrong size, and `InternalError` reads right for "our own invariant broke".

One consequence worth a line in the PR notes, since the failure is now user-visible where before it was silent. Returning here skips `register_single_file_clean_completion`, so:

- the file is never registered in the session shard — correct, we don't want the bad entry
- but `completion_tracker` has been taking `increment_file_size` on every `add_data` call, and nothing marks this file complete or removes it

I went looking for whether that wedges the session and it doesn't: `assert_complete()` only runs under `#[cfg(debug_assertions)]` *and* only when `done == total`, so an incomplete file just leaves `is_complete()` returning false rather than tripping an assert, and `finalize_impl` doesn't gate on it. Any xorbs already registered for the file still upload — wasted bytes, but content-addressed, so not harmful.

What the caller sees is an `Err` from `finish()` plus a progress report where that file sits permanently partial. That's a reasonable outcome, I'd just rather it be a documented one than something the next person rediscovers. Could the "Behaviour change" note in the description mention it?

xet_data/tests/test_defrag_prevented_dedup_metrics.rs
```diff
@@ -0,0 +1,98 @@
...
+    for i in 0..repeats {
+        data.extend_from_slice(&random_data(i as u64, 24 * 1024));
+        data.extend_from_slice(&repeated_block);
+    }
+    data
+}
+
+/// Builds a session whose defrag prevention rejects essentially every dedup match: the estimator
+/// is ready after a single range, and the target chunks-per-range is set far above anything the
+/// data can produce, so `allow_dedup_on_next_range` always takes the reject path.
```
rajatarya, line 51: nit: "so `allow_dedup_on_next_range` always takes the reject path" is a touch stronger than what the config guarantees — `file_data_sequence_continues_current(&fse)` short-circuits the `||` before `allow_dedup_on_next_range` is consulted, so a range continuing the previous one is still accepted regardless of these settings.

Doesn't weaken the test: `assert!(metrics.defrag_prevented_dedup_bytes > 0, ...)` establishes the path was exercised, and that guard is the right way to keep the test from silently going vacuous if the estimator's tuning shifts. Just "rejects essentially every dedup match that isn't a continuation" would match what the code does.

While I'm here — using `.expect("nranges config path should exist")` on the `with_config` calls is a good touch. A silently-ignored config key is exactly how this kind of test stops testing anything.

rajatarya (commented):
Nice! This is a good find. I'm overriding what Claude came up with - it was wrong on both claims and too verbose in its comments - deleting them. Approved!

rajatarya (approved):
Looks good - Claude was way wrong in its review.

seanses (approved):

PR #941, opened 2026-08-12
Cap untrusted shard parser memory allocations

seanses:
## Summary
- Cap shard file/xorb section parsing with an optional `size_limit` so untrusted headers cannot force huge allocations.
- Move `MDB_SHARD_TARGET_SIZE` into `test_configurable_constants` (remove unused shard config `target_size` / `max_target_size`) because the max shard size is a fixed production choice, and should not be easily modified by env vars.

---

> [!NOTE]
> **Medium Risk**
> Touches untrusted shard ingestion and changes how max shard size is configured (no longer overridable via `HF_XET_SHARD_*` env vars), but behavior is intentionally bounded and covered by new parser tests.
> 
> **Overview**
> **Hardens metadata shard parsing** against malicious or corrupt headers by threading an optional `size_limit` through sync/async file and xorb section processors. Claimed payload sizes are validated with checked arithmetic (`num_info_entry_following`, `checked_section_bytes`) *before* `Vec::with_capacity`, returning `CoreError::InvalidShard` when limits or overflows are hit. `MDBFileInfo::num_bytes` and in-memory `recalculate_shard_size` propagate those errors.
> 
> **Centralizes shard sizing** by introducing `MDB_SHARD_TARGET_SIZE` (64 MiB) in `test_configurable_constants` and removing `shard.target_size` / `shard.max_target_size` from `XetConfig` and related env aliases. Merge, consolidation, and session upload paths now use the constant (tests can override via `test_set_constants!`).
> 
> Adds unit tests for oversized sections, entry-count overflow with verification, and remaining-budget accounting.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 750a476ef003ded0201960bcac3d49186d056c6d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


xet_core_structures/src/metadata_shard/shard_in_memory.rs
```diff
@@ -285,9 +286,13 @@ impl MDBInMemoryShard {
...
 
-        process_shard_file_info_section(reader, |view| shard.add_file_reconstruction_info(MDBFileInfo::from(&view)))?;
+        process_shard_file_info_section(
+            reader,
+            |view| shard.add_file_reconstruction_info(MDBFileInfo::from(&view)),
+            None,
+        )?;
 
-        process_shard_xorb_info_section(reader, |view| shard.add_xorb_block(MDBXorbInfo::from(&view)))?;
+        process_shard_xorb_info_section(reader, |view| shard.add_xorb_block(MDBXorbInfo::from(&view)), None)?;
```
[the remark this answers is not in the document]
seanses: This is not the targeted user case: global dedupe shards are returned by CAS servers and are always trusted.

xet_core_structures/src/metadata_shard/file_structs.rs
```diff
@@ -7,6 +7,7 @@ use serde::Serialize;
 
 use super::shard_file::MDB_FILE_INFO_ENTRY_SIZE;
 use super::xorb_structs::{XorbChunkSequenceEntry, XorbChunkSequenceHeader};
+use crate::error::{CoreError, Result as CoreResult};
```
assafvayner: Do we use these import names everywhere, I don't like CoreResult, would rather inline crate::error::Result

xet_runtime/src/config/groups/shard.rs
```diff
@@ -2,20 +2,6 @@ use crate::utils::ByteSize;
...
-    ///
-    /// Use the environment variable `HF_XET_SHARD_TARGET_SIZE` to set this value.
-    ref target_size: u64 = 64 * 1024 * 1024;
-
-    /// Maximum shard size; small shards are aggregated until they are at most this.
-    ///
-    /// The default value is 67108864.
-    ///
-    /// Use the environment variable `HF_XET_SHARD_MAX_TARGET_SIZE` to set this value.
-    ref max_target_size: u64 = 64 * 1024 * 1024;
```
assafvayner, line 17: why remove the config values and make these constants, I think leaving it as a configurable constant is fine.

assafvayner: your agent commented this is a production server choice, it is, but also the clients may want a smaller shard size for example, or an alternative server in could enforce different values.

assafvayner (commented):
double check that your agent didn't go too far beyond what you were expecting.

xet_runtime/src/config/groups/shard.rs
seanses: This is actually what I instructed the agent to do. The max shard size, like the max xorb size, is our prod choice. So I don't think we should make it easy to change by env vars in an official client release. Moving them under `test_configurable_constants!` make them not configurable in non-debug builds.

If any server deployment may prefer a different max shard and xorb size, there should be a protocol to communicate that expectation, instead of relying on users to manually set that value.

assafvayner (approved):

assafvayner:
approved; my only concern was agent overstepping.

xet_core_structures/src/metadata_shard/file_structs.rs
seanses: sounds good, changed from `CoreResult<...>` to `Result<...,CoreError>`

rajatarya (commented):
D'oh, was _about_ to approve!

My main feedback was around wasm and sync methods.

wasm in-memory reader doesn't support setting a size - it always passes None. Does it make sense to specify `*MDB_SHARD_TARGET_SIZE` for those?

In `MDBMinimalShard::from_reader` there isn't a size method - (unlike from_reader_async) so cannot have size restriction on the sync cases.

seanses:
> wasm in-memory reader doesn't support setting a size - it always passes None. Does it make sense to specify *MDB_SHARD_TARGET_SIZE for those?
> In MDBMinimalShard::from_reader there isn't a size method - (unlike from_reader_async) so cannot have size restriction on the sync cases.

We don't have a wasm based server and the server doesn't use the sync methods so that's not a problem. Trying to keep the PR scoped only for necessary changes, and if the server decides to switch to another API we will revisit this.

PR #943, opened 2026-08-19
Derive download buffer defaults from usable, cgroup-aware system memory

### LINKED ISSUE #927: Download buffer defaults are constants: 2.7x left on a 20 Gbit/s host, OOM in a 1 Gi container
### Summary

The reconstruction download buffer is sized from constants, not from the machine. On a large host that leaves most of the link unused; in a small container it is the reported cause of OOM kills. Both ends would be served by deriving the default from available memory, and the client already has everything it needs to do so.

### The measurement

192 cores, 1996 GiB RAM, ~20 Gbit/s link, hf-xet 1.5.2, `huggingface_hub` 0.36.2. A 96.8 GB GGUF repo downloaded in full per arm, each in its own `HF_HOME` so no run inherits another's dedup state, order interleaved with the baseline re-measured at both ends (drift 6%).

| configuration | Mbit/s | vs default | peak RSS |
|---|---:|---:|---:|
| defaults | 6709 | 1.00x | 2.0 G |
| + `MIN`/`MAX_RECONSTRUCTION_FETCH_SIZE` at their high-performance values | 6810 | 1.02x | 2.3 G |
| + `MIN_PREFETCH_BUFFER` raised | 6789 | 1.01x | 2.1 G |
| + the adaptive-concurrency band at its high-performance values | 6984 | 1.04x | 2.2 G |
| + the three `DOWNLOAD_BUFFER_*` knobs only | 18114 | **2.70x** | 14.2 G |
| the whole high-performance preset | 19930 | 2.97x | 20.1 G |

The three `DOWNLOAD_BUFFER_*` knobs account for essentially the entire difference. Concurrency alone is inside the drift band, which is consistent with #709 and #711 both being abandoned for insufficient impact.

### Why the defaults leave that on the table

#637 lowered `download_buffer_size` from 8gb to 2gb on the basis of benchmarks run on `i4i.xlarge` and `m5d.xlarge`, both capped at 10 Gbit/s. That is a sound result inside the range it was measured over. The host above is roughly twice that link speed, and there the 2 GB default costs 2.7x. #666 then made the buffer scale, but with the number of concurrent files rather than with the machine, so the ceiling is still a constant 8gb regardless of whether there are 8 GB or 2 TB of RAM.

### The other end

The same constants are too large in a small container. huggingface_hub#3300 (open since 2025-08) reports an OOM kill in a Kubernetes init container with a 1 Gi limit, ~10 GB used against ~200 MB on an older `huggingface_hub`. A default that is simultaneously too small for a 2 TB host and too large for a 1 Gi container is the signature of a constant that wants to be a function of available memory.

### Suggestion

Derive the buffer defaults from usable memory, keeping the current values as the clamps. Something as simple as a fraction of total RAM, floored at today's small-machine values and ceilinged at the existing high-performance numbers, reproduces the current behaviour in the middle of the range and fixes both ends. Two details that mattered when we tried it downstream:

- **Read the cgroup limit, not the host total.** Inside a container `sysinfo` and `psutil` both report the host's memory, which is exactly how a 1 Gi pod ends up sizing a multi-GB buffer. The v2 limit is at the path named in `/proc/self/cgroup` rather than at `/sys/fs/cgroup` itself, and on v1 a nested path (a Slurm step, for instance) reads "unlimited" at the controller root.
- **Keep `download_buffer_size + max_concurrent_file_downloads * download_buffer_perfile_size` inside `download_buffer_limit`**, so the three numbers describe one allocation rather than three that can overshoot together.

`xet_runtime/src/logging/system_monitor.rs` already pulls in `sysinfo`, so the memory figure is available without a new dependency; today it is only used for opt-in logging.

Happy to share the harness or run further arms on this host if that would help, and equally happy to be told the constant is deliberate and the answer is for callers to set the variables themselves. In that case the documentation issue is the one worth fixing, since the variables are currently discoverable only by reading the Rust.

rajatarya:
# Memory-derived download buffer defaults

Fixes #927. Related: huggingface_hub#3300.

## Problem

The three `HF_XET_RECONSTRUCTION_DOWNLOAD_BUFFER_*` defaults are compile-time constants (2 GB / 512 MB / 8 GB). Issue #927 measured the cost on both tails: a 20 Gbit/s, 2 TB host leaves **2.7x throughput** on the table (6709 → 18114 Mbit/s from raising only these three knobs), while a 1 Gi Kubernetes container gets **OOM-killed** (huggingface_hub#3300). A default that is simultaneously too small for a 2 TB host and too large for a 1 Gi container should be a function of available memory, not a constant.

## Design overview

Three cooperating changes, all inside `xet_runtime`:

1. **A memory probe** (`utils/system_memory.rs`, new): usable memory = min(host total, effective cgroup limit), cgroup-aware in the way the issue prescribes — resolving the path from `/proc/self/cgroup` rather than reading the controller root, via sysinfo 0.39's `Process::cgroup_limits()`. This requires **sysinfo 0.38.4 → 0.39.6** and a **Rust toolchain bump 1.94.1 → 1.95.0** (sysinfo 0.39's MSRV), both part of this PR.
2. **Derived defaults**: each buffer knob becomes `clamp(fraction × usable, floor, ceiling)`, anchored so a 32 GB machine reproduces today's defaults and a ≥256 GB machine reproduces today's high-performance preset. Explicit env vars override, exactly as before.
3. **Coherence fixes** in the same machinery: setting `SIZE > LIMIT` no longer panics (normalized with a warning); the HP preset becomes memory-aware and no longer silently stomps explicit env vars.

### The memory probe

```mermaid
flowchart TD
    A(["USABLE_MEMORY probe"]) --> B{"LazyLock cached?"}
    B -->|yes| Z(["return cached UsableMemory"])
    B -->|no| C["host_total via sysinfo<br/>RefreshKind memory only"]
    C --> D["refresh own PID entry then<br/>Process cgroup_limits<br/>from sysinfo 0.39"]
    D --> E{"limits returned?<br/>Linux only"}
    E -->|"no: macOS via Mach sysctl<br/>Windows via GlobalMemoryStatusEx<br/>wasm stub - no info"| N["cgroup_limit = None"]
    E -->|yes| F["sysinfo resolves the path in<br/>/proc/self/cgroup and takes the<br/>min limit across ancestor cgroups<br/>v2 memory.max or v1 limit_in_bytes"]
    F --> G{"limit below host total?"}
    G -->|"no - unlimited"| N
    G -->|yes| H["cgroup_limit = limit"]
    N --> L
    H --> L["usable = min of host_total<br/>and cgroup_limit"]
    L --> M["cache UsableMemory struct<br/>host_total / cgroup_limit / usable"]
    M --> Z
```

Why sysinfo 0.39: the previously pinned sysinfo 0.38.4 only offered `System::cgroup_limits()`, which reads controller-root paths (`/sys/fs/cgroup/memory.max`) — fine under Docker/k8s cgroup namespaces but blind to nested cgroups (systemd slices, Slurm steps), the exact failure modes the issue reporter hit downstream. sysinfo 0.39.0 added `Process::cgroup_limits()`, which resolves the process's own cgroup path from `/proc/<pid>/cgroup` and takes the tightest bound across ancestor cgroups, for both cgroup v1 and v2, treating `max`/PAGE_COUNTER_MAX as unbounded. Depending on the maintained implementation beats maintaining a hand-rolled `/proc` parser here; sysinfo carries its own fixture tests for the walk. sysinfo still supplies the host total; the probe adds no new dependency.

The probe is `LazyLock`-cached (the `config_group!` macro evaluates default expressions on every `XetConfig::new()`, which runs per session, in the legacy runtime, twice in git_xet, and in xtool). `Process::cgroup_limits` returns `None` off-Linux by contract, so the probe needs no OS-specific gating of its own; wasm gets a stub returning no memory info, which falls back to today's static defaults. When no cgroup limit binds, the reported limit equals the host total and is filtered out, keeping `cgroup_limit = None` meaning "no limit set" in logs.

### Platform behavior

The probe has two independent inputs — host total and container limit — and each platform fills in what it can. `usable = min(present values)`; if neither is available, derivation is skipped entirely and today's static defaults apply (fail-safe, never worse than current behavior).

| Platform | Host total | Container/limit awareness | Net effect |
|---|---|---|---|
| Linux (bare, k8s, Docker, WSL2 distro) | sysinfo (`/proc/meminfo`) | Full via `Process::cgroup_limits` (sysinfo 0.39): cgroup v2 + v1, nested paths, namespaced roots | Both tails fixed — the issue's target |
| macOS | sysinfo (Mach kernel API / `sysctl hw.memsize`) — physical RAM | None (no cgroup concept; Docker Desktop runs a Linux VM, inside which the Linux path applies) | Laptops scale by RAM: 16 GB → 1 GB/248 MB/4 GB; 128 GB Mac Studio → 8 GB/2 GB/32 GB |
| Windows | sysinfo (`GlobalMemoryStatusEx`) — physical RAM | **Not probed**: Job Object / Windows-container memory limits are not read (sysinfo has no API for it; rare deployment; documented limitation — recourse is the env vars or kill switch). WSL2 goes through the Linux path. | Same RAM-proportional scaling as macOS |
| wasm | stub → `None` | `None` | Static defaults (2 GB/512 MB/8 GB), exactly today |
| Any platform, sysinfo returns 0/fails | `None` | — | Static defaults |

The derivation formula itself is platform-independent and unit-tested everywhere. cgroup resolution is delegated to sysinfo, which is tested upstream (including nested-ancestor fixtures); this repo carries no OS-specific probe code beyond the wasm stub.

### The derivation

For standard mode: `limit = clamp(u/4, 264 MB, 64 GB)`, `size = clamp(u/16, 64 MB, 16 GB)`, `perfile = clamp(u/64, 16 MB, 2 GB)`, each then rounded down to a multiple of 8 MB (decimal units throughout, matching `ByteSize`). The floors are the 1 GiB derivation values; the ceilings are the historical HP constants. High-performance mode uses 2x-aggressive fractions with the same floors/ceilings: `u/2`, `u/8`, `u/32`.

| usable memory | size (u/16) | perfile (u/64) | limit (u/4) | note |
|---|---|---|---|---|
| ≤ 1 GiB | 64 MB | 16 MB | 264 MB | floors; **the #3300 container: was 2 GB base / 6.1 GB target → OOM** |
| 8 GB | 496 MB | 120 MB | 2 GB | small VM / CI runner |
| 16 GB | 1 GB | 248 MB | 4 GB | typical laptop |
| **32 GB** | **2 GB** | **496 MB** | **8 GB** | **≈ today's defaults (512 MB → 496 MB from rounding)** |
| 64 GB | 4 GB | 1 GB | 16 GB | workstation |
| 128 GB | 8 GB | 2 GB (ceil) | 32 GB | |
| ≥ 256 GB | 16 GB (ceil) | 2 GB (ceil) | 64 GB (ceil) | **= today's HP preset = the issue's 2.70x arm** |

Properties, verified analytically at every region boundary for both fraction sets:
- **Coherence invariant**: `size + max_concurrent_file_downloads(8) × perfile ≤ limit` holds everywhere (mid-range: 3u/16 ≤ u/4; floors: 64+128=192 ≤ 264 MB; ceilings: 16+16=32 ≤ 64 GB; HP: 3u/8 ≤ u/2). The three numbers describe one allocation, as the issue requested.
- **Monotonic** in usable memory; anchored to reproduce current behavior at 32 GB and current HP at ≥256 GB (the issue's measured 2.70x configuration).
- The 32 GB anchor is exact for size and limit; perfile lands at 496 MB vs today's 512 MB (u/64 kept clean rather than u/62.5).

Floors note: per review, the floors are the 1 GiB derivation values — environments below ~1 GiB get the same triple rather than a smaller dedicated tier. At the floors a single maximum-size term (one unpacked xorb block, ≤ 64 MiB) fits the budget once one file download is active (64 + 16 = 80 MB); as a backstop, `acquire_many` clamps an oversized single term to total permits, so it proceeds serially rather than deadlocking.

### Startup flow — how system info reaches the consumers

```mermaid
sequenceDiagram
    autonumber
    participant PY as Python hf_xet
    participant XS as PyXetSession
    participant CFG as XetConfig
    participant MEM as system_memory probe
    participant CTX as XetContext
    participant COM as XetCommon
    participant SEM as AdjustableSemaphore
    participant FR as FileReconstructor

    PY->>XS: XetSession constructor
    XS->>CFG: XetConfig::new
    CFG->>CFG: Default::default builds groups
    CFG->>MEM: USABLE_MEMORY (first read)
    Note over MEM: cgroup walk then sysinfo host total<br/>cached in LazyLock for process lifetime
    MEM-->>CFG: usable bytes
    Note over CFG: derived defaults<br/>size = clamp of usable/16<br/>perfile = clamp of usable/64<br/>limit = clamp of usable/4
    CFG->>CFG: with_high_performance if HP flag<br/>memory-aware HP fractions
    CFG->>CFG: with_env_overrides<br/>explicit env vars always win
    CFG-->>XS: XetConfig
    XS->>CTX: XetContext::with_config
    CTX->>CTX: normalize reconstruction group<br/>ensure limit >= size then freeze as Arc
    CTX->>COM: XetCommon::new
    COM->>SEM: new with initial size<br/>floor size and ceiling limit
    Note over SEM: one permit equals one byte<br/>caps in-flight unwritten bytes
    XS-->>PY: session ready
    PY->>FR: download file (later)
    FR->>SEM: grow target to min of<br/>size plus n_active times perfile<br/>and limit
    SEM-->>FR: seed permit for growth
    FR->>SEM: acquire_many term_size per term
    Note over FR,SEM: permits released only after<br/>bytes are written to disk
```

The same `XetConfig::new()` → `XetContext::new()` funnel serves every entry point (Python `XetSession`, the legacy `download_files` runtime still used by huggingface_hub, git_xet's LFS agent, xtool), so all of them inherit derived defaults with no per-caller changes. `XetContext::new` is the single choke point where the normalized config is frozen into an `Arc` — guaranteeing the semaphore bounds (built once in `XetCommon::new`) and the per-download #666 target formula (recomputed in `FileReconstructor`) always read identical values.

### Ordering and preset fixes

`XetConfig::new()` becomes **defaults → HP preset → env overrides** (today env runs before HP, so `HF_XET_HP` silently discards explicit `HF_XET_RECONSTRUCTION_*` settings — the issue's benchmark matrix would literally have been impossible to run as intended with both set). Enabling this requires `apply_env_overrides` to fall back to the *current* field value rather than re-evaluating the macro default — a one-line macro change (`let default_value = self.$name.clone()`), audited safe: all existing callers outside `new()` are tests invoking it immediately after construction, and it halves default-expression evaluations as a bonus.

`with_high_performance()` keeps its static fetch-size and concurrency literals (metadata-scale, not RAM-scale) but takes its three buffer values from the HP derivation — so `HF_XET_HP` on an 8 GB laptop now yields 1 GB / 248 MB / 4 GB instead of a physically impossible 16 GB base buffer.

**Panic fix**: `AdjustableSemaphore::new` release-asserts `min ≤ max`, so `HF_XET_RECONSTRUCTION_DOWNLOAD_BUFFER_SIZE=16gb` alone crashes today (default limit 8 GB). Normalization (`if limit < size { warn!; limit = size }`) runs once in `XetContext::new`; `XetCommon::new` additionally applies `limit.max(base)` locally since it is `pub` and must never panic. Derived values can never trigger it by construction.

### Small containers and large machines

- **1 Gi container** (the #3300 report): derived 64 MB / 16 MB / 264 MB. The semaphore caps in-flight-not-yet-written bytes, and the permit is released only after bytes hit disk, so buffer RSS stays ≤264 MB vs ~6.1 GB target today. Remaining known slack outside the semaphore: up to one decompressed xorb block (≤64 MiB) per active connection may be resident while only partially permitted — pre-existing behavior, unchanged, bounded by download concurrency; documented in the api_changes note.
- **2 TB / 20 Gbit/s host** (the #927 measurement): derived values equal the HP preset = the issue's isolated 2.70x arm (18114 Mbit/s, 14.2 G peak RSS on a 2 TB machine).
- **Prefetch/fetch knobs** (`min_prefetch_buffer`, `min/max_reconstruction_fetch_size`) are deliberately **not** derived: traced to be metadata-scale — they size reconstruction-term prefetch (term descriptions, URLs), not data buffers; actual data bytes acquire a buffer permit before the download task spawns. The issue's own benchmark confirms: those knobs moved throughput ≤2%, within drift.

### Escape hatches and observability

- Every `HF_XET_RECONSTRUCTION_DOWNLOAD_BUFFER_*` env var still overrides its derived default (macro semantics, unchanged).
- New kill switch `HF_XET_DISABLE_MEMORY_DERIVED_DOWNLOAD_BUFFERS=1` (bool, unset by default) restores the static 2 GB / 512 MB / 8 GB (and static HP) values — read per-call (cheap) so it is testable with `EnvVarGuard`; only the probe and derived numbers are cached.
- One `info!` line emitted by `logging::init` after the subscriber is installed: host total, cgroup limit, usable, and the three derived values. (It cannot be logged where the values are computed — the LazyLock is first forced while building the logging config itself, before any subscriber exists, so an event there would be silently dropped and never fire again.) Values are also visible in Python via `XetConfig.__repr__` and per-field `Config: name = value (default)` log lines.

### Risks

- Defaults become machine-dependent: nothing in xet-core asserts the literals (verified repo-wide), but downstream (huggingface_hub CI) is flagged in the PR/api_changes note.
- Larger default in-flight budget on big hosts (up to the 64 GB ceiling ≥256 GB RAM): ceilings equal the existing opt-in HP values; kill switch available.
- `with_env_overrides` no longer resets env-absent fields to macro defaults — documented; audited zero behavioral change for existing callers.
- Users who set both `HF_XET_HP` and explicit buffer env vars previously got HP values; now env wins (the intended fix, but a behavior change).

---

## Testing

- TDD throughout: derivation-table tests (11 memory sizes x standard/HP fraction sets), a multiplicative 256MB-2.5TB sweep asserting the coherence invariant, 8MB rounding, clamps, and monotonicity; integration tests for env-beats-derived, env-beats-HP, kill switch, and the panic fix. cgroup-walk correctness is covered by sysinfo's own upstream fixture tests (nested ancestors, v1/v2, unlimited sentinels).
- On Rust 1.95.0 with sysinfo 0.39.6: `cargo test -p xet-runtime` (232 passed) and `cargo test -p xet-data` (340 passed), zero failures; `cargo clippy` clean on both libs; `hf_xet` and both wasm sub-workspaces (`hf_xet_thin_wasm`, `hf_xet_wasm` on wasm32) compile, exercising the wasm stub path.
- Toolchain: all CI workflow pins moved 1.94.1 → 1.95.0 (sysinfo 0.39 MSRV); the wasm sub-workspace stays on its pinned nightly.
- Per review: sysinfo is a `cfg(not(target_family = "wasm"))` target dependency of `xet_runtime`, so wasm builds no longer compile it at all (`cargo tree --target wasm32-unknown-unknown` shows zero sysinfo entries; all three consumers — `system_monitor`, `logging::init`, the memory probe — were already wasm-gated). Side benefit: wasm consumers are not subject to sysinfo's 1.95 MSRV.

## Note for downstream

`huggingface_hub` CI or downstream tests asserting the literal 2GB/512MB/8GB defaults will see machine-derived values; `HF_XET_DISABLE_MEMORY_DERIVED_DOWNLOAD_BUFFERS=1` restores the static defaults. User-facing documentation for the changed env vars is updated in huggingface/hub-docs#2731 (draft until this ships in an `hf_xet` release).

This implementation was AI-assisted (design and code reviewed and directed by @rajatarya).

https://claude.ai/code/session_01MGuU1K7rpjFfBGrA3SvCxL

---

> [!NOTE]
> **Medium Risk**
> Machine-dependent defaults may surprise downstream tests; config ordering and HP/env interaction changed. Memory sizing affects OOM risk in small containers and in-flight I/O on large hosts, though env vars and a kill switch remain.
> 
> **Overview**
> Reconstruction download buffer defaults (`HF_XET_RECONSTRUCTION_DOWNLOAD_BUFFER_*`) are no longer fixed 2 GB / 512 MB / 8 GB. They are computed from **usable memory** (min of host RAM and effective cgroup limit via **sysinfo 0.39** `Process::cgroup_limits`), with floors/ceilings so ~32 GB hosts match old defaults and large hosts match the old HP preset. **`HF_XET_DISABLE_MEMORY_DERIVED_DOWNLOAD_BUFFERS=1`** restores the static values.
> 
> **Config behavior changes:** `XetConfig::new()` applies **defaults → high performance → env overrides** so explicit env vars beat `HF_XET_HP`; `with_env_overrides` keeps preset-adjusted fields when an env var is absent. **`normalize()`** raises `download_buffer_limit` to at least `download_buffer_size` (in `XetConfig::new` and `XetContext::new`); **`XetCommon`** avoids panics on incoherent configs.
> 
> **Infra:** Rust **1.94.1 → 1.95.0** in CI/release workflows; **`sysinfo`** is a non-wasm-only dependency of `xet_runtime`. Lockfiles pick up transitive updates (e.g. `chacha20`, objc2 stack for sysinfo).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 242f99a7aec76ca2fa8d319f4735b3c7046b3756. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rajatarya:
Hold off on review @assafvayner and @seanses : I am updating the PR to use latest sysinfo so we don't have to hand-roll our /proc file reading. This will also bump our Rust toolchain to 1.95, but I think is worth it.

assafvayner:
let's make the rust version used in CI consistent (all the same)

api_changes/update_260819_memory_derived_download_buffers.md
assafvayner, line 1: I don't think this change merits checking this in, IIUC we keep api_changes for really significant changes.

wasm/hf_xet_thin_wasm/Cargo.lock
assafvayner, line 1: we may want to feature gate the sysinfo usage so it's not attempted to be picked up into the wasm impls.

xet_runtime/tests/memory_derived_defaults_test.rs
```diff
@@ -0,0 +1,137 @@
+use serial_test::serial;
+use xet_runtime::config::XetConfig;
+use xet_runtime::core::{XetCommon, XetContext};
+use xet_runtime::utils::system_memory::{default_download_buffer_sizes, high_performance_download_buffer_sizes};
+use xet_runtime::utils::{ByteSize, EnvVarGuard};
+
+fn clear_buffer_env_vars() {
```
assafvayner: can we leverage EnvVarGuard's instead of this wipe per test.

xet_runtime/src/utils/system_memory.rs
```diff
@@ -0,0 +1,347 @@
...
+/// `size + 8 * perfile <= limit` just like the fractions do.
+const SIZE_FLOOR: u64 = 32_000_000;
+const SIZE_CEILING: u64 = 16_000_000_000;
+const PERFILE_FLOOR: u64 = 8_000_000;
+const PERFILE_CEILING: u64 = 2_000_000_000;
+const LIMIT_FLOOR: u64 = 128_000_000;
+const LIMIT_CEILING: u64 = 64_000_000_000;
+
+/// Derive the three buffer values from a usable-memory figure.
+fn derive(usable: u64, fractions: &Fractions) -> DownloadBufferDefaults {
```
assafvayner: might want a better fn name?

api_changes/update_260819_memory_derived_download_buffers.md
rajatarya: Sounds good, removing.

wasm/hf_xet_thin_wasm/Cargo.lock
rajatarya: Sounds good, adjusting.

rajatarya:
> let's make the rust version used in CI consistent (all the same)

Sounds good, will make related PRs in the other repos that directly depend on xet-core.

api_changes/update_260819_memory_derived_download_buffers.md
rajatarya: Removed in aa1292ad. The behavior change is still documented user-facing — the env-var tables in hub-docs are updated in huggingface/hub-docs#2731 (draft until this ships in an hf_xet release), and the PR description here carries the downstream notes.

wasm/hf_xet_thin_wasm/Cargo.lock
rajatarya: This was already in the PR, should be feature-gated on build.

rajatarya:
@assafvayner ready for re-review. Also, cargo audit is fixed in #950.

xet_runtime/src/core/context.rs
```diff
@@ -24,7 +24,12 @@ pub struct XetContext {
 
 impl XetContext {
     /// Creates a context from a pre-built thread pool and configuration.
-    pub fn new(config: XetConfig, runtime: Arc<XetRuntime>) -> Self {
+    ///
+    /// The reconstruction buffer values are normalized here, before the config is
+    /// frozen, so the buffer semaphore bounds and the per-download scaling target
+    /// always read identical, coherent values.
+    pub fn new(mut config: XetConfig, runtime: Arc<XetRuntime>) -> Self {
```
seanses, line 34: I don't like `XetContext::new` taking a `mut XetConfig` and change configurations internally (even if it "normalizes" numbers). Let's conduct all sanity checks inside `XetConfig::new` and fix problems, so what the internal uses is consistent with what the caller knows.

And that means we don't need to call `config.reconstruction.normalize()` below and we don't need to double check in `XetCommon::new`

xet_runtime/tests/memory_derived_defaults_test.rs
seanses: There's a test unit `EnvVarGuard` specifically for this purpose

xet_runtime/src/utils/system_memory.rs
```diff
@@ -0,0 +1,347 @@
...
+#[cfg(target_family = "wasm")]
+fn probe_usable_memory() -> UsableMemory {
+    UsableMemory {
+        host_total: None,
+        cgroup_limit: None,
+    }
+}
+
+/// Probe the system once and cache the result for the process lifetime.
+pub fn usable_system_memory() -> &'static UsableMemory {
```
seanses: I don't see benefit of wrapping the LazyLock statics inside a function, can we use the static variables directly? Same for `derived_defaults` below, which returns a tuple and is consumed as `derived_defaults().0` and `derived_defaults().1`, seems unconventional

xet_runtime/src/utils/system_memory.rs
```diff
@@ -0,0 +1,347 @@
...
+    perfile_divisor: 32,
+    limit_divisor: 2,
+};
+
+/// Floors keep tiny environments functional; ceilings are the historical
+/// high-performance values. All are multiples of `ROUND_TO`, so rounding a clamped
+/// value down never violates the clamp. The floors satisfy
+/// `size + 8 * perfile <= limit` just like the fractions do.
+const SIZE_FLOOR: u64 = 32_000_000;
+const SIZE_CEILING: u64 = 16_000_000_000;
```
seanses: Let's not repeat the "CEILING" constants, either use the `STATIC_HP_DEFAULTS` numbers or let `STATIC_HP_DEFAULTS` use this.

xet_runtime/src/utils/system_memory.rs
```diff
@@ -0,0 +1,347 @@
...
+    size_divisor: 8,
+    perfile_divisor: 32,
+    limit_divisor: 2,
+};
+
+/// Floors keep tiny environments functional; ceilings are the historical
+/// high-performance values. All are multiples of `ROUND_TO`, so rounding a clamped
+/// value down never violates the clamp. The floors satisfy
+/// `size + 8 * perfile <= limit` just like the fractions do.
+const SIZE_FLOOR: u64 = 32_000_000;
```
seanses: I think these `FLOOR` values are misleading. Note that a file term can be as large as 64 MiB, so setting the size to 32 MB either leads to potential deadlock because no semaphore can be acquired, or (if the memory semaphore is not that strictly enforced, which I don't believe so) implying we don't respect this configuration. I don't think we need to accomadate for the "≤ 512 MB" case, "1 GiB" usable memory is small enough.

xet_runtime/src/utils/system_memory.rs
```diff
@@ -0,0 +1,347 @@
...
+                (STATIC_DEFAULTS, STATIC_HP_DEFAULTS)
+            },
+        }
+    });
+    &DERIVED
+}
+
+/// Whether memory-derived defaults are enabled. Read on every call (cheap) so the
+/// switch is testable in-process; only the probe and derived values are cached.
+fn memory_derived_defaults_enabled() -> bool {
```
seanses, line 220: I feel this is not needed. If users want to configure memory manually they just set the base, perfile and limit env vars, which override everything anyway.

xet_runtime/src/core/context.rs
rajatarya: Added normalize() at the end of XetConfig::new() in c8bba29f, so the config a caller gets back is coherent from the start. I kept the call in XetContext::new and the guard in XetCommon::new, though: configs can be mutated after construction — the Python binding's `with_config` (and the Rust builder method of the same name) set fields on an already-built config, and that flow reaches AdjustableSemaphore::new, where min > max panics via `u64::clamp` even in release builds. XetContext::new is the one choke point every entry path funnels through before the config freezes, and XetCommon::new is pub, so it keeps its local guard. In the normal env-only flow both are silent no-ops.

xet_runtime/tests/memory_derived_defaults_test.rs
rajatarya: Switched in c8bba29f — clear_buffer_env_vars now returns EnvVarGuard::unset guards, which also restores the developer's environment after each test instead of permanently removing the vars.

xet_runtime/src/utils/system_memory.rs
rajatarya: Restructured in c8bba29f: the probe and derived values are now module-level statics (USABLE_MEMORY, DERIVED_DEFAULTS) used directly, and the tuple is a struct with named standard/high_performance fields.

The two public accessors stay functions because they consult the HF_XET_MEMORY_DERIVED_DOWNLOAD_BUFFERS kill switch on every call — a static would freeze the env read at first access, making the switch untestable in-process. The LazyLock underneath is what keeps the sysinfo probe (syscalls + /proc walk) from re-running on every XetConfig construction (the config_group! default expressions evaluate three of these per construction) and guarantees all three knobs derive from a single memory reading, so the size + 8·perfile ≤ limit invariant is computed against one consistent figure. If the kill switch gets dropped (see the other thread), these can collapse into plain statics too.

xet_runtime/src/utils/system_memory.rs
rajatarya: Done in c8bba29f — the ceilings are now defined as STATIC_HP_DEFAULTS.\*.as_u64(), so the historical HP constants are the single source.

xet_runtime/src/utils/system_memory.rs
rajatarya: Raised the floors to the 1 GiB derivation values (64 MB / 16 MB / 264 MB) in c8bba29f, dropping the ≤512 MB tier as suggested. The smallest configuration now genuinely holds a maximum-size term once one file download is active (size 64 + perfile 16 = 80 MB ≥ 64 MiB unpacked).

One correction on the deadlock half: acquire_many clamps a single request to total permits (to_physical_acquire in adjustable_semaphore.rs), so an oversized term acquires everything currently available and proceeds serially rather than deadlocking. But at the old floors that backstop was doing the real work — the configured budget wasn't actually being held — hence the change.

xet_runtime/src/utils/system_memory.rs
rajatarya: The three env vars override values, but the switch turns off the new probing/derivation heuristic wholesale, which I think is worth having for the first releases of this behavior: if the cgroup/memory probe misdetects on some platform we didn't anticipate, the support answer is one documented var set to 0 rather than three vars whose old values users would have to look up — and it also restores the static HP preset under HF_XET_HP in the same stroke. It's likewise a one-line restore for downstream CI that asserts the old literals (huggingface_hub). I'd keep it at least until derivation has soaked in a release or two, then we can drop it. WDYT?

xet_runtime/src/config/groups/reconstruction.rs
```diff
@@ -25,26 +25,33 @@ crate::config_group!({
     /// file downloads times the per file size up to the global limit of
     /// download_buffer_limit.
     ///
-    /// The default value is 2GB.
+    /// The default is derived from usable memory (the minimum of host RAM and the
+    /// effective cgroup limit): usable/16, clamped to [32MB, 16GB]. If memory cannot
```
assafvayner: Claude generated

Stale bounds: the code floors are 64MB / 16MB / 264MB, not 32MB / 8MB / 128MB. Same for the two docstrings below.

xet_runtime/src/utils/system_memory.rs
```diff
@@ -0,0 +1,363 @@
...
+}
+
+/// Computed once and logged on first use.
+static DERIVED_DEFAULTS: LazyLock<DerivedDefaults> = LazyLock::new(|| {
+    let memory = &*USABLE_MEMORY;
+    match memory.usable() {
+        Some(usable) => {
+            let standard = derive(usable, &STANDARD_FRACTIONS);
+            let high_performance = derive(usable, &HIGH_PERFORMANCE_FRACTIONS);
+            info!(
```
assafvayner, line 201: Claude generated

This line is lost in hf_xet: `xet_pkg::init_logging` calls `XetConfig::new()` (which forces this LazyLock) before `logging::init`, so there is no subscriber yet and it never fires again. It needs to be logged after `logging::init`.

assafvayner (approved):
Approving conditional on the docs+log fix

xet_runtime/src/config/groups/reconstruction.rs
rajatarya: Fixed in 7bd14271 — the three docstring clamp bounds now read [64MB, 16GB] / [16MB, 2GB] / [264MB, 64GB]. These went stale in the floors change because my post-change sweep grepped for the numeric literals (32_000_000 etc.), which misses prose forms like "32MB" in doc comments.

xet_runtime/src/utils/system_memory.rs
rajatarya: Confirmed and fixed in 7bd14271. The LazyLock initializer no longer logs; a new log_derived_defaults() is called by logging::init right after the subscriber is installed (next to the existing version-info line), so the ordering is guaranteed by construction for every consumer of logging::init. Verified against the real init path: calling xet_pkg::init_logging with HF_XET_LOG_DEST pointing at a file now writes the version line followed by the derived-defaults line (host_total, cgroup_limit, usable, and the three values). When the kill switch disables derivation, the function logs that static defaults are in use instead.

rajatarya:
Manual container verification of the probe (7bd14271, Docker Desktop Linux VM with 8.3 GB, throwaway binary reproducing the hf_xet init ordering — `XetConfig::new()` while building the logging config, then `logging::init`):

| scenario | cgroup_limit detected | derived size/perfile/limit |
|---|---|---|
| `--memory=1g` | `Some(1073741824)` | 64MB / 16MB / 264MB |
| `--memory=512m` | `Some(536870912)` | 64MB / 16MB / 264MB (floors) |
| no limit | `None` | 512MB / 128MB / 2.072GB (from 8.3 GB host) |
| `--memory=2g`, `--cgroupns=host` (deep path `/docker/<id>`, no root memory.max) | `Some(2147483648)` | 128MB / 32MB / 536MB |
| nested 768M child cgroup inside a 2g container (systemd-scope-style nesting) | `Some(805306368)` — min across ancestors | 64MB / 16MB / 264MB |

In every case `host_total` reported the VM's 8.3 GB — the figure a naive `total_memory`/psutil probe would have used — while `usable` picked up the container limit. The `info!` line lands in the log file after the version line in all runs.

One edge found while testing: a process placed in a cgroup whose **memory controller is not enabled at the leaf** (no `memory.max` file on its own node — requires manually created cgroups without controller delegation; Docker/k8s/systemd all enable it) gets `cgroup_limit = None` from sysinfo even when an ancestor has a limit. The fallback is host-total-derived values, i.e. parity with today's behavior, so no action taken — noting it for the record.

This verification was AI-assisted (run and reviewed by @rajatarya).

xet_runtime/src/utils/system_memory.rs
rajatarya: Per our discussion: kept as a disable flag, renamed to HF_XET_DISABLE_MEMORY_DERIVED_DOWNLOAD_BUFFERS (=1 restores the static defaults; unset means derivation on) in e4667f0c.

seanses (approved):
LGTM!

seanses (approved):
LGTM!

PR #951, opened 2026-08-27
Python bindings for XetSession.new_range_upload()

lhoestq:
Required for https://github.com/huggingface/huggingface_hub/pull/4752
Follows https://github.com/huggingface/xet-core/pull/717

This PR is mostly LLM generated (I'm sorry I'm still too slow at writing rust :x). I tried to make sure it follows the similar coding pattern as for file uploads, and I made sure it has good test coverage. I double checked it myself, wrote some tests myself, and used it a lot to build the python PR in `huggingface_hub` and it works great. But some aspects require a bit more reviews like progress and abort / cancel handling.

The CI needs a HF_TOKEN secret for some of the e2e tests to work with the Hub (using bucket https://huggingface.co/buckets/hf-internal-testing/test-xet-core), let me know who i should ping to take care of this 

Or if you are ok with only keeping the e2e tests with local CAS and removing the e2e ones with the Hub

---

> [!NOTE]
> **Medium Risk**
> New CAS upload/composition path with auth and cancel/progress behavior; mistakes could corrupt composed files or mishandle tokens, though patterns mirror existing upload commits and tests cover main edit shapes.
> 
> **Overview**
> Adds **dirty/range upload** support end-to-end so clients can patch an existing CAS file by uploading only changed byte ranges instead of re-uploading the whole object.
> 
> **Rust (`xet_pkg`)** introduces `XetSession::new_range_upload()` with `XetRangeUploadCommit` / `XetRangeUploadEdit` (`edit`, `insert`, `delete`, `append`, `commit`/`abort`, progress hooks). Commit sorts pending edits and calls `upload_ranges`.
> 
> **`xet_data::upload_ranges`** now accepts an optional shared `FileUploadSession` so range uploads can report progress through the same session the commit holds.
> 
> **Python (`hf_xet`)** wires `session.new_range_upload(original_hash, original_size, …)` to `XetRangeUploadCommit` / `XetRangeUploadEdit`, including context-manager auto-commit/abort, optional progress callbacks, and GIL-friendly blocking commit. New pytest coverage exercises edit/insert/delete/append and multi-edit (including unsorted edit order) against a local CAS.
> 
> Also ignores `**/__pycache__/` and extends lockfile/dev-deps for Hub-backed Rust e2e tests (`HF_TOKEN`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 653f33e7af7b82d3abeaf716800bd39320b3704c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


xet_pkg/src/xet_session/range_upload_edit.rs
```diff
@@ -0,0 +1,126 @@
...
+
+impl std::error::Error for RangeUploadEditError {}
+
+// ── XetRangeUploadEditInner ─────────────────────────────────────────────────
+
+pub(super) struct XetRangeUploadEditInner {
+    pub(super) original_range: Range<u64>,
+    pub(super) new_length: u64,
+    /// The accumulated data for this edit. None means it has been consumed.
+    pub(super) data: Mutex<Option<Vec<u8>>>,
```
assafvayner, line 29: I don't love this as the data type. Should at least be an enum Consumed/Accumulating(buf)

xet_pkg/src/xet_session/range_upload_edit.rs
```diff
@@ -0,0 +1,126 @@
...
+}
+
+impl XetRangeUploadEditInner {
+    fn write(&self, data: &[u8]) {
+        let mut guard = self.data.lock().unwrap();
```
assafvayner, line 34: no unwraps, goes for whole file/PR. unwraps are only fine in tests and in specific cases with good reason. here we should be returning a Result. we should also consider implementing this as Write trait with flush, if that's possible.

assafvayner (changes requested):
Hey @lhoestq this PR is a bit too prototype-y and doesn't follow best practices/style in some places.

would it be ok if you tell us the interface that you want hf_xet to export and then we can implement it for you?

I assume it's something like this, but preferably give me what you want or the huggingface_hub usage now. also make sure this interface can be made future-proof i.e. that we can add arguments/modes if you need later on.

```python
# get session...
commit = session.new_edit_commit(repo_info etc.)
file_edit = new_file_edit(file_information_params)
file_edit.append(...)
file_edit.replace_section(byte_range, new_data)
file_edit.cut(byte_range)
file_edit.insert(offset, new_data)
commit.edit_file(file_edit)
commit.finalize()
```
note that there is a big interface surface area and need to take some care to get it right.

lhoestq:
Arf sorry about that ! So the implem should simply be 

```python
# get session...
commit = session.new_edit_commit(repo_info etc.)
file_edit = new_file_edit(file_information_params)
file_edit.replace_section(byte_range, new_data)
commit.edit_file(file_edit)
commit.finalize()
```

since all the operations can be implemented using `replace_section` and that's what we'll use on the python client side. We can drop features like progress bars / callbacks for now

(if you are low on bandwidth and have some guidelines I can pass to my agent I'd be happy to babysit a llm generated second draft of the implem btw - but at the same time I don't want to waste your time reviewing llm generated code)

xet_pkg/src/xet_session/range_upload_edit.rs
lhoestq: good point, sry about that

PR #953, opened 2026-08-30
perf(xet_client): stream multipart byte-range responses

### LINKED ISSUE #929: Multirange responses are buffered, stream them?
Querying a multirange request buffers the whole response before splitting it on boundary

Currently improving huggingface.js, and it showed streaming had better perf than buffering the whole response: https://github.com/huggingface/huggingface.js/pull/2345#issuecomment-5204027473

(to investigate deeper, didn't bench on `xet-core` itself - possibly `xet-core` is already optimally handling this)

This is in contrast to the single-range requests that stream data currently in `xet-core`

sliortega295-ops:
## Summary

- parse `multipart/byteranges` responses incrementally from `Response::bytes_stream()`
- deserialize each completed Xorb part directly into the final output buffer
- avoid an extra full-buffer reorder for the normal in-order response while preserving range ordering for out-of-order servers
- retain the public whole-body parser's zero-copy `Bytes::slice` behavior and existing EOF compatibility

This removes the explicit whole compressed-response aggregation that previously happened before any part could be processed. The final decompressed output still remains resident by design; the streaming parser's logical payload buffer holds only the incomplete current part, in addition to the incoming transport chunk and allocator capacity.

Closes #929.

## Compatibility

The new parser is crate-private. Existing public API and multipart ordering semantics are unchanged, including out-of-order responses and the prior exact-boundary EOF behavior.

## Validation

- `cargo test -p xet-client multipart --lib`: 11 passed
- staged local HTTP test: the first part is decoded and reports progress before the server sends the second part or EOF; the old whole-body path fails this negative control
- `cargo test -p xet-client --all-targets`: 212 passed, 4 ignored
- `cargo test -p xet-client --features simulation --all-targets`: 239 passed, 4 ignored, plus 2 integration tests passed
- `cargo clippy -r -p xet-client --features simulation -- -D warnings`: passed
- `cargo +nightly fmt --manifest-path ./Cargo.toml --all -- --check`: passed
- `cargo check -p xet-client --target wasm32-unknown-unknown`: passed with eight pre-existing warnings in untouched code
- `git diff --check`: passed

All validation used CPU execution or a local deterministic HTTP simulation; no GPU or production CAS service was required.

## AI disclosure

The implementation, tests, and this description were generated with OpenAI Codex at the account owner's request. Codex re-checked the final diff against repository code and ran the validations listed above. No separate human line-by-line code review was performed before submission.

---

> [!NOTE]
> **Medium Risk**
> Changes core multi-range download parsing and Xorb reassembly; behavior is heavily tested but mistakes could corrupt reconstructed file data or progress reporting.
> 
> **Overview**
> Multi-range downloads no longer buffer the full compressed **multipart/byteranges** body before work begins. **`RemoteClient::get_file_term_data`** now reads **`bytes_stream()`**, feeds chunks into a new crate-private **`StreamingMultipartParser`**, and deserializes each completed range with **`deserialize_chunks_to_writer`** as it arrives—so transfer progress can fire before EOF.
> 
> The parser keeps only the in-flight part (plus boundary search state), matching the old whole-body semantics for EOF-after-boundary and out-of-order **`Content-Range`** parts; in-order responses avoid an extra reorder copy. The public **`parse_multipart_byteranges`** API is unchanged aside from shared **`parse_part`** helpers and expanded unit tests, including a staged local HTTP test that progress happens before the second chunk is sent.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 086f117c77612c3b6418a0b4db2a071a4d07228f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/xet_client/src/cas_client/multipart.rs
+++ b/xet_client/src/cas_client/multipart.rs
@@ -1,4 +1,4 @@
-use bytes::Bytes;
+use bytes::{Buf, Bytes, BytesMut};
 
 use crate::cas_types::HttpRange;
 use crate::error::{ClientError, Result};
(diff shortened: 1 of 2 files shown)


assafvayner:
Claude generated

Thanks for putting this together — the implementation does what it says, but we're closing it along with #929.

The buffering this fixes only happens on the `client.enable_multirange_fetching = true` path, which defaults to `false` (`xet_runtime/src/config/groups/client.rs:285`) and is currently set only in tests. In the default configuration V2 reconstruction's multi-range fetch entries are split in `file_term.rs` into one `XorbBlock` per individual range, each carrying a single `HttpRange`, so `get_file_term_data` sends `bytes=S-E`, the server answers `206` with `application/octet-stream`, and the existing `bytes_stream()` path already streams and reports progress incrementally. `parse_multipart_byteranges` is unreachable there.

So this adds ~500 lines of incremental-parser complexity on a code path that's off by default, in exchange for no change in shipped behavior. Details in #929.

PR #958, opened 2026-09-04
feat(simulation): let the simulation store S3-style object tag sets on xorbs

sirahd:
Adds `get_xorb_tag_set` / `set_xorb_tag_set` to `DeletionControlableClient`, carrying an `ObjectTagSet` of ordered `(key, value)` pairs. Writes replace the whole set, as S3 `PutObjectTagging` does.

Rename all references of  the existing 32-byte `ObjectTag` to ETag to avoid confusion with the new `ObjectTagSet`

- `MemoryClient` holds the sets beside `xorbs`; `LocalClient` writes a `<canonical>.tagset` JSON sidecar, matching the existing `.gctag` convention of keeping GC state next to the object rather than in it
- Exposed as `GET`/`PUT /simulation/xorbs/{hash}/tag_set`, so `SimulationControlClient` picks it up like the other deletion controls
- Tagging an absent xorb errors rather than leaving an orphan sidecar

---

> [!NOTE]
> **Medium Risk**
> Changes conditional-delete and etag semantics used by GC simulation (stable etags vs re-upload detection), with concurrent delete/upload race handling; scope is simulation clients and HTTP control plane, not production CAS.
> 
> **Overview**
> Adds **S3-style object tagging** to the CAS simulation: `get_xorb_tag_set` / `set_xorb_tag_set` on `DeletionControlableClient`, with ordered `(key, value)` pairs that **replace wholesale** on write (like `PutObjectTagging`). Uploads stamp **`last-upload`** (unix seconds) so callers can distinguish re-uploads from content-stable **ETags**.
> 
> Renames the old 32-byte conditional-delete value from **`ObjectTag` → `ObjectETag`** and renames listing/conditional-delete APIs (`list_xorbs_and_etags`, `delete_*_if_etag_matches`, etc.). **ETag derivation** moves from filesystem mtime/ctime to **key + stored length** (and payload hash in memory), so identical content keeps the same etag after delete/re-upload; **`last-upload`** carries write freshness instead.
> 
> **`LocalClient`** persists tag sets in a **`.tagset` JSON sidecar** (atomic write/rename); deletes **claim** the sidecar before removing data to avoid racing uploads. **`MemoryClient`** merges xorbs, lifecycle tags, and tag sets under **one lock** and drops generation-based etag bumps.
> 
> Simulation HTTP routes are renamed (`/xorbs_with_etags`, `/etag_delete`) and expose **`GET`/`PUT /xorbs/{hash}/tag_set`** via `SimulationControlClient`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 61c2511eac975abd3af5175d8c84cfb3db0fb636. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/xet_client/src/cas_client/simulation/local_client.rs
+++ b/xet_client/src/cas_client/simulation/local_client.rs
@@ -436,22 +436,22 @@ impl LocalClient {
     ///
     /// We hash multiple metadata fields to increase entropy and reduce false
     /// matches during rapid rewrite/delete races.
-    fn object_etag_from_path(path: &Path) -> Result<ObjectETag> {
-        let meta = std::fs::metadata(path).map_err(ClientError::internal)?;
-        let modified = meta.modified().map_err(ClientError::internal)?;
-        let modified_nanos = modified.duration_since(std::time::UNIX_EPOCH).unwrap_or_default().as_nanos();
-        let created_nanos = meta
-            .created()
-            .ok()
-            .and_then(|ts| ts.duration_since(std::time::UNIX_EPOCH).ok())
-            .map_or(0u128, |d| d.as_nanos());
-
-        let mut entropy = Vec::with_capacity(16 + 16 + 8 + 1);
-        entropy.extend_from_slice(&modified_nanos.to_le_bytes());
-        entropy.extend_from_slice(&created_nanos.to_le_bytes());
-        entropy.extend_from_slice(&meta.len().to_le_bytes());
-        entropy.push(u8::from(meta.permissions().readonly()));
-
+    /// The object's etag, derived from its key and stored length rather than
+    /// from filesystem timestamps.
+    ///
+    /// S3's ETag is a function of content, so a byte-identical re-upload leaves
+    /// it unchanged. Xorbs and shards are content-addressed, so the key already
+    /// fixes the content and the length is what distinguishes a differently
+    /// serialized rewrite of the same key. Hashing mtime/ctime instead made
+    /// every write look like different content — including a re-upload, and
+    /// including the rename onto a temp path that the conditional deletes do.
+    fn object_etag_for(prefix: &[u8], key: &MerkleHash, path: &Path) -> Result<ObjectETag> {
+        let len = std::fs::metadata(path).map_err(ClientError::internal)?.len();
+        let key_bytes: [u8; 32] = (*key).into();
+        let mut entropy = Vec::with_capacity(prefix.len() + key_bytes.len() + 8);
+        entropy.extend_from_slice(prefix);
+        entropy.extend_from_slice(&key_bytes);
+        entropy.extend_from_slice(&len.to_le_bytes());
         Ok(compute_data_hash(&entropy).into())
     }
 
@@ -1017,7 +1017,7 @@ impl super::DeletionControlableClient for LocalClient {
             if let Some(pos) = name.rfind('.') {
                 let hex = &name[(pos + 1)..];
                 if let Ok(hash) = MerkleHash::from_hex(hex) {
-                    let etag = Self::object_etag_from_path(&path)?;
+                    let etag = Self::object_etag_for(b"xorb", &hash, &path)?;
                     ret.push((hash, etag));
                 }
             }
@@ -1036,7 +1036,7 @@ impl super::DeletionControlableClient for LocalClient {
             return Err(ClientError::XORBNotFound(*hash));
         }
 
-        let current_etag = match Self::object_etag_from_path(&tmp_path) {
+        let current_etag = match Self::object_etag_for(b"xorb", hash, &tmp_path) {
             Ok(t) => t,
             Err(e) => {
                 Self::restore_from_tmp(&tmp_path, &file_path);
@@ -1093,7 +1093,7 @@ impl super::DeletionControlableClient for LocalClient {
     async fn list_shards_with_etags(&self) -> Result<Vec<(MerkleHash, ObjectETag)>> {
         let mut ret = Vec::new();
         for (hash, path) in self.shard_file_paths()? {
-            let etag = Self::object_etag_from_path(&path)?;
+            let etag = Self::object_etag_for(b"shard", &hash, &path)?;
             ret.push((hash, etag));
         }
         Ok(ret)
@@ -1107,7 +1107,7 @@ impl super::DeletionControlableClient for LocalClient {
             return Err(ClientError::Other(format!("Shard not found: {}", hash.hex())));
         }
 
-        let current_etag = match Self::object_etag_from_path(&tmp_path) {
+        let current_etag = match Self::object_etag_for(b"shard", hash, &tmp_path) {
             Ok(t) => t,
             Err(e) => {
                 Self::restore_from_tmp(&tmp_path, &path);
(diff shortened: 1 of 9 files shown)


seanses (commented):
.

xet_client/src/cas_client/simulation/local_client.rs
```diff
@@ -1050,35 +1076,63 @@ impl super::DeletionControlableClient for LocalClient {
...
+        if !self.get_path_for_entry(hash).exists() {
+            return Err(ClientError::Other(format!("XORB not found: {}", hash.hex())));
+        }
+        let path = self.tag_set_xorb_path(hash);
+        let raw = serde_json::to_vec(&tags).map_err(|e| ClientError::Other(format!("serialize tag set: {e}")))?;
+        #[cfg(windows)]
+        if path.exists() {
+            Self::clear_readonly(&path);
+        }
+        std::fs::write(&path, raw)?;
```
seanses, line 1106: This is not atomic write, is there a possible case that the same file is read at the same time, in the above `get_xorb_tag_set` function?

To atomic write, write to a temp file and then atomic mv, i.e. rename. There is also a utility for the entire process: `SafeFileCreator`

xet_client/src/cas_client/simulation/memory_client.rs
```diff
@@ -1204,62 +1230,75 @@ impl super::DeletionControlableClient for MemoryClient {
         } else {
             self.xorbs.write().await.remove(hash);
         }
+        self.clear_xorb_tag_set(hash).await;
```
seanses, line 1233: Note that removing the hash from the `xorbs` map and clearing the xorb tag is not one atomic operation, so re-uploading can race with delete, leading to a situation where a new xorb is uploaded, but its tag is erased.

xet_client/src/cas_client/simulation/local_client.rs
sirahd: Fixed!

xet_client/src/cas_client/simulation/memory_client.rs
sirahd: Fixed by acquiring all of the locks in the critical section within the same scope

xet_client/src/cas_client/simulation/local_client.rs
```diff
@@ -1663,6 +1721,23 @@ impl Client for LocalClient {
...
+        // the whole tag set, so the stamp both records this write and clears
+        // whatever was there.
+        let tag_set_path = self.tag_set_xorb_path(&hash);
+        match serde_json::to_vec(&last_upload_tag_set_now()) {
+            Ok(raw) => {
+                #[cfg(windows)]
+                if tag_set_path.exists() {
+                    Self::clear_readonly(&tag_set_path);
+                }
+                if let Err(e) = std::fs::write(&tag_set_path, raw) {
```
seanses: Just want to call out another place where we should use atomic file write

xet_client/src/cas_client/simulation/memory_client.rs
```diff
@@ -961,7 +948,15 @@ impl Client for MemoryClient {
         let bytes_written = serialized_data.len();
 
         {
+            // The write, the un-tag and the `last-upload` stamp are one critical section:
+            // a delete interleaving between them would erase the tag of the xorb this call
+            // just wrote. Locks are taken tagged-before-xorbs everywhere, so readers that
+            // hold both cannot deadlock against this.
+            let mut tagged = self.gc_tagged_xorbs.write().await;
```
seanses: Nit: I want to raise a maintenance concern that acquiring these locks separately requires all other places acquiring any more than two of them following the same order, or risk a deadlock. So maybe consider combining them into a single lock or split the locking logics into functions to ensure all places follow the same order.

xet_client/src/cas_client/simulation/local_client.rs
```diff
@@ -1050,35 +1076,67 @@ impl super::DeletionControlableClient for LocalClient {
         } else {
             std::fs::remove_file(&tmp_path)?;
         }
+        self.clear_tag_set_xorb(hash);
```
seanses: I think we have the similar concern here that delete racing with upload and erase a concurrent upload's fresh tags. We should apply the same fix as done for the "memory_client.rs" counterpart

seanses (approved):
Approving to unblock but fix the above issues before merge

xet_client/src/cas_client/simulation/local_client.rs
sirahd: Fixed in 352f334d. Both sidecar writers now share one `write_tag_set_file` helper that goes through `SafeFileCreator`, rather than each having its own write — having two copies is what let this one drift out of sync with the fix to `set_xorb_tag_set`.

xet_client/src/cas_client/simulation/local_client.rs
sirahd: Fixed in 352f334d, in both `delete_xorb` and `delete_xorb_if_etag_matches`.

There are no locks here, so the `MemoryClient` fix does not port directly — but this file already had the right idiom for the data: it renames the xorb out of the namespace first, so it only ever acts on bytes it exclusively claimed. The sidecar was simply left out of that scheme and unlinked by path at the end. It is now claimed the same way via `claim_tag_set_xorb`, restored on every path that backs out, and only the claimed copy is unlinked, so a sidecar written by an upload that overtakes the delete is never touched.

One limit worth stating: this makes `delete_xorb_if_etag_matches` correct, because the etag guard backs it. Unguarded `delete_xorb` still races with a concurrent upload by nature, as S3 DeleteObject does against PutObject, so there the change narrows the damage to "never unlink a sidecar we did not claim" rather than removing the race.

`clear_tag_set_xorb` is dead now and removed, same as its `MemoryClient` counterpart last round.

xet_client/src/cas_client/simulation/memory_client.rs
sirahd: Went with your first suggestion: `xorbs`, `gc_tagged_xorbs` and `xorb_tag_sets` are now one `RwLock<XorbState>`, so there is no order left to get wrong and no comment anyone has to notice.

The tradeoff is read concurrency: xorb reads now block tag writes, where before they were independent.

xet_client/src/cas_client/simulation/memory_client.rs
```diff
@@ -121,7 +137,18 @@ impl MemoryClient {
...
-        self.gc_tagged_xorbs.read().await.contains(hash)
+        self.xorb_state.read().await.tagged.contains(hash)
+    }
+
+    /// Errors unless the xorb is present and not condemned. A lifecycle-tagged
+    /// xorb reads as gone everywhere else here, and `LocalClient` errors on it
+    /// because the canonical file has been renamed away.
+    #[cfg(not(target_family = "wasm"))]
+    async fn require_readable_xorb(&self, hash: &MerkleHash) -> Result<()> {
+        if !self.xorb_state.read().await.xorbs.contains_key(hash) || self.xorb_is_tagged(hash).await {
```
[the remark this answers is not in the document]
sirahd: Correct, and a regression from folding the three xorb locks into one in the previous commit — those two reads used to be on different locks, so the nesting was harmless until it wasn't. Fixed in 61c2511e: one guard now serves both checks.

Audited the rest of the file for the same shape rather than just this instance: no other bound guard is held across a self-method await, the five inline uses are each a single self-contained statement, and both `require_readable_xorb` callers acquire sequentially after the `?`. This was the only one.

xet_client/src/cas_client/simulation/local_client.rs
```diff
@@ -1026,15 +1076,27 @@ impl super::DeletionControlableClient for LocalClient {
             return Err(ClientError::XORBNotFound(*hash));
         }
 
-        let current_tag = match Self::object_tag_from_path(&tmp_path) {
+        // Claim the sidecar the same way, and put it back on every path that backs out.
+        // An upload that overtakes this delete rewrites both canonical paths, and neither
+        // of those is what gets unlinked below.
+        let claimed_tag_set = self.claim_tag_set_xorb(hash);
```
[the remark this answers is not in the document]
sirahd: Right on both counts, including that `delete_xorb` doesn't have the hole — I applied the ordering correctly in one path and not the other. Fixed in 61c2511e by claiming the sidecar before the data rename, with a restore added on the not-found path.

The comment I had written there actually asserted the opposite ("neither of those is what gets unlinked below"), which was wrong for exactly the reason you give: the claim happened after the window opened. Replaced it with one that says why the order matters. Restoring stays safe against a concurrent upload because `restore_from_tmp` declines to clobber a path that has been recreated.

PR #960, opened 2026-09-05
fix: reject degenerate RTT predictions in adaptive concurrency

### LINKED ISSUE #875: Adaptive concurrency: RTT predictor emits degenerate predictions from near-singular fits
The adaptive concurrency RTT predictor (`xet_client/src/cas_client/adaptive_concurrency/`) can emit physically-impossible predictions when its weighted linear regression is near-singular — too few effective samples, or near-uniform transfer sizes at a single concurrency level. Filing to confirm whether this is a real controller bug: do these values feed concurrency decisions, or only reporting/logging?

### Mechanism
- **`predicted_bandwidth()` ≈ 9.5 TiB/s is the `1e-6` floor, not a measurement.** When `predicted_rtt(10 MiB, conc=1)` comes out negative it is clamped to 0 (`rtt_prediction.rs:94`), then `10 MiB / min_rtt.max(1e-6)` gives `10·1024·1024 / 1e-6 = 9.5367 TiB/s` (`rtt_prediction.rs:115-121`) — a data-independent constant.
- **`predicted_max_rtt` reaches tens of seconds**: the spurious regression slope extrapolated out to the 64 MiB reference query point.
- **The degenerate-fit signal is dropped.** `df = sw − 2 ≤ 0` returns `(Some(mean), None)` (`exp_weighted_olr.rs:86-87`), and the controller forwards the mean anyway via `.unwrap_or(0.)` (`controller.rs:143-145`).
- The singularity guard is **absolute** (`delta.abs() < 1e-12`, `exp_weighted_olr.rs:75`), so a tiny-but-nonzero `delta` (near-collinear x) passes and yields an exploding slope / large-negative intercept.

### Trigger
Low x-variance regime: early in a transfer, or uniform-size transfers at one concurrency level → the regression x-axis has almost no spread, so RTT jitter drives a near-vertical slope that is then extrapolated far past the observed size range.

### Suggested direction
Don't emit predictions outside the identifiable regime:
- Gate `predict()` on a **relative** condition number / minimum x-spread (not the absolute `1e-12`); return `None` for the mean when `df ≤ 0`.
- Have `predicted_bandwidth()` return `None` when the underlying RTT was clamped to 0, instead of dividing by the `1e-6` floor.
- Optionally cap extrapolation to near the observed size range.

MuhammadMaazA:
Fixes #875.

This change makes the RTT regression reject near-singular fits using a relative conditioning check instead of an absolute determinant threshold. This prevents very small variation in transfer sizes from producing unstable slopes and unrealistic extrapolated RTTs.

It also prevents a clamped zero RTT from being converted into a synthetic ~9.5 TiB/s bandwidth estimate by returning None for non-positive predicted RTTs.

Tests cover both the near-singular regression case and the zero-RTT bandwidth case.

---

> [!NOTE]
> **Medium Risk**
> Changes RTT regression gating and concurrency-increase inputs in adaptive upload/download control, which can alter throughput behavior but not auth or data integrity.
> 
> **Overview**
> Fixes degenerate RTT modeling that could **inflate concurrency** or report **absurd bandwidth** when transfer sizes barely vary.
> 
> **Exp-weighted OLR** now treats a fit as unidentifiable using a **relative** normal-matrix check (`delta` vs `sw * sxx`) instead of a fixed absolute cutoff, so near-collinear size samples no longer yield unstable slopes. **`predicted_bandwidth`** returns `None` when the RTT estimate is non-positive, instead of dividing by a tiny floor after negative predictions are clamped to zero.
> 
> New tests cover near-singular regression rejection, zero-RTT bandwidth, and an end-to-end case where the controller must **not** raise permits when the RTT fit is unreliable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c3a0d74fab69207306dada96c6e4792e5066665d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/xet_client/src/cas_client/adaptive_concurrency/exp_weighted_olr.rs
+++ b/xet_client/src/cas_client/adaptive_concurrency/exp_weighted_olr.rs
@@ -65,15 +65,17 @@ impl ExpWeightedOnlineLinearRegression {
     /// Predict the mean and standard deviation of the *fitted mean* at x.
     ///
     /// Returns:
-    ///   - (None, None) if the model is not identifiable yet (normal matrix singular).
+    ///   - (None, None) if the model is not identifiable yet (normal matrix singular or ill-conditioned).
     ///   - (Some(mean), None) if coefficients can be estimated but df <= 0, so we can't compute a standard deviation
     ///     yet.
     ///   - (Some(mean), Some(std_dev)) otherwise.
     pub fn predict(&self, x0: f64) -> (Option<f64>, Option<f64>) {
-        // Need a well-conditioned normal matrix to estimate beta.
+        // Need a well-conditioned normal matrix to estimate beta. Use a relative
+        // threshold so near-collinear x values are rejected independent of scale.
         let delta = self.sw * self.sxx - self.sx * self.sx;
-        if delta.abs() < 1e-12 {
-            // Can't estimate beta0/beta1 at all.
+        let scale = (self.sw * self.sxx).abs();
+        if scale == 0.0 || delta <= 1e-12 * scale {
+            // Can't estimate beta0/beta1 reliably.
             return (None, None);
         }
 
@@ -114,7 +116,8 @@ impl ExpWeightedOnlineLinearRegression {
     #[allow(dead_code)]
     pub fn coefficients(&self) -> Option<(f64, f64)> {
         let delta = self.sw * self.sxx - self.sx * self.sx;
-        if delta.abs() < 1e-12 {
+        let scale = (self.sw * self.sxx).abs();
+        if scale == 0.0 || delta <= 1e-12 * scale {
             return None;
         }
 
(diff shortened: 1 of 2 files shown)


seanses:
I'm not sure if the referenced issue is valid or needs fix. Will double check. Low priority.

MuhammadMaazA:
Thanks. I traced this through the controller path and added a controller-level regression test to make the reported behaviour easier to evaluate.

The test seeds the RTT model with successful 32 MiB transfers whose sizes differ by only a few bytes. With the previous absolute determinant guard, RTT jitter produces an unstable negative slope. Extrapolation to the reference size at the next concurrency level is then clamped to zero, causing the controller to increase concurrency from 2 to 3. With the relative conditioning guard, the fit is treated as unidentifiable and the controller remains at 2 until it has usable data.

The regression test fails with the previous guard and passes with this change. The full `xet-client` library suite also passes with 209 tests passed and 4 ignored.

To clarify the scope, `predicted_rtt()` is the part that directly affects the concurrency decision. The `predicted_bandwidth()` change only prevents the same zero RTT result from becoming a synthetic throughput value in the reported model state.

PR #962, opened 2026-09-09
fix(simulation): store uploaded shards the way CAS does for LocalClient

sirahd:
The simulation rebuilt each uploaded shard through `MDBInMemoryShard`, while CAS re-serializes through the streaming serializer and keys the object by the hash of those bytes. Only the streaming path stamps `shard_creation_timestamp`, so identical content collapsed onto a single hash forever in the simulation but not in production. That let a re-upload land back on a hash GC had already deregistered from the global dedup table and still had queued for deletion, leaving dedup rows pointing at a condemned shard — `stress_test_high_fp` failed on it 7 runs out of 7.

Two consequences of storing what CAS stores:
- shards carry no lookup tables, so `get_file_chunk_hashes` reads its file entry from the authoritative table, as the reconstruction path already did
- they record a zero key expiry, so this server-side store is no longer scanned on open as if it were a client cache — that scan prunes by exactly that field

Test plan: full GC stress suite against this branch with no GC-side change, all three variants green and `integrity=36ok/0FAIL` each (`stress_test_high_fp` 14/14 epochs, `stress_test`, `stress_test_convergence`), at unchanged throughput and unchanged collection rates.

---

> [!NOTE]
> **Medium Risk**
> Changes simulation shard hashing, on-disk layout, and file lookup paths that GC stress tests depend on; behavior is intentional parity with CAS but affects dedup/GC edge cases in tests only.
> 
> **Overview**
> Aligns **simulation `LocalClient` shard persistence** with production CAS so GC stress tests match real dedup/GC behavior.
> 
> **Shard uploads** no longer rebuild via `MDBInMemoryShard`; validated input is re-serialized through the streaming path (`HashedWrite` + `minimal_shard.serialize`), keyed by the hash of those bytes (client-supplied hash ignored), written atomically via temp file + rename. Identical logical content therefore gets a **new shard hash** when the footer stamps a creation timestamp—fixing cases where re-upload reused a hash already deregistered/queued for deletion while dedup rows still pointed at it.
> 
> **Shard manager init** disables directory scan on open (`scan: false`) because CAS-form shards use zero key expiry; scanning would incorrectly prune them as if this were an expiring client cache.
> 
> **`get_file_chunk_hashes`** resolves file metadata from **`FILE_TO_SHARD_TABLE`** (`get_file_info_from_table`) instead of shard-manager reconstruction lookup, consistent with shards stored without embedded lookup tables.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5cec0e0586d2efafae5dfa588abe6886188e7404. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/xet_client/src/cas_client/simulation/local_client.rs
+++ b/xet_client/src/cas_client/simulation/local_client.rs
@@ -15,13 +15,14 @@ use redb::{ReadableDatabase, ReadableTable, TableDefinition};
 use tempfile::TempDir;
 use tokio::time::{Duration, Instant};
 use tracing::{error, info, warn};
-use xet_core_structures::merklehash::{MerkleHash, compute_data_hash};
+use xet_core_structures::merklehash::{HashedWrite, MerkleHash, compute_data_hash};
 use xet_core_structures::metadata_shard::file_structs::{FileDataSequenceHeader, MDBFileInfo, MDBFileInfoView};
-use xet_core_structures::metadata_shard::shard_file_reconstructor::FileReconstructor;
 use xet_core_structures::metadata_shard::shard_format::MDB_FILE_INFO_ENTRY_SIZE;
+#[cfg(test)]
 use xet_core_structures::metadata_shard::shard_in_memory::MDBInMemoryShard;
 use xet_core_structures::metadata_shard::streaming_shard::MDBMinimalShard;
 use xet_core_structures::metadata_shard::utils::{parse_shard_filename, shard_file_name};
+#[cfg(test)]
 use xet_core_structures::metadata_shard::xorb_structs::MDBXorbInfo;
 use xet_core_structures::metadata_shard::{MDBShardFile, MDBShardFileHeader, ShardFileManager};
 use xet_core_structures::serialization_utils::read_u32;
@@ -325,8 +326,11 @@ impl LocalClient {
             write_txn.commit().map_err(map_redb_db_error)?;
         }
 
-        // Open / set up the shard lookup
-        let shard_manager = ShardFileManager::new_in_session_directory(&ctx, shard_dir.clone(), true).await?;
+        // Open / set up the shard lookup. Not scanned on open: the scan prunes shards whose
+        // key expiry has passed, and shards here are stored as CAS stores them, which records
+        // a zero expiry. This directory is a server-side store, not an expiring client cache,
+        // and nothing reads it back through the manager.
+        let shard_manager = ShardFileManager::new_in_session_directory(&ctx, shard_dir.clone(), false).await?;
         #[cfg(feature = "fd-track")]
         report_fd_count("LocalClient::new_internal after shard manager init");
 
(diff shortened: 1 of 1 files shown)


xet_client/src/cas_client/simulation/local_client.rs
```diff
@@ -1522,34 +1526,33 @@ impl Client for LocalClient {
...
-            in_memory_shard.add_file_reconstruction_info(MDBFileInfo::from(file_view))?;
-        }
+        // Store the shard exactly as CAS does: re-serialize the validated shard through the
+        // streaming serializer and store those bytes under their own hash, discarding the
+        // hash the client sent (see `validate_shard_from_async_read` in cas_server). That
+        // footer stamps a creation timestamp, so re-uploading identical content produces a
+        // distinct object rather than reusing one hash forever.
+        let mut cas_form = Vec::with_capacity(minimal_shard.serialized_size(false));
+        let mut hashed_writer = HashedWrite::new(&mut cas_form);
+        minimal_shard.serialize(&mut hashed_writer, false)?;
```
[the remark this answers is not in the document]
sirahd: Good catch — confirmed and fixed in f62f815f.

The chain is real: `From<&MDBFileInfoView> for MDBFileInfo` preserves verification and `MDBFileInfo::serialize` writes it, so the previous in-memory path did store those entries, and re-serializing with verification off dropped them.

Worth noting why matching CAS exactly is not the fix here. CAS trims verification too, but it serves file info from its own database (`cas_db.get_file_reconstruction`), which retains it — in the simulation the stored shard *is* what file lookups read, so trimming loses data production never loses. So verification is now kept on write.

One wrinkle: `serialize` returns `missing requested verification info` unless every file carries verification, so the flag is computed from the shard rather than hardcoded. A shard mixing files with and without verification still stores none, since the API has no per-file control — same as before this PR.

Added `test_upload_shard_keeps_file_verification`, which fails without the fix with `lost its verification entries on upload`.

xet_client/src/cas_client/simulation/local_client.rs
```diff
@@ -1522,34 +1526,40 @@ impl Client for LocalClient {
...
-
-        // Add file info from the views
-        for i in 0..minimal_shard.num_files() {
-            let file_view = minimal_shard.file(i).unwrap();
-            in_memory_shard.add_file_reconstruction_info(MDBFileInfo::from(file_view))?;
-        }
+        // Store the shard exactly as CAS does: re-serialize the validated shard through the
+        // streaming serializer and store those bytes under their own hash, discarding the
+        // hash the client sent (see `validate_shard_from_async_read` in cas_server). That
+        // footer stamps a creation timestamp, so re-uploading identical content produces a
```
seanses, line 1532: What's the intent of this PR? I thought we actually don't want this behavior, i.e. re-uploading identical content produces a distinct object. And we can fix it by enforcing the timestamp in the serialize shard footer to be, for example, all zeros.

seanses (approved):

xet_client/src/cas_client/simulation/local_client.rs
seanses: Communicated offline.

PR #965, opened 2026-09-11
Add AVX 512 support for bg4 predictor

ajanon:
Add AVX 512 support for the bg4 predictor.

On my host with a Ryzen 9 7950X:
- SWAR ~3000 MB/s
- AVX-512 ~3300-3400 MB/s

This is a modest increase in performance (10%) because the path is mostly bound on the scalar histogram scatter. Improving the scatter could be a good path forward for further performance increases.

Make the bg4 predictor also more generic over the word size, to allow using larger word sizes to fit 64-byte AVX 512 words, or possibly for other SIMD implementations later on. The AVX 512 implementation is runtime-detected, to allow making a single x86_64 build, and still get the performance increase where supported.

The `debug_assert`s enforce the word-size contract documented on `add_data_impl` for any future word type. The current u128 and __m512i instantiations always satisfy it. A compile-time check would need [`generic_const_exprs`](https://doc.rust-lang.org/beta/unstable-book/language-features/generic-const-exprs.html), which is still nightly-only.

---

> [!NOTE]
> **Low Risk**
> Localized performance change to BG4 heuristic histograms with reference-equivalence tests and CPU feature fallbacks; no API or security-sensitive behavior changes beyond choosing a faster implementation of the same logic.
> 
> **Overview**
> Adds a **runtime-detected** x86-64 fast path for BG4 histogram building using **`vpopcntb`** (`avx512bitalg`), with **SWAR fallback** when the extension is missing. **`add_data`** on x86-64 now routes through this path instead of SWAR-only.
> 
> Refactors the inner driver to be **generic over SIMD word size** (`T: Pod` via **bytemuck**), so the same alignment/head/body/tail loop works for **16-byte SWAR/Neon** and **64-byte AVX-512**, including 16-byte chunking for histogram scatter on wider words.
> 
> Tests and the **bg4_prediction** benchmark now also exercise **`add_data_avx512`** and assert histogram parity with the reference implementation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cde23c6aa34d7d41ad0fd5aa23e81199a76630a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/xet_core_structures/Cargo.toml
+++ b/xet_core_structures/Cargo.toml
@@ -59,6 +59,10 @@ blake3 = { workspace = true }
 # without changing native builds.
 blake3 = { workspace = true, features = ["wasm32_simd"] }
 
+[target.'cfg(target_arch = "x86_64")'.dependencies]
+# Pod impls for the __m512i word type used by the AVX-512 BG4 popcount path.
+bytemuck = { workspace = true, features = ["avx512_simd"] }
+
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
 tokio = { workspace = true, features = [
     "time",
--- a/xet_core_structures/Cargo.toml
+++ b/xet_core_structures/Cargo.toml
@@ -29,6 +29,7 @@ xet-runtime = { version = "1.6.0", path = "../xet_runtime" }
 
 async-trait = { workspace = true }
 base64 = { workspace = true }
+bytemuck = { workspace = true }
 bytes = { workspace = true }
 clap = { workspace = true, optional = true }
 countio = { workspace = true }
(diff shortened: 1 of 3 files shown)


seanses:
Thank you for contribution! 

Notes for other reviewers: Running the bg4 predictor occupies only a tiny fraction of the upload runtime, so this is low-priority or not-planned.

ajanon:
> Thank you for contribution!
> 
> Notes for other reviewers: Running the bg4 predictor occupies only a tiny fraction of the upload runtime, so this is low-priority or not-planned.

Thanks for taking a look!

Yeah, that matches what I've seen too. I thought it would be good to have parity between amd64 and arm64. Feel free to close this if you think it's not worth the added complexity.

PR #967, opened 2026-09-16
Check each downloaded file against its hash after it is written

humuhumu33:
## Summary

Currently a full-file download checks that the written size is right, but not the content. In #604, about 9% of large files downloaded with the right size and the wrong bytes while the download reported success. #641 fixed that cause (thanks again to @volodymyrjf for tracking it down). This PR adds a content check next to the existing size check in `download_file_with_id`, so a problem like that fails the download instead of leaving a bad file. Range and stream downloads are unchanged.

- While terms are fetched, each chunk's length is recorded in file order (`ChunkLayout`).
- After writing, the file is read once and hashed along those chunks with `compute_data_hash` and `file_hash`. Reading it back, rather than hashing in memory, is what covers the write itself. A file of the wrong length returns the existing `DataError::SizeMismatch`; a different hash returns the new `DataError::HashMismatch`, which `xet_pkg` maps to `XetError::DataIntegrity`.

The chunk lengths come from the download instead of re-chunking the file, because files chunked before #487 hash differently with today's chunker. I checked five Hub files; each matched `X-Xet-Hash` only along the chunks the download delivered.

Checking downloads was set aside before because re-reading large files is slow (huggingface/huggingface_hub#3643). This reads the file once, right after writing it: 0.16 to 0.21 s per 512 MiB here, about 4 to 5% of a gigabit download. It is on by default; `HF_XET_DATA_VERIFY_DOWNLOADED_FILES=false` turns it off, as an escape hatch in case it ever refuses a file it should accept. Hash mismatches are counted under the `format` class in download telemetry, so a rise would be visible rather than sitting in `other`.

## Test plan

- [x] `cargo test -p xet-data --lib download_file_rejects_corrupted_content`: flips a byte in a local xorb; `reconstruct_to_file` writes wrong bytes and succeeds, `download_file` fails
- [x] `verify_written_file_reads_what_is_on_disk`: a changed byte, and an extra byte at the end, after writing
- [x] `chunk_layout_reproduces_file_hash`: small fetch batches that start inside a chunk
- [x] `chunk_layout_records_terms_served_from_a_fetched_block`
- [x] `an_empty_chunk_is_not_merged_with_the_next` and `download_empty_file_passes_the_check`
- [x] With `verify_downloaded_files = false`, the corrupted download goes through as before
- [x] `cargo test -p xet-data --lib` (285 passed), `cargo clippy -p xet-data -p hf-xet -p xet-runtime -- -D warnings`, `wasm32-unknown-unknown`, `cargo +nightly fmt --all`
- [x] Six real Hub files (safetensors, GGUF, parquet, tokenizer) downloaded through the changed session, all accepted

<details>
<summary>Cost</summary>

Release build, 512 MiB file, local store, median of five rounds in each of two runs, AMD Ryzen AI MAX 390, Windows 11, NVMe. x86_64 only.

| | Run 1 | Run 2 |
|---|---|---|
| Reconstruct only | 0.489 s, 1,046 MiB/s | 0.627 s, 816 MiB/s |
| Download with check | 0.667 s, 767 MiB/s | 0.966 s, 530 MiB/s |
| Check alone | 0.157 s, 3,263 MiB/s | 0.214 s, 2,391 MiB/s |

The added time is one read of a file the page cache has just seen, so it grows with file size, not with network speed. Against a local store it is +36 to +54%. At 125 MiB/s the same 0.16 to 0.21 s is about 4 to 5%. A slow disk, or a file larger than free memory, will pay more.

Memory is 16 bytes per chunk while downloading, then 48 bytes per chunk during the check, about 8 MB for a 10 GiB file. Upload holds the same chunk list.

</details>

---

> [!NOTE]
> **Medium Risk**
> Adds a post-download integrity gate on the main file-to-disk path with an extra full-file read; incorrect chunk recording could reject valid files, though behavior is configurable and covered by new tests.
> 
> **Overview**
> Full-file downloads now verify **content** after write, not just byte count. When `verify_downloaded_files` is on (default), reconstruction records per-chunk lengths via **`ChunkLayout`** as terms are fetched, then **`verify_written_file`** re-reads the file on disk and checks the merkle hash using those boundaries (so hashing matches how the file was originally chunked). Mismatches surface as **`DataError::HashMismatch`** (mapped to **`XetError::DataIntegrity`**); size issues still use **`SizeMismatch`**. Stream/range/`download_to_writer` paths are unchanged.
> 
> **`FileReconstructor`** / **`FileTerm`** accept an optional chunk layout; verification can be disabled with **`HF_XET_DATA_VERIFY_DOWNLOADED_FILES=false`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 54e1289dd5e8f5eeb44b73300460bdcfc2b52c9b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


seanses:
Thank you for contributing but I don't think we want this feature.

PR #968, opened 2026-09-16
perf: avoid zero-filling read buffers on the xorb chunk decode path

assafvayner:
Profiling showed a significant share of server-side CPU going to the allocator zeroing recycled large extents and the page faults taken when writing to them again. Two of the largest contributors sit on the xorb download/deserialize path, both bottoming out in `vec![0u8; n]` buffers in this crate that are immediately fully overwritten by `read_exact` right after allocation, so the zero-fill is pure waste.

- Switched the four hot allocation sites (compressed-chunk and chunk-data reads in `xorb_chunk_format/deserialize_async.rs` and `xorb_object_format.rs`) to `Vec::with_capacity` + `set_len`, each with a comment on why every byte is guaranteed written before any read.
- Left the two `.resize(n, 0)` sites at `xorb_object_format.rs` test helpers alone — they're under `#[cfg(test)]` and don't fully overwrite the buffer, so they're out of scope.
- Added a round-trip test (`test_roundtrip_byte_exact_with_differing_compressed_length`) that serializes an LZ4-compressed chunk whose compressed and uncompressed lengths differ, deserializes it, and asserts byte-exact equality — so an under-filled buffer would fail loudly.

---

> [!NOTE]
> **Medium Risk**
> Uses `unsafe` uninitialized buffers on the core chunk decode path; behavior is unchanged only if every call site always completes a full `read_exact` before reading the buffer.
> 
> **Overview**
> This PR cuts allocator work on the **xorb download/deserialize** path by stopping unnecessary zero-fills on large read buffers that are immediately overwritten by `read_exact`.
> 
> It introduces a local **`alloc_uninit`** helper (`Vec::with_capacity` + `unsafe set_len`) in `deserialize_async.rs` and `xorb_object_format.rs`, with comments that every byte is filled by a full read before use. **Four** former `vec![0u8; n]` sites now use it: async compressed-chunk reads, range reads in `get_range`, and compressed payload reads in `reconstruct_xorb_with_footer`.
> 
> A new async test **`test_roundtrip_byte_exact_with_differing_compressed_length`** LZ4-round-trips data where compressed size differs from uncompressed size, so any under-read leaving garbage in the buffer would fail decompression or byte equality.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1d53c56370e944c4b882667bcb163eff18946e5c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


seanses (changes requested):
> significant share of server-side CPU going to the allocator zeroing recycled large extents and the page faults taken when writing to them again

Hmm... this statement sounds flimsy. Handling page faults indeed takes time, but only at the first access to a page. So either the cost is expensed at zeroing out the memory (first access) or if not zeroing, delayed to the actual first access. Either way, there's no page faults when writing to them again. 

For large allocations that requires getting a new frame from the OS, those frames are guaranteed by OS to be zeroed out before mapping to virtual pages, to avoid data leak across processes. And Rust's `vec![0; N]` macro, essentially `std::alloc::alloc_zeroed`, benefits from that guarantee so will not `memset` at all.

However regarding the specific cases in this PR, the memory allocations are for compressing / decompressing one chunk, which is roughly 64 KiB. These small allocations may come from recycled memory chunks from its heap freelist and indeed allocator runs an optimized SIMD `memset` to zero out the bytes.

But I think there are better ways to handle this inefficiency than introducing unsafe memory allocating functions:
1. Our use cases don't typically compress / decompress one chunk, but a long list of chunks. We can refactor the code to reuse the read buffer for all these chunks, so that's only one allocation instead of multiple.
2. Regarding `deserialize_chunk_with_header_to_writer` specifically, it is nested under three levels to deserialize one chunk (`deserialize_chunk` -> `deserialize_chunk_to_writer` -> `deserialize_chunk_header` + `deserialize_chunk_with_header_to_writer`), where each level allocates some buffer. It's worthwhile to refactor the code to avoid all these small and separate allocations.

I think reducing the number of allocations will contribute more than just not zeroing out the memory. These refactors may make these low level functions inelegant, but since they are a time critical part of our infrastructure, we can sacrifice the readability.

PR #977, opened 2026-09-22
feat: add default-on `logging` feature to gate the tracing subscriber stack

assafvayner:
`xet_runtime::logging::init` is the only user of `tracing-subscriber` and `tracing-appender`. Putting it behind a `logging` feature — opt-in on xet-runtime, on by default in hf-xet — lets consumers that install their own subscriber drop **16 crates** from the build (`hf-xet`: 205 → 189, nothing added). Instrumentation is unaffected; the `tracing` macros and the spans and events these crates emit stay exactly as they are. `LoggingConfig`, `SystemMonitor`, and the rest of the module stay ungated; only `init` / `wait_for_log_directory_cleanup` and `xet::init_logging` sit behind the feature.

- `logging` is deliberately **not** a xet-runtime default. If it were, every path through xet-data / xet-client / xet-core-structures would turn it on, and cargo has no way to un-enable a default — the switch would never reach a consumer unless each sub-crate carried `default-features = false`.
- `tracing-subscriber` moves to the non-wasm target table. `init` was already `#[cfg(not(target_family = "wasm"))]`, so wasm builds were compiling it for nothing; both crates are now gone from the `hf_xet_wasm` tree.
- `tokio-console` implies `logging`; so do `log_test_executable`, its integration test, and the `simulation` crate, which calls `logging::init` directly.
- The new CI job checks the tree, not just the build — `cargo check` alone would keep passing if someone re-added a non-optional `tracing-subscriber`, so it also greps `cargo tree` and fails if the stack reappears.

Verified: `cargo check --workspace --all-targets`; `cargo check -p xet-runtime --all-targets` and `-p hf-xet --no-default-features --features rustls-tls --all-targets`; clippy clean with `-D warnings` in all four configs; `cargo test -p xet-runtime` 220 passed, 323 with hf-xet selected (the log tests run whenever anything enables the feature, which the workspace test job does); `cargo machete` clean; wasm32 builds; no new rustdoc warnings; no Cargo.lock churn, including the two wasm lockfiles CI checks.

Expect a trivial conflict with #975 on hf-xet's `default = [...]` line and the README feature section — whichever lands second should end up documenting that `default-features = false` drops both `upload` and `logging`.

---

> [!NOTE]
> **Medium Risk**
> Changes default Cargo features and public API availability of `init_logging`, so downstream crates using `default-features = false` must opt back into `logging` or provide their own subscriber.
> 
> **Overview**
> Adds a **`logging`** feature so **`tracing-subscriber`** / **`tracing-appender`** (and related setup like **`init_logging`**) compile only when consumers want the built-in subscriber, while **`tracing`** instrumentation stays always available.
> 
> **`hf-xet`** turns **`logging`** on by default and documents how **`default-features = false`** drops it (and how to re-enable with TLS-only configs). **`xet-runtime`** exposes opt-in **`logging`** (not a default there) and gates **`logging::init`** / cleanup helpers; **`tokio-console`**, log test binaries, and **`simulation`** explicitly enable **`logging`** where they call init.
> 
> CI adds **`check-no-logging`** (plain **`xet-runtime`** check, **`hf-xet`** with **`--no-default-features`**, plus a **`cargo tree`** guard against **`tracing-subscriber`** sneaking back) and tightens workflow **permissions**; a few action pin comments are updated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0fd77a4488eeb3332b81fbf2fa277f0a7b8aee59. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/ci.yml
assafvayner, line 1: some extraneous changes here consistent with some dependabot usual auto updates

